PR #731, opened 2014-12-31
NPCs: Introduce Expansion Restriction Support

starlightknight:
This commit introduces support for restricting npcs by expansion, using the same approach that was used for spells/abilities/traits

This gets rid of npcs that are not relevant to the era that the server wants to play in - ex: Cavernous Maws without WOTG, Geo stuff from SOA, etc

The approach for tagging npcs is the same as for the other files - the expansion tagged is the latest expansion out at the time the npc was introduced

I only hit obvious ones and tagged expansion zones with their expansion - I am sure they are more. If there are specific npcs that should be mass tagged with a specific expansion, I have a script that can do this by polutils name (will commit it in my next batch update, coming soon)

This makes things more accurate to the era a server wants to play in, as well as reduces memory and resource waste for those servers by not loading unwanted npcs.

I have tested disabling all expansions and re-enabling and the functionality appears to work. I don't get any startup warnings with all expansions disabled. If you find any issues though feel free to tag me and I will look into it.

### DIFF
--- a/scripts/globals/conquest.lua
+++ b/scripts/globals/conquest.lua
@@ -1160,21 +1160,24 @@ function SetRegionalConquestOverseers(region)
     local nation  = GetRegionOwner(region);
 
     for i = 1, table.getn(npclist), 2 do
-        if(npclist[i+1] == nation) then
-            GetNPCByID(npclist[i]):setStatus(0);
-        else
-            GetNPCByID(npclist[i]):setStatus(2);
-        end
-
-        if(npclist[i+1] == OTHER) then
-            if(nation ~= BEASTMEN) then
-                GetNPCByID(npclist[i]):setStatus(0);
+    	local npc = GetNPCByID(npclist[i]);
+    	
+    	if(npc ~= nil) then
+            if(npclist[i+1] == nation) then
+                npc:setStatus(0);
             else
-                GetNPCByID(npclist[i]):setStatus(2);
+                npc:setStatus(2);
             end
-        end
+        
+            if(npclist[i+1] == OTHER) then
+                if(nation ~= BEASTMEN) then
+                    npc:setStatus(0);
+                else
+                    npc:setStatus(2);
+                end
+            end
 	end
-
+    end
 end;
 
 -----------------------------------
@@ -1311,25 +1314,3 @@ function conquestUpdate(zone, player, updateType, messageBase)
         end
     end
 end;
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
--- a/scripts/globals/settings.lua
+++ b/scripts/globals/settings.lua
@@ -27,6 +27,12 @@ ENABLE_ASA     = 0;
 ENABLE_ABYSSEA = 0;
 ENABLE_SOA     = 0;
 
+-- Setting to lock content more accurately to the expansions you have defined above
+-- This generally results in a more accurate presentation of your selected expansions
+-- as well as a less confusing player experience for things that are disabled (things that are disabled are not loaded)
+-- This feature correlates to the required_expansion column in the SQL files
+RESTRICT_BY_EXPANSION = 1;
+
 -- CHARACTER CONFIG
 INITIAL_LEVEL_CAP = 50; --The initial level cap for new players.  There seems to be a hardcap of 255.
 MAX_LEVEL = 75; -- Level max of the server, works by disabling Maat quests.
--- a/scripts/globals/zone.lua
+++ b/scripts/globals/zone.lua
@@ -23,9 +23,7 @@ function SetFieldManual(manuals)
     if (FIELD_MANUALS == 1) then
         for i,id in ipairs(manuals) do
             local npc = GetNPCByID(id);
-            if(npc == nil) then
-                printf("'SetFieldManual' Error trying to load undefined npc (%d)", id);
-            else
+            if(npc ~= nil) then
                 npc:setStatus(0);
             end
         end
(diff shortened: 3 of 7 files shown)


teschnei:
fix the merge conflicts and i'll merge it

gedads:
The only main problem i see is that set aside jap users, no one ever played ffxi basic, the rest of the world played ffxi roz cop.
Also, base ffxi is pre abyssea. (up to wotg and even walk of echoes was 75 content). For woe, the difficulty was so high it has to be done with a large amount of players (like einjerhar, dynamis, ans the other event in aht urghan where you go naked at first and need items to unlock jobs ability etc).
Also, in retail, you can see planar rift , maws, ship to aht urghan, vortex for cop, waypoints... you just aren't able to use those and probably just have  a message displayed (can't check since i have all addons enabled). 
In zitah, mobs area wandering same in romaeve, for asa addon... they're not hidden to players.
SOA is filled with mobs and npcs... what's the point in specifying which expansion it belongs to since a player can't normally go there?

gedads:
One last thing
It's impossible to claim we want a old school retail server:
Game balance is enhanced through the life of the game, some jobs were unplayable, (ninja can't tank, pld being useless, war used to rely on dual axes for years because of poor ga accuracy attack power, monk dying in a second, drk... what is that job? who plays it in old school ffxi? drg? i discovered this job years after, and i always thought it was a special event when one was in a party.) you can't reject new stuff because seing npcs granting quests or access to other areas are there and take all the progress up to adding jobs like pup or blu while toau missions don't even work yet. (dnc as well).
I've played for years ffxi, there's still event i don't do (reives, dynamis, salvage, einerjhar, pankration chocobo racing etc...) but i wouldn't understand seing them removed from retail because i don't use them or seeing them could be a nuisance to me.

R3P0FFXI:
If you merge this I hope you are taking the time to label every NPC for every expansion through SOA

starlightknight:
I can add a setting to disable this expansion restriction system for people that would rather not have it active, its only an extra line of code. While its true we can't get a perfectly accurate old school implementation due to client issues, I think you can get close.

@R3P0FFXI if you have major merge conflicts from custom stuff in-line let me know and I can help you - see merge advice in previous comments. If that is not sufficient it is easy to write a script to fill in remaining custom columns by counting commas to see if you need to add the new column

R3P0FFXI:
Star please add that line, we have a lot of custom content that we are still trying to get to work right from the ZMQ merge and variable changes that were recently done, with all the pending changes as is from the packet and moghouse stuff I am really not looking forward to our mass February update.

gedads:
another idea could be to finish "base" ffxi "quickly" (can be done if we work together on missions events dynamis and jailers) with base+rotz+cop up to the last verse which was the main project.
then branch and keep the db as it is for people willing to go farther?

starlightknight:
@gedads Ongoing merging between branches is a lot of manual work. Even though people are going to be annoyed at short term merge conflicts with this, they'll only have to do it one time. Otherwise someone has to keep that branch up to date constantly,.

The restrict-by-expansion set of patches tries to provide out of box support for both ways depending on your preference - the only time it really gets in your way is if you have a ton of custom stuff

R3P0FFXI:
gedads we have completed most of abyssea, the problems we have now are getting the mobs to have variable HP which no one on our project except maybe Teo who has been assisting us would know how to do, as well as getting the items to drop from chests properly as the containers would need to be hardcoded into DSP.

gedads:
can't add something like a mob hp boost?
anyway my main interest is still in that order base missions, cop zm, toau wotg.

TeoTwawki:
@R3P0FFXI the issue with chests, you mean placing items from box into loot pool yeah? I can make everything work except that so far. That and displaying more than 2 augments on an item inside a chest (the latter I just need some captures of the params I think).

gedads:
abyssea also have random sized mobs

gedads:
like 9 normal size one big sized

xdemolish:
discussions belong either on the forums or irc
especially when an email is being sent out for each reply

starlightknight:
@teschnei Conflicts are fixed, it was with elevators starting to be moved out of the sql

I also added a setting RESTRICT_BY_EXPANSION in attempt to make everyone happy as well (1 extra line of code to the shared function). People who don't want this because it interferes with their customizations can just toggle that off.

teschnei:
please keep your offtopic conversations out of github comments, I woke up to 20 emails this morning...

@gedads the main reason really is because it doesn't make sense to have these npcs spawned for servers that have the expansion turned off: as a player, it's a little strange to see abyssea maws when you're supposed to be playing on a 75 era server

@starlightknight i don't really see the need for the RESTRICT_BY_EXPANSION flag:  it'll work fine with custom content, especially since the default value of null makes it always spawn.  
anyways, i really don't think it should apply to elevators/transports - i feel that really doesn't make sense.  I think it should really only apply to visible and interactable NPCs

starlightknight:
@teschnei I did that to suppress warnings. Since those are currently loaded by id, if any of the associated npcs are filtered out by the setting then it will error because it can't find the reference to the npc. This should go away when we have them initialized completely by scripts, which seems to already be in progress by @xdemolish 

I can drop the setting, but it seemed like an easy way to mitigate some of the complaining for barely any effort. Some of the complaints were that they would have to maintain null versus what is in upstream on any npc they highjack for custom content that is not normally part of the expansion, creating merge conflicts (although these values shouldn't change once set, so I don't think repeated conflicts should ever happen unless there was a mistake)

starlightknight:
@teschnei I think this should be good now.

I did the following

1) Dropped transport/elevator changes and the tag on those transport npcs (no startup warnings now), even without the filter

2) Used std::string + append as you suggested

3) Cached the results with a std::unordered_map so that it doesn't hit the lua file every time

No leaks related to this are recorded in the DSP.leaks file now. With expansion restriction enabled and all expansions disabled, the memory usage goes down from an average of about 575MB to about 532MB for me, making it somewhere in the neighborhood of 40MB savings. So I guess you can save a chunk of memory this way for expansions you don't want as well as get a more accurate representation of your selections :-)

Please let me know if I need to do anything else. Apologies for being a C++ noob, its been over 10 years since I've really done anything with it other than messing with this project. I am primarily a Java developer :-P

PR #743, opened 2015-01-03
Simplify currency manipulation commands

TeoTwawki:
@teschnei I'm unable to test this due to not being able to build or to even run ffxi on this machine, am concerned as to if I handled the strings in luabaseentity correctly or not. I tried to get someone to test for me but they didn't build the core and only checked that my scripts flowed. 

Once I'm sure those 3 commands work as intended I can convert some of the existing lua's to use them and axe any redundant bindings you want gone (I can use a 2nd pull req or amend this, whichever you wish).


scripts/commands/addcurency.lua
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    if (targ == nil) then
+        player:PrintToPlayer( string.format( "Player named '%s' not found!", target ) );
+        return
+    end
+
+    -- Making sure currency type valid before adding.
+    if (targ:getCurrency(currency) ~= nil) then
+        targ:addCurrency(amount);
```
bope12: are you missing the type in the call?

TeoTwawki: yup copy paste error, thanks

PR #749, opened 2015-01-05
New conf setting to allow adjusting delay before engaging next mob.

TeoTwawki:
Someone had shared a setting to eliminate the engage delay before but it had a prob with being exploitable. I had some of my offline friends test this for me over the last 4 months with settings of 0.50,  0.75 and my own server has been using 0.85, everyone reports it functioning as I intended to reduce but not eliminate the delay without effecting delay between melee attacks after engaging.

Reason I think this has a shot at being in DarkStar: 
SE actually has changed this in the past. 

Reason it might not:
Engage delay used to be shorter in retail but was exploitable by a specially formed macro while using a high delay weapon. My implementation attempts to allow you to avoid reopening that exploit without enforcing the huge pause SE used, but this can be misused so I understand if you choose to close/reject this without merge.

### DIFF
--- a/conf/map_darkstar.conf
+++ b/conf/map_darkstar.conf
@@ -76,6 +76,10 @@ level_sync_enable: 0
 #Enable/disable jobs other than BST and RNG having widescan
 all_jobs_widescan: 1
 
+#Acts as multiplier for the delay between engaging targets. 0.80 would be a 20% reduction from retail. Does not change attack round delay.
+#WARNING: even seemingly small changes can potentially be exploitable (especially with high delay weapons).
+engage_delay_mod: 1.00
+
 #Modifier to apply to player speed. 0 means default speed of 40, where 20 would mean speed 60 or -10 would mean speed 30.
 speed_mod: 0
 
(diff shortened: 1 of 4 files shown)


teschnei:
how does this avoid that exploit?

TeoTwawki:
if you only lessen the extra delay slightly rather than undo it (Its longer than just the normal delay till next swing) its still long enough to prevent the macro attack, what SE did was overkill. But dropping the setting too far does make it possible again which is why I caution against large adjustments.

teschnei:
i think it might still be exploitable via packet injection to disengage/reengage instantly, would be worth testing if you know how

TeoTwawki:
I dunno how. I am interested in ways to improve this though. When changed the setting just shortens 1 of the 2 values used to determine how much time passed.

I think that injection method still works on retail itself. I remember some guys on alla and saying something about still being able to but not what the method involved was - thats probably it.

teschnei:
nah, as we can see here the delay is server side anyways (so injecting bypasses nothing)

at any rate, i still don't think a configuration value to change one line is useful to many (or a good use of a config option)

TeoTwawki:
Alrighty, np.

PR #759, opened 2015-01-06
Use proper port 54230 for several noname zones

tagban:
This is simple, was causing error 3001 if a character logged out in one of these zones. Note: only affected noname zones, and character would get error 3001 and hang on downloading screen . Don't forget to re-update ZoneIP after pull!


atom0s:
Players should not be accessing these zones to begin with since they are technically not zones they should be able to enter without the use of zone command or being forced there by someone else.

tagban:
The port not being set doesn't stop them from accessing it via @zone command. I agree with you, but its rather trivial not to use the right ports given it doesn't prevent accessing, only gets people stuck there if they accidently logout, or server shutsoff.

atom0s:
The intent of the port being 0 is so that the zone does not load and take up pointless resources on your server.

PR #763, opened 2015-01-07
CORE: fix seigan counter rate (zanshin proc /4), cleanup obsolete attack...

maxtherabbit:
...round vars


teschnei:
should check if seigan is giving the third eye counter chance as well

maxtherabbit:
looks like it's not, I'll implement it tomorrow

PR #769, opened 2015-01-07
Typo that somehow went unnoticed or unreported.

TeoTwawki:


starlightknight:
@TeoTwawki You need to find replace this value in traits/spells/abilities as well

TeoTwawki:
Man..I miss 1 little s months ago, and all this trouble results.

Pisses me off that I know of at least 8 servers that had to have known about bugs from that variable and said nothing.

PR #770, opened 2015-01-07
Fix Abyssea Tagged Spells/Traits/Abilities

starlightknight:
Find/Replace ABYSEA -> ABYSSEA per commit 1306105c155da354307841a68e132fbb308d3fc9


TeoTwawki:
Sorry for the trouble I had done a search of the script and src dirs for "ABYSEA" after seeing the reason a few maws were malfunctioning verifying which spelling was used on wiki and completely forgotten the sql or I'd have hit that at same time.

starlightknight:
@TeoTwawki No problem :-)

PR #777, opened 2015-01-09
Cavernous Maw fixes

TeoTwawki:
Fixed all Abyssea maws and scripted them up through "Dawn of Death". In zones where more than one Cavernous Maw exists in npc_list.sql the Abyssea entry maw will be "Cavernous_Maw_2" to separate it from WotG's "Cavernous_Maw". Fixed the expansion tagging that had some marked for wrong expansion.

Todo: move quest complete cs check into the zone.lua for each zone, because I'm pretty sure the maw isn't used to trigger that. I'll do it later after the NMs actually work and the quest can actually be done.

There are also some minor changes just for consistencies sake.

@gedads, noticed during cs that some of the maws pos are slightly off, could you check the retail positioning later after this is merged? Thanks. (Example: in Buburimu when the player char is supposed to be looking at the maw before being approached by an NPC, he's staring at the wall and the maw is off to the side).

### DIFF
--- a/scripts/zones/Abyssea-Altepa/npcs/Cavernous_Maw.lua
+++ b/scripts/zones/Abyssea-Altepa/npcs/Cavernous_Maw.lua
@@ -1,0 +1,44 @@
+-----------------------------------
+-- Area: Abyssea Altepa
+--  NPC: Cavernous Maw
+-- @pos 444.000 -0.500 320.000 218
+-- Notes Teleports Players to South Gustaberg
+-----------------------------------
+package.loaded["scripts/zones/Abyssea-Altepa/TextIDs"] = nil;
+-----------------------------------
+
+-----------------------------------
+-- onTrade Action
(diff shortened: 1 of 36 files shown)


teschnei:
you'll have to pull again to get gedads latest commit (they conflict with yours, probably because he changed the pos of a maw and you changed the name)

TeoTwawki:
.....

teschnei:
really it's only like 30 seconds of work: git pull, says you have conflicts, open the resolve window, double click npc_list, edit it in there (use gedads chunk probably and then rename it again manually), commit, done

TeoTwawki:
Doesn't matter. I'm not able to keep a workflow today. Constant interpositions pissing me off.

gedads:
sorry

TeoTwawki:
Fixed. Didn't rebase, gave you an evil merge.

PR #782, opened 2015-01-10
Update Derrick.lua

JoeCT:
Fixes a minor bug where when buying an airship pass directly from Derrick in Lower Jeuno the message 'Um, you don't have enough money.' is displayed irregardless of whether you do or not.


teschnei:
the if statements conditions shouldn't need to be changed, delGil should return a bool if the gil was deleted or not

JoeCT:
I changed the if statement condition because it causes a debug message to be printed when delGil is used that way.  "[Debug] UpdateItem: Trying to move too much quantity." Didn't really look into why delGil was throwing the message, just worked around it.

teschnei:
oh, yeah, that's ok - it's flagged under the debug header so it's totally ok for that message to be shown (anyone who isn't actively trying to debug something should have debug messages off anyways)

PR #783, opened 2015-01-11
Dynamis entry timer changes

bope12:
Makes entering dynamis only set your dynamis wait timer the first time you enter.
Changes how it handles players who log out inside dynamis and log in after their wait timer is up


xdemolish:
should local cs instead of leaving it as a global var

teschnei:
onZoneIn doesn't have a zoning param (only onGameIn has that)
also, the diff lists the shantotto commits you pull requested a few months ago so it's not automatically mergeable (using the same branch probably)

PR #788, opened 2015-01-12
Reworked onMobDeathEx

TeoTwawki:
Made onMobDeathEx run for all party and alliance members, added "isKillShot" parameter to enable restricting scripts to only the person who landed killing blow. "isWeaponSkillKill" seems to be broken before I arrived, I am not fixing it because I intended to obsolete that anyway as soon as possible.

### DIFF
--- a/scripts/globals/mobs.lua
+++ b/scripts/globals/mobs.lua
@@ -15,27 +15,28 @@ require("scripts/globals/status");
 --
 -----------------------------------
 
-function onMobDeathEx(mob, killer, isWeaponSkillKill)
-
+-- function onMobDeathEx(mob, killer, isKillShot, killType)
+function onMobDeathEx(mob, killer, isKillShot, isWeaponSkillKill)
 	-- DRK quest - Blade Of Darkness
 	local BladeofDarkness = killer:getQuestStatus(BASTOK, BLADE_OF_DARKNESS);
 	local BladeofDeath = killer:getQuestStatus(BASTOK, BLADE_OF_DEATH);
 	local ChaosbringerKills = killer:getVar("ChaosbringerKills");
 
 	if (BladeofDarkness == QUEST_ACCEPTED or BladeofDeath == QUEST_ACCEPTED) then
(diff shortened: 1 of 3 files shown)


TeoTwawki:
@teschnei 
Found a related issue with trunk that predates my change, and is likely why people have been complaining about drk quest not working for last 3-4 weeks. If you can tell me how to fix it, I'd like to add it into this pull request. The issue seems to be that scripts\globals\mobs.lua is not loaded at server start anymore. If I use @reloadglobal to load it, everything works. Again, thats not a bug created by this pull request. I added
```lua
killer:PrintToPlayer("test!")
```
to current trunk version fo mobs.lua and got nothing till I used the gm command to reload the file. After that every kill resulted in my test message printing. No console / log errors at all.

teschnei:
Oh? I fixed that already. Its not a global load, it's just loaded only for
onMobDeathEx like any other script (ie. using the prepscript stuff). What
are you seeing not work?

TeoTwawki:
Well people were complaining about drk not giving them the kill counts on their quest, killing hundreds of mobs for nothing. So I had tossed in a print that displayed the number. And right away saw nothing. So at the top of the function I tossed in the print I showed you above, and didn't see that either. Then I did 
```
@reloadglobal mobs
```
And poof it starts printing with every kill.
The entire function had not been executing at all.

teschnei:
I'm sure the  function isn't opening the file like it's supposed to then.
Probably something small in the section that loads the file (before
lua_getglobal).  I'll look if you can't figure it out by the time I'm home

TeoTwawki:
I think there was a getWhatever where it needed a setWhatever, I'll check in a bit.

TeoTwawki:
@teschnei all fixed now, will work on isWeaponSkillKill conversion to killType in a separate pull request later.

teschnei:
just one last thing: remove the "loaded" bool and replace the "loaded = false" lines with returns like the other luautils functions

TeoTwawki:
k, be on that in a sec :)

TeoTwawki:
@teschnei Got distracted for a few hrs, but the bool is nuked now. All good now?

Issue #791, opened 2015-01-13
Character appears unarmed at logon even though weapon actually is equiped.


TeoTwawki:
On latest trunk when you logon your character looks as if they have no weapon and will use the h2h attack animation and use h2h skill (I even got skillups) until you re-equip. 

@Detsero Caused by b377297375d445d68060824ae3de04f8a6fa50d7 ?
I just ask because its the only recent change to equip stuffs.

TeoTwawki:
Reverted that change on my local copy and prob went away, so looks like ya done goofed :P

PR #792, opened 2015-01-13
CORE: Correcting an equipment slot comparison

Detsero:
In response to issue #791 - correcting the SQL column which is used for comparison and making use of the SLOT_MAIN enum value.  An edge case bug still exists such that the appropriate unarmed weapon item will not be equipped when the user logs in with no equipment at all.  I will investigate this bug further.

Performed a few tests by logging in with a Monk unarmed with a body piece equipped and then attacking something, logging in as a Monk with nothing equipped and then attacking something, logging in as a Monk with a weapon equipped and attacking something, and logging in as a Dark Knight with a weapon equipped and attacking something.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -840,10 +840,11 @@ void LoadInventory(CCharEntity* PChar)
 		{
 			if (Sql_GetUIntData(SqlHandle, 1) < 16)
 			{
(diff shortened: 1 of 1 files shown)


TeoTwawki:
That was rather fast. Thanks!

Detsero:
No problem, I didn't want to leave something so high profile broken for long! I'll try to be a little more careful with future changes.

PR #797, opened 2015-01-14
Update Regain for Nightfall to match new TP values

korietsu:
Set from 10 to 100 in item_mod.sql


teschnei:
actually, it was always supposed to be in this format

PR #800, opened 2015-01-15
Added Functionality for Past Event Watcher NPCs Dalba(Port Bastok), Lamepaue(Bastok Markets), and Durogg(Northern San d'Oria). Adds Quest: Bastok - Wish Upon a Star

JoeCT:
Adds functionality to past event watchers Dalba, Lamepaue and Durogg. Assuming that the cut scenes for multi-cut scene quests/missions all become available at the completion of the quest/mission then these scripts mimic retail behavior. A few of the cut scene event ids are unknown to me and also a few of the quests/missions are not yet defined in either quests.lua or missions.lua; those particular quests/missions have been commented out and clearly documented.

Adds quest: Bastok - Wish Upon a Star. The quest requires a rare/ex item, "Fallen Star," which can only be obtained by logging in the Yuhtunga Jungle or the Yhoator Jungle. The drop rate is unknown; given that this is a rare item I set the drop rate to 5%


scripts/globals/logging.lua
```diff
@@ -39,8 +39,8 @@ local drop = {2,{0x1198,0.0900,0x02B0,0.1800,0x02BA,0.2700,0x039B,0.3600,0x02B7,
 		101,{0x02B6,0.0500,0x02B8,0.1130,0x023E,0.1760,0x02B6,0.2640,0x02B3,0.4760,0x02B0,0.7380,0x02BA,1.0000},
 		104,{0x02BB,0.0120,0x1198,0.0600,0x039B,0.1320,0x02BA,0.2590,0x02B0,0.4220,0x02B7,0.5970,0x02B8,0.7900,0x02B5,1.0000},
 		118,{0x02BC,0.0170,0x02BD,0.0430,0x02BE,0.0770,0x023E,0.1290,0x039B,0.2070,0x02B9,0.3020,0x1197,0.4050,0x115D,0.5600,0x02B0,0.7580,0x02B1,1.0000},
-		123,{0x02B0,0.2350,0x02D1,0.4650,0x02B1,0.7000,0x02B9,0.7700,0x03AC,0.8400,0x0390,0.9100,0x02BD,0.9400,0x02BE,0.9700,0x04D5,1.0000},
-		124,{0x02BE,0.0120,0x02BC,0.0520,0x03AC,0.0920,0x039B,0.1620,0x0390,0.3070,0x02B1,0.4780,0x02B0,0.7290,0x02D1,0.9800,0x04D5,1.0000},
+		123,{0x04A8,0.0500,0x02B0,0.2350,0x02D1,0.4650,0x02B1,0.7000,0x02B9,0.7700,0x03AC,0.8400,0x0390,0.9100,0x02BD,0.9400,0x02BE,0.9700,0x04D5,1.0000},
+		124,{0x04A8,0.0500,0x02BE,0.0120,0x02BC,0.0520,0x03AC,0.0920,0x039B,0.1620,0x0390,0.3070,0x02B1,0.4780,0x02B0,0.7290,0x02D1,0.9800,0x04D5,1.0000},
```
teschnei: i do believe these must be in ascending order (or else it's impossible to log a 0x02BE)

teschnei:
every variable you use should be declared as local unless you specifically need it to be global (you probably shouldn't be using global variables if that's the case, either)

another thing: when contributing code, you should always take a look at the style of code there already and conform to it as best you can - in this case, it means putting the space before the brackets instead of after.  keeping code style consistent makes it easier to read, and nobody wants to go and change other peoples code to be consistent

JoeCT:
In response to your comments logging.lua has been corrected; the values are now in ascending order.  
The code style has been changed to reflect the generally used style in the project.  And, the majority of the variables declared have been defined as locals.  One exception is the playCutscene variable defined in each of the three past event watcher scripts that I've modified.  The variable is defined in the 
onEventUpdate function and is evaluated in the onEventFinish function.  Is it more preferable to set a player variable in the database then to use a global? I'm curious, what is the scope of global variables in Lua? Are they available to any script being ran on the server or are they only available to the particular script in which they are defined?  I've been trying to stay away from setting player variables in the database as I want to keep the server overhead down, but if the global variables could cause an issue I'd rather not use them.

teschnei:
they are available to every script.  the main reason that the global wouldn't work in this case is that it's not per character - if two people start the cutscene, the first guy to finish will get the second guys value. 

this is a perfect example to utilize something new that I added a month ago, called localVars (as opposed to playerVars) - they are not saved to the database, are destroyed on zone/logout, but do not have the overhead of db accessing (they are just stored in a std::map server side).  you can use player:getLocalVar(name) and player:setLocalVar(name, value) to get/set them.  they're available on any entity in case you ever need them for anything else in the future, but for characters the primary use case would be for events and stuff that does reset on logout/zone (another example being getting multiple events in a row for different quests from one NPC)

JoeCT:
Thanks, for the info.  The three event watcher scripts have been updated to use localVars as you suggested. - No more globals.

scripts/globals/logging.lua
```diff
@@ -39,8 +40,8 @@ local drop = {2,{0x1198,0.0900,0x02B0,0.1800,0x02BA,0.2700,0x039B,0.3600,0x02B7,
 		101,{0x02B6,0.0500,0x02B8,0.1130,0x023E,0.1760,0x02B6,0.2640,0x02B3,0.4760,0x02B0,0.7380,0x02BA,1.0000},
 		104,{0x02BB,0.0120,0x1198,0.0600,0x039B,0.1320,0x02BA,0.2590,0x02B0,0.4220,0x02B7,0.5970,0x02B8,0.7900,0x02B5,1.0000},
 		118,{0x02BC,0.0170,0x02BD,0.0430,0x02BE,0.0770,0x023E,0.1290,0x039B,0.2070,0x02B9,0.3020,0x1197,0.4050,0x115D,0.5600,0x02B0,0.7580,0x02B1,1.0000},
-		123,{0x02B0,0.2350,0x02D1,0.4650,0x02B1,0.7000,0x02B9,0.7700,0x03AC,0.8400,0x0390,0.9100,0x02BD,0.9400,0x02BE,0.9700,0x04D5,1.0000},
-		124,{0x02BE,0.0120,0x02BC,0.0520,0x03AC,0.0920,0x039B,0.1620,0x0390,0.3070,0x02B1,0.4780,0x02B0,0.7290,0x02D1,0.9800,0x04D5,1.0000},
+		123,{0x02B0,0.2350,0x02D1,0.4650,0x02B1,0.7000,0x02B9,0.7700,0x03AC,0.8400,0x0390,0.9100,0x02BD,0.9400,0x02BE,0.9700,0x04A8,0.0500,0x04D5,1.0000},
+		124,{0x02BE,0.0120,0x02BC,0.0520,0x03AC,0.0920,0x039B,0.1620,0x0390,0.3070,0x02B1,0.4780,0x02B0,0.7290,0x02D1,0.9800,0x04A8,0.0500,0x04D5,1.0000},
```
teschnei: sorry if i was unclear on this, I meant ascending order of rarity (0.0120 < 0.0500 < 0.0520, etc)
the script generates one number and then steps through them one by one until it hits one where the randomly generated number is less than the number it stepped to, then immediately returns - so they have to be in order of rarity or they may not ever get picked.

teschnei:
rest looks good now

JoeCT:
Logging.lua corrected. data is now in ascending order by rarity.

Issue #799, opened 2015-01-14
charmed pet crash


TeoTwawki:
https://yadi.sk/d/HoenCaIYdyhge

"isWeaponSkillKill" hasn't even been working but this crash lands there when a players charmed pet dies...

Someone told me they think a similar or same crash happens on killing a bst mob while its pet is still alive. I have not confirmed that.

TeoTwawki:
Fixed by  #801

PR #801, opened 2015-01-15
Fix charmed pet crash #799 

TeoTwawki:
So, apparently Pkiller was not already certain to be a player entity. I thought it was. Later I'll probably need to do some work to make pet kills exec onMobDeathEx on their master for those danged magian trials. :/

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -2331,7 +2331,10 @@ int32 OnMobDeath(CBaseEntity* PMob, CBaseEntity* PKiller)
 
     CLuaBaseEntity LuaMobEntity(PMob);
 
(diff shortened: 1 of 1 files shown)


src/map/lua/luautils.cpp
```diff
@@ -2331,7 +2331,10 @@ int32 OnMobDeath(CBaseEntity* PMob, CBaseEntity* PKiller)
 
     CLuaBaseEntity LuaMobEntity(PMob);
 
-    luautils::OnMobDeathEx(PMob, PKiller, true);
+    if (PChar)
```
teschnei: this doesn't work the way you think it does - casting it won't let you know if the object type is actually that type, it only forces the program to interpret that memory location as if it were a character - it'll still crash unless you do PKiller->objtype == TYPE_PC

PR #803, opened 2015-01-15
January Update: NPC ID Updates

starlightknight:
Using the new version of the npc job. Several zones had large shifts.

There are probably mob shifts in these zones too. I can try to gather
some data later w/ my experimental mobid job but its not 100% complete yet


starlightknight:
@gedads A couple of zones had large shifts where you may want to recap. See diff for details. There are probably mob changes in these zones too. 

I can scrape some data later with my work-in-progress job for mobs, but they may need to be recapped, esp to get new mobs. I would be able existing mobs id shifting, and if some got deleted - but if some got deleted within a group (like theres 3 Mob X, and 1 got deleted) I may not be able to tell for sure which one was the deleted one, as POLUtils only gives name, not pos.

gedads:
Is there a log or something that directly points me to those areas?

starlightknight:
@gedads 

These are the zones w/ NPC ID shift

Cape Teriggan
Tahrongi Canyon
The Sanctuary of Zi'Tah
Ro'Maeve (shift of 21)
Yuhtunga Jungle (shift of 9)
Western Altepa Desert
Eastern Adoulin
Rala Waterways
Yorcia Weald
Mog Garden

Some of these had mobs in a large chunk of space where npcs used to be, but i did not detect any mob shifts when rescraping mob names - maybe that means there are a handful of new mobs? Ro'Maeve in particular had a huge shift.

Issue #802, opened 2015-01-15
Summoner Unlock Quest: I Can Hear a Rainbow


korietsu:
Looks like folks aren't getting the CS triggers while on the SMN flag quest. We've tried about every zone and forcing weather and zoning players and still not registering.

JoeCT:
I actually did some work on this quest about 2 years ago. I'll take a look and see what I can figure out.  I know it worked back then...

JoeCT:
When entering a zone which is related to the quest, a function is called which determines whether or not the player has Carbuncle's Ruby and whether the player has accepted the I Can Hear a Rainbow quest.  The problem is that hasItem(), when called in function onZoneIn(), is returning false irregardless of whether the item is in inventory or not.

This is repeatable, add item 1125 to inventory and paste the following into the onZoneIn() function for any zone.

if( player:hasItem( 1125, 0)) then   -- Player has Carbuncle's Ruby?
printf("True");
else
printf("False");
end

Anyone have any ideas why this function is no longer working properly?

JoeCT:
I took a look at charutils.cpp to see if I could determine a cause for this issue and in the LoadChar() function near the end you can see that the line "PChar->m_event.EventID = luautils::OnZoneIn(PChar);" is before the line "charutils::LoadInventory(PChar);" Placing the latter line before the former and recompiling solves this issue.

teschnei:
unfortunately, it breaks some other stuff which is why I moved it there

JoeCT:
I've been looking into using AfterZoneIn() to trigger this quest.  It's used in six zones to to send visual packets.  I placed a printf() inside of one to see when it triggered, but never got any output.  Anyone know what I'm doing wrong?

korietsu:
Is it possible to use onGameHour to call the function check?

JoeCT:
I have submitted a pull request (#811) to resolve this issue.

teschnei:
inventory should be available in onZoneIn now

PR #811, opened 2015-01-17
Fix to Summoner Flag Quest: I Can Hear a Rainbow. (Issue #802)

JoeCT:
Moved startEvent() functions to afterZoneIn() so that a test for an item in inventory could run properly.


teschnei:
i don't want to use afterZoneIn for this: it's not how retail does it.  I'll fix zone.lua to run after inventory is loaded but before gear equips are calculated in order to keep this working and allow level caps to be applied on zone without disconnecting the client

PR #817, opened 2015-01-18
Move roamflag dependent of mob family to its own column in database

santssoft:
SQL: Added column roamflag to the table mob_family_system.

CORE: Removed references to roamflag dependent of mob family. Load roamflag from database.


teschnei:
better for pools than mob_family (BCNMs, NMs, etc)

atom0s:
After a discussion with kjLotus we feel this should be moved to mob pools instead of mob family. Since not all groups of mobs in a total family roam the same in every zone etc. So it would be better suited added to mob_pool sql etc. Otherwise the rest of the pull is fine if you dont mind moving it to mob pools.

santssoft:
OK, I have no problem in moving to mob_pool. I'll create a new pull request when ready.

PR #819, opened 2015-01-19
Replaced obsolete script commands for currency manipulation with the new ones.

TeoTwawki:


TeoTwawki:
@teschnei Sorry It didn't hit me that in the [at]command version of that, I don't have to use quotes but in npc scripts I do. Its fixed now, and I logged in and checked some of the npc's to be sure I didn't derp anything else. I will make sure to check actual npc's first next time to make sure I used things correctly and not just test the function I make 'em use.

PR #830, opened 2015-01-22
Home Point Teleports - Corrections and Additions

JoeCT:
Added remaining homepoints, corrected a few bugs.  Previously, the bitmasks were not being stored in the database properly and free teleports were still charging a fee.  Very similar to retail except for the retail map feature.


scripts/globals/homepoint.lua
```diff
@@ -1,361 +1,248 @@
...
+		local HpTeleportMask1 = bit.bor( bit.lshift( player:getVar("HpTeleportMask1a"), 16), player:getVar("HpTeleportMask1b"));
+		local HpTeleportMask2 = bit.bor( bit.lshift( player:getVar("HpTeleportMask2a"), 16), player:getVar("HpTeleportMask2b"));
+		local HpTeleportMask3 = bit.bor( bit.lshift( player:getVar("HpTeleportMask3a"), 16), player:getVar("HpTeleportMask3b"));
+		local HpTeleportMask4 = bit.bor( bit.lshift( player:getVar("HpTeleportMask4a"), 16), player:getVar("HpTeleportMask4b"));
+
+		-- Register new homepoint?
+		local newHp = 0;
+		if( homepoints[hpIndex][1] == 1) then
+			if( bit.rshift( bit.lshift( HpTeleportMask1, 32 - homepoints[hpIndex][2]), 31) == 0) then
+				newHp = 65536; -- This value causes the "You have registered a new home point!" dialog to display
```
teschnei: 0x10000

obviously it's not a huge deal, but it's always a good idea to put the numbers in the format that is most readable - the numbers are never random!

scripts/globals/homepoint.lua
```diff
@@ -1,361 +1,248 @@
...
-		local hps = player:getVar("hpmask3")
-		if (bit.band(hps, bit.lshift(1, (num))) == 0)then
-			player:setVar("hpmask3",bit.bor(hps, bit.lshift(1, (num))))
-			return 65536
-		end
+end;
+
+function hpTeleport( player, option)
+
+	if( option == 2 or option > 65536 and option < 8323075) then
```
teschnei: 0x7F0003

scripts/globals/homepoint.lua
```diff
@@ -1,361 +1,248 @@
...
+		local hpIndex = (option - 2) / 65536; -- Calculate hpIndex based on option selected
```
teschnei: use bit operations instead of dividing (rshift I think is what you're looking for here)

JoeCT:
I looked into the bit.rshift like you suggested, took about 5 minutes then realized what you meant, never thought to do it that way before.  I like that implementation better myself.  Also, I did convert those decimal numbers to hex; I prefer the hex numbers, especially with larger values, but used the decimal numbers to try and stay with the general style of the scripts already in the project.  Seems that most of the scripts when evaluating 'option' stick with decimal, I'll happily ignore that from now on.

teschnei:
yeah.. some of them are quite old - most even before bit operators were in our lua library (which was pretty recent, truth be told)

PR #847, opened 2015-01-25
NPCs, as retail as I can currently get them using wiki as a guide.

TeoTwawki:
Helped the Legion server with their Legion NPCs they are customizing, here are the retail bits of that.

### DIFF
--- a/scripts/zones/Rolanberry_Fields/npcs/Mayuyu.lua
+++ b/scripts/zones/Rolanberry_Fields/npcs/Mayuyu.lua
@@ -1,0 +1,66 @@
+-----------------------------------
+-- Area: Rolanberry Fields
+--  NPC: Mayuyu
+-- Legion NPC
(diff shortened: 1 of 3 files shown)


scripts/zones/Rolanberry_Fields/npcs/Mayuyu.lua
```diff
@@ -0,0 +1,66 @@
...
+
+-----------------------------------
+-- onEventFinish
+-----------------------------------
+
+function onEventFinish(player,csid,option)
+    -- print("CSID:", csid);
+    -- print("RESULT:", option);
+
+    -- Event needs work, also the Legion Pass item is "tagged" via fields not yet implemented in core.
```
teschnei: it's all part of m_extra of CItem, you should be able to find everything you need for that at https://github.com/Windower/Lua/blob/4.1-dev/addons/libs/extdata.lua

PR #848, opened 2015-01-25
Fixed 3 busted GM commands

TeoTwawki:
1. getzone changed, changed again, then we got getzoneID so using that now.
2. spawnmob no longer works for NPC objects, it used to.

Added rot in there while I was at it, semi useful if moving NPCs around. Tested these, however sometimes you have to zone to see the NPC change even though I used status update. Might be just me though, I had a lot open during testing so my PC has that laggy feel.


TeoTwawki:
1 sec made a minor flub...

PR #856, opened 2015-01-26
Fix ZM14 and IDs

console08:
Updated new shimmering circle IDs in La'Loff Amphitheater and removed comment blocks that were (inadvertently?) added a few months ago which broke Ark Angels (ZM14)


teschnei:
that comment block looks exactly like how I test ark angels!  wait a sec..

PR #863, opened 2015-01-28
Added Functionality to NPCs Bhagirath and Sahn (On Ships to Selbina/Maura). Added Functionality to Airship Time NPCs and Randomized Player Starting Point on Airships.

JoeCT:
Added functionality to NPCs Bhagirath and Sahn.  These NPCs are the captains of the ships bound for either Mhaura or Selbina.  They now accurately relay approximate times to destination.

### DIFF
--- a/scripts/zones/Ship_bound_for_Selbina/npcs/Bhagirath.lua
+++ b/scripts/zones/Ship_bound_for_Selbina/npcs/Bhagirath.lua
@@ -51,13 +51,23 @@ function onTrigger(player,npc)
 		vHour = 0;
 	end
 
-	local minutes = math.floor((2.4 * ((vHour * 60) + 40 - vMin)) / 60);
+	local vMinutes = (vHour * 60) + 40 - vMin;
+
+	vHour = math.floor( vMinutes / 60 + 0.5);
+
+	local message = ON_WAY_TO_SELBINA;
+
+	if (vMinutes <= 30) then
+		message = ARRIVING_SOON_SELBINA;
+	elseif (vMinutes < 60) then
+		vHour = 0;
+	end
 
 	if (vHour > 7) then -- Normal players can't be on the boat longer than 7 Vanadiel hours. This is for GMs.
 		vHour = 7;
 	end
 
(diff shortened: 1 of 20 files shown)


teschnei:
space before brackets, not after! (yes, i'm anal about this :p)

scripts/zones/Ship_bound_for_Selbina/npcs/Bhagirath.lua
```diff
@@ -21,7 +21,33 @@ end;
...
-	player:messageSpecial(ON_WAY_TO_SELBINA,0,0); -- Earth Time, Vana Hours. Needs a get-time function for boat?
+
+	local vHour = VanadielHour();
+	local vMin  = VanadielMinute();
+
+	while vHour >= 6 do
+		vHour = vHour - 8;
+	end
+
+	if(     vHour == -2) then vHour = 8;
```
TeoTwawki: Are you trying to vertically align your var names or "then"s or something?
```lua
if (vHour == -2) then
    vHour = 8;
elseif (vHour == -1) then
    vHour = 7;
elseif (vHour == 0) then
    vHour = 6;
elseif (vHour == 1) then
    vHour = 5;
elseif (vHour == 2) then
    vHour = 4;
elseif (vHour == 3) then
    vHour = 3;
elseif (vHour == 4) then
    vHour = 2;
elseif (vHour == 5) then
    vHour = 1;
end
```
This is the style usually used, though not strictly enforced (I wish it were).

JoeCT:
Updated both scripts.  Should be more to your liking.  - A while back I used to work as a programmer and we were evaluated based on lines of code written.  They had very strict guidelines concerning style  which is where I picked up the majority of mine.  Anyway, we were required to condense simple conditionals like the ones in these scripts down to as few lines as possible.  If not, the extra lines were actually counted against us.  Personally, I don't mind which style I use.  Just gotta break my old habits.

JoeCT:
Added functionality to NPCs Dereck and Michele. These NPCs are on the Bastok-Jeuno Airship. They now accurately relay approximate times to destination.

PR #865, opened 2015-01-29
Updated weapon skill damage calculations and added missing effects. Also...

Fr0zt:
... added an lua for wildfire and made it so that RNG could actually use it. All changes were taken from both wiki's. I have cross referenced all my changes and tested them in-game. All damage should reflect retail (except for magic weapon skills still being counted as physical damage). Some information on ffxiclopedia is still inaccurate for a couple of weapon skills, so if it looks incorrect while referring to that, check bg-wiki's page for that specific weapon skill.

Some posts that back up my changes:
June 2014 changes:
http://www.bluegartr.com/threads/121610-Rehauled-Weapon-Skills-tier-lists?p=6142188&viewfull=1

December 2014 changes:
http://www.bluegartr.com/threads/124376-December-10th-Update-Weaponskill-Adjustments-testing


scripts/globals/weaponskills/spirits_within.lua
```diff
@@ -19,21 +18,20 @@ require("scripts/globals/utils");
 -----------------------------------
 
 function onUseWeaponSkill(player, target, wsID)
-	local D = player:getWeaponDmg();
```
maxtherabbit: this line is still needed...

scripts/globals/weaponskills/spirits_within.lua
```diff
@@ -19,21 +18,20 @@ require("scripts/globals/utils");
 -----------------------------------
 
 function onUseWeaponSkill(player, target, wsID)
```
maxtherabbit: this one is not needed

scripts/globals/weaponskills/spirits_within.lua
```diff
@@ -19,21 +18,20 @@ require("scripts/globals/utils");
 -----------------------------------
 
 function onUseWeaponSkill(player, target, wsID)
-	local D = player:getWeaponDmg();	
 	local HP = player:getHP();
```
maxtherabbit: HP is still needed, D is not

maxtherabbit:
since tornado kick and dragon kick were the only weapon skills using params.kick and you debunked that, you can remove it entirely from weaponskills.lua line 61-3

teschnei:
since you're already here, why don't you make the merit WS wsc scale on merit instead of hardcoding it to the maximum value?

Fr0zt:
I added merit scaling to wsc and fixed the merit.sql as requested.

teschnei:
note that getMerit will return 3 (not 0.03), so you have to do wsc = (70 + getMerit(MERIT_WHATEVER)) / 10

Fr0zt:
WSC calculations should be correct now.

teschnei:
i'll just point out that you don't need one commit per file, if they all do the same thing it should just be one commit (group logically, not by file) - i'll go through the rest now

Fr0zt:
Ok, sorry about the spam commits. This is my first time really using github so I'm still learning all the ins and outs. I will keep that in mind for next time.

teschnei:
actually, I will hold off on this for now - I need to decide how I want to handle the big changes to certain WS (looking at you, one handed "damage varies with TP"), since they will have huge balance changes pre-adoulin.  a config option is probably best here, but it will be a pain to get that set up..

Cloudef:
@Fr0zt I suggest googling git rebase command, you could use git rebase -i and squash all those similar commits together, and then force push on your branch. Github will update this pull request okay.

teschnei:
ok, this is how i'd like to see it: add a new setting in settings.lua for adoulin weaponskill mods (call it whatever you want) - then, in every script that has had big changes to either wsc or damage/tp scaling, include settings.lua, and have the old stats as the default, and then overwrite them with the new stats in an if block checking that setting being true (or 1, whichever is the standard in settings.lua)

Fr0zt:
Ok, I will begin working on changes you have specified and commit it once I am finished.

starlightknight:
@Fr0zt you can combine your commits into one by doing the following

git rebase -i master

<Window pops up, listing commits in your default text editor>

Replace all but the first instance of "pick" with "s" or "squash". Save and close. Squash combines that commit into the previous one. You will be prompted for a final commit message afterward. At that point, if you force push your branch to github, it will automatically update this pull request to only have that one commit:

git push -u -f origin yourbranchname

@teschnei It seems like there has been tons of balance changes for SOA on retail. As these start to get implemented, do you think it makes sense to just tie them to ENABLE_SOA so that we don't end up with a billion settings around these? 

Some of the other stuff like the traits and such already get tied to that with the expansion restriction system, and I'm not sure why you'd want to use the setting without SOA as it'd just break the balance of your game completely as you noted (and there are no valid retail scenarios for doing so).

teschnei:
ah, nah, it's not up to me to decide for someone else whether they want to use OP weaponskills on a 75 era server

TeoTwawki:
Can just do like we did for cure math, I would think.

Fr0zt:
Finished making the requested changes for toggling new and old weapon skill calculations.

teschnei:
one final change: merit WS old wsc should be 85 + merit level * 3 (so it scales up to 100), and see if you can rebase to squash these commits

Fr0zt:
Went from 45 to 2 commits, also completed the requested changes.

PR #877, opened 2015-02-03
Converted item ID checks for occasionally does increased damage into a modifier check

TeoTwawki:
Added item_mods.sql entries for relic/mythic already in table that have this property. 

Acceptable change? The other way to do this would have been 2x, 2.5x, and 3x as 3 separate mods.

Might have done it that way in the first place had I seen the lv 95+ Mythics before I had 99% finished this, although I think this is more flexible. Let me know if I should change it. Did not yet add missing post lv75 relics. Will get to those soon unless someone beats me to it, just wanted to make sure this logic was ok first.

I probably didn't need the extra decimal place for the % proc rate but this way if it turns out something out there that is something point something percent, bases are covered already.

### DIFF
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -1177,8 +1177,9 @@ MOD_FERAL_HOWL_DURATION       =0x1F7 -- +20% duration per merit when wearing aug
 MOD_MANEUVER_BONUS            =0x1F8 -- Maneuver Stat Bonus
 MOD_OVERLOAD_THRESH           =0x1F9 -- Overload Threshold Bonus
 
--- MOD_SPARE =0x1FA -- (modId = 506)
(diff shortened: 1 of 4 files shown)


teschnei:
gastraphetes does not occ. deal bonus damage (only in aftermath 3)

TeoTwawki:
So previous code in the switch I removed should not have had gastraphetes in there then. Fixing.

teschnei:
yep (its the same as death penalty too.. dunno why there was one but not the other)

teschnei:
i wonder how you will get those to work (aftermath, both ranged mythics and all empyreans :D)

TeoTwawki:
Looking into those big if/else piles in the aftermath scripts will be "fun" /airquotes

PR #878, opened 2015-02-03
Made SAM meditate a bit more retail (now stacks with regain).

TeoTwawki:


teschnei:
don't go under 768 with custom effects: that's the absolute minimum (because the effect icons in the packet can go up to 768)

TeoTwawki:
ok, editting both ID and the comment I left then.

PR #882, opened 2015-02-04
Added Functionality to the 8 Working Strange Apparatuses.

JoeCT:
Added functionality to the 8 working strange apparatuses.  I have yet to research and implement the 3 non-working strange apparatuses, furthermore the security systems do spawn elementals, but I have yet to determine if they are the correct NPCs (I'm sure three are correct).  These scripts required packet 0x60 to be implemented in the map server, my implementation is not totally correct, but does serve to emulate the server fairly closely, closely enough in my opinion to work until a better implementation is written.  I hope to verify the remaining NPC IDs for the elementals within the week.  Please let me know if you have suggestions for improvement.

### DIFF
--- a/scripts/globals/settings.lua
+++ b/scripts/globals/settings.lua
@@ -83,7 +83,6 @@ HARVESTING_RATE         = 0.50; --% chance to recieve an item from haresting.  S
 EXCAVATION_RATE         = 0.50; --% chance to recieve an item from excavation.  Set between 0 and 1.
 LOGGING_RATE            = 0.50; --% chance to recieve an item from logging.  Set between 0 and 1.
 MINING_RATE             = 0.50; --% chance to recieve an item from mining.  Set between 0 and 1.
-STRANGE_APPARATUS_RATE  = 0.50; --% chance to recieve an item other than a crystal cluster from a strange apparatus.  Set between 0 and 1.
 
 -- SE implemented coffer/chest illusion time in order to prevent coffer farming. No-one in the same area can open a chest or coffer for loot (gil, gems & items)
 -- till a random time between MIN_ILLSION_TIME and MAX_ILLUSION_TIME. During this time players can loot keyitem and item related to quests (AF, maps... etc.)
--- a/scripts/globals/settings.lua
+++ b/scripts/globals/settings.lua
@@ -83,6 +83,7 @@ HARVESTING_RATE         = 0.50; --% chance to recieve an item from haresting.  S
 EXCAVATION_RATE         = 0.50; --% chance to recieve an item from excavation.  Set between 0 and 1.
 LOGGING_RATE            = 0.50; --% chance to recieve an item from logging.  Set between 0 and 1.
 MINING_RATE             = 0.50; --% chance to recieve an item from mining.  Set between 0 and 1.
+STRANGE_APPARATUS_RATE  = 0.50; --% chance to recieve an item other than a crystal cluster from a strange apparatus.  Set between 0 and 1.
 
 -- SE implemented coffer/chest illusion time in order to prevent coffer farming. No-one in the same area can open a chest or coffer for loot (gil, gems & items)
 -- till a random time between MIN_ILLSION_TIME and MAX_ILLUSION_TIME. During this time players can loot keyitem and item related to quests (AF, maps... etc.)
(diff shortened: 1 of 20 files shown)


scripts/globals/settings.lua
```diff
@@ -83,6 +83,7 @@ HARVESTING_RATE         = 0.50; --% chance to recieve an item from haresting.  S
 EXCAVATION_RATE         = 0.50; --% chance to recieve an item from excavation.  Set between 0 and 1.
 LOGGING_RATE            = 0.50; --% chance to recieve an item from logging.  Set between 0 and 1.
 MINING_RATE             = 0.50; --% chance to recieve an item from mining.  Set between 0 and 1.
+STRANGE_APPARATUS_RATE  = 0.50; --% chance to recieve an item other than a crystal cluster from a strange apparatus.  Set between 0 and 1.
```
TeoTwawki: With doctor status on retail I have never got a cluster at all, ever. Pretty sure it outright removes them from possible results. Since most people are going to calc their password and pretty much always have that status, the setting to adjust the cluster rate will not often figure in and be mostly unused.

As for other items [ffxiclopedia](http://wiki.ffxiclopedia.org/wiki/Strange_Apparatus) likely a pretty accurate list of what you can get if you haven't already referenced that.

teschnei:
hm.. should have asked me about that packet beforehand.  i can set it up to pass the actual string to the script

JoeCT:
If you'd be willing to return the string I'd be happy to modify these scripts to work with your implementation.

Fiocitrine:
aparati btw ^^

JoeCT:
Actually apparati would be wrong. The correct plural form of the word apparatus is apparatus or apparatuses. Look here: http://www.oxforddictionaries.com/us/definition/american_english/apparatus and here: http://www.merriam-webster.com/dictionary/apparatus . 

An explanation of why apparati is not correct is available here: http://english.stackexchange.com/questions/147989/what-is-the-proper-plural-form-of-apparatus

To be honest the only reason why I know this is that before I titled this pull request, being unsure myself and thinking apparati to not sound correct, I googled it to be sure.

teschnei:
now that i've finally got internet back, try using that (just use onEventUpdate like normal, and the third param is a string instead of a number)

JoeCT:
I really appreciate your help Tyler; using the tool you gave me earlier I was able to troubleshoot why I was having problems with the events completing on their own. My earlier implementation of packet 0x60 actually skipped the event after password entry and then I used calls to messageSpecial() to emulate the rest of the event as close to retail as possible.  Using your new implementation I was still having the same problem (the password is handled a lot better though). After some testing on the retail server today and viewing the packets sent I noticed that the retail server is sending a different release type than what is currently defined in the project.  I added RELEASE_UNKNOWN to the release types defined in release.h and modified your implementation to send the new type after sending a RELEASE_EVENT.  This mimics retail behavior exactly and allows the events to progress as they should.  I just wanted to verify that these changes are acceptable for you and to also ask what the new release type should be called..

In release.h:

enum RELEASE_TYPE
{
	RELEASE_STANDARD	= 0,
	RELEASE_EVENT		= 1,
	RELEASE_SKIPPING	= 2,
	RELEASE_UNKNOWN    = 3,
	RELEASE_FISHING	= 4
};

in packet_system.cpp:

void SmallPacket0x060(map_session_data_t* session, CCharEntity* PChar, int8* data)
{
    uint32 charid = RBUFL(data, 0x04);
    int8* string = data + 12;
    luautils::OnEventUpdate(PChar, string);

	PChar->pushPacket(new CReleasePacket(PChar, RELEASE_EVENT));
	PChar->pushPacket(new CReleasePacket(PChar, RELEASE_UNKNOWN));

	return;
}

teschnei:
that's fine, you can leave is as RELEASE_UNKNOWN and put a comment saying it's somehow related to the text entry events

JoeCT:
Tyler, All scripts have now been updated to use your implementation of packet 0x060.

TeoTwawki:
@JoeCT You have merge conflicts which means need to either rebase or pull the new changes and fix any edit conflicts before re-pushing (the devs have been pretty tolerant of "evil merges" from what I have seen).

JoeCT:
Ok, I'll take care of that later this evening. I'm still new to this, thanks for pointing that out.

JoeCT:
Replaced by PR #917.

PR #883, opened 2015-02-04
Made magic weapon skills calculate as magic damage.

Fr0zt:
Made magic weapon skills calculate as magic damage instead of physical.


whasf:
Can't auto-commit. Please rebase and resubmit!

PR #884, opened 2015-02-04
ACP mobskills

TeoTwawki:
Added the missing mobskills for ACP mobs. Still need to do the Seed Goblin's Teleport and sneak attack buff, and confrontation status for the Seed Mandragora fight.

I accidentally included a separate thing in same pull request, but its both small and correct - I fixed 2 typos, in the spell script for curse (lua uses EFFECT_CURSE_I not EFFECT_CURSE) and in spell_list.sql (animation ID on Meteor was mistakenly Kaustra).

https://github.com/DarkstarProject/darkstar/pull/884/files?w=1         <- easier read


TeoTwawki:
P.S.
The BCNM itself is still inaccessible because its commented out in the global, this is intentional.

PR #887, opened 2015-02-05
Made magic weapon skills calculate as magic damage.

Fr0zt:
Made magic weapon skills calculate as magic damage instead of physical.


Fr0zt:
Not exactly sure how my other pull request got canceled... but here it is again.

teschnei:
i don't think doRanged has support for numhits yet (and i know for sure neither have the capability of duplicating ftp across all hits), but that's something that can be done at a later point

PR #894, opened 2015-02-07
fixed error causing bcnm to not complete

sisternERA:
the statement conflicted and causes the bcnm to sit idle after the mobs were killed

### DIFF
--- a/scripts/zones/The_Garden_of_RuHmet/bcnms/when_angels_fall.lua
+++ b/scripts/zones/The_Garden_of_RuHmet/bcnms/when_angels_fall.lua
@@ -62,8 +62,8 @@ end;
 	
 function onEventFinish(player,csid,option)
 -- print("bc finish csid "..csid.." and option "..option);
-  if(csid== 0x7d01)then
-    player:setPos(420,0,445,192);
-end
+    if(csid== 0x7d01)then
+        player:setPos(420,0,445,192);
+    end
  
 end;


scripts/zones/The_Garden_of_RuHmet/bcnms/when_angels_fall.lua
```diff
@@ -62,8 +62,8 @@ end;
 	
 function onEventFinish(player,csid,option)
 -- print("bc finish csid "..csid.." and option "..option);
-  if(csid== 0x7d01)then
-    player:setPos(420,0,445,192);
-end
+    if(csid== 0x7d01)then
+        player:setPos(420,0,445,192);
+    end
```
TeoTwawki: All I see changed is whitespace, not actual code.

sisternERA: the end was in the wrong spot and was calling an end in the wrong spot

TeoTwawki: You didn't move the end. You changed the whietspace. Lua doesn't care about whats between words it is not a whitespace sensative laguage. It could be written as
```lua
if(csid== 0x7d01)then player:setPos(420,0,445,192); end
```
or
```lua
if(csid== 0x7d01)then
player:setPos(420,0,445,192);
end
```
and still have the same behavior (would make me cry to have to read, but it wouldn't stop it from working).

sisternERA:
-- print("bc finish csid "..csid.." and option "..option);
  if(csid== 0x7d01)then
    player:setPos(420,0,445,192);
end
 
end;  i thought that the ends stacked like that would cause problems
sorry i dont have alot of experience yet.  but i soon as i did that though something changed and it would work.  Before it was throwing out errors that had alot of ===== in a row though

teschnei:
whitespace is not part on syntax in lua

PR #895, opened 2015-02-08
Added items

nasomi:


scripts/globals/items/bravura.lua
```diff
@@ -0,0 +1,26 @@
...
+-- onAdditionalEffect Action
+-----------------------------------
+
+function onAdditionalEffect(player,target,damage)
+    local chance = 10;
+
+    if (math.random(0,99) >= chance) then
+        return 0,0,0;
+    else
+        target:delStatusEffect(EFFECT_EVASION_DOWN)
```
xdemolish: might be better to check the status effect first before attempting to remove it (get rid of the not and stick this line inside the check, move the adding of the status effect outside of it)

scripts/globals/items/gungnir.lua
```diff
@@ -0,0 +1,26 @@
...
+-- onAdditionalEffect Action
+-----------------------------------
+
+function onAdditionalEffect(player,target,damage)
+    local chance = 10;
+
+    if (math.random(0,99) >= chance) then
+        return 0,0,0;
+    else
+        target:delStatusEffect(EFFECT_DEFENSE_DOWN);
```
teschnei: gungnir doesn't overwrite all defense downs (it will not overwrite angon)

maxtherabbit:
pull rq #912 supersedes this, can close

PR #900, opened 2015-02-08
Missing 2hr IDs, commented out because I didn't make retail accurate scripts.

TeoTwawki:
These will mostly pertain to future content nobody has worked on yet, like missing limit break quests.


teschnei:
most of those will have a) an animation for non-equip type modelIDs (since they don't animate any bones) and an animation for equip type modelIDs (which just looks the same as players)

PR #905, opened 2015-02-08
crafting belt correction..

eraffxi:
correcting craft belts, giving 55 skill, supposed to give 2.


teschnei:
they should all give 3, not 2 (except fishing, which gives 2 and lasts 2 hours)

PR #906, opened 2015-02-09
s'more Additional effect stuffs.

TeoTwawki:
https://github.com/DarkstarProject/darkstar/pull/906/files?w=1

Fixed missing / incorrect animation for MP Drain sub effect. Also a handful of lower lv things and that didn't make it into my last batch before doing the rest.


TeoTwawki:
P.s.
MP draining behavior is retail, if target is undead, no proc at all. If target has no mp, drain for zero.

PR #908, opened 2015-02-09
Fix missing require.

TeoTwawki:
Worked when before without the require during testing because:
1) If something else just loaded status.lua then those variables would be defined even without a require.
2) Status normal had an ID of zero anyway.


TeoTwawki:
nvm, need to take another look at wtf has changed with npc's this worked when I tested before and now it does strange things...

TeoTwawki:
For the curious, see #909

PR #911, opened 2015-02-09
Update mob_skill.sql

nesstea:
Gave the Antlions the pit ambush ability.


nesstea:
My bad. Still learning GitHub.

PR #912, opened 2015-02-09
Item changes

nesstea:
Updated the SQL with some files for additional effects and/or weapon skills.


scripts/globals/items/excalibur.lua
```diff
@@ -0,0 +1,29 @@
...
+        chance = utils.clamp(chance, 5, 95);
+    end
+    if (math.random(0,99) >= chance) then
+        return 0,0,0;
+    else
+		local finalDMG = math.floor(player.getHP(player)/4);
+        if (finalDMG >0) then
+			local physicalResist = target:getMod(MOD_SLASHRES)/1000;
+			finalDMG=finalDMG*physicalResist;
+			target:delHP(finalDMG);
```
teschnei: still need a call to target:physicaldmgTaken to get the rest of the physical damage modifiers

nesstea: Will do.

maxtherabbit:
@TeoTwawki that's because finalMagicNonSpellAdjustments() *does* remove hp from the victim

scripts/globals/items/shiranui.lua
```diff
@@ -0,0 +1,43 @@
...
+			params.bonusmab = 0;
+			params.includemab = true;
+			lightDamage = addBonusesAbility(player, ELE_LIGHT, target, lightDamage, params);
+			lightDamage = lightDamage * applyResistanceAddEffect(player,target,ELE_LIGHT,0);
+			lightDamage = adjustForTarget(target,lightDamage,ELE_LIGHT);
+			if (lightDamage < 0) then
+				lightDamage = 0
+			end
+			lightDamage = finalMagicNonSpellAdjustments(player,target,ELE_LIGHT,lightDamage);
+			if (lightDamage >35) then
```
maxtherabbit: which reminds me...you need to cap this valuebefore calling finalMagicNonSpellAdjustments()

maxtherabbit:
and we need to handle the victim's phalanx/stoneskin for the excalibur hits

add in something like this before the delHP() call:

finalDMG = finalDMG - target:getMod(MOD_PHALANX);
utils.clamp(finalDMG, 0, 99999);
finalDMG = utils.stoneskin(target, finalDMG);

scripts/globals/items/shiranui.lua
nesstea: Nrrh is getting the actual cap from retail.

teschnei:
oh yeah, you have to remove all the delStatusEffect lines in the additional effects of your weapon procs

TeoTwawki:
eva down removed eva boost, def down removed def boost. And take out the if(not effect check, it will overwrite or not by power lv

TeoTwawki:
I made same mistake earlier, bad copy paste

TeoTwawki:
#1536 I believe I fixed all the erasing of wrong status effects (def down will remove def boost, rather than itself, unless the previous instance of it was weaker)

PR #916, opened 2015-02-11
Updated Augment table with new info and added some missing modifiers and values.

TeoTwawki:


TeoTwawki:
Interested parties who'd like to help make all these unimplemented augments work should contact me on IRC.

PR #921, opened 2015-02-12
Added drop ID for Kotan-kor Kamuy to drop Cradle Horn

Fiocitrine:


Fiocitrine:
I don't know what that pull request #2 thingy is.
Someone, give me an idea?

Fiocitrine:
ok, I think it's because I did a pull from darkstar to my fork then did a pull request with the newly merged commits to darkstar. hmmm I don't think it should affect it.

PR #924, opened 2015-02-12
CORE: postpone calls to UpdateHealth() until equipment swaps are complete

maxtherabbit:
this allows for correct calculation of HP/MP when swapping from 2 equipment items in the same slot that both have +HP/MP

i.e.  you have full mp and swap from MP+24 item to MP+14 item, now you only lose 10 mp, before your max mp was adjusted correctly but you end up 14 shy of max


teschnei:
you don't need to put a comment for anything thats pretty immediately obvious - comments should be used to help explain code, not repeat what the code says

maxtherabbit:
It's obvious to you and me, but I put it in the header so ppl could see it without reading the function

Issue #899, opened 2015-02-08
onAdditionalEffect always overrides Sambas


nesstea:
I added the item in the db to have an onAdditionalEffect and got it working perfectly, except it will never drain hp with drain or aspir sambas. I am trying to figure it out, but if anything could lend a hand, I can make the edit and commit.

nesstea:
Do I have to code the samba effect in the onAdditionalEffect script when it doesn't proc (if the mob has daze)?

teschnei:
no, either the effect is supposed to return something specific if it didn't
proc or the core is missing something to check if it didn't proc

TeoTwawki:
Core is missing the check. The guy that originaly coded the scripted add effects never made it check sambas after a failed additional effect proc so if add effects are checked at all no other effect can happen after it.

teschnei:
Yeah, just checked. Not a hard fix, I should have internet finally in an
hour or two for fixing

TeoTwawki:
Cool, I just got going on the next heap of scripts for the items.

nesstea:
@TeoTwawki I have all but two relics coded for additional effects if you'd like me to submit them.

TeoTwawki:
yes please, no point is multiple people doing same ones more than once. I'm hitting items in a mostly level ascending order, did up to lv39 last batch so starting there and working my way up now.

nesstea:
They currently delete the previous effect on the mob without checks and reapply it. Just a temporary fix in the meantime.

PR #926, opened 2015-02-12
CORE: scripted addeffect weapons will not overwrite sambas if they don't proc

maxtherabbit:
fixes issue #899


nesstea:
max you are a god. thanks

Issue #367, opened 2014-07-01
Cure for zero on non undead monster target adds player to hate list?


TeoTwawki:
Pretty sure you can't pull mobs with cure on retail unless the mob is undead.

TeoTwawki:
@maxtherabbit 
Thank you max! <3

PR #928, opened 2015-02-13
CORE: issue #367

maxtherabbit:


teschnei:
i dont think thats right, pretty sure resisted spells still generate enmity (like bind etc)

should double check what tookEffect does

teschnei:
also a cure probably still aggroes the mob (it must generate at least 0 enmity)

maxtherabbit:
with this change, resisted spells will still generate enmity unless they are cures

the condition is if !isHeal() *or* tookEffect()

maxtherabbit:
and in the comments of #367 you said the cure doesn't aggro the mob so I'm lost now

teschnei:
yes, that's why i wanted to check what tookEffect actually does (because tookEffect was not in this conditional before)

if i said that after testing it, then that's probably right

maxtherabbit:
all tookEffect does is test the spell action message

maxtherabbit:
wait I think I get what you're saying now... if you cast cure on an undead and it resists it should still generate enmity? I can change that, you were just confusing me talking about resists b/c when you cure a non-undead the message is "no effect"

maxtherabbit:
better?

teschnei:
yeah, but other things do "no effect" and should claim and generate enmity as well

maxtherabbit:
uh they do already don't they? the conditional I added tests true for all spells that aren't cures so it wouldn't affect other spells that have no effect

PR #932, opened 2015-02-14
Added Functionality to the two Non-Working, But Responsive, Strange Apparatuses

JoeCT:
Added functionality to the two non-working, but responsive, strange apparatuses.  They now mimic retail behavior.


JoeCT:
Lua Function TellToPlayer(), GM Command @setpoint

TellToPlayer() is used to display messages to the player without the
message "----== SystemMessage ==----" preceding each line.

@setpoint allows the user to save up to 9 different locations with a
short (44 char) description and to selectively 'warp' to them as needed.
I use this when testing scripts involving several different locations.
The @setpoint command makes use of the function above, TellToPlayer(),
to display usage instructions and a small table listing the descriptions
of the currently saved set points.

xdemolish:
the command should be part of a separate pull request

JoeCT:
I agree, but I'm not very familiar with git and I mistakenly added it to my fork prior to this pull request being merged.  How can easily I remove it?

starlightknight:
@JoeCT If its the latest change in your branch, you can do git reset --soft HEAD~1 to pull it back into uncomitted state, where you can then move it to another branch. 

If it isnt, you can make a new branch and use git cherry-pick to move it to another branch. After that, you can do git rebase -i master on your old branch and drop it off the commit list that pops up.

JoeCT:
Figured it out. Thanks.

starlightknight:
Reverting in git leaves it in history, so unless it has already become part of the main repo its better to drop it out, otherwise it can be confusing when you go to reintroduce it later. 

Similarly its better to always rebase your branches instead of pulling the latest or merging it into your pull request - it makes history cleaner, and clearer as to what happened when. 

Merges will cause a divergance point in history when browsing it, if you rebase the changes will come in in a straight line

JoeCT:
I believe I understand what you're saying.  I'm going to have to study up on git in order to get better at it.  I've just learned the bare minimum to get by, i'm still trying to get a good grasp on everything in the server, kind of pushed git to the back burner.

starlightknight:
No problem at all, just offering some advice :-) 

If you want to understand the difference between merge vs rebase w/ illustrations, this page in the reference book has some good ones - http://git-scm.com/book/en/v2/Git-Branching-Rebasing

PR #934, opened 2015-02-14
These are my first .lua files also my first commit, be gentle with me.

Yuchemey:
Signed-off-by: Yuchemey <Yuchemey@hotmail.com>

### DIFF
--- a/scripts/globals/effects/etude.lua
+++ b/scripts/globals/effects/etude.lua
@@ -20,11 +20,13 @@ end;
 -----------------------------------
 
 function onEffectTick(target,effect)
+	-- the effect loses modifier of 1 every 10 ticks.
+	local song_effect_size = effect:getPower();
 	if(effect:getTier() == 2 and effect:getPower() > 0) then
-	effects:setpower(effect:setpower(), 1)
-	target:delMod(effect:getSubPower(), 1)
+		effect:setPower(song_effect_size -1)
+		target:delMod(effect:getSubPower(), 1);
+		
 	end
-
 end;
 
 -----------------------------------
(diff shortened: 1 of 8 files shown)


teschnei:
title your pull request with what you actually changed, you can mention the rest in the description

Yuchemey:
OH sorry, i've added T2  Etude bard song.

teschnei:
don't forget to make their effect decrease by 1 every 10 seconds (if you don't know how, just ask)

Yuchemey:
I don't know how ^^;, i'll look into it.

teschnei:
the third param to addBardEffect is tick time (how often onEffectTick is called) and the last param is tier (which you should probably change to 2, since it's the second tier)

then you just have to modify effects/etude.lua to lower the stat if it's a) a tier 2 etude, b) is still giving more than 0 of the stat, and c) set the power to the new value of the mod

Yuchemey:
Should the effect be in script/globals/effect/etude.lua or should it be in script/global/spells/<insertspellname>.lua?

teschnei:
find the one that exists already (it's the same one that tier 1 etudes use)

Yuchemey:
ok i made Bewitching Etude i used teotwacki's/temp effect. I forgot i could browse threw the site to help me. is that pretty much it?

Yuchemey:
"if not (target:addBardSong(caster,EFFECT_ETUDE,power,10,duration,caster:getID(), 7, 2)) then"
what is the seven? i know that 10 is the power and 2 is the tier. but the 7?

Yuchemey:
Ok so i modified the effect of the spells as requested. 
-Sub power
-T2 instead of T1
-Tick at 10 instead of 0 
from my previous Push. Let me know what you think.

teschnei:
ok, a couple things here: the second to last param (subpower) is the mod that the etude adds.  so herculean etude should be MOD_STR, etc.

second thing, you still have to modify scripts/globals/effects/etude.lua to tick the effect downwards if it's a tier 2 etude

teschnei:
now just waiting for you to test it so you can fix the typos :)

Yuchemey:
hehe awesome.

Issue #822, opened 2015-01-20
Latent Effect: Song/Roll active not implemented


Fiocitrine:
I tried to script the Melody Earring/ +1's Latent effect to activate on song/roll but I guess the Latent ID isn't implemented. The Latent ID is 25 according to latent_effect.h. https://github.com/DarkstarProject/darkstar/blob/master/src/map/latent_effect.h

maxtherabbit:
all the bard songs and cor rolls need to have the proper EFFECTFLAG set before this can work

PR #937, opened 2015-02-14
CORE / SQL: added latents for rolls/songs

xdemolish:
CORE: status effect flags are now uint32 (don't know why they were still uint16 with EFFECTFLAG_SYNTH_SUPPORT at 0x40000)

needs testing


maxtherabbit:
looks solid, will close issue #822

PR #942, opened 2015-02-15
Lua Function TellToPlayer(), GM Command @setpoint

JoeCT:
TellToPlayer() is used to display messages to the player without the
message "----== SystemMessage ==----" preceding each line.

@setpoint allows the user to save up to 9 different locations with a
short (44 char) description and to selectively 'warp' to them as needed.
I use this when testing scripts involving several different locations.
The @setpoint command makes use of the function above, TellToPlayer(),
to display usage instructions and a small table listing the descriptions
of the currently saved set points.

### DIFF
--- a/scripts/commands/setpoint.lua
+++ b/scripts/commands/setpoint.lua
@@ -1,0 +1,716 @@
+---------------------------------------------------------------------------------------------------
+-- func: @setpoint
+-- auth: Joe
+-- desc: @setpoint allows players to save up to 9 positions with a 44 char description and then return to them as needed.
+---------------------------------------------------------------------------------------------------
+
+cmdprops =
+{
+    permission = 1,
+    parameters = "ssssssssssssssss"
+};
+
+local function delPointNum(player,pointNumber)
+
+	player:setVar("@setpoint" .. pointNumber .. "_XPos", 0);
+	player:setVar("@setpoint" .. pointNumber .. "_YPos", 0);
+	player:setVar("@setpoint" .. pointNumber .. "_ZPos", 0);
+	player:setVar("@setpoint" .. pointNumber .. "_Zone", 0);
+	player:setVar("@setpoint" .. pointNumber .. "_DescA", 0);
+	player:setVar("@setpoint" .. pointNumber .. "_DescB", 0);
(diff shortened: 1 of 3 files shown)


teschnei:
useful for sure, but not sure if it needs to be in repository's code

JoeCT:
I had originally intended to just upload this command to the forum, but because I use a modified version of PrintToPlayer in the script I felt that most wouldn't bother to add it in.

starlightknight:
I don't think having good debug commands like this in the repo is a bad thing, as long as appropriate GM level is set by default. I'm always fumbling with positions when testing missions and tbh i shouldve invented something like this by now. Having an easy way built in to store the points for a mission to do a final runthrough may help reduce bugs. It can encourage doing a final end-to-end runthrough - its easy to accidentally break something last minute when tweaking things, but a pain to keep track of all the positions to do a full run fast.

teschnei:
i don't script missions, so if these are useful for developers then that's good (i just dont really care much for putting server management stuff in)

TeoTwawki:
I think we should just modify print to player to take a message type param, and default it to the current behavior when none is specified.

TeoTwawki:
sort of like so but without the spoofy object stuff: 
https://github.com/LegionXI/darkstar/blob/TestServer/src/map/lua/lua_baseentity.cpp#L8583
https://github.com/LegionXI/darkstar/blob/TestServer/src/map/packets/chat_message.h#L53

Tagban had posted something similar to the forums awhile back.

xdemolish:
@TeoTwawki  it's been done before, just not in the repo yet https://forums.dspt.info/viewtopic.php?f=19&t=2402

TeoTwawki:
@xdemolish I literally just mentioned that post lol

xdemolish:
i hadn't refreshed so didn't see til after i posted my comment :p

JoeCT:
I hadn't seen Tagban's PrintToPlayer () yet, I really like that I can use it to change the color of the text.  I'm going to go ahead and close this pull request.  I've already modified my script to use Tagban's version of PrintToPlayer () and will upload it to the forums for anyone who may be interested in using it.

PR #943, opened 2015-02-15
Implement retaliation

TeoTwawki:
https://github.com/DarkstarProject/darkstar/pull/943/files?w=1

### DIFF
--- a/scripts/globals/abilities/retaliation.lua
+++ b/scripts/globals/abilities/retaliation.lua
@@ -1,0 +1,23 @@
+-----------------------------------
+-- EFFECT_RETALIATION
+-- Allows you to counterattack but reduces movement speed.
+-- Unlike counter, grants TP like a regular melee attack.
+-----------------------------------
+
+require("scripts/globals/status");
+
+-----------------------------------
+-- onAbilityCheck
+-----------------------------------
+
+function onAbilityCheck(player,target,ability)
+    return 0,0;
+end;
+
+-----------------------------------
+-- onUseAbility
+-----------------------------------
+
+function onUseAbility(player, target, ability)
+    player:addStatusEffect(EFFECT_RETALIATION,1,0,180);
+end;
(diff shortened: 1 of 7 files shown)


maxtherabbit:
looks right

scripts/globals/abilities/retaliation.lua
```diff
@@ -0,0 +1,24 @@
...
+function OnAbilityCheck(player,target,ability)
+    return 0,0;
+end;
+
+-----------------------------------
+-- OnUseAbility
+-----------------------------------
+
+function OnUseAbility(player, target, ability)
+    player:delStatusEffectSilent(EFFECT_RETALIATION);
```
teschnei: not needed (that's what all the flags in status_effects.sql are for)

src/map/utils/battleutils.cpp
```diff
@@ -551,7 +551,56 @@ bool HandleSpikesDamage(CBattleEntity* PAttacker, CBattleEntity* PDefender, apAc
...
+                dmg = dmg + bonus;
+
+                // FINISH HIM! dun dun dun
+                // TP and stoneskin are handled inside TakePhysicalDamage
+                Action->spikesMessage = 536;
+                Action->spikesParam = battleutils::TakePhysicalDamage(PDefender, PAttacker, dmg, false, SLOT_MAIN, 1, NULL, true);
+                PAttacker->addHP(-Action->spikesParam);
+            }
+            /*
+            else
```
teschnei: just put the skill up code in the if TYPE_PC block? pretty sure the rest of that is all fine for mobs, then you don't have to worry about it being not implemented later

teschnei:
a good idea would be to write a new function that calculates the damage of one attack and have all this just call that (get weapon dmg, fstr, ratio, check for null damage, check for absorbed damage, etc)

TeoTwawki:
Updated. Couldn't do the new function, need to much needed that is unrelated to this JA. Would like to revisit that idea later though.

PR #945, opened 2015-02-15
CORE: LoadChar(): assign hp/mp values from db after loading equipment

maxtherabbit:
this allows players to retain hp/mp added from gear when zoning/logging in

tested working, but I want @teschnei to look at it before merge


src/map/utils/charutils.cpp
```diff
@@ -706,13 +708,15 @@ namespace charutils
...
+        PChar->animation = (HP == 0 ? ANIMATION_DEATH : ANIMATION_NONE);
         charutils::LoadInventory(PChar);
-        PChar->m_event.EventID = luautils::OnZoneIn(PChar);
-
         charutils::LoadEquip(PChar);
+        PChar->health.hp = HP;
+        PChar->health.mp = MP;
+        PChar->UpdateHealth();
+
+        PChar->m_event.EventID = luautils::OnZoneIn(PChar);
```
teschnei: can't move this after loadequip

PR #954, opened 2015-02-17
Linkshell Audit

RunAwayDSP:
Include linkshell name in audit useful for web based scripts aswell


whasf:
FINALLY getting to this.... It should have its own db column, not be concatenated into the string.
I'm thinking about adding the zone from where they sent the message in a not-too-distant future update

PR #956, opened 2015-02-17
CORE/SCRIPTS: create battleutils::TakeWeaponskillDamage() and apply dama...

maxtherabbit:
...ge class reductions in scripts

### DIFF
--- a/src/map/ai/ai_char_normal.cpp
+++ b/src/map/ai/ai_char_normal.cpp
@@ -2527,15 +2527,7 @@ void CAICharNormal::ActionWeaponSkillFinish()
     //incase a TA party member is available
     CBattleEntity* taChar = NULL;
 
-    //trick attack agi bonus for thf main job
-    if (m_PChar->GetMJob() == JOB_THF && m_PChar->StatusEffectContainer->HasStatusEffect(EFFECT_TRICK_ATTACK))
-    {
-        taChar = battleutils::getAvailableTrickAttackChar(m_PChar, m_PBattleTarget);
-        if (taChar != NULL && damage >= 0) damage += m_PChar->AGI();
-    }
-
-    //check if other jobs have trick attack active to change enmity lateron
-    if (taChar == NULL && m_PChar->StatusEffectContainer->HasStatusEffect(EFFECT_TRICK_ATTACK))
+    if (m_PChar->StatusEffectContainer->HasStatusEffect(EFFECT_TRICK_ATTACK))
         taChar = battleutils::getAvailableTrickAttackChar(m_PChar, m_PBattleTarget);
 
     if (!battleutils::isValidSelfTargetWeaponskill(m_PWeaponSkill->getID()))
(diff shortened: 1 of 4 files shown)


src/map/utils/battleutils.cpp
```diff
@@ -1904,6 +1904,129 @@ uint32 TakePhysicalDamage(CBattleEntity* PAttacker, CBattleEntity* PDefender, in
...
+    if (PDefender->objtype == TYPE_PC)
+        charutils::UpdateHealth((CCharEntity*)PDefender);
+
+    if (PAttacker->objtype == TYPE_PC && !isRanged)
+        PAttacker->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_ATTACK);
+
+    if (damage < 0)
+        return -damage;
+    else
+        return damage;
```
maxtherabbit: @teschnei looking at this today, I'm thinking maybe I should change the new function to returned signed damage in case the mob absorbs the magical weaponskill. thoughts?

teschnei: Ya.. I think at this point all damage calcs should be signed 32 bit ints

maxtherabbit:
ok that should do it, I've tested pretty thoroughly

I didn't bother to convert TakePhysicalDamage() to signed return type. I can if you want, but I don't know any situation where a true physical attack can be absorbed

teschnei:
There are many many mobs that absorb physical attacks. Even before abyssea,
mammets in CoP fights do it

maxtherabbit:
ok I'll get to it eventually in a different pull rq, it's not relevant to this branch and TakePhysicalDamage() is called all over the place

teschnei:
indeed. I'll look at this PR when I get home

src/map/ai/ai_char_normal.cpp
```diff
@@ -2531,22 +2531,15 @@ void CAICharNormal::ActionWeaponSkillFinish()
     if (m_PChar->GetMJob() == JOB_THF && m_PChar->StatusEffectContainer->HasStatusEffect(EFFECT_TRICK_ATTACK))
     {
         taChar = battleutils::getAvailableTrickAttackChar(m_PChar, m_PBattleTarget);
-        if (taChar != NULL) damage += m_PChar->AGI();
+        if (taChar != NULL && damage >= 0) damage += m_PChar->AGI();
```
teschnei: er.. this isn't even close to right (not your change, but what was there before...)

lets just get rid of it.. (it should be in the script, because it has to get multiplited by ftp - maybe it needs a binding)

src/map/ai/ai_char_normal.cpp
```diff
@@ -2636,14 +2629,14 @@ void CAICharNormal::ActionWeaponSkillFinish()
...
-    if (Action.reaction == REACTION_HIT && (m_PWeaponSkill->getPrimarySkillchain() != 0))
+    if (Action.reaction == REACTION_HIT && m_PWeaponSkill->getPrimarySkillchain() != 0 && damage >= 0)
     {
         // NOTE: GetSkillChainEffect is INSIDE this if statement because it
         //  ALTERS the state of the resonance, which misses and non-elemental skills should NOT do.
         SUBEFFECT effect = battleutils::GetSkillChainEffect(m_PBattleSubTarget, GetCurrentWeaponSkill());
         if (effect != SUBEFFECT_NONE)
         {
-            uint16 skillChainDamage = battleutils::TakeSkillchainDamage(m_PChar, m_PBattleSubTarget, damage);
+            uint16 skillChainDamage = battleutils::TakeSkillchainDamage(m_PChar, m_PBattleSubTarget, (uint16)damage);
```
teschnei: while you're here, should modify this so it a) accepts values up to 99999, and b) still works for absorbed WS (it would just treat a -3000 WS as a 3000 WS)

maxtherabbit:
ok so:
1) take out the AGI bonus 
2) add to script (I'll have to look up the proper formula unless you know it offhand?)
3) expand takeSCdamage param to 32bit

so you're saying if you close a skillchain with an absorbed WS that does -3000dmg, the skillchain will still go off and do +3000dmg correct?

maxtherabbit:
AGI bonus for trick attack was already being added in script

teschnei:
Well then.. Just do 1 and 3 then and were good. 3 is just changing param to
int32 and using the positive number in sc calcs if it was negative

maxtherabbit:
done and done, tested, should be mergeable

PR #958, opened 2015-02-17
SQL: add enspell dmg bonus to hollow earring

maxtherabbit:


nesstea:
same with enhancing sword.

maxtherabbit:
enhancing sword already works...

korietsu:
Does Apoc Nigh even work to use the earring?

maxtherabbit:
not sure, but the other mods for the earring are in the db already

PR #961, opened 2015-02-17
King Behemoth Defending Ring

RunAwayDSP:
http://wiki.ffxiclopedia.org/wiki/King_Behemoth
Either Pixie Earring or Defending Ring will drop 100% of the time;
Defending Ring being the much rarer drop.


RunAwayDSP:
oops wrong location for code... fix coming

PR #964, opened 2015-02-18
Clamming, Modifications to Several NPCs to Mimic Retail Behavior

JoeCT:


scripts/zones/Bibiki_Bay/npcs/Clamming_Point.lua
```diff
@@ -0,0 +1,176 @@
...
+			5123, 11, 0.628, 0.892, -- Jacknife  
+			5122,  3, 1.000, 1.000  -- Bibiki Slug
+		      };
+
+-----------------------------------
+-- Local Functions
+-----------------------------------
+
+local function giveImprovedResults(player)
+	local legs = player:getEquipID(SLOT_LEGS);
```
TeoTwawki: I think making a modifier to check would be better than checking static itemIDs. 

And if clamming is possible in any other zone ever, can move clamming code to a global (I think your are correct in doing it here because noplace else has clamming in retail right now? but not 100% sure thought I'd mention just in case).

scripts/zones/Bibiki_Bay/npcs/Pohka_Chichiyowahl.lua
```diff
@@ -0,0 +1,54 @@
...
+end;
+
+-----------------------------------
+-- onTrigger Action
+-----------------------------------
+
+function onTrigger(player,npc)
+	player:showText(npc,POHKA_SHOP_DIALOG);
+
+	stock = {
```
TeoTwawki: local stock

JoeCT:
The reason why I didn't create a modifier and jut checked the static itemIDs is because that is the way it is currently implemented in harvesting, mining, and logging. 

Currently clamming is only available in Bibiki Bay on Purgonorgo Isle, but if they ever do expand it, it should be relatively easy to transfer the code to a global.  I sincerely doubt they will though, I don't remember too many people that were actually interested in clamming.

Thanks for pointing that out about stock not being a local var.  I just corrected it.

TeoTwawki:
ew...Well, leave that like it is if devs ok with it, and I'll catch it when I do the existing ones since they have same thing and I plan to go on a mod creation spree for that sort of thing soon. Tons of stuff doing item checks in stuff like ability scrpts atm.

teschnei:
it should be a modifier at some point - if teo says he's going to do it then that's alright since i'm sure he'll get it done, but ideally all of them should be modifiers eventually

JoeCT:
Closing this pull request in lieu of pull request #965.

PR #965, opened 2015-02-19
Clamming (Now Using Item Mods), Modifications to Several NPCs to Mimic Retail

JoeCT:
Clamming now uses item mods instead of checking static itemIDs.


JoeCT:
@TeoTwawki Look this over and let me know if this is how you thought the item mods should be implemented.  -- Thanks

TeoTwawki:
The way you have it should be fine since each race currently has only 1 item each that can be equipped at one time to grant each of these 2 mods making there no way to stack it. For the other ones I gotta do it slightly diff, since there are multiple items granting the same boost.

JoeCT:
@TeoTwawki  Thanks for looking it over. Hadn't dealt with item mods before, took me a little bit to figure out how they're implemented.

nesstea:
great work joe

JoeCT:
@nesstea Thanks.

PR #968, opened 2015-02-19
February Update: NPC ID / Lua Shifts

starlightknight:
Usual batch work. NPC batch has been becoming more stable.

Usual note on the lua files - If the were off before this update, they will be off by the same amount post-batch. This is because it correlates it to npc id shifts, where npcs were already correct.


starlightknight:
@gedads - here are the changes for followup, marked which ones need to be recapped for sure:

Uleguerand Range (Whole zone)
Attohwa Chasm (Whole zone)
Lufaise Meadows (Whole zone, Reverify 16875789 16875790, NPC[x] now appears to refer to a mob)
Misareaux Coast (Whole zone, Reverify 16879910 16879911, NPC[x] now appears to refer to a mob)
East Sarutabaruta (1 change)
The Eldieme Necropolis (Reverify 17576303, it used to not have a label in POLUtils but is now labeled Strange Apparatus)
Provenance (2 changes)
Selbina (1 change)
Western Adoulin (Reverify 17825943)
Eastern Adoulin
Ceizak Battlegrounds
Foret de Hennetiel (Midzone partial zone shift, may want to recap)
Yorcia Weald [U] (The 1 npc in the db now has a label in POLUtils and didn't previously. May want to verify it - it has an internal name so I can't tell)
Mog Garden (Couple of shifts)

I will try and scrape for mob changes in a bit - that code is still a work in progress but I may be able to identify changed zones for you, if any.

xdemolish:
is it ok to merge, or would you rather wait on gedads to recap some zones first?

starlightknight:
@xdemolish Usually we merge it so he can base his patch on this easier. With the exception of the handful of noted npcs above, they all check out. No startup warnings with the latest build with this change

gedads:
With unity system they added a lot of mobs, so shifting in some areas isn't abnormal i think also some areas were modified (mog garden), i need to recapture those zones soon.

starlightknight:
@gedads I rescraped mob names from POLUtils w/ the new dump and did not find any changes to the mobs in our db. So I didn't see any shifts on existing mobs, so they must all be new.

PR #969, opened 2015-02-19
Quests and missions

nesstea:
Dangruf stone ??? is supposed to only show up during sunny/clear weather (day or night). This change allows players to get the stone for one of 3 missions.


nesstea:
These changes did not change the ??? showing up or not. That is handled somewhere else. :smile:

PR #973, opened 2015-02-20
Changed Wildfire to calculate as magic damage

Fr0zt:


teschnei:
need to do a revision on the doMagicWeaponskill function since Wildfire's dSTAT is AGI - INT (not INT - INT) and I guess the club ones are probably MND as well

TeoTwawki:
@teschnei  Shining Blade and Seraph Blade would be MND too

maxtherabbit:
do you want to just add a dSTAT parameter to doMagicWS() like we have with applyResistance() ?

teschnei:
That or use params to send the stats (that way we can do int-int default
and not have to change them all. Then again, there's not that many..
Anyways, I prefer params over another arg though. That's what its for, not
bloating args

PR #976, opened 2015-02-20
Implemented Colibri Magic Parroting (often called "reflect").

TeoTwawki:
https://github.com/DarkstarProject/darkstar/pull/976/files?w=1

The same method can be used to handle Jailer of Fortitude but I leave that to someone more familiar with his retail behavior.

This is based mainly on this wiki page http://wiki.ffxiclopedia.org/wiki/Colibri#notes

1) Wiki claims that bird unable to cast will stay in mime mode for "about a minute or so". Not exactly hard data so I went with 30 seconds, adjust as needed.
2) Wiki claims that if you cast cure on them, they will cast cure on a random party member. Doesn't work (yet).
3) That delay IS needed, or they won't animate properly.
4) Didn't need the MOBMOD at all, that ID is now freed up for future use.
5) In the core onMagicHit is return to expect 1 return but does nothing at all with it. I can think of a few uses for that, but nothing we really need for anything yet. Rainy day?
6) Wiki says any spell that either does damage, or is not outright resisted will be copied unless it is a blue magic spell or an avatar blood pact (which aren't really spells). I got that part working flawlessly.


TeoTwawki:
@korietsu tag, your it!

korietsu:
This looks fantastic. I'll start taking a look at it to see about JoF. 

Finally got my new rig up, so I should be able to start working on sea again, not to mention im not being flown half way across the world for work anymore!

Scavenge:
i also have been working on sea and have most jailers coded to the limit of what is currently possible, i am willing to share

scripts/zones/Bhaflau_Remnants/mobs/Colibri.lua
```diff
@@ -0,0 +1,58 @@
...
+-- onMobFight Action
+-----------------------------------
+
+function onMobFight(mob, target)
+    local delay = mob:getLocalVar("delay");
+    local LastCast = mob:getLocalVar("LAST_CAST");
+    local spell = mob:getLocalVar("COPY_SPELL");
+    if (mob:AnimationSub() == 1) then
+        if (mob:getBattleTime() - LastCast > 30) then
+            target:PrintToPlayer("In timer check");
```
xdemolish: dat debug print tho

TeoTwawki:
...Removed the debug prints I missed.

src/map/ai/states/magic_state.cpp
```diff
@@ -665,6 +661,11 @@ void CMagicState::FinishSpell()
 
         action.messageID = msg;
 
+        if (PTarget->objtype == TYPE_MOB && msg != 31)
```
teschnei: hello magic number, you seem to be missing an enum or a comment defining what you are supposed to be

TeoTwawki: Comment it is, because same thing happens with the utsu message.

PR #986, opened 2015-02-22
Ability changes

nesstea:
Allows steps to consider the mythic Terpsichore.

### DIFF
--- a/scripts/globals/abilities/box_step.lua
+++ b/scripts/globals/abilities/box_step.lua
@@ -160,7 +160,7 @@ function onUseAbility(player, target, ability)
 			if (daze > 2) then
 				daze = 2;
 			end;
-			player:addStatusEffect(EFFECT_FINISHING_MOVE_3+daze + MythicModifier,1,0,7200);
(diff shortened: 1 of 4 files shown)


maxtherabbit:
why not just do 

player:addStatusEffect(EFFECT_FINISHING_MOVE_1+daze + player:getMod(MOD_STEP_FINISH),1,0,7200);

and do away with the modifier variable?

nesstea:
If I was going to change it, I'd rewrite the whole thing. I was going to just add one to the daze variable, but the daze variable is being set again in the nested sluggish daze if-thens.

teschnei:
you're better off getting rid of the variable and tacking on the getMod to the "daze" variable in the first half of the script (except when the target has max daze already), and then capping daze in the second half of the script - the way its currently written isn't really extensible and will break if there are more than one source of the mod

nesstea:
ok but there is only one source of this mod. are you talking about future content?

teschnei:
yes, i am always keeping future content in mind (as the proper developer should)

nesstea:
Ok I'll do that today.

PR #995, opened 2015-02-24
Adjusted offset so the mob opens the right door.

nesstea:
The gargoyles open the door on death with this change.


nesstea:
Redoing it for whasf

Issue #918, opened 2015-02-11
nation mission 5-1 (all 3 nation) very much not retail


TeoTwawki:
All mobs these mobs should not be up and super linked. Only 2 helpers should be up at a time with the archlich. Also a single whm banishaga should not be near a near one hit kill for every skeleton type mob in the game.

gedads:
https://www.youtube.com/watch?v=boEOjTIxfkA
watch near 6:00 up to 7:00
they're not all there at start and are summoned as the fight goes on and come back if defeated. If left alone you can totally get surronded.

TeoTwawki:
Still, right now the second you agro/attack the Archlich, you instantly have ALL of them on you, which is not how retail handles that. Also, the 2 sorcerers never come back, only the 4 warriors. There are 4, but they only pop 2 at a time. So I misspoke saying up at once, its 2 pop at once.

Yuchemey:
Yes I did the fight on era - I also remember that fight that you can't kite the helpers. But I I was able to.
I remember that "a kite" would take a mob even though after a little while it would ve. Loose enmit and the tank would get pwn d

maxtherabbit:
@TeoTwawki is this closable now?

TeoTwawki:
@maxtherabbit I made the skeleton spawns and repops as close to retail as I could figure, but didn't address anything else. 

Did retail severely bump up whm nuke power and we got that or something? I dunno whats up with banish power. Or cure power for that matter: whm have discovered they can near 1 shot IT bones in about every zone for fast exp so was seeing them alternate banishga's and cures to take out groups for quick chains. On legions server they me help them add cure resists to mobs in gusgen so ppl would stop trying to abuse it there with a lv sync alt.

PR #1001, opened 2015-02-27
Attempt to make nation mission 5-1 less retarded and more like retail. ...

TeoTwawki:
...Not supposed to have all 7 mobs on you at battle start. See issue #918

Pop timing may need adjusted, I have gone with 30 seconds. Neither wiki has an exact duration between pops specified. This does not address everything I brought up in issue #918 just the skeleton repops and initial agro state.


teschnei:
pretty sure you dont need to script mobs to link in bcnms anymore, it should be done automatically

TeoTwawki:
then I can replace spawnmob(id):updateemnity with just spawnmob(id) and make sure that the warriors do not start pre-popped. Amending.

TeoTwawki:
Updated and fully tested even multi boxed in more characters to fix the 2nd and 3rd battlefields that had mobs and players not arriving in same battlefield.

TeoTwawki:
The other mobs didn't need scripts at all, so I deleted them since i have everything handled in the Archlich's script (and sorcerers already link to him). The warriors I had to updateEmnity() in their spawn command though or they'd stand there doing nothing.


https://github.com/DarkstarProject/darkstar/pull/1001/files?w=1

Issue #297, opened 2014-06-05
Telling zoning apart from log out/in


TeoTwawki:
There are some things on retail where you can log out and not lose progress or an effect, but will lose them on zone. Example include the the sheep's meat places in the campfire in South Gustaberg and the recently added Prowess effect.

The script in South Gustaberg makes use of needToZone() and EFFECT_PROWESS uses a flag in its database entry. When I asked about it on IRC Kj said he thinks needToZone() does it wrong then. I do not know how to change this behavior to be more retail, so currently prowess buffs wear off on logout/connection loss and will cause you to lose your progress on that galkan sausage quest if you were doing that. 

This likely effects or will effect other things these are just two examples that come to mind.

maxtherabbit:
is this still a thing?

TeoTwawki:
Yes. needToZone() stipp has its state changed on relogging into same zone and shoudl not. Along with my failure to work aroudn this issue for buffs that need to be removed *only* on zoning, but are stripped by logout.

xdemolish:
https://github.com/DarkstarProject/darkstar/commit/203a30332312803e4b9fb8f7504a88b81bf3a50d

xdemolish:
0 duration effects is another issue, which i will probably fix later (once i come up with a non-ugly way to do it)

TeoTwawki:
Thanks demo.

PR #1012, opened 2015-03-02
Made GoV prowess buff behave more retail (only lost on zoning out, not from dc or even logoff).

TeoTwawki:


TeoTwawki:
(Apologies for repeat, I engrished the hell out of initial comment).
I think we still need the behavior of needToZone() changed to not trigger on re-logging or getting dc'd, and then I think issue #297 can be closed (although the flag for zone still also trips on logging out or dc, really should be 2 diff flags).

The places needToZone() is already used at will be or are perfectly retail if this is done. Unless that was already changed and I didn't notice, we're already all set. Last I knew it was still triggering on entering game.

teschnei:
Now that the server can differentiate logging in and zoning again, it
shouldn't be any work at all (just have to figure out how needtozone works
in the first place)

TeoTwawki:
@teschnei related: I was just helping gedads test something, and found out that if trip cs during login that retail only has happen on zone in, you crash out the client. So I guess we need to preven cs in onZoneIn from happening when its a login as well.

TeoTwawki:
ok..Shits broken that I did not touch in this pull request, but make sthsi pull request 100% useless. Not one word about testing. There is no way this should have failed and a simple flag change shouldn't need to be tested rigorously: I set the effect flags so that it would NOT be removed form zoning, since that trips on login, and then manually removed the effects in player.lua ONLY when an actual zone occurs, since that is the only way to tell them apart right now. That should have worked fine. Except the #$%^&ing server is STILL removing effects that are not flagged to be removed.

PR #1015, opened 2015-03-03
Added validation to ensure that the lobby will only allow new characters...

joncloud:
... to start using the starting jobs.


teschnei:
use dsp_max/dsp_min instead of max/min (or better yet, dsp_cap)

PR #1026, opened 2015-03-04
Replace a few hard-coded integers with their enum names

Bapabooiee:
To anyone curious, the respective enums can be found in `basic_message.h` and `battleentities.h`.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -4123,7 +4123,7 @@ inline int32 CLuaBaseEntity::canUseCostume(lua_State *L)
         lua_pushinteger(L, 445);
         return 1;
     }
-    lua_pushinteger(L, (m_PBaseEntity->loc.zone->CanUseMisc(MISC_COSTUME) ? 0 : 316));
+    lua_pushinteger(L, (m_PBaseEntity->loc.zone->CanUseMisc(MISC_COSTUME) ? 0 : MSGBASIC_CANT_BE_USED_IN_AREA)); //316
     return 1;
 }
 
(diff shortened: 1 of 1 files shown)


src/map/ai/ai_mob_dummy.cpp
```diff
@@ -894,7 +894,7 @@ void CAIMobDummy::ActionAbilityStart()
 		Action.speceffect = SPECEFFECT_HIT;
 		Action.animation  = 0;
 		Action.param	  = m_PMobSkill->getMsgForAction();
-        Action.messageID = m_PMobSkill->getMsg() == 0 ? 0 : 43; //readies message
+        Action.messageID = m_PMobSkill->getMsg() == 0 ? 0 : SKILLTYPE::SKILL_BLU; //readies message (43)
```
teschnei: i think you meant something else here? skill types aren't related in any way to message IDs

Bapabooiee:
I couldn't find any kind of MSG's that looked like 43 appropriately mapped to, so I assumed that `SKILL_BLU` from `battleentities.h` was what it was referring to. But I'll defer to you and assume it's incorrect, since I'm not familiar with the code yet.

I also removed the `MSGBASIC_ID::` prefix because other parts of the codebase don't use it when referring to messages anyway.

Bapabooiee:
I couldn't find any kind of MSG's that looked like 43 appropriately mapped to, so I assumed that `SKILL_BLU` from `battleentities.h` was what it was referring to. But I'll defer to you and assume it's incorrect, since I'm not familiar with the code yet.

I also removed the `MSGBASIC_ID::` prefix because other parts of the codebase don't use it when referring to messages anyway.

starlightknight:
There are multiple enum sequences that  can have the same values - just cause that one is also 43 doesnt mean its right descriptively. See status.lua for a good reference. 

Not all possible values are enumerated yet. Some things in status.lua dont have a backend enumeration either, but you could make one that matches if you want to do code cleanup for readability. For things that don't have an enumeration you could add one if you make the name correctly descriptive

teschnei:
usually for messages, they don't get enumerated until someone uses them - there's like 500 of them and most of them won't be used by their enumerated name anyways, but there's no issue with adding one

teschnei:
you can update that branch and open the PR again if you choose to do that

PR #1029, opened 2015-03-05
AF+1 fixes

RunAwayDSP:
Fixed DNC AF+1  for Male and Female versions
Fixed Bug where ABC would not be subtracted on AF+1 upgrade


RunAwayDSP:
now includes the brygid fix that was reverted earlier

PR #1038, opened 2015-03-08
WotG scripts relating to joining the Irons Rams and the Sigil granting NPC in Southern San d'Oria [S]

TeoTwawki:
Also fixed the exit in Windurst Waters [S]

I think I was holding off on this because of the parts not yet completed (all commented) but I figured if I don't get what I have already done shared back soon I am going to forget it again for another 6 months because my memory is terrible.

Its all tested, was running on Annwn server for quite awhile and I shared it with the Legion server like a week ago, the bits not commented out all work and are as retail accurate as it gets without working campaign region  control stuffs.

### DIFF
--- a/scripts/globals/campaign.lua
+++ b/scripts/globals/campaign.lua
@@ -10,6 +10,99 @@ PAST_SANDORIA = 5;
 PAST_BASTOK = 6;
 PAST_WINDURST = 7;
 
+-- -------------------------------------------------------------------
+-- getMedalRank()
+-- Returns the numerical Campaign Medal of the player.
+-- -------------------------------------------------------------------
+
+function getMedalRank(player)
+    local rank = 0;
+    local medals =
+    {
+         0x039C, 0x039D, 0x039E, 0x039F, 0x03A0, 0x03A1, 0x03A2,
+         0x03A3, 0x03A4, 0x03A5, 0x03A6, 0x03A7, 0x03A8, 0x03A9,
+         0x03AA, 0x03AB, 0x03AC, 0x03AD, 0x03AE, 0x03AF
+    }
+    while (player:hasKeyItem(medals[rank + 1]) == true) do
+        rank = rank + 1;
+    end;
+    return rank;
+end;
+
+-- -------------------------------------------------------------------
+-- get[nation]NotesItem()
+-- Returns the item ID and cost of the Allied Notes indexed item
+-- (the same value as that used by the vendor event)
+-- Format:
+-- ListName_AN_item[optionID] = itemID; -- ItemName
+-- ListName_AN_price[optionID] = cost; -- ItemName
+-- -------------------------------------------------------------------
+
+function getSandOriaNotesItem(i)
+    local SandOria_AN =
+    {
+        [2] = {id = 15754, price = 980}, -- Sprinter's Shoes
+        [258] = {id = 5428, price = 10}, -- Scroll of Instant Retrace
+        [514] = {id = 14584, price = 1500}, -- Iron Ram jack coat
+        [770] = {id = 14587, price = 1500}, -- Pilgrim Tunica
+        [1026] = {id = 16172, price = 4500}, -- Iron Ram Shield
+        [1282] = {id = 15841, price = 5000}, -- Recall Ring: Jugner
+        [1538] = {id = 15842, price = 5000}, -- Recall Ring: Pashow
+        [1794] = {id = 15843, price = 5000}, -- Recall Ring: Meriphataud
+        [2050] = {id = 10116, price = 2000} -- Cipher: Valaineral
+    }
+    local item = SandOria_AN[i];
+    return item.id, item.price;
+end;
+
+function getBastokNotesItem(i)
+    local Bastok_AN =
+    {
+        [2] = {id = 15754, price = 980}, -- Sprinter's Shoes
+        [258] = {id = 5428, price = 10}, -- Scroll of Instant Retrace
+        -- [514] = {id = ?, price = ?}, --
+        -- [770] = {id = ?, price = ?}, --
+        -- [1026] = {id = ?, price = ?}, --
+        [1282] = {id = 15841, price = 5000}, -- Recall Ring: Jugner
+        [1538] = {id = 15842, price = 5000}, -- Recall Ring: Pashow
+        [1794] = {id = 15843, price = 5000}, -- Recall Ring: Meriphataud
+        [2050] = {id = 10116, price = 2000} -- Cipher: Valaineral
+    }
+    local item = Bastok_AN[i];
+    return item.id, item.price;
+end;
+
+function getWindurstNotesItem(i)
+    local Windurst_AN =
+    {
+        [2] = {id = 15754, price = 980}, -- Sprinter's Shoes
+        [258] = {id = 5428, price = 10}, -- Scroll of Instant Retrace
+        -- [514] = {id = ?, price = ?}, --
+        -- [770] = {id = ?, price = ?}, --
+        -- [1026] = {id = ?, price = ?}, --
+        [1282] = {id = 15841, price = 5000}, -- Recall Ring: Jugner
+        [1538] = {id = 15842, price = 5000}, -- Recall Ring: Pashow
+        [1794] = {id = 15843, price = 5000}, -- Recall Ring: Meriphataud
+        [2050] = {id = 10116, price = 2000} -- Cipher: Valaineral
+    }
+    local item = Windurst_AN[i];
+    return item.id, item.price;
+end;
+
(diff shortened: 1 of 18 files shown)


scripts/globals/effects/sigil.lua
```diff
@@ -16,6 +28,26 @@ end;
...
+    local MPP = target:getMP()/target:getMaxMP();
+    local power = effect:getPower();
+    local subPower = effect:getSubPower();
+    -- subPower sets % required to trigger regen/refersh.
+    -- Campaign Area's controlled sets subPower at time of effect gain.
+
+    if (target:getCurrentRegion() >= 33 and target:getCurrentRegion() <= 40) then -- Only works in WotG zones
+        if (HPP < subPower) then
+            if (power == 1 or power == 3 or power == 5 or power == 7 or power == 9 or power == 11 or power == 13 or power == 15) then
+                target:addHP(1);
```
teschnei: not quite something i like seeing

scripts/globals/effects/sigil.lua
```diff
@@ -16,6 +30,50 @@ end;
...
+
+        if (MPP < subPower and bit.band(tier, 2) == 0) then
+            if (power == 2 or power == 3 or power == 6 or power == 7 or power == 10 or power == 11 or power >= 14) then
+                target:addMod(MOD_REFRESH, 1);
+                effect:setTier(bit.bor(tier, (2)));
+            end
+        end
+    end
+
+    if ((inWotGzone == false or HPP >= subPower+10) and bit.band(tier, 1) > 0) then
```
TeoTwawki: @teschnei can't for the life of me see where I goofed on setting the bits back to zero.

whasf:
??

teschnei:
i will fix it before it gets merged

TeoTwawki:
hold on, git just chewed up my branch..repairing

TeoTwawki:
Had to rebuild it and then force push. I think i recovered everything. I had screwed up a rebase attempt.

scripts/globals/campaign.lua
```diff
@@ -10,6 +10,133 @@ PAST_SANDORIA = 5;
...
+-- Windurst_AN_price[514] =
+-- Windurst_AN_price[770] =
+-- Windurst_AN_price[1026] =
+Windurst_AN_price[1282] = 5000;-- Recall Ring: Jugner
+Windurst_AN_price[1538] = 5000;-- Recall Ring: Pashow
+Windurst_AN_price[1794] = 5000;-- Recall Ring: Meriphataud
+Windurst_AN_price[2050] = 2000;-- Cipher: Valaineral
+
+function getWindurstNotesItem(i)
+    return Windurst_AN_item[i],Windurst_AN_price [i];
```
xdemolish: dat rogue [i]

scripts/zones/Garlaige_Citadel_[S]/npcs/Randecque.lua
```diff
@@ -0,0 +1,55 @@
...
+
+function onTrade(player,npc,trade)
+end;
+
+-----------------------------------
+-- onTrigger Action
+-----------------------------------
+
+function onTrigger(player,npc)
+    if (player:getVar("Used_RED_RECOMMENDATION_LETTER") == 1) then
```
Comment (xdemolish) [inline scripts/zones/Garlaige_Citadel_[S]/npcs/Randecque.lua]: can probably mark the keyitem as 'seen' instead of using a separate var for it

scripts/globals/campaign.lua
```diff
@@ -10,6 +10,133 @@ PAST_SANDORIA = 5;
...
+
+SandOria_AN_price[2] = 980; -- Sprinter's Shoes
+SandOria_AN_price[258] = 10; -- Scroll of Instant Retrace
+SandOria_AN_price[514] = 1500; -- Iron Ram jack coat
+SandOria_AN_price[770] = 1500; -- Pilgrim Tunica
+SandOria_AN_price[1026] = 4500;-- Iron Ram Shield
+SandOria_AN_price[1282] = 5000;-- Recall Ring: Jugner
+SandOria_AN_price[1538] = 5000;-- Recall Ring: Pashow
+SandOria_AN_price[1794] = 5000;-- Recall Ring: Meriphataud
+SandOria_AN_price[2050] = 2000;-- Cipher: Valaineral
```
xdemolish: should probably stick these inside the function itself if they're not used anywhere else and do the same with the others (cause readability)

scripts/zones/Windurst_Waters_[S]/npcs/Ajen-Myoojen.lua
```diff
@@ -39,7 +36,9 @@ end;
 -----------------------------------
 
 function onEventFinish(player,csid,option)
-	-- printf("CSID: %u",csid);
-	-- printf("RESULT: %u",option);
-end;
-
+    -- printf("CSID: %u",csid);
+    -- printf("RESULT: %u",option);
+    if (csid == 0x0C8 and option == 1) then
```
Comment (helixhamin) [inline scripts/zones/Windurst_Waters_[S]/npcs/Ajen-Myoojen.lua]: Is this correct? Shouldn't it be 0x00C8, not 0x0C8?

scripts/globals/campaign.lua
TeoTwawki: Exact same thing besieged.lua was doing for Imperial Standing item list :P

scripts/globals/campaign.lua
TeoTwawki: Exact same thing besieged.lua was doing for Imperial Standing item list :P

scripts/globals/campaign.lua
TeoTwawki: If this and the [i] need changed then so does the stuff Kj has in beseiged.lua
..I just copied without really thinking digging into it, just checked if it worked and called it a day. Will fix (both) if needs fixed though.

scripts/zones/Garlaige_Citadel_[S]/npcs/Randecque.lua
Comment (TeoTwawki) [inline scripts/zones/Garlaige_Citadel_[S]/npcs/Randecque.lua]: Nope. "seen" trips if you enter menu and click it, not if its been used. And I could just check quest status either, as its possible to do the quest without the KI at all. I hate dangling vars though (player variables that get set and left forever, never being zeroed out) so I am looking to change this.

..I think perhaps checking campaign allegiance along with having/not having the KI here would be enough. I don't think you can get any more letters once you've joined a nation, but I have to check on that in retail 1st. I don't trust wiki on it. I know you can get all 3 beforehand though and use them each after the fact way later.

Comment (TeoTwawki) [inline scripts/zones/Garlaige_Citadel_[S]/npcs/Randecque.lua]: Before a certain someone shows up to talk about a certain mission: the mission is not required to trigger any of this. Your mission log change is from the fact you can't pass through a maw to reach these NPC's without triggering that first mission, these NPCs have nothing to do with it and require no change. **Servers allowing players to use the maw before missions exist are the fault of the inconstancy, not the NPC scripts.**

scripts/zones/Windurst_Waters_[S]/npcs/Ajen-Myoojen.lua
Comment (TeoTwawki) [inline scripts/zones/Windurst_Waters_[S]/npcs/Ajen-Myoojen.lua]: Either one works. Even 0xC8 would work.

TeoTwawki:
@teschnei I had a better idea about the regen/refresh effect I want to run by you on IRC later.

TeoTwawki:
@teschnei apologies for the million emails you just got.

xdemolish:
it's ok he has nothing better to do during weekdays anyways (even though it's saturday today, but still)

scripts/globals/campaign.lua
```diff
@@ -10,6 +10,133 @@ PAST_SANDORIA = 5;
...
+
+function getSandOriaNotesItem(i)
+    return SandOria_AN_item[i],SandOria_AN_price [i];
+end;
+
+local Bastok_AN_item = {}
+
+Bastok_AN_item[2] = 15754; -- Sprinter's Shoes
```
teschnei: i think it's easier to understand if you just do 0,1,2.. etc (it's just bit.rshift(option,8)) instead of 2,258, etc

also, combining the price and id in the same table is also a good idea
Bastok_AN_item = { {id=xxx, price=xxx},{  ... etc

TeoTwawki:
@teschnei All set now. Will revisit other part after the I've seen the rest of the option IDs for the event, but did get itemId and itemPrice tables combined like you wanted.

helixhamin:
@TeoTwawki if you prefer to take my pull and include in yours after making sure it is all up to standard that is no problem. I notice it needs to wait fir your pull anyways. The campaign alliance bit fails the process in mine since I don't have that. I had to comment it out to test my submission.

TeoTwawki:
@helixhamin you able to get on IRC atm? I have same name on rizon as I do here

PR #1045, opened 2015-03-09
Additional Effects

Krynasty:
Over 190 Additional Weapon effects. Enjoy.

### DIFF
--- a/scripts/globals/items/bokuto_+1.lua
+++ b/scripts/globals/items/bokuto_+1.lua
@@ -1,0 +1,32 @@
+-----------------------------------------
+-- ID: 16926
+-- Item: bokuto +1
+-- Additional Effect: blind Damage
+-----------------------------------------
+require("scripts/globals/status");
+require("scripts/globals/magic");
+
+-----------------------------------
+-- onAdditionalEffect Action
+-----------------------------------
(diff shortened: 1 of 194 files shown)


scripts/globals/items/bokuto_+1.lua
```diff
@@ -0,0 +1,32 @@
...
+    end
+    if (math.random(0,99) >= chance) then
+        return 0,0,0;
+    else
+        local duration = 25;
+        if (target:getMainLvl() > player:getMainLvl()) then
+            duration = duration - (target:getMainLvl() - player:getMainLvl())
+        end
+        utils.clamp(duration,1,25);
+        --duration = duration * applyResistanceAddEffect(player,target,EFFECT_BLINDNESS,0);
```
maxtherabbit: why did you comment these out?

Krynasty: the reason why some are commented out is because the server will crash upon reading them. i think that most mobs already handle the resistance checks. so reading it twice causes an break in debug. or crash when not on debug.

teschnei: they do not already handle resistance checks

PR #1050, opened 2015-03-09
additional effects

Krynasty:
corrected 106 elemental efffect weapons

PR #1054, opened 2015-03-10
Additional Effects V1.3

Krynasty:
fixed weapons effects calling globals constantly.
fixed sub effect for darkness damage weapons.

### DIFF
--- a/scripts/globals/items/algol.lua
+++ b/scripts/globals/items/algol.lua
@@ -4,7 +4,7 @@
 -- Additional Effect: Fire Damage
 -----------------------------------------
 
-package.loaded["scripts/globals/magic"] = nil;
+
 require("scripts/globals/status");
 require("scripts/globals/magic");
 
--- a/scripts/globals/items/algol.lua
+++ b/scripts/globals/items/algol.lua
@@ -1,0 +1,37 @@
+-----------------------------------------
+-- ID: 18385
+-- Item: Algol
+-- Additional Effect: Fire Damage
+-----------------------------------------
+
+package.loaded["scripts/globals/magic"] = nil;
+require("scripts/globals/status");
+require("scripts/globals/magic");
+
+-----------------------------------
+-- onAdditionalEffect Action
+-----------------------------------
+
+function onAdditionalEffect(player,target,damage)
+    local chance = 10;
+
+    if (math.random(0,99) >= chance) then
+        return 0,0,0;
+    else
+        local dmg = math.random(3,10);
+        local params = {};
+        params.bonusmab = 0;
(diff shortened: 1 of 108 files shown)


nesstea:
wanna do the sirroco dagger while you are at it? it's the wind elemental one.

TeoTwawki:
Like I said on IRC, you don't have to open a new pull request, just pushing to the same branch automatic updates the existing one

TeoTwawki:
@nesstea sirroco kukri has already been in trunk for like a year

TeoTwawki:
Don't forget to check the changes to base dmg and proc% now that you've closed #1050 which was same branch. When you add new stuff, avoid changing old stuff unless you can safely say the change was more retail accurate than the old value.

Krynasty:
ill bear that in mnd. thank you.

maxtherabbit:
so I see you're still changing dmg and proc rates? got sources on that stuff?

also why are there two of these?

Krynasty:
i have closed the other one now. this will be the only one that gets updated.
still getting used to github and stuff. sorry about this.

maxtherabbit:
Status effect resistances must be checked in the weapon scripts. I realise you don't have any in this particular pull rq but regarding the other one you closed: there are no other places that apply resistance for the mob it must be done here. I suggest you look at other additional (status) effect weapon scripts to see the correct way to do the resist chance. I'm not 100% sure how the element is determined for the status effect but I would assume it would be the same as the element of the spell which would normally apply it. @TeoTwawki may know more about that. 

Also I will be shortly modifying applyResistanceAddEffect() to apply specific status resistance bonuses. With this in mind, the correct format for the call is: applyResistanceAddEffect(player,target,ELE_WHATEVER,EFFECT_WHATEVER)

The 4th param is currently a unused param for a bonus, so I ask that any new additional status effect scripts added to the trunk replace the '0' with the appropriate EFFECT_X as of now :D

xdemolish:
@maxtherabbit get modifying

maxtherabbit:
there's no need to me to fix that prior to this getting merged, he can go ahead and make the requested changes anytime now :fast_forward:

TeoTwawki:
Some of the changes were pointed out in the other 2 pull requests he had open, and it got all confusing with the comments scattered among those because of pull requests being closed then reopened and diff one closed so I'm repeating some of what I can remember here:

pre-existing data like dmg amount and proc rate should not be being changed unless the change is for sure making it more retail accurate than the previous values.

xdemolish:
reopen when changes have been made (i keep accidentally checking this pull req thinking it was opened recently)

PR #1061, opened 2015-03-12
Add some missing value to augments table, ...

TeoTwawki:
...and fix some mistakes left in modifier.h and status.lua

NULL mistakenly replaced with nullptr in mod names, and dangling mods you ceased using which hopefully are not still left any scripts now that we have a table for those - I **did** do a quick search of the item scripts, didn't get any results.

### DIFF
--- a/src/map/modifier.h
+++ b/src/map/modifier.h
@@ -51,7 +51,6 @@ enum MODIFIER
     MOD_CHR                       = 0x0E,  // Charisma (modId = 14)
 
     // Elemental Defenses
-
     MOD_FIREDEF                   = 0x0F,  // Fire Defense (modId = 15)
     MOD_ICEDEF                    = 0x10,  // Ice Defense (modId = 16)
     MOD_WINDDEF                   = 0x11,  // Wind Defense (modId = 17)
@@ -76,7 +75,6 @@ enum MODIFIER
     MOD_MEVA                      = 0x1F,  // Magic Evasion (modId = 31)
 
     // Magic Accuracy and Elemental Attacks
-
     MOD_FIREATT                   = 0x20,  // Fire Damage (modId = 32)
     MOD_ICEATT                    = 0x21,  // Ice Damage (modId = 33)
     MOD_WINDATT                   = 0x22,  // Wind Damage (modId = 34)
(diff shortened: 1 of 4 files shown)


src/map/modifier.h
```diff
@@ -526,18 +514,18 @@ enum MODIFIER
...
 
-    //Elemental nullptr Chance
-    MOD_FIRE_nullptr                 = 0x1D3, // (modId = 467)
-    MOD_EARTH_nullptr                = 0x1D4, // (modId = 468)
-    MOD_WATER_nullptr                = 0x1D5, // (modId = 469)
-    MOD_WIND_nullptr                 = 0x1D6, // (modId = 470)
-    MOD_ICE_nullptr                  = 0x1D7, // (modId = 471)
-    MOD_LTNG_nullptr                 = 0x1D8, // (modId = 472)
-    MOD_LIGHT_nullptr                = 0x1D9, // (modId = 473)
-    MOD_DARK_nullptr                 = 0x1DA, // (modId = 474)
```
xdemolish: @teschnei r the best

teschnei:
too bad visual studios replace all symbol isn't until 2015.. Could have
avoided all these nullptr things. Oh well, changing the name of an enum
everywhere isn't gonna break anything

maxtherabbit:
I'm using a table for the null chance in battleutils::MagicDmgTaken so they need to be fixed there as well

teschnei:
surely he attempted do compile after changing enum names. surely!

TeoTwawki:
Surely I would compile after putting back the obviously correct stuff! No not really. I have to much to do to waste my time with the trivial. But I will go fix now that its pointed out, and make sure it works.

TeoTwawki:
"Build succeeded" so there shouldn't be any more to find.

maxtherabbit:
Thanks bro

teschnei:
I mean, if this wasn't an enum defined in a header then I'd agree with you.
But its an enum defined in a header that currently compiled, and all it
takes is 2 seconds to hit f7 to make sure nothing else referenced them.

TeoTwawki:
Its done now anyway, but I see this as code review doing what its supposed to do. No offense meant, I'm just not that good a coder to think of those things as 100% of my learning has come from searching and asking questions - no schooling on any language at all, so the important of it being an enum in a header and thinking beyond "is thing I did **here** correct" is lost on me until things like this happen.

TeoTwawki:
tl;dr
I was being lazy, I was being ignortant.

TeoTwawki:
tl;dr
I wasn't being lazy, I was being ignortant.

teschnei:
I think it's still a good idea to compile if you changed any core files.
Don't be like me, I'm a bad influence

TeoTwawki:
I'll remember to do at least a build if I even touch them from now on.

maxtherabbit:
why'd you take out the pet mods from status.lus?

teschnei:
Because I forgot to in my commit

maxtherabbit:
so it's straight to merge now?

PR #1068, opened 2015-03-13
Added 2hours to Dyna Mobs

nesstea:
Xarc needs more data if the small weapons 2hour.


teschnei:
are you sure they don't 2hour already

nesstea:
Yeah I am sure they don't 2hour.

nesstea:
I need to do this differently. I will figure it out.

Issue #1065, opened 2015-03-13
Retaliation poorly implemented


vickycat:
http://wiki.ffxiclopedia.org/wiki/Retaliation

Issues:
1. You can retaliate attacks from enemies you are not facing.
2. 100% proc rate (basically Hundred Fists against Charybdis even though you'll die. try it on test server it's funny).
3. Procs even when using "defensive job abilities" such as Defender
4. If you take the first hit from an enemy with multiple attacks (such as a triple attack from bunnies), you will retaliate against all following attacks even if they are misses.

teschnei:
3 is correct (you can retaliate with defender up)

TeoTwawki:
1. when i tested I didn't get a single proc with my back to the enemy,

2. I tested that quite a bit and definitely was not seeing anywhere close to 100% proc rate. retaliation is based on hitrate, and there is not even a log message when it doesn't proc. If you go up againast charby at lv99 I wouldn't be surprised to find your hitrate capped.

3. that does not mean what you think it does. It means if another defensive ability (parried the attack) happens, you cannot retaliate.

4. That may be a problem. Didn't think of that.

TeoTwawki:
someone broke something, just went out and retested on latest trunk...I don't understand that, far as I know hasn't been touched since I implemented it.

TeoTwawki:
https://github.com/DarkstarProject/darkstar/blob/cc409d74a1059f2db98efdc6e12cffcd4010a796/src/map/utils/battleutils.cpp#L545

That was working before..

maxtherabbit:
isFaceing(PAttacker->loc.p, PDefender->loc.p, 40)

the arguments are reversed - PAttacker is the mob here

TeoTwawki:
it was working back when I pull requested, and I didn't change it max. And counter, 3rd eye, etc have those arguments in same order.

TeoTwawki:
After testing it, max is right and I've fixed that part. Got several hits in my log with no retaliation proc proving ist not 100%, and saw several instances where it didn't proc on all hits of a multi attack.

maxtherabbit:
any remaining issues with this are attributable to mobs having low evasion, the retaliation implementation is good to go

PR #1070, opened 2015-03-13
Fix only part of issue #1065 that could be verified to exist on latest trunk. ...

TeoTwawki:
... Verified no problems with multi hit melee and that proc rate was not 100%. Did seem a bit high, but acc on DSP in general seems high - which retaliation rate is based on.


teschnei:
the formula for accuracy is quite known, if "acc on DSP in general seems high" then that is probably not very difficult to troubleshoot

TeoTwawki:
prolly, but not my goal here. if I take that on it'd get its own pull request.

maxtherabbit:
Yeah I'll look into it this weekend

PR #1078, opened 2015-03-15
Couple Item Luas

nesstea:
For max to look over


maxtherabbit:
ok 2 things

1) the if (dmg < 0) part needs to go, adjustForTarget() will negate the damage amount if the target absorbs light magic

2) instead of hardcoding item id's for the staves, delete that whole section and use:

dmg = dmg * AffinityBonus(caster, ELE_LIGHT);

maxtherabbit:
disregard the first email, I edited the comment

PR #1081, opened 2015-03-16
Created MOBMOD_NO_DROPS to disable all drops including seals and made negative values of MOBMOD_EXP_BONUS result in exp reduction.

TeoTwawki:
https://github.com/DarkstarProject/darkstar/pull/1081/files?w=1


teschnei:
there's already an exp multiplier mobmod, why not just use that? (it currently ignores amounts that are less than 0, so it would be trivial to change that)

src/map/ai/ai_mob_dummy.cpp
```diff
@@ -408,18 +408,18 @@ void CAIMobDummy::ActionDropItems()
 				blueutils::TryLearningSpells(PChar, m_PMob);
 				m_PMob->m_UsedSkillIds.clear();
 
-				if(m_PMob->m_giveExp)
-				{
-	                charutils::DistributeExperiencePoints(PChar, m_PMob);
-				}
+                if(m_PMob->m_giveExp && m_PMob->getMobMod(MOBMOD_EXP_BONUS) >= 0)
```
teschnei: yeah.. i'm pretty sure this will make every mob that doesn't have this mod give 0 exp

TeoTwawki: is m_PMob->getMobMod(MOBMOD_EXP_BONUS)  null instead of zero when they don't have it?

teschnei: nope.. but even if it was, null is not greater than 0 either (in fact, it is exactly 0)

TeoTwawki: was trying to check that it was zero or greater, under the belief it'd be zero when they don't have it.

TeoTwawki: just double checked...I get exp just fine off the water elemental next to my test subject (a flamingo I set to not give exp via the mod)

TeoTwawki:
@teschnei amended, works like you wanted now. And is tested on multiple mobs with diff amounts of both positive and negative mod, and with no mod set.

@gedads now that I changed it again...this:
```lua
function onMobInitialize(mob)
    mob:setMobMod(MOBMOD_GIL_MAX, -1); -- disables gil
    mob:setMobMod(MOBMOD_EXP_BONUS, -100); -- negative 100% exp
    mob:setMobMod(MOBMOD_NO_DROPS, 1); -- disables all loot, even seals
end;
```
will do what you needed.

ashdam:
In which part of the game is used?

TeoTwawki:
@ashdam gedads said he needed to make certain mobs drop nothing, no gil no exp no seals etc. So I did it.

TeoTwawki:
@xdemolish ty for helpin me with my derp

ashdam:
@teotwawki ohh so it's just for fun.  Hehe, I thought it was used in the
new expansion or something like that.  Long time no play retail :(

teschnei:
Not quite, but used to get rid of gil drops and experience from mobs in
mission BCNMs.

PR #1084, opened 2015-03-16
Small tweak for nation mission 5-1.

TeoTwawki:
(Despawn the other mobs, because BCNM cleanup is so slow that the adds will kill you while cs loads. And sorcerer kill is not a win requirement, only archlich is.)

### DIFF
--- a/sql/bcnm_battlefield.sql
+++ b/sql/bcnm_battlefield.sql
@@ -398,15 +398,30 @@ INSERT INTO `bcnm_battlefield` VALUES ('482', '3', '17608728', '3');
 INSERT INTO `bcnm_battlefield` VALUES ('484', '1', '17608744', '3');
 INSERT INTO `bcnm_battlefield` VALUES ('484', '2', '17608745', '3');
 INSERT INTO `bcnm_battlefield` VALUES ('484', '3', '17608746', '3');
-INSERT INTO `bcnm_battlefield` VALUES ('512', '1', '17620993', '3');
-INSERT INTO `bcnm_battlefield` VALUES ('512', '1', '17620994', '3');
-INSERT INTO `bcnm_battlefield` VALUES ('512', '1', '17620995', '3');
-INSERT INTO `bcnm_battlefield` VALUES ('512', '2', '17621000', '3');
-INSERT INTO `bcnm_battlefield` VALUES ('512', '2', '17621001', '3');
-INSERT INTO `bcnm_battlefield` VALUES ('512', '2', '17621002', '3');
-INSERT INTO `bcnm_battlefield` VALUES ('512', '3', '17621007', '3');
-INSERT INTO `bcnm_battlefield` VALUES ('512', '3', '17621008', '3');
-INSERT INTO `bcnm_battlefield` VALUES ('512', '3', '17621009', '3');
+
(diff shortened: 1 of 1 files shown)


TeoTwawki:
Note: I keep seeing ppl say in irc that this mission is broken. I just ran through entire mission before making this pull, works fine for me not seeing any missing mob issues etc...

teschnei:
sounds like a problem elsewhere... Asked demo or max to look at them not
despawning instantly?

teschnei:
They should, unless im missing some specific case they shouldn't

TeoTwawki:
They depop, its just slow. Massive lag as cs loads.

TeoTwawki:
Running the depop command in onMobDeath happens instantly, so I put it back in (it was removed last time I touched this).

TeoTwawki:
@xdemolish  @maxtherabbit either of you wanna volunteer to make this pull request uneeded?

teschnei:
Yeah, they should depop as soon as the conditions are met. (Also players in
a full cs are supposed to lose aggro but that's something else)

xdemolish:
can't depop if the warriors are spawned outside the battlefield handler!

teschnei:
yeesh, this is gonna be fun. No way to tell if a mobs part of a bcnm either
I don't think... I'd like to do this in a way that doesn't involve changing
the scripts. Maybe check the mobs zone in SpawnMob, see if its a bcnm zone?
I suppose that doesn't really work for bahamut bcnm, or any other stuff
that might use that system (voidwatch maybe?).

gedads:
INSERT INTO `bcnm_battlefield` VALUES ('512', '1', '17620993', '3');
INSERT INTO `bcnm_battlefield` VALUES ('512', '1', '17620994', '3');
INSERT INTO `bcnm_battlefield` VALUES ('512', '1', '17620995', '3');

Maybe it's because other adds if left alive (17620996 ,97,98,99) aren't in the bcnm_battlefield so aren't cleared on win?

TeoTwawki:
I gather whatever is being done to bcnm system is going to take awhile. How about we allow the 4 line fix that works without probs just for now? Can always be removed (again) later.

teschnei:
Figuring out if the bcnm system even needs changes yet or if you just need
to add them to the bcnm table

TeoTwawki:
Can't add the warriors to bcnm table, that would be non retail behavior - they are NOT supposed to be up at start, they are spawned repeatedly by Archlich during fight..

teschnei:
that's why I added that extra column like 2 years ago. you don't see both
phases of shadow lord spawned at the beginning of 5-2 ya?

TeoTwawki:
.........Didn't even know he was 2 IDs, or about this extra column. >.>

I'll test and amend later today.

teschnei:
Its a bit packed number (only 2 bits ATM..), one is for spawned at the
start, the other is for required to be killed for victory. I don't remember
the name, but I assume if you put the skeletons in there as 0 they should
despawn when the bcnm cleans up.

TeoTwawki:
Amended and tested. The whole end of bcnm is still quite slow, but they stop attacking and depop properly so its good. I never knew what that last field in the table did or where it was enum'd till now, just kept what was already there.

PR #1098, opened 2015-03-18
CORE: Foundational work for chocobo digging

Ex0r:


src/map/utils/diggingutils.cpp
```diff
@@ -0,0 +1,550 @@
...
+*	Return the Chocobo Digging delay									*
+*																		*
+************************************************************************/
+
+	int8 ReturnDigDelay(CCharEntity* PChar)
+	{
+		uint8 delay		= 16;								// Base Delay of 16
+		uint8 skillRank = PChar->RealSkills.rank[58];		// Grab Digging Rank
+		switch(skillRank)
+		{
```
xdemolish: should be able to just         delay = dsp_cap(delay - (skillRank * 5), 0, 16); or something instead of using a switch here

src/map/utils/diggingutils.cpp
```diff
@@ -0,0 +1,550 @@
...
+*	Return the Area Chocobo Digging delay								*
+*																		*
+************************************************************************/
+
+	int8 ReturnAreaDigDelay(CCharEntity* PChar)
+	{
+		uint8 delay		= 60;								// Base Delay of 60
+		uint8 skillRank = PChar->RealSkills.rank[58];		// Grab Digging Rank
+		switch(skillRank)
+		{
```
xdemolish: same here, can just    delay = 60 - (skillRank * 5); instead of switch

src/map/utils/diggingutils.cpp
Ex0r: In this instance, the delay is only applied in the first 3 ranks. After that, it no longer applies.

src/map/utils/diggingutils.cpp
```diff
@@ -0,0 +1,550 @@
...
+				(skillRank == 1 && charSkill >= 200) ||
+				(skillRank == 2 && charSkill >= 300) ||
+				(skillRank == 3 && charSkill >= 400) ||
+				(skillRank == 4 && charSkill >= 500) ||
+				(skillRank == 5 && charSkill >= 600) ||
+				(skillRank == 6 && charSkill >= 700) ||
+				(skillRank == 7 && charSkill >= 800) ||
+				(skillRank == 8 && charSkill >= 900) ||
+				(skillRank == 9 && charSkill >= 1000) ||
+				(skillRank == 10 && charSkill >= 1100))
```
xdemolish: can probably be shortened to use a for loop from 0 to 10, checking if charSkll >= ((skillRank * 100) + 100) and then increment rank and exit the loop

Ex0r: I agree with this, I don't know why I didn't think of this when I was building it.

src/map/utils/diggingutils.cpp
```diff
@@ -0,0 +1,490 @@
...
+				else
+				{
+					// Beat the Dig chance but not the Item chance
+					PChar->pushPacket(new CMessageTextPacket(PChar, MessageOffset2));
+				}
+			}
+			CalculateSkillUp(PChar);
+		}
+		UpdatePlayerDigCount(PChar);
+		SetLastDig(PChar);
```
xdemolish: stick these two in the same function (UpdatePlayerDigCount), not really a need to separate them since they're getting updated at the same time

src/map/utils/diggingutils.cpp
```diff
@@ -0,0 +1,490 @@
...
+/************************************************************************
+*																		*
+*	Check if the character has the Egg Helm equipped					*
+*																		*
+************************************************************************/
+
+	bool EggHelm(CCharEntity* PChar)
+	{
+		CItemArmor* PHead = (CItemArmor*)PChar->getStorage(LOC_INVENTORY)->GetItem(PChar->equip[SLOT_HEAD]);
+		if(PHead->getID() == 16109)
```
xdemolish: can move these to a bool in the main dig function instead of creating a new one for a one line check

Ex0r: CanPlayerBore and CanPlayerBurrow were seperated because 'eventually', when chocobo raising is implemented, chocobos are able to learn an ability, either burrow or bore (but not both), and this will check to see if the chocobo knows either one when digging.

src/map/utils/diggingutils.h
```diff
@@ -0,0 +1,90 @@
...
+
+/************************************************************************
+*																		*
+*	Enumerators Digging Requirements									*
+*																		*
+************************************************************************/
+
+enum REQUIREMENTS
+{
+	NONE		= 0,
```
xdemolish: REQUIREMENTS_ cause ambiguous, same with QUALITY enums

src/map/packet_system.cpp
```diff
@@ -785,14 +786,25 @@ void SmallPacket0x01A(map_session_data_t* session, CCharEntity* PChar, int8* dat
...
+			bool canDig = diggingutils::CanPlayerDig(PChar);
+
+			if (canDig)
+			{
+				charutils::UpdateItem(PChar, LOC_INVENTORY, slotID, -1);
+
+				PChar->pushPacket(new CInventoryFinishPacket());
+				PChar->pushPacket(new CChocoboDiggingPacket(PChar));
+				diggingutils::Dig(PChar);
+				Sleep(1000);
```
xdemolish: don't think this is needed

Ex0r: I can't see what you are referring to. The sleep? If it's not in there the animation for digging plays back to back when you dig (even while it's still calculating)

teschnei: absolutely not - you don't want to pause the entire server for a second every time someone digs
also, Sleep is a windows function and wouldn't compile on linux

src/map/utils/diggingutils.cpp
```diff
@@ -0,0 +1,490 @@
...
+*																		*
+*	Chocobo Digging Utilities											*
+*																		*
+************************************************************************/
+
+namespace diggingutils
+{
+
+	// Get message offset for  current zone
+	uint16 m_MessageOffset[3][MAX_ZONEID];
```
teschnei: im sure you dont need this for *every* zone.. it's probably better to make this a std::map of tuples (std::map<uint16, std::tuple<uint16, uint16, uint16> > )

src/map/utils/diggingutils.cpp
```diff
@@ -0,0 +1,490 @@
...
+		int8 Chance = WELL512::irand() % 100;
+
+		if (Chance <= 15)	// Check if we lose chance
+		{
+			PChar->pushPacket(new CMessageTextPacket(PChar, MessageOffset2));
+			CalculateSkillUp(PChar);
+		}
+		else				// Calculate item received and display message
+		{
+			int32 ItemID = 4545; // Temporary item for testing purposes, will give Gysahl Greens (to be removed before committing)
```
teschnei: you didn't remove before committing yet :)

src/map/utils/diggingutils.cpp
```diff
@@ -0,0 +1,490 @@
...
+*																		*
+*	Return the character Zone In time									*
+*																		*
+************************************************************************/
+
+	int32 GetZoneInTime(CCharEntity* PChar)
+	{
+		int32 value = 0;
+
+		const int8* fmtQuery = "SELECT value FROM char_vars WHERE charid = %u AND varname = 'zoneintime' LIMIT 1;";
```
teschnei: you certainly don't need to use the database for this (it is exactly the opposite of needing the db, so use localvars)

Ex0r: I think demo tried to do this, and the local vars reset on zone?

src/map/utils/diggingutils.cpp
teschnei: if you have to check a specific item ID in the core, you should have made a modifier

src/map/utils/diggingutils.cpp
xdemolish: database usage was intentional to prevent exploits e.g. zoning/relogging

src/map/utils/diggingutils.h
```diff
@@ -0,0 +1,90 @@
...
+
+#include "../../common/cbasetypes.h"
+
+/************************************************************************
+*																		*
+*	Enumerators Digging Requirements									*
+*																		*
+************************************************************************/
+
+enum REQUIREMENTS
```
teschnei: change the name to make this less ambiguous (remember that these enums are in global space, so it should be obvious what the type is referring to - for example, spell requirements use SPELLREQ)

src/map/utils/itemutils.cpp
```diff
@@ -30,6 +30,7 @@
 CItem *		g_pItemList[MAX_ITEMID];    // глобальный массив указателей на игровые предметы
 DropList_t* g_pDropList[MAX_DROPID];    // глобальный массив списков выпадающих предметов
 LootList_t* g_pLootList[MAX_LOOTID];
+DigList_t* g_pDigList[MAX_ZONEID]; // Diglist
```
teschnei: again, not every zone needs a diglist, so there's no reason to use all that memory for each one

src/map/utils/itemutils.cpp
```diff
@@ -236,9 +237,25 @@ namespace itemutils
 		     return g_pDropList[DropID];
 	    }
         ShowWarning(CL_CYAN"DropID %u too big\n" CL_RESET, DropID);
-	    return nullptr;
+	    return NULL;
```
teschnei: change that back to nullptr

src/map/entities/battleentity.h
```diff
@@ -143,7 +143,8 @@ enum SKILLTYPE
 	SKILL_ALC			= 55,
 	SKILL_COK			= 56,
 	SKILL_SYN			= 57,
-	SKILL_RID			= 58
+	SKILL_RID			= 58,
+	SKILL_DIG			= 59 // Chocobo digging
```
TeoTwawki: I suppose it doesn't really matter since neither display in game, but you sure those aren't really the same skill? 

I remember ages ago (early 2004 ish it got changed) if you opened the menu fast enough you could see hidden skills like riding and goldfish scooping (and several "blank" slots between them) and then they'd get hidden, back before SE patched the client to not have those menu text strings anymore. IIRC, choco skill was the very last menu entry back when it existed.

Ex0r: I'm not 100% positive that it's NOT the same, but demo looked it up in polutils earlier and said it was 59.

TeoTwawki: Cools, was just curious.

teschnei: don't think either of them are in polutils (they have no name, so why would they be in the client?), but i don't think it really matters.. i would have used riding because I don't think there is actually a skill for riding anyways

src/map/modifier.h
```diff
@@ -553,6 +552,9 @@ enum MODIFIER
     MOD_HARVESTING_RESULT         = 0x201, // Improves harvesting results (modId = 513)
     MOD_LOGGING_RESULT            = 0x202, // Improves logging results (modId = 514)
     MOD_MINNING_RESULT            = 0x203, // Improves mining results (modId = 515)
+	MOD_ABSORB_DMG_TO_MP		  = 0x204, // (modId = 516)  Unlike PLD gear mod, works on all damage types (Ethereal Earring)
```
teschnei: put this back where it was (they are not intended to be sorted)

src/map/utils/diggingutils.cpp
teschnei: zoneintime has nothing to do with getting more items - it, by definition, resets when you zone.  there is no reason to have it in the database

src/map/modifier.h
Ex0r: I moved it to be consistent in the file. I made a new modifier just to find out 204 was taken somewhere else. I'll move it back to where it was at.

teschnei:
all in all, i like where it started, however, i really think you should consider moving the deciding of items into scripts - you don't need the diglists in the core, you won't need any db tables for it, you won't have a bunch of extra shit for zones that you can't dig in anyways, it would be (in my opinion) all around cleaner if you put the deciding of items in the zones Zone.lua

Ex0r:
I will get with you later tonight with the best method to accomplish that, and also hopefully sort out a formula for dig chance and conditional. Thanks for the input.

maxtherabbit:
i agree with kj on the diglists

other than that I have nothing to add, looks like a good start

xdemolish:
moved to lua https://github.com/DarkstarProject/darkstar/pull/1114

PR #1107, opened 2015-03-19
New Weather System (Very similar to retail)

santssoft:
The weather cycle is 2160 Vana'diel days
    
Each day have three types of weather:
    
     * Normal
     * Common
     * Rare

Zones with no weather data (you can add weather info in database of these zones):
     * All zones of the SOA expansion
     * Some zones of the WOTG expansion

SCRIPTS: Implement Weather Reporters NPC's
    
SQL: Set SPAWNTYPE_FOG to mob_groups
SQL: Update zone_weather with weather data

CORE: exposed Vanadiel Time to use by the weather reporters NPC's
CORE: New/Update weather functions:
   	* Load weather from database to std::vector
    	* Initialize weather on server start
    	* Update weather as task in a range of 3-30 minutes (each zone update individually)
CORE: Remove TIME_FOG and logic to spawn fog mobs
CORE: Add logic to spawn fog mobs when weather change to fog
CORE: cleanup and formatting modified files in this pull

To see changes in files without modified spaces use https://github.com/DarkstarProject/darkstar/pull/1107/files?w=1


teschnei:
why did you choose to make a weather .dat instead of using the database?

santssoft:
Because I got the data from a file and keep it considered so much information, but it is better database i'll change

teschnei:
ok, I was just thinking it'd be really hard to read/modify binary data compared to the database, plus everything else is already in the database (which makes it easier when everything is in the same spot)

santssoft:
ok, i'll move to database.

santssoft:
I move all weather data to database

PR #1113, opened 2015-03-21
Curses, Foiled...Again!? - didn't consider a year change.

nesstea:


maxtherabbit:
I know that it was already like that but... why does this crap even involve year?

nesstea:
because it's a clock?

PR #1126, opened 2015-03-22
Updated wyvern healing breath to include party members as potential targets

joncloud:


teschnei:
i don't think you needed a switch statement for that

PR #1128, opened 2015-03-22
Adjust some annoying default settings. Sick of changing them every time I have to test something they effect.

TeoTwawki:
Also made some commenting clearer and removed reference to something that does not exist in trunk.
https://github.com/DarkstarProject/darkstar/pull/1128/files?w=1


TeoTwawki:
Servers that have these 2 settings in the opposite position are choosing to be non retail, thats fine, but they should have the burden of changing their setting (all of one whole time) rather then me every time I make a new branch.

PR #1133, opened 2015-03-23
Implemented stealth shot.

joncloud:
I'm not sure if LowerEnmityByPercent is the right function call to reduce.  It seems like the skill isn't supposed to reduce by a percentage.

### DIFF
--- a/src/map/ai/ai_char_normal.cpp
+++ b/src/map/ai/ai_char_normal.cpp
@@ -1188,21 +1188,6 @@ void CAICharNormal::ActionRangedFinish()
         // only remove detectables
 		m_PChar->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_DETECTABLE);
 
-		// Handle stealth shot.
-		CStatusEffect* stealthShot = m_PChar->StatusEffectContainer->GetStatusEffect(EFFECT_STEALTH_SHOT);
-		if (stealthShot != nullptr) {
-
-			// Make sure that the target is a mob.
-			if (m_PBattleTarget->objtype == TYPE_MOB) {
-				uint16 power = stealthShot->GetPower();
-
-				((CMobEntity*)m_PBattleSubTarget)->PEnmityContainer->LowerEnmityByPercent(m_PChar, power, nullptr);
-			}
-
-			// Delete the status effect.
-			m_PChar->StatusEffectContainer->DelStatusEffect(EFFECT_STEALTH_SHOT);
-		}
-
         // Try to double shot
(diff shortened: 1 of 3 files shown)


src/map/ai/ai_char_normal.cpp
```diff
@@ -1186,7 +1186,22 @@ void CAICharNormal::ActionRangedFinish()
...
+
+		// Handle stealth shot.
+		CStatusEffect* stealthShot = m_PChar->StatusEffectContainer->GetStatusEffect(EFFECT_STEALTH_SHOT);
+		if (stealthShot != nullptr) {
+
+			// Make sure that the target is a mob.
+			if (m_PBattleTarget->objtype == TYPE_MOB) {
+				uint16 power = stealthShot->GetPower();
+
+				((CMobEntity*)m_PBattleSubTarget)->PEnmityContainer->LowerEnmityByPercent(m_PChar, power, nullptr);
```
teschnei: "reduces emnity of your next shot"

this would be reduce total emnity

joncloud: I am not sure if I follow completely.  Are you suggesting adding a comment to describe why the statement is being executed, or something else?

teschnei: sorry,
what you wrote: reduce all enmity by a %
what stealth shot is supposed to do: reduce emnity of next attack

joncloud:
That makes quite a bit more sense.  I probably should have read the ability more thoroughly.

src/map/enmity_container.cpp
```diff
@@ -102,6 +103,27 @@ float CEnmityContainer::CalculateEnmityBonus(CBattleEntity* PEntity){
...
+			action == ACTION_WEAPONSKILL_FINISH) {
+
+			CStatusEffect* stealthShot = PEntity->StatusEffectContainer->GetStatusEffect(EFFECT_STEALTH_SHOT);
+			if (stealthShot != nullptr) {
+
+				// Ranged attacks, and all ranged weapon skills besides Trueflight should get a bonus applied.
+				if (action == ACTION_RANGED_FINISH ||
+					((PEntity->PBattleAI->GetCurrentWeaponSkill()->getType() == 25 ||
+					PEntity->PBattleAI->GetCurrentWeaponSkill()->getType() == 26) &&
+					PEntity->PBattleAI->GetCurrentWeaponSkill()->getID() != 217)) {
```
teschnei: this is a pretty good indication that this should be handled in the script level instead of here (which may not be fully supported yet - in which case it's better to leave this unimplemented until it is)

PR #1152, opened 2015-03-25
Fixed syntax errors and a bad name

nesstea:


xdemolish:
; is optional in lus (except when one-lining multiple things)

Scavenge:
the last end in a function seems to get pretty upset if i don't add one

PR #1160, opened 2015-03-26
Implemented Aggressive Aim.

joncloud:


scripts/globals/effects/aggressor.lua
```diff
@@ -11,6 +11,12 @@ require("scripts/globals/status");
 -----------------------------------
 
 function onEffectGain(target,effect)
+   -- Add Ranged Attack bonus.
+   local racc = effect:getPower();
+   if racc > 0 then
```
teschnei: you don't need to check for greater than zero, just do 
target:addMod(MOD_RACC, effect:getPower());
looks nicer on one line and is functionally no different

teschnei:
by the way, whenever i request a change and you fix it, please come back here and comment that you fixed it - i don't get any notifications unless you comment and i'll probably forget about it for a few days if you

PR #1161, opened 2015-03-26
Relic Iron drops in Dynamis

Joeyjohjjo:
This lets Relic Iron drop in Dynamis.
Regular mob list taken from here: http://www.ffxidb.com/items/1466
NM list taken from ffxiclopedia.
Regular mob drop rate is 8%; NM rate is 20%.


teschnei:
hmm.. none of the mobs on the ffxidb page have a drop rate over 4%, where did the 8% come from?

Joeyjohjjo:
The 8% is consistent with the DSP rate of other dynamis crafting material drops, like Lancewood Logs, Colossal Skulls, and Sparkling Stones. Also, since it takes two Relic Irons to make one Relic Steel, I figured it wouldn't be good for the drop rate to be too low.

PR #1163, opened 2015-03-26
scripts/globals/homepoint.lua - Corrected values

Xaver-Entropia:
Some homepoints were not giving the correct teleport option.

Btw, would a pull request with retail mp cost, cast time and recast time of the spell.sql be acepted or are they not updated on purpose?


teschnei:
they are not updated on purpose until someone also does the formulas for them - because it makes no sense (balance wise) to have current mp costs/cast times and old damage formula

teschnei:
that being said, if the sql work is done, it can be left somewhere (in another sql file, or another branch) until the formulas ever get done

PR #1165, opened 2015-03-26
Implemented some random functions.

Bluezeon:


src/common/dsprand.h
```diff
@@ -32,13 +39,98 @@ class WELL512
         return state[index];
     }
 
-    // generates double floating point numbers in the half-open interval [0, 1)
```
maxtherabbit: stick with using interval notation for the comments, it is actually *more* readable than spelling it out (e.g. between and inclusive zero and the specified max value)

maxtherabbit:
after reading over the comments in your other pull request again, I feel the need to explain something

you may or may not know that most of our contributions come in the form of lua scripts, especially those from inexperienced coders

the vast majority of the game content is scripted (and luajit actually includes its own prng functions with min, max params)

the core is not really something that gets new content added to it regularly, so making the functions within user-friendly to the novice is not super important. that said I do not *oppose* improvements to the core that increase readability, but I can never personally endorse the notion that because a operator / function / whatever is not widely understood, using it is bad practice

all and all I like the idea of adding these wrapper functions

PR #1171, opened 2015-03-27
Implemented Enchainment.

joncloud:


teschnei:
eh.. not really how it should be done. you don't really want people popping enchainment to get tp bonus on all their WS right?
i would check to see where blue magic gets their TP from (hopefully its in one function...) and check for enchainment there

joncloud:
I updated the core blue magic calculation.  It looks like Chain Affinity also did not adjust based on the current caster's TP.  I implemented it using BluefTP like the TODO comment suggested.

scripts/globals/bluemagic.lua
```diff
@@ -101,9 +101,19 @@ function BluePhysicalSpell(caster, target, spell, params)
...
     --print("wsc val is ".. WSC);
 
-    -- TODO: If under CA, replace multiplier with fTP(multiplier, tp150, tp300)
+    local multiplier = params.multiplier;
+    
+    -- If under CA, replace multiplier with fTP(multiplier, tp150, tp300)
+    local chainAffinity = caster:getStatusEffect(EFFECT_CHAIN_AFFINITY);
+    if chainAffinity ~= nil then
+        -- Calculate the total TP available for the fTP multiplier.
+        local tp = caster:getTP() + caster:getMerit(MERIT_ENCHAINMENT);
```
teschnei: just need to make sure tp doesn't go over the maximum and then it's good

joncloud:
I did not catch anywhere that declared a constant, so I put a literal in for capping.

PR #1175, opened 2015-03-28
Corrected Esuna to only remove status effects when the caster's party me...

joncloud:
...mbers within range have the same effects.


teschnei:
i'm having trouble understanding what the problem was, and how it was fixed

joncloud:
From my understanding the spell is supposed to remove a Status Effect so long as the caster and at least one party member within range has it.  It also appears to remove it from all party members that have the Status Effect (including the caster).  The previous revision appeared to only remove it from the target.

teschnei:
it removes it a debuff as long as the caster has it.  and also removes it from all party members in range, but having party members with it is not required

joncloud:
After further testing it looks like the existing code works just fine when removing Status Effects, and cast upon a party member.  When the spell is cast upon the caster it removes the Status Effect, but indicates it has no effect.

A tangential problem is that the spell can be cast on targets other than the caster.  It looks like the spell is intended to be cast on the caster itself.

It looks like the framework is executing the Lua script for each target, but it needs to execute in a specific order (party members first, caster last).  Can you recommend a technique to re-implement this?

teschnei:
why does it need a specific order?

joncloud:
If the first target is the caster, then the effects will be removed prior to the script executing for the other party members.  This will result in the local variable "has" having all false values.

teschnei:
is that not what the has array was for?

PR #1177, opened 2015-03-28
Move Mission check up

Turttle:
Moved Mission check up, before Fenrir's quest to be able to proc the
quest.

### DIFF
--- a/scripts/zones/Windurst_Waters/npcs/Leepe-Hoppe.lua
+++ b/scripts/zones/Windurst_Waters/npcs/Leepe-Hoppe.lua
@@ -29,6 +29,18 @@ function onTrigger(player,npc)
 		elseif(MissionStatus >= 1 and MissionStatus < 3) then
 			player:startEvent(0x008e); -- Keep displaying the instructions
 		end
+	elseif(player:getCurrentMission(WINDURST) == AWAKENING_OF_THE_GODS and player:getVar("MissionStatus") == 0) then
+		player:startEvent(0x02DE);
+	elseif(player:getCurrentMission(WINDURST) == AWAKENING_OF_THE_GODS and player:getVar("MissionStatus") == 1) then
+		player:startEvent(0x02DF);
+	elseif(player:getCurrentMission(WINDURST) == AWAKENING_OF_THE_GODS and player:getVar("MissionStatus") == 2) then
+		player:startEvent(0x02E3);
+	elseif(player:getCurrentMission(WINDURST) == AWAKENING_OF_THE_GODS and player:getVar("MissionStatus") == 5 and player:hasKeyItem(BOOK_OF_THE_GODS)) then
+		player:startEvent(0x02E6);
+	else
+		player:startEvent(0x0159); -- Standard Dialogue?
+	end
+	---------------------------
 	elseif(player:getQuestStatus(WINDURST,FOOD_FOR_THOUGHT) == QUEST_ACCEPTED) then
 		player:startEvent(0x0137);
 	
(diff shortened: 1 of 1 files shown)


scripts/zones/Windurst_Waters/npcs/Leepe-Hoppe.lua
```diff
@@ -29,6 +29,18 @@ function onTrigger(player,npc)
...
+		player:startEvent(0x02DE);
+	elseif(player:getCurrentMission(WINDURST) == AWAKENING_OF_THE_GODS and player:getVar("MissionStatus") == 1) then
+		player:startEvent(0x02DF);
+	elseif(player:getCurrentMission(WINDURST) == AWAKENING_OF_THE_GODS and player:getVar("MissionStatus") == 2) then
+		player:startEvent(0x02E3);
+	elseif(player:getCurrentMission(WINDURST) == AWAKENING_OF_THE_GODS and player:getVar("MissionStatus") == 5 and player:hasKeyItem(BOOK_OF_THE_GODS)) then
+		player:startEvent(0x02E6);
+	else
+		player:startEvent(0x0159); -- Standard Dialogue?
+	end
```
TeoTwawki: "end" in wrong spot for what you are trying to do.

TeoTwawki: Basically everything after that 
```lua
else
    player:startEvent(0x0159); -- Standard Dialogue?
end
```
is going to not exec because 1) the else condition will be met (because it has no conditions) and the entire conditional branch ends at "end" so the next elseif cannot be entered.

Turttle: Just moved up the mission check, at the moment is giving an endless loop of Fenrir's quest.

TeoTwawki: you can't "just move it up" like this. This is a broken script the way you have it.

scripts/zones/Windurst_Waters/npcs/Leepe-Hoppe.lua
```diff
@@ -29,6 +29,18 @@ function onTrigger(player,npc)
...
+	---------------------------
 	elseif(player:getQuestStatus(WINDURST,FOOD_FOR_THOUGHT) == QUEST_ACCEPTED) then
 		player:startEvent(0x0137);
```
TeoTwawki: This here
player:getQuestStatus(WINDURST,FOOD_FOR_THOUGHT
will never be possible to exec as you have it.

scripts/zones/Windurst_Waters/npcs/Leepe-Hoppe.lua
Turttle: ok i will revise it.

scripts/zones/Windurst_Waters/npcs/Leepe-Hoppe.lua
Turttle: I see.  would have to use a case check then?

TeoTwawki: Compare this with what you have, but most importantly, test it out. The mapserver console log will throw out errors with a line number and its best guess about what was wrong, helping you catch and correct these syntax errors.

https://www.privatepaste.com/download/990b5952d7

This is coming form a guy who's been chewed out a few times about not thoroughly testing small stuff :P

Turttle: ok going to close this pull and make a new one, after testing your suggestion it work as intended. Thx!

xdemolish: remember to test and make sure whatever it is you're sharing works before submitting a pull request :)

PR #1180, opened 2015-03-28
The Fighting Fourth questline scripting (Matched to @TeoTwawki pull, ready for pull.)

helixhamin:
This scripts the fighting fourth up until the last step. I have not put in the last step to complete, since you become part of Bastok Campaign division. Also, I (or someone) will need to add in Sandy and Windy ability to do this quest.
Completed fighting fourth. It adds in the rewards, title change, closes the questline, and sets the var for joining the Bastok fighting fourth.

### DIFF
--- a/scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
+++ b/scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
@@ -39,7 +39,7 @@ function onEventFinish(player,csid,option)
 --printf("CSID: %u",csid);
 --printf("RESULT: %u",option);
 
-	if(csid == 0x0066) then
+	if(csid == 0x0066 and player:getQuestStatus(CRYSTAL_WAR,THE_FIGHTING_FOURTH) == QUEST_ACCEPTED) then
 			player:delKeyItem(BATTLE_RATIONS);
 			player:setVar("THE_FIGHTING_FOURTH",1);
 	end		
--- a/scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
+++ b/scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
@@ -39,7 +39,7 @@ function onEventFinish(player,csid,option)
 --printf("CSID: %u",csid);
 --printf("RESULT: %u",option);
 
-	if(csid == 0x0066 and player:getQuestStatus(CRYSTAL_WAR,THE_FIGHTING_FOURTH) == QUEST_ACCEPTED) then
+	if(csid == 0x0066) then
 			player:delKeyItem(BATTLE_RATIONS);
 			player:setVar("THE_FIGHTING_FOURTH",1);
 	end		
--- a/scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
+++ b/scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
@@ -34,6 +34,7 @@ end;
 -----------------------------------
 -- onEventFinish
 -----------------------------------
+
 function onEventFinish(player,csid,option)
 --printf("CSID: %u",csid);
 --printf("RESULT: %u",option);
--- a/scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
+++ b/scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
@@ -1,0 +1,45 @@
+-----------------------------------
+-- Area: North Gustaberg (S) (I-6)
+-- NPC: Gebhardt
+-- Involved in Quests: The Fighting Fourth
+-----------------------------------
+package.loaded["scripts/zones/North_Gustaberg_[S]/TextIDs"] = nil;
+package.loaded["scripts/globals/quests"] = nil;
+-----------------------------------
+
+require("scripts/globals/quests");
+require("scripts/zones/North_Gustaberg_[S]/TextIDs");
+require("scripts/globals/keyitems");
+
+-----------------------------------
+-- onTrade Action
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end;
+
+-----------------------------------
+-- onTrigger Action
+-----------------------------------
+
+function onTrigger(player,npc)
+
+	if (player:getQuestStatus(CRYSTAL_WAR,THE_FIGHTING_FOURTH) == QUEST_ACCEPTED and player:hasKeyItem(917)) == true then
+		player:startEvent(0x0066)
+	else
+		player:startEvent(0x006E)
+	end
+end;
+
+-----------------------------------
+-- onEventFinish
+-----------------------------------
+function onEventFinish(player,csid,option)
+--printf("CSID: %u",csid);
+--printf("RESULT: %u",option);
+
+	if(csid == 0x0066) then
+			player:delKeyItem(BATTLE_RATIONS);
+			player:setVar("THE_FIGHTING_FOURTH",1);
+	end		
+end;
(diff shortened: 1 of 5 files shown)


scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
```diff
@@ -0,0 +1,46 @@
...
+
+-----------------------------------
+-- onEventFinish
+-----------------------------------
+
+function onEventFinish(player,csid,option)
+--printf("CSID: %u",csid);
+--printf("RESULT: %u",option);
+
+	if(csid == 0x0066 and player:getQuestStatus(CRYSTAL_WAR,THE_FIGHTING_FOURTH) == QUEST_ACCEPTED) then
```
Comment (xdemolish) [inline scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua]: dont need to check the quest status again since you've already checked that before starting the event (should probaly check the 'option' they chose is the one which accepts the quest, though

scripts/zones/North_Gustaberg_[S]/npcs/Roderich.lua
```diff
@@ -0,0 +1,46 @@
...
+
+-----------------------------------
+-- onEventFinish
+-----------------------------------
+
+function onEventFinish(player,csid,option)
+--printf("CSID: %u",csid);
+--printf("RESULT: %u",option);
+
+	if(csid == 0x0068 and player:getVar("THE_FIGHTING_FOURTH") == 1) then
```
Comment (xdemolish) [inline scripts/zones/North_Gustaberg_[S]/npcs/Roderich.lua]: same here, except the option (result) doesn't need checking if the result is always the same or /there's no yes/no option in the cutscene

xdemolish:
not merging for now since it can't be completed at this time

scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
Comment (helixhamin) [inline scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua]: Concern here is that, as before, I want to make sure that completion of a stage of the quest cannot be triggered by someone using the cs command.

Comment (xdemolish) [inline scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua]: don't need to add gm safeguards, it's not our concern if someone broke anything by use of gm commands

helixhamin:
xdemolish, question for you. Since I need to finish the quest before merging, I need to try to figure out how to flag memberships.How do we find out how to join a given allegiance? Or do I just need to set a var? Because after this quest, you will become campaign allied to Bastok, and that is where I am having the issues. I can figure out the title, the reward, and the advancement. But, I cannot figure out where to flag nations. I checked one of the developed, like checking to see if you are a member of a nation before talking to a guard in sandy, but that doesn't seem to be the correct way to do it here (Using the 
local guardnation = BASTOK; -- SANDORIA, BASTOK, WINDURST, JEUNO
local guardtype   = 1;      -- 1: city, 2: foreign, 3: outpost, 4: border
)

xdemolish:
prolly need to add a new field in chars/char_profile for campaign allegiance (wait for @teschnei before going any further with this, though)

helixhamin:
Yes, that is another reason I haven't moved to finish the quest. Since it relies on his work, and I only know about his Sandy S work.

gedads:
Nice but wotg mission 1 still can't be made retail

scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
Comment (TeoTwawki) [inline scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua]: Just viewing the cs via GM command shouldn't be able to actually trigger the script here anyway.

Comment (TeoTwawki) [inline scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua]: Only place I've seen that prob is in cs that are scripted in the zone.lua (and I have no idea how/why that happens).

TeoTwawki:
In my own pull req for the San d'Orea version of this I have temp used
```lua
local Allegiance = player:getVar("Campaign_Nation");
-- 0 = none, 1 = San d'Oria Iron Rams, 2 = Bastok Fighting Fourth, 3 = Windurst Cobras
-- Todo: change this player var into a new db field for Campaign Nation Allegiance?
```
Until its given a perm. home alongside regular nation affiliation. Will be asking Kj where he'd like that and amending soon, if any other places (like this bastok one) need same change I'll be happy to take care of that.

helixhamin:
Got it TeoTwawki. I will follow your example so the quest can be finished, because I have others I want to release waiting on this, since I am unsure how to submit separately. I have started adding in the normal npc conversation scripts, so that it seems like they are more interactive. Only focusing on the ones that aren't involved in missions/quests/etc. The little conversations that bring the area more to life, instead of motionless dolls.

scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
Comment (helixhamin) [inline scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua]: Exactly. That is why I added the other check. Then you won't trigger by simply typing (at)cs number. This could also be a problem if the event viewer npc is in the same zone.

Comment (TeoTwawki) [inline scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua]: You misunderstand. The event viewer is an entirely diff script though, its impossible for code in this script to fire there. The at command should also be completely unable to make it happen (but seems to currently happen *only* for zone in cs, which needs fixed).

helixhamin:
@TeoTwawki Quick question on this. How would I remove a temporary stored var in the database? Not sure if I would need to yet for the fighting fourth, since there is also the parallel nation quest stuff that needs to be taken into account, but for cleaning up vars on any other quest I may do, I want to figure out how to do that.

TeoTwawki:
@helixhamin setting a variable's value to zero deletes it.

helixhamin:
@TeoTwawki
I want to check to see if I set the var properly:
				player:setVar("Campaign_Nation",2);
So, I set the nation for 2 based on your feedback:
-- 0 = none, 1 = San d'Oria Iron Rams, 2 = Bastok Fighting Fourth, 3 = Windurst Cobras

Let me know if there is any mistake.

scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua
Comment (helixhamin) [inline scripts/zones/North_Gustaberg_[S]/npcs/Gebhardt.lua]: Okay, got it. So, I can remove those checks? Or should I do them slight different?

TeoTwawki:
@helixhamin after talking with Kj I've now created new commands that work like this:
```lua
player:setCampaignAllegiance()
 -- will return current database value, which are same as what we had in the variable before
```
and
```lua
player:setCampaignAllegiance(nation)
 -- nation is a number, same ones we were using before :)
```
They won't exist till my pull request gets merged though. If yours gets merged first, I'll switch your variable over to the new command and tack that change into mine. :)

TeoTwawki:
Btw if you'd like to, its possible to collapse your commits down into 1. 

I've been just using the amend feature but rebase (or reset with option "mixed" even) can set let you put them all into one commit.

teschnei:
what kind of heathen would pick bastok
guys please, we all know theres no point in scripting anything other than windurst missions

helixhamin:
@TeoTwawki lol, wish I knew how to put it in 1 commit. Also wish I knew how to submit other stuff separately. A month or two ago, and I wouldn't know much about any of this. Still learning to help grow DSP project. 

@teschnei Actually, on retail, my home city was Windy, and always Windy, and my home city for WOTG was Windy [S]. Basically, picked Bastok because it was the easiest at that time.

TeoTwawki:
@helixhamin catch me on IRC when I'm awake, and I can teach. (I'm passing out just now, long night x_X)

helixhamin:
Replaced by #1230

PR #1184, opened 2015-03-29
Fixed changes to npc animation via [at]animatenpc not being visible at time of command use.

TeoTwawki:


teschnei:
Don't think thats right. Animate should go regardless of status. If it
doesn't, it means set animation is missing an update flag. Can you check
the binding

TeoTwawki:
Sure thing.

TeoTwawki:
Its got 
```c++
m_PBaseEntity->updatemask |= UPDATE_HP;
```
Whats it need?

teschnei:
That's weird. what's status have?

TeoTwawki:
```c++
m_PBaseEntity->status = (STATUSTYPE)lua_tointeger(L, 1);
m_PBaseEntity->loc.zone->PushPacket(m_PBaseEntity, CHAR_INRANGE, new CEntityUpdatePacket(m_PBaseEntity, ENTITY_UPDATE, UPDATE_COMBAT));
```

TeoTwawki:
incidentally, I have to send the same setStatus if I change an NPC's pos via script.

teschnei:
Okay.. That's a different problem then. NPC are missing the update in their
AI I think, then

TeoTwawki:
I think I found what I need to change, but will be an hour or 2 before I can test and amend this. I'll let you know.

TeoTwawki:
@teschnei do updatemask lines need to be before pushPacket lines or does that not matter?

teschnei:
They replace CEntityUpdatePackets. Other packets do not matter. We can talk
about it when I'm at my PC next

PR #1185, opened 2015-03-29
Bastok 8.1

Turttle:
Iron Eater typo fix. Was not loading the CS after 
the CS with Zeid in Quick Sand Cave.

### DIFF
--- a/scripts/zones/Metalworks/npcs/Iron_Eater.lua
+++ b/scripts/zones/Metalworks/npcs/Iron_Eater.lua
@@ -45,9 +45,9 @@ function onTrigger(player,npc)
 		player:startEvent(0x02cc);
 	elseif(currentMission == THE_CHAINS_THAT_BIND_US and missionStatus == 0) then
 	    player:startEvent(0x02ff); -- First cutscene of mission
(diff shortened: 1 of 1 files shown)


whasf:
how are those two lines different than line 46?

Turttle:
The statement check was split in to having (statament A) and (statementB) instead of (statementA and statementB)

scripts/zones/Metalworks/npcs/Iron_Eater.lua
```diff
@@ -45,9 +45,9 @@ function onTrigger(player,npc)
 		player:startEvent(0x02cc);
 	elseif(currentMission == THE_CHAINS_THAT_BIND_US and missionStatus == 0) then
 	    player:startEvent(0x02ff); -- First cutscene of mission
-	elseif (currentMission == THE_CHAINS_THAT_BIND_US) and (missionStatus == 2) then
+	elseif (currentMission == THE_CHAINS_THAT_BIND_US and missionStatus == 2) then
 		player:showText(npc, 8596); -- Dialogue after first cutscene
```
TeoTwawki: While its def something I'd change too (original line was bad styling) it wouldn't actually cause any functional difference, so something else had to be causing the problem.

Turttle:
derp... found the issue. Human reasons.

PR #1198, opened 2015-03-30
Packet base class refactoring

z16:
This time with testing! Everything seems to work fine although I'd still appreciate a second pair of eyes on it.

A drop-in replacement for the current packet base class. Allows for specific type casting of its contents as illustrated below while still allowing older code to use it as before, allowing for gradual replacement over time.

#### Accessors
Before:
```c++
    WBUFL(data,(0x04)-4) = PSender->id;
```

Problems:
1. `WBUFB`, `WBUFW`, `WBUFL`, `WBUFF` and `WBUFP` are easy to mix up (especially when adjusting older code)
2. `WBUFx` and `RBUFx` confusion
3. Subtracting -4 to adjust for the header offset

After:
```c++
    ref<uint32>(0x04) = PSender->id;
```
* Problem 1 fixed by explicit type declaration. Type is immediately visible
* Problem 2 fixed by providing only one type of accessor, returns reference to the buffer at the specified position
* Problem 3 fixed by being explicit about the offset, since the header is included in the buffer

#### Header adjustment

Before:
```c++
    this->type = 0x19;
    this->size = 0x7F;
```

Problem: Since `type` and `size` are both `uint8` they cannot contain all the information they need, because the type of a packet is 9 bit long. This results in having to manually convert that information and the above packet `0x119` results in `type = 0x19`, which is confusing and makes searching for it harder.

After:
```c++
    id(0x119);
    size(0x7E)
```

Problem avoided by allowing the full 9 bit to be assigned to.

#### Struct assignment (bidirectional)

Assuming the following server-side struct:
```c++
struct foo
{
    bool a;
    uint8 b;
    uint16 c;
    float d;
} some_foo;
```

Before:
```c++
    WBUFB(data, (0x0A) - 4, some_foo.a);
    WBUFB(data, (0x0B) - 4, some_foo.b);
    WBUFW(data, (0x0C) - 4, some_foo.c);
    WBUFF(data, (0x10) - 4, some_foo.d);
```

Problems
1. Tedious, 4 separate assignments
2. Need to consider alignment issues
3. Not type-safe, since `WBUFB` must be used for both `bool` and `uint8`

After:
```c++
    ref<foo>(0x0A) = some_foo;
```

Avoided all problems while being explicit about what it does. In addition it saves maintenance overhead, because only `foo` needs to be adjusted and it will continue working as expected.


teschnei:
Hurray for explanations :)

maxtherabbit:
this looks good to me

xdemolish:
*cough* https://github.com/DarkstarProject/darkstar/commit/27b916c71e8b20a40e9214332423e21863421b5d#diff-41f727a0cfa74040eaa900d63b19fb61R85

z16:
Nevar 4get.

maxtherabbit:
rip in pain ls item packet 3/9/15 - 3/30/15

PR #1209, opened 2015-04-02
1. Moved some variable out of global and into the only place in entire game that actually uses them.

TeoTwawki:
Btw The sanction NPC's have multiple issues, I'm not touching it nor changing the tables in this pull request, if I do that it'll be its own pull request.

https://github.com/DarkstarProject/darkstar/pull/1209/files?w=1

The change to Sanraku, he'll actually work if the code were to be un-commented out, its just commented out because the camera stuff doesn't exist yet. I think i was the one to briefly break him in a previous commit by local'ing something that was global and I'm just fixing it here by making it not need to be global anymore (it never should have been global'd).

### DIFF
--- a/scripts/globals/besieged.lua
+++ b/scripts/globals/besieged.lua
@@ -19,22 +19,20 @@ NYZUL_ISLE_ASSAULT_POINT = 5;
 -----------------------------------
 
 function hasRunicPortal(player,portal)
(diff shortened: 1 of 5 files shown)


TeoTwawki:
Changed my mind on part of it - now its included.

PR #1219, opened 2015-04-04
Halvier

Turttle:
Mission 2-3 fix, cleanup.

### DIFF
--- a/scripts/zones/Chateau_dOraguille/npcs/Halver.lua
+++ b/scripts/zones/Chateau_dOraguille/npcs/Halver.lua
@@ -39,73 +39,73 @@ function onTrigger(player,npc)
 	local WildcatSandy = player:getVar("WildcatSandy");
 	local MissionStatus = player:getVar("MissionStatus");
 	
+	-- Lure of the Wildcat San d'Oria
 	if(player:getQuestStatus(SANDORIA,LURE_OF_THE_WILDCAT_SAN_D_ORIA) == QUEST_ACCEPTED and player:getMaskBit(WildcatSandy,16) == false) then
 		player:startEvent(0x022e);
-
-elseif(player:getVar("Flagsando") == 1)then
-		if(player:getFreeSlotsCount() == 0) then 
-		player:messageSpecial(ITEM_CANNOT_BE_OBTAINED,181);
-		else
-		player:setVar("Flagsando",0);
-		player:addItem(181);
-		player:messageSpecial(ITEM_OBTAINED,181);
-		end
 	-- Blackmail quest
 	elseif(player:getQuestStatus(SANDORIA, BLACKMAIL) == QUEST_ACCEPTED and player:hasKeyItem(SUSPICIOUS_ENVELOPE)) then
 		player:startEvent(0x0225);
 		player:setVar("BlackMailQuest",1);
-		player:delKeyItem(SUSPICIOUS_ENVELOPE);
-
-elseif(player:hasCompletedMission(SANDORIA,THE_HEIR_TO_THE_LIGHT)) then player:startEvent(0x001f);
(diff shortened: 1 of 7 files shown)


scripts/zones/Chateau_dOraguille/npcs/Halver.lua
```diff
@@ -162,7 +166,7 @@ function onEventFinish(player,csid,option)
 --printf("RESULT: %u",option);
 
 	if(csid == 0x01f5) then
-		player:addMission(BASTOK,THE_EMISSARY_SANDORIA);
+		player:addMission(1, 6);
```
TeoTwawki: Wut.

scripts/zones/Chateau_dOraguille/npcs/Halver.lua
```diff
@@ -187,24 +191,23 @@ function onEventFinish(player,csid,option)
...
-	player:setTitle(295);
-	player:setVar("SandoEpilogue",1);
+			if(player:getFreeSlotsCount() == 0) then 
+				player:messageSpecial(ITEM_CANNOT_BE_OBTAINED,181);
+				player:setVar("Flagsando",1);
+			else	player:addItem(181);
+				player:messageSpecial(ITEM_OBTAINED,181);
+			end
+		player:setVar("MissionStatus",0);
+		player:completeMission(0, 23);
```
TeoTwawki: Oh theres another.
Don't change them from named global vars as defined in missions.lua to being just number please.

Turttle: Ok, that is what I thought, also is what i put 2 part on  the cleanup.

TeoTwawki: Ty for daring to tackle that s.o.b. Halfassver, I mean Halver!

TeoTwawki: Just so you know quests have the same thing going on as missions, but their stuff is defined in quests.lua instead

Turttle:
well i was inclined to use the numbers for the mission since the Halvier.lua uses require("scripts/globals/missions");
 And when calling "player:getCurrentMission()" or"player:getQuestStatus()"  it will look it up on that table. My logic was making the process faster by already having the "index" number for each mission. Would not that make thing better?

TeoTwawki:
The numbers are already translated to text in missions.lua and quest.lua for a reason. As long as the require line is used, we can use human readable text instead of a number that we have to go look up to see which mission/quest this is messing with.

TeoTwawki:
@xdemolish I choose you! *throws pokeball*

xdemolish:
if the mission checks aren't already nested in one that checks the nation, you should probably do so cause getCurrentMission(player:getNation()) would work for all nations instead of just sandy's or w/e

scripts/zones/FeiYin/Zone.lua
```diff
@@ -101,7 +104,7 @@ function onEventFinish(player,csid,option)
         player:addKeyItem(LETTER_FROM_DALZAKK);
         player:messageSpecial(KEYITEM_OBTAINED,LETTER_FROM_DALZAKK);
     elseif (csid == 0x0017) then
-        player:setVar("SANDO92",3);
+        player:setVar("MissionStatus",3);
```
ashdam: The name of the variable isn't a bit generic and create a conflict if someone uses it in case you have several missions active from different expansions?

Turttle: No, this variable is used for the player nation Mission only, it gets resettled every time the lest mission ended. Also you can only accept one Nation mission at the time.

Turttle:
All the mission checks are nested by nations.

Turttle:
Just notice i have 2 pull in 1. Going to put just Halvier in 1 and Mission 9-2 variable name update in another.

PR #1227, opened 2015-04-04
Added Spell Distract

madwraith:


teschnei:
Mark it as SOA in the spell list table. It did not exist prior to adoulin.

madwraith:
k I got it marked as SOA now

PR #1229, opened 2015-04-04
Tombstone CS fix sandy 1-2

Turttle:
This quest is repeatable, if the check contains a Boolean for the
"BathuntComplete" then next time this quest is repeated the player will
be stuck. The MissionStatus check is set to 0.


nesstea:
Thanks for doing that turttle

PR #1235, opened 2015-04-05
new GM command

TeoTwawki:
Prints the ID of the currently selected target under the cursor (mobs npcs and even players, anything). 

This has made ID related bug fixing and working with multiple objects that have the same name go much faster for me. Also saves me having to look up a players charid when I need to find their stuff in a database table.

### DIFF
--- a/scripts/commands/getid.lua
+++ b/scripts/commands/getid.lua
@@ -1,0 +1,20 @@
+---------------------------------------------------------------------------------------------------
+-- func: @getid
+-- auth: TeoTwawki
+-- desc: Prints the ID of the currently selected target under the cursor
+---------------------------------------------------------------------------------------------------
+
+cmdprops =
+{
+    permission = 1,
+    parameters = ""
+};
+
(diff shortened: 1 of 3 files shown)


teschnei:
Nah. Use a plugin, there are already ones for that

TeoTwawki:
whasf already merged it on you. You might find other used for the binding used for the command. I'd rather have stuff used for this sorta thing built in anyway, need to know what the *server* thinks the ID is, not the client, and in the event someone is using just client+darkstar+xiloader they are not going to have any plugins.

teschnei:
It's not possible for them to disagree on ID. I see that he merged it, but
next time look into plugins first for client based info retrieval

TeoTwawki:
Ok look, I get that you think this shouldn't have existed, but I have to disagree even if there is a plugin that will fetch an id of anything without any object type restriction. Not everyone is even using the same launcher anymore, relying on a plugin for a 3rd party app I might not even have is preferable? I don't agree.

PR #1236, opened 2015-04-06
Command gotonpc + add text event to  Estiliphire (South san d'oria)

Chronos77:
gotonpc => teleport to an npc with id


scripts/zones/Southern_San_dOria/npcs/Estiliphire.lua
```diff
@@ -0,0 +1,42 @@
...
+package.loaded["scripts/zones/Southern_San_dOria/TextIDs"] = nil;
+-----------------------------------
+
+-----------------------------------
+-- onTrade Action
+-----------------------------------
+
+function onTrade(player,npc,trade)
+	if (FlyerForRegine == 1) then
+		count = trade:getItemCount();
```
xdemolish: should make this a local var

scripts/commands/gotonpc.lua
```diff
@@ -0,0 +1,25 @@
...
+
+cmdprops =
+{
+    permission = 1,
+    parameters = "s"
+};
+
+function onTrigger(player, target)
+    if (target == nil) then
+        player:PrintToPlayer("You must enter a valid npc id.");
```
TeoTwawki: If I have to know the ID to use this, its not so usefull..The spawn info is litterly right there when you look up the NPC in the db. Its also in the header of almost every npc script in conveniant copy and past form with (at)pos.
https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Bastok_Markets/npcs/Brygid.lua#L7
Should actually be in all NPC's headers but some got forgotten.

Your print after this implies its by name not ID, but the code before that is clearly expecting an ID with the exception of the comand info paramters expectinga  string instead of an integer.

Have you actually tried to use this command in game?

Chronos77: Thanks for your return. I corrected this.

Yes I am currently using because I find it easier to write an id as writing a position no ?
Basically I wanted to do with the name of npc but we have many npc with the same name.

scripts/zones/Southern_San_dOria/npcs/Estiliphire.lua
Chronos77: I corrected that

scripts/commands/gotonpc.lua
TeoTwawki: Yeah its pretty impractical (near impossible from all the caveats involved) to do it by name.

Chronos77:
i corrected filename :
 /scripts/zones/Quicksand_Caves/mobs/Antican_Priceps.lua => /scripts/zones/Quicksand_Caves/mobs/Antican_Princeps.lua.
It's ok for that fix and for all ?

xdemolish:
there's not *really* much benefit to adding this command into the main repo (since the coordinates are listed in npc_list where you find the id anyways)
guess it's useful for warping to npcs used with GetNPCByID(), though

Chronos77:
wait i make new pull request

PR #1248, opened 2015-04-07
Sandy 9 2

Turttle:
Updated Variable name from SANDO92 to MissionStatues for Heir to the
light Sandy 9-2. Added Variable check to the BCNM else if a player with
lower variable enter when a player with the correct variable activates
the Fight, such person will be able to skip previews steps.


Turttle:
for some reason the Bastok and sandy files got mixed up...

PR #1253, opened 2015-04-08
Sando 9 2

Turttle:
Sandy mission 9-2 update. Added a check to verify that the player is on the correct step of the mission before advancing on the mission, after the bcnm win.


Turttle:
Also update variable name from SANDO92 to MissionStatus, on all related scripts.

PR #1228, opened 2015-04-04
Bastok 8-2 fix

Turttle:
Now all mission variable start at 0. The change is required for player to progress.

PR #1265, opened 2015-04-10
Drake Fang Bastok 8-2

Turttle:
Fixed correct CS IDs, Bastok 8-2 patch, script cleanup


Turttle:
Tested all the mission on the scrip are working as intended. Only thing i notice is that event "0x00cc" will leave a black screen, does that means I need to update?

Turttle:
#1228

gedads:
Hi, 
For the black screen i have it too at some point, then it was lost (i added a parameter i think don't remember).
If the blackscreen is there maybe it's because you're transported to bastok mine and the cs is resuming from there but i don't remember too well if you have to walk by yourself outside or you're being warped

PR #1268, opened 2015-04-10
Ix'Aern (MNK) Encounter

nesstea:
Kid tested, mother approved.
Includes the 2hours, Ix'aern immunities, ??? Respawn locations, and Loot adjustments.


scripts/zones/Grand_Palace_of_HuXzoi/mobs/Ix_aern_mnk.lua
```diff
@@ -0,0 +1,82 @@
...
+
+function onMobInitialize(mob)
+end;
+
+-----------------------------------
+-- onMobSpawn Action
+-----------------------------------
+
+function onMobSpawn(mob)
+    local chance = GetServerVariable("[SEA]IxAern_DropRate"); -- Adjust drop rate for the items based on the organs traded.
```
xdemolish: can't this be moved to a localVar? might not *need* to set it as a server variable (would prefer if it wasn't, anyway)

nesstea: The mob won't be spawned for a local var. I've already tried doing that.

xdemolish: tried setting it onMobEngage? could do a check before setting it
can also have the var set in the npc's script instead of the mob's and make use of GetNPCByID to get/set the var
aaand can also add a new mobmod: MOBMOD_DROP_RATE or something

(you don't really *have* to do any of the above, i just hate server vars)

teschnei: It's easy to reset local vars on death instead of spawn

nesstea: Im setting a server var because the ??? spawns the nm, and based on the trade, adjusts the drop rate.

teschnei: A server vars purpose is to save data between restarts. If that's not what
you're doing (it's not) then I will fix local vars for what you need.

nesstea: I'm not. I was using it for temporary storage.

teschnei: Yes, that's what local vars are for :p

nesstea: I know that but the mob doesn't exist yet when I call spawnmob! :octopus:

xdemolish: it does, just isn't spawned to the player

teschnei: Also, everything can have local vars. Even the ??? can!

xdemolish: *cough*
see third comment

nesstea: Why didn't I think of that... I'll make the changes soon.

xdemolish: o and have you checked those mods apply correctly in onMobInitialize?
i don't remember what it was exactly, but there was something like a mobmod or w/e that couldn't be applied to the mob on initialise and had to be applied on spawn instead

though im probably getting it confused with something else

scripts/zones/Grand_Palace_of_HuXzoi/npcs/qm1.lua
```diff
@@ -0,0 +1,83 @@
...
+-----------------------------------
+
+function onTrade(player,npc,trade)
+    local IxAern = 16916815; -- Ix'Aern (MNK). This is the ID base for Ix and his minions. 
+    local chance = 0; -- Rate in percent in which an item will drop.
+    local validTrade = 0;
+    -- Trade Organs
+	if(GetMobAction(IxAern) == 0) then
+        if (trade:hasItemQty(1900,1) and trade:getItemCount() == 1) then -- 1 HQ Aern Organ (33%)
+            chance=33;
```
Turttle: The drop rates are set in a 1000 scale, 33 would translate into 3.3%.

nesstea: Check the ix_aern_mnk.lua. I knew this.

nesstea:
I have Ix'DRG waiting for this merge.

PR #1269, opened 2015-04-10
bugfix: level sync

m241dan:
When the person that was synced to died and homepointed, the level sync was not removed and the party could set their level sync on someone who already had a synced level.


teschnei:
Is this only for home point in the same zone?

m241dan:
Home point from death. I think things like warps and zoning already work.

teschnei:
Are you saying it doesn't remove if you homepoint and your HP is in another
zone?

m241dan:
Ooooooohhh, yes, if you home point is in another zone. I can add a check to for a different zone. The way it is currently, it removed level sync on any home point.

teschnei:
The thing is that home pointing should go through the generic zone code,
which disables level sync (same as hitting the zoneline OE warping). Can
you check why HP is not disabling it there? It'd be preferable to
duplicating the code

m241dan:
Sure, I'll do that.

m241dan:
I'm not sure home pointing uses a zone out mechanic.

From, what I can see, the client sends 0x01A. The server responds by updating some stats, removing weakness(and levelsync which is why I used this section), clears packets and uses the charutils::SendToZone, which doesn't do anything except update the database and pushes a packet back to the client.

Does the client then push another back and call some zone mechanic? I think there is a reason someone put the level sync in this section.

teschnei:
yep, the client will send back a 0x0D (which tells the server the client is out of there)

m241dan:
0x00D :P

xdemolish:
you're like 3 years l8 m8 i r8 2/8

m241dan:
xdemolish: Huh?

Also, the problem with having one code is that the Home Point stuff needs the LEVEL_SYNC effect off that character so it can reset stats and stuff. But, the code that does this, removes the "checks" that trigger the zone out code to remove from the rest of the party.

So, people could home point with the wrong HP or have code in two places.

m241dan:
Specifically the SmallPacket0x01A. When its case is for a homepoint. It removes the level sync affect then so it can set their HP.

With level sync removed, that players m_PRestriction is 0 so when it gets to CharZoneOut(), it doesn't pass the if check to unsync the party.

So, if you don't unsync that player specifically there, the HP/MP/etc(which is set in this homepoint case) would not be accurate but the party will unsync. Or, the way it is now, the HP is correct but the party won't unsync due to the homepoint, just that one member.

Though... on reflection I can fix this by putting the unsync stuff at the delStatusEffect level. Either in the lua script or the DelStatusEffect with a special case. If either of these solutions work for you, let me know which is prefered.

teschnei:
certainly not in the delstatuseffect function as a special case, but the script itself calls disableLevelSync (and that's the only thing that calls it), so it might be a good place to put it.  it might also be possible to move the other sync disabling code into there, but it would be annoying to preserve the correct message to display (as you'd probably not know from the lua binding why the sync was disabled, just that it was)

m241dan:
I haven 't tested that yet.

teschnei:
I think the best solution is to just extract that logic into its own
function and have both zoning and HP call it. There are too many other
small things to bother with a better fix for now.. extract it to the party
class (DisableSyncFromZone perhaps ) and I think that'll be fine

teschnei:
the main problem with the way you did it is that you're disabling sync under message 551, but the function has no idea why level sync was disabled (thus it has no idea which message it should use)

m241dan:
That 551 is just sort of a reckless copy and paste, but yes. I see what you mean. I didn't know it would update my pull request when I pushed this code. First time doing a pull request.

teschnei:
ah yeah, it does that.

it's kind of a pain because of how level sync works: disabling it doesnt remove it, it just puts a 30 second timer on it (and zoning the leader or sync target would do the same thing)

m241dan:
oops, ignore that, I will get to this one day. Extremely busy.

PR #1275, opened 2015-04-11
[SEA] Ix'DRG

nesstea:


whasf:
Only concern I have is if the mob IDs can/will be automatically updated by @gedads script when SE issues a patch? Otherwise that could potentially be some manual work.

xdemolish:
scripts/zones/The_Garden_of_RuHmet/MobIDs.lua
stick the ids in there, require that script, use the variables for id references so you dont have to manually update each one in case there are any id shifts (doubt there are gonna be any in this zone at least)

xdemolish:
AwAernGroups =
redefined vars with exactly the same ids in like 3 different files

PR #1278, opened 2015-04-12
Add in blu pinecone_bomb.lua

helixhamin:
Spell already in sql.

### DIFF
--- a/scripts/globals/spells/bluemagic/pinecone_bomb.lua
+++ b/scripts/globals/spells/bluemagic/pinecone_bomb.lua
@@ -52,16 +52,15 @@ function onSpellCast(caster,target,spell)
     damage = BluePhysicalSpell(caster, target, spell, params);
     damage = BlueFinalAdjustments(caster, target, spell, damage, params);
 
-	if (damage > 0 and target:hasStatusEffect(EFFECT_SLEEP_I) == false) then
-		local typeEffect = EFFECT_SLEEP_I;
+	if (damage > 0) then
 		local tp = player:getTP();
 		local getBlueEffectDuration = (tp/100 * 5);
-        local pStrAgi = caster:getStat(MOD_STR+MOD_AGI);
-        local mStrAgi = target:getStat(MOD_STR+MOD_AGI);
+        local pStrAgi = caster:getStat(MOD_STR) + caster:getStat(MOD_AGI);
+        local mStrAgi = target:getStat(MOD_STR) + target:getStat(MOD_AGI);
         local dStrAgi = (pStrAgi - mStrAgi);
         local resist = applyResistance(caster,spell,target,dStrAgi,BLUE_SKILL,0);
-		target:delStatusEffect(typeEffect);
-		target:addStatusEffect(typeEffect,1,0,getBlueEffectDuration);
+		target:delStatusEffect(EFFECT_SLEEP_I);
+		target:addStatusEffect(EFFECT_SLEEP_I,1,0,getBlueEffectDuration);
 	end
     return EFFECT_SLEEP_I;
 end;
--- a/scripts/globals/spells/bluemagic/pinecone_bomb.lua
+++ b/scripts/globals/spells/bluemagic/pinecone_bomb.lua
@@ -29,7 +29,6 @@ end;
 -- OnSpellCast
 -----------------------------------------
 
-local resist = duration*applyResistance();
 function onSpellCast(caster,target,spell)
     local params = {};
     -- This data should match information on http://wiki.ffxiclopedia.org/wiki/Calculating_Blue_Magic_Damage
--- a/scripts/globals/spells/bluemagic/pinecone_bomb.lua
+++ b/scripts/globals/spells/bluemagic/pinecone_bomb.lua
@@ -1,10 +1,16 @@
 -----------------------------------------
--- Bluemagic: Pinecone Bomb
--- Additional effect: sleep. Duration of effect varies with TP.
+-- Spell: Pinecone Bomb
+-- Additional effect: sleep. Duration of effect varies with TP
+-- Spell cost: 48 MP
 -- Monster Type: Plantoids
 -- Spell Type: Physical (Piercing)
+-- Blue Magic Points: 2
 -- Stat Bonus: STR+1
--- lvl: 36 MP Cost: 48 Blue Points: 2
+-- Level: 36
+-- Casting Time: 2.5 seconds
+-- Recast Time: 26.5 seconds
+-- Skillchain property: Liquefaction (can open Scission or Fusion)
+-- Combos: None
 -----------------------------------------
 
 require("scripts/globals/magic");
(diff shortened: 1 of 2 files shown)


scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,44 @@
...
+	params.numhits = 1;
+	params.multiplier = 2.25; params.tp150 = 2.25; params.tp300 = 2.25; params.azuretp = 2.25; params.duppercap = 37;
+	params.str_wsc = 0.2; params.dex_wsc = 0.0; params.vit_wsc = 0.0; params.agi_wsc = 0.2; params.int_wsc = 0.0; params.mnd_wsc = 0.0; params.chr_wsc = 0.0;
+    damage = BluePhysicalSpell(caster, target, spell, params);
+    damage = BlueFinalAdjustments(caster, target, spell, damage, params);
+
+	if (damage > 0 and target:hasStatusEffect(EFFECT_SLEEP_I) == false) then
+		local typeEffect = EFFECT_SLEEP_I;
+		local tp = player:getTP();
+		local getBlueEffectDuration = (tp/100 * 5); -- Unsure if this is correct, based on shockwave calc for estimation
```
teschnei: this is not really what you were probably supposed to be doing

also, i'm not sure if the sleep will actually apply because the spell deals damage

helixhamin: Sorry, not sure which one you meant that this is not what I am supposed to be doing. You mean the status effect sleep part?

Should I move damage to before the the second if statement?

helixhamin:
After discussing with Teo, he pointed out that in bluemagic.lua, line 221: function BlueFinalAdjustments(caster, target, spell, dmg, params)
Should apply damage before applying effect, meaning the mob will take damage, then will have the sleep effect applied at the end by the return function.

teschnei:
There is more to it than just that.  Needs a little bit of work in the core
as well. Not too much though if the damage is applied in the script

scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,67 @@
...
+        params.mnd_wsc = 0.0;
+        params.chr_wsc = 0.0;
+    damage = BluePhysicalSpell(caster, target, spell, params);
+    damage = BlueFinalAdjustments(caster, target, spell, damage, params);
+
+	if (damage > 0 and target:hasStatusEffect(EFFECT_SLEEP_I) == false) then
+		local typeEffect = EFFECT_SLEEP_I;
+		local tp = player:getTP();
+		local getBlueEffectDuration = (tp/100 * 5);
+        local pStrAgi = caster:getStat(MOD_STR+MOD_AGI);
```
TeoTwawki: I think you meant to do
```lua
local pStrAgi = caster:getStat(MOD_STR) + caster:getStat(MOD_AGI);
```
There.

helixhamin:
Thanks for the help on the math Teo. I fixed this problem. Wasn't aware I needed to split, but know now.

TeoTwawki:
@helixhamin was because getStat() checks a single ID and returns an amount, so adding 2 ID's together would have had it checking the single ID of whatever value resulted from that addition.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,67 @@
...
+
+	if (damage > 0 and target:hasStatusEffect(EFFECT_SLEEP_I) == false) then
+		local typeEffect = EFFECT_SLEEP_I;
+		local tp = player:getTP();
+		local getBlueEffectDuration = (tp/100 * 5);
+        local pStrAgi = caster:getStat(MOD_STR) + caster:getStat(MOD_AGI);
+        local mStrAgi = target:getStat(MOD_STR) + target:getStat(MOD_AGI);
+        local dStrAgi = (pStrAgi - mStrAgi);
+        local resist = applyResistance(caster,spell,target,dStrAgi,BLUE_SKILL,0);
+		target:delStatusEffect(typeEffect);
```
TeoTwawki: Just noticed above you have the conditional check for if they do not have sleep, but you are trying to delete sleep here before adding it. 

Also, mostly cosmetic but: defined typeEffect as EFFECT_SLEEP_I after using that already, could move that local var up and then have your check use it **or** could just use EFFECT_SLEEP_I in place of typeEffect (whichever, doesn't really matter) just to be consistent throughout.

helixhamin: Hmm, on the wiki: Because the Sleep effect is applied after the damage is dealt, it is possible to overwrite any Sleep effect, including its own.
So, guess I need to remove that target conditional, correct?
And yes, it seems like a better idea to just use EFFECT_SLEEP_I throughout.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,66 @@
...
+
+function onMagicCastingCheck(caster,target,spell)
+    return 0;
+end;
+
+-----------------------------------------
+-- OnSpellCast
+-----------------------------------------
+
+local resist = duration*applyResistance();
```
bendangelo: Why is this outside the function?

scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,66 @@
...
+    damage = BlueFinalAdjustments(caster, target, spell, damage, params);
+
+	if (damage > 0) then
+		local tp = player:getTP();
+		local getBlueEffectDuration = (tp/100 * 5);
+        local pStrAgi = caster:getStat(MOD_STR) + caster:getStat(MOD_AGI);
+        local mStrAgi = target:getStat(MOD_STR) + target:getStat(MOD_AGI);
+        local dStrAgi = (pStrAgi - mStrAgi);
+        local resist = applyResistance(caster,spell,target,dStrAgi,BLUE_SKILL,0);
+		target:delStatusEffect(EFFECT_SLEEP_I);
```
bendangelo: This should remove all sleep effects, lullaby, etc. I thought we had a wakeup method?

teschnei: We do. But the script really doesn't matter right now because it can still
never function properly until someone changes the point of sleep removal
for spells to be during the script instead of after

scripts/globals/spells/bluemagic/pinecone_bomb.lua
helixhamin: Because I am a noob that barely understands what I  am doing. Seems to be a relic from whatever spell I copied the code from. There is also a resist inside. I will delete when I get home.

whasf:
So what are we doing with this

teschnei:
It can't work until there are some core changes, so nothing for now

helixhamin:
lol, well, I am not dead and gone yet, so just patiently waiting on this while I play around with other stuff xD
Or get time to actually test my implementation to see if it works (Without increasing my character's BLU to the appropriate level and giving myself the spell...still want to play it somewhat official like.)

teschnei:
we'll revisit this once there is a way to sleep targets with a damaging spell/ability

Handbananax:
Have there been anymore advancements towards this?

nesstea:
@teschnei and for weaponskills too, like the greatsword aoe sleep WS. forgot name

RwNigma:
Kartstrahl.lua Would be a good place to start. It deals damage and you stay slept.

teschnei:
unfortunately kartstrahl only works because the mobskills themselves deal
the damage, whereas spells and weaponskills return the damage for the core
to handle

the scripts should be the ones dealing the damage (not the core), but it'll
be a bit of work to move it all to the scripts (probably requiring changing
every weaponskill and spell script)

PR #1279, opened 2015-04-12
Fixed Dyna var being set on enter

nesstea:


TeoTwawki:
@dr01d3r nice catch

dr01d3r:
Glad to contribute.  Bigger projects to come soon. :)

PR #1285, opened 2015-04-12
Updated wyvern types for Geomancer and Runefencer.

joncloud:


whasf:
GEO and RUN can summon wyverns?

joncloud:
No, but they can be the support job for a Dragoon.

teschnei:
tabs -> spaces

joncloud:
Fixed - Is there a preference, or should it just be consistent with the code around the scope?

teschnei:
If you put spaces in a tabbed file I'd be okay with it (since spaces are
the preference). So I suppose that makes spaces preferred in all situations

teschnei:
all files will be spaces eventually, I swear it! (just not right now because I don't want to convert it all)

PR #1290, opened 2015-04-13
[SEA] ??? having a 0 y coord prevents it from spawning.

nesstea:


teschnei:
it only doesn't spawn if all three coordinates are 0

PR #1294, opened 2015-04-14
Added columns to persist pet zoning information to allow pets to follow ...

### LINKED ISSUE #1286: Pets do not respawn when zoning.
packet_system.cpp SmallPacket0x00C does appears to handle respawning pets upon zoning, however the original PChar gets deleted, and the current PChar is a new instance thus petZoningInfo.respawnPet is false.

Is there a specific technique that the PetInfo_t data should be persisted with?  It seems like it would be necessary to put it into the database, because the character could be zoning into a completely different server.

joncloud:
...masters while zoning.  Fixed #1286.


teschnei:
is respawn_pet really necessary? could you just check the hp not being 0?

joncloud:
Nope - HP definitely makes sense for representing the same flag.

PR #1295, opened 2015-04-14
Standardizing format of all BLU spells.

helixhamin:
Please note, this does not change any spells, or fix any spells, it is only to standardize the spell template.


teschnei:
change your editor to use spaces instead of tabs :p

helixhamin:
Okay, lol, I just added the tabs myself.
How many spaces would you prefer for each nested statement? Or should I just align the spaces to match a single tab in my notepad++?

teschnei:
settings -> preferences -> tab settings -> tab size: 4, check replace by space

helixhamin:
Okay, tabs poofed, all tabs ctrl+hed with 4 spaces in all 110 files.

PR #1298, opened 2015-04-15
getAlliance() for lua scripting use.

TeoTwawki:
Someone told me he needs a function for ix'aern(somejob) to grab alliance members in lua, instead of party members. I happened to have one I use for custom things and didn't have a retail excuse to pull request it till now. It just throws all members of all alliance party into a table like getParty does for a single party. If there's a simpler/shorter way to do this, feel free to close this pull request and implement it. :)


nesstea:
TY Teo

PR #1300, opened 2015-04-15
Added logic to define the log file via command line parameter.  Fixed #1...

joncloud:
...296.

### DIFF
--- a/src/common/showmsg.cpp
+++ b/src/common/showmsg.cpp
@@ -20,24 +20,8 @@ This file is part of DarkStar-server source code.
 #ifdef WIN32
 	#define WIN32_LEAN_AND_MEAN
 	#include <windows.h>
-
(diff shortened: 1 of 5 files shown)


maxtherabbit:
2 issues with this:

1) if you do not #define DEBUGLOGwhatever (it's not defined for search by default) you get a crash on your call to InitializeLog() using an undefined pointer

2) need to move the InitializeLog calls for map and login into main function in kernel.cpp otherwise it does not log the first couple console messages and trips "log_file not defined" warnings

teschnei:
i think it's easier to just name the file based on the passed IP and port rather than specifying one (saves a lot of work, etc)

maxtherabbit:
don't really think that would save much work

you'd still have to set things up the same way, just using different args

teschnei:
i'd make that the default then instead of a warning

PR #1302, opened 2015-04-15
Prevent completeMission() from attempting to complete non current missions

TeoTwawki:
When completeMission() tries to complete non current missions it winds up saving null data into the characters database info.


nesstea:
Ty Teo.

PR #1305, opened 2015-04-15
change priority,because don't launch the cs for mission san d'oria 6.1

Chronos77:
He don't launch the CS for rank 6.1 san d'oria when finish. he proposes quest


Chronos77:
qm1.lua is for pop Adamantoise or Aspidochelone not for the rank 9.1 you have actually 2 in 1 .
https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Valley_of_Sorrows/npcs/qm1.lua
the true "???" is qm2.lua and this is for rank 9.1 san d'oria and this point return message with id = 6381 but in this page we can see is 6382
https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Valley_of_Sorrows/TextIDs.lua

scripts/zones/Valley_of_Sorrows/npcs/qm2.lua
```diff
@@ -0,0 +1,55 @@
...
+
+function onTrade(player,npc,trade)
+--printf("CSID: %u",csid);
+--printf("RESULT: %u",option);
+end;
+
+-----------------------------------
+-- onTrigger Action
+-----------------------------------
+local KEYITEM_OBTAINED = 6382; -- Put this as a local until the TextIDs have all been fixed.
```
Turttle: " KEYITEM_OBTAINED = 6382;" should be placed in the "TextIDs.lua".

Chronos77: yes, i see that i can delete it too is already in TextIDs.lua . corrected !

Turttle:
Also the 6382 ("Obtained:...") is not the correct one, need to use "6385" ("Obtained key item:...") as you showed in your comment. https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Valley_of_Sorrows/TextIDs.lua

Chronos77:
yes it's ok now. i include this file so i use "KEYITEM_OBTAINED" WHo is equal to 6385 :)

Chronos77:
it's ok for me i get my rank 10.

Chronos77:
it's ok ! :)

PR #1306, opened 2015-04-15
Fixing some homepoint glitches & "save my sister" lantern order

b03605079:


b03605079:
ID corrections

xdemolish:
for any future pull requests: you don't need to make a new commit for every file - just one commit with all related changes

PR #1310, opened 2015-04-16
stoneskin.lua was always setting SS to 350. Set to correct value.

ProjectTako:
Stoneskin was clamping (1, 350), which was making the base stoneskin value always set to 350 rather than taking enhancing magic/mnd into account.


teschnei:
pAbs = utils.clamp(pAbs, 1, STONESKIN_CAP);

teschnei:
make sure you leave a comment when you update after a comment (i get notifications for comments but not commits) :)

PR #1318, opened 2015-04-18
Pso lift patch

Turttle:
PsoXja lift lua files.


teschnei:
closing until the server crash is fixed

Issue #752, opened 2015-01-05
Pso'Xja and Fort Ghelsba missing elevators


takhlaq:
need moving to scripts

nesstea:
Teo said he was going to rewrite the elevator function in the core.

zynjec:
This can also be closed, both have working elevators now.

PR #1322, opened 2015-04-18
Psoxja lift patch

Turttle:
Updated and tested the scripts. All of the lifts are working as intended.

### DIFF
--- a/scripts/zones/PsoXja/npcs/@090.lua
+++ b/scripts/zones/PsoXja/npcs/@090.lua
@@ -1,0 +1,24 @@
+-----------------------------------
+-- Area: PsoXja
+-- NPC:  TOWER_A_Lift_0 @pos 300 15.450 -60
+-----------------------------------
+
+-----------------------------------
+
+
+-----------------------------------
+-- onSpawn
+-----------------------------------
+function onSpawn(npc)
+    local lift = 090;
+    local elevator = {
+                        id = 0,                      -- id is usually 0
+                        lowerDoor = npc:getID() + 2, -- lowerDoor's npcid
+                        upperDoor = npc:getID() + 1, -- upperDoor usually has a smaller id than lowerDoor
+                        elevator = npc:getID(), 	 -- actual elevator npc's id is usually the smallest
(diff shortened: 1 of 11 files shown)


TeoTwawki:
Grats on successfully working with the current ridiculous elevator code.

Turttle:
I'm not 100% sure how it works, but it does ^.^

nesstea:
tytytytyty

xdemolish:
once you fix fort ghelsba we can close https://github.com/DarkstarProject/darkstar/issues/752 :p

ashdam:
Hahaha ;)  {yes, please}

Turttle:
does the Fort Ghelsba lift has a special id, like the one id Port Bastok?

TeoTwawki:
Fort Ghelsba one has a lever at top and bottom. Can actually be done much simpler than the others. To use existing elevator framework, copy what palborough miens one does and change the IDs.

Turttle:
Palborough mines one does have a unique id = ELEVATOR_PALBOROUGH_MINES_LIFT.
If I change the id = ELEVATOR_FORT_GHELSBA_LIFT then server crashes and i get error DSP_DEBUG_BREAK_IF(lua_isnil(L,1) || !lua_isnumber(L, 1)); <~ breakpoint

TeoTwawki:
What you likely ran into is one the reason I hate the current implementation: its done on spawn, and in some cases the elevator code can't be in the actual elevator NPC or its lever. It has to be the NPC that gets loaded first (lowest ID value of all involved NPCs).

Also from your crash I can tell you aren't ion teh most recent version of DarkStar because Kj made a change the other day that has it abort instead of crashing when that happens.

TeoTwawki:
We need to switch IRC or forum thread (or forum PMs) because ppl get emails every time we post here.

TeoTwawki:
@xdemolish I summon thee! @Turttle has the scripts for elevator and lever looking correct, and I just made sure eh defined the elevator ID in status.lua with the others. His crash is fixed but elevator didn't go.

PR #1323, opened 2015-04-18
Corrected a bug that would let a character cast a spell they started cas...

joncloud:
...ting prior to level sync, but do not have access to after level sync is applied.


teschnei:
i think it would be better to check in the AI rather than when level sync is applied (in CMagicState::Update)

joncloud:
That is quite a bit more logical.

teschnei:
sorry, put it in CheckInterrupt :p

PR #1324, opened 2015-04-19
[CORE] and [Attowah] Sekhmet

nesstea:
Yes, I tested it @xdemolish


xdemolish:
looks gud
just need to change vs to use 4 spaces for tabs and fix dat indentation

xdemolish:
actually, i don't even care since i'll never win this war on tabs

PR #1328, opened 2015-04-19
[Sky]Gods Additional Effects

nesstea:


scripts/zones/RuAun_Gardens/mobs/Suzaku.lua
```diff
@@ -36,5 +46,30 @@ function onMonsterMagicPrepare(mob, target)
...
+    local dmg = statDiff + levelDiff + math.random(0, 15); -- INT modifier + difference in level + variance
+    local params = {};
+    params.bonusmab = 0;
+    params.includemab = false;
+    
+    dmg = addBonusesAbility(mob, ELE_FIRE, target, dmg, params);
+    dmg = dmg * applyResistanceAddEffect(mob, target, ELE_FIRE, 0);
+    dmg = adjustForTarget(target, dmg, ELE_FIRE);
+    
+    if (dmg < 0) then
```
TeoTwawki: If less than zero, change the message id (with the right gear, you can in fact get healed by the "damage").

TeoTwawki:
To view diff without the whitespace changes: https://github.com/DarkstarProject/darkstar/pull/1328/files?w=1

scripts/globals/items/antarctic_windy.lua
```diff
@@ -0,0 +1,19 @@
+-----------------------------------------
+-- ID: 18164
+-- Item: Antarctic Wind
```
TeoTwawki: @nesstea  *D'oh!* I typo'd the file name. "Antarctic_windy.lua"  ... /FAIL

nesstea: LOL ok 1 sec

teschnei:
you don't need to check dmg < 0 and change the messageid if it is, the core automatically does that for you,  so remove those and it'll be good

(you changed it to the wrong messageid anyways)

TeoTwawki:
@teschnei unless something was recently changed, it won't auto change that because Action->addEffectMessage is set by that return. Which is why all pre-existing scripts did that. As for the ID, guessing when I assisted I went and used message ID for dmg/heal on mob target instead of the message for on player target.

teschnei:
i don't want to make every script writer remember to do that manually, so if there is a problem, then it needs to be fixed in the core and not in 100 different scripts

TeoTwawki:
@teschnei not it :)

nesstea:
Doing another pull req because I derped. Changes in next.

PR #1341, opened 2015-04-22
Adjusted login name and password to allow any printable ASCII characters, minor refactoring

z16:
The entire *DSConnect* solution needs a bit of an upgrade to become a bit less last millennium, but that would require more of an overhaul than I'm personally willing to commit to.

The changes in here are very simple, I adjusted the log in check for name and password. The original issue before now was that only alphanumeric passwords were allowed. This meant that people who mistakenly felt that alphanumeric passwords were not secure enough would not be able to choose a password of their liking.

This adjustment allows people to choose any password they want so long as it consists of printable ASCII characters. This could be adjusted to allow all UTF-8 characters, but would be more work than I was willing to invest.

I also adjusted the name check similarly, and the reasoning behind it is that the name is never visible to anyone else (unless the server admins feel like manually browsing the database). It's not their character name, as such it doesn't need to conform to any character naming rules and conventions, it will only be used for their own log in process.

## On passwords

I decided to allow any length for the password (except for empty passwords, i.e. zero-length). Since this is a somewhat controversial decision I would like to explain the reasoning behind it. Some people here already know a lot about the issue, but many people don't, and even worse, some people are misinformed on the subject. So I figured I'd leave this for people here to read and think about and evaluate for themselves if my decisions make sense.

### Technical strength

Technically passwords are harder to crack if they are longer and contain a wider range of and more random characters. This is due to password breakers trying certain techniques to get to the real password quickly. The most popular is to limit brute force to alphanumeric characters, because many people are too lazy to use special characters. This gives them a much smaller range of characters to iterate through, as such it speeds up the speed of the hack significantly. Another popular technique is called a dictionary attack, which tries combinations of common words and phrases, because it is somewhat likely for certain people to use known words as their passwords, so they can remember it easily.

There are many other techniques and combination of techniques employed that make brute forcing easier, but all of them rely on trying to guess the structure of the password and limit their search to passwords within that domain.

Here is an example of the typical argument made for inclusion of numbers and special characters into the password:
If a password consists of 8 alphabetical characters, and a password hacker could guess 1 billion passwords per second, it would, at most, only take them 3.5 minutes to guess the right one, due to a limited number of combinations.

The argument is very valid, but the suggested fix is awful. Forcing users to massacre their wanted password is the worst of all possible solutions to that problem and shifts the burden of security completely on the user instead of the developer.

The proper way to guard against this is to time password entries and implement a threshold for number of attempts per second. If you limit your system to allow only two password check attempts per second, even a purely alphabetical lower case password of length 8 will take thousands of years to guess right, and the user won't experience any inconvenience. The only drawback of this method is if the password guessing rate is very high, but that is not the case for us. This mostly only affects devices with commonly known initial passwords (like wi-fi routers, which often used initial passwords like "0000" in the past, although they are also switching away from that model).

The key to this is to have a well designed system that only allows password entry through a single entry point that is well guarded. But it stops *any* kind of brute force attack dead in its tracks, and password guessing is the only argument in favor of forcing all kinds of weird password selection rules.

There are still various other types of security vulnerabilities, but none relating to password strength itself. You don't need to know the actual password for rainbow table attacks, and other methods are there to prevent against that (like using a salt on password hashes). You also don't need to guess the password for man-in-the-middle attacks, this requires secure communication with the server to prevent. The password guessing was the only attack that this is supposed to prevent, and such a system is much simpler than having complicated password validation checks.

Note that I did *not* actually implement such a timed check, although it can be done, if requested.

### Philosophical argument

This one is very debatable, but personally I think that the *responsibility* for password security lies *entirely* on the user, and not at all on the developer. I don't think any person these days is dumb enough to not know the risks in choosing "abcd" as their password. If they still want to do it, I don't want to insult their intelligence by assuming that they were too dumb to realize the weakness of the password, but instead would rather assume that they simply don't care much for the password to be strong. And if they don't care, why should we?

I know everyone has a different opinion on this, and for different kinds of businesses the answer to this problem may be different. I can absolutely understand if banks employ a forced structure on passwords to make them more secure by default (although most don't and prefer other types of password security, for example the timed method mentioned above, as well as a limited number of guesses, which I don't think would make a great addition in our case, because it would require a backup system if they screw up). But for this, I don't think that we need to be bothered even *if* this lack of enforced security results in someone getting their account compromised.

I think everyone should be responsible for choosing their own password, and when they do so, I hope (only for them) that they [are reasonable about it](https://xkcd.com/936/). In this vein, it might make sense to increase the maximum length for passwords, but that would also require a somewhat deeper recode than what I did here.

### DIFF
--- a/src/common/socket.cpp
+++ b/src/common/socket.cpp
@@ -856,12 +856,12 @@ int32 realloc_fifo(int32 fd, uint32 rfifo_size, uint32 wfifo_size)
 		return 0;
 
 	if( session[fd]->max_rdata != rfifo_size && session[fd]->rdata_size < rfifo_size) {
-		RECREATE(session[fd]->rdata, unsigned char, rfifo_size);
+		RECREATE(session[fd]->rdata, char, rfifo_size);
 		session[fd]->max_rdata  = rfifo_size;
 	}
 
 	if( session[fd]->max_wdata != wfifo_size && session[fd]->wdata_size < wfifo_size) {
-		RECREATE(session[fd]->wdata, unsigned char, wfifo_size);
+		RECREATE(session[fd]->wdata, char, wfifo_size);
 		session[fd]->max_wdata  = wfifo_size;
 	}
 	return 0;
@@ -887,7 +887,7 @@ int32 realloc_writefifo(int32 fd, size_t addition)
 	else // no change
 		return 0;
 
-	RECREATE(session[fd]->wdata, unsigned char, newsize);
+	RECREATE(session[fd]->wdata, char, newsize);
 	session[fd]->max_wdata  = newsize;
 
 	return 0;
@@ -1103,8 +1103,8 @@ void set_eof(int32 fd)
 int create_session(int fd, RecvFunc func_recv, SendFunc func_send, ParseFunc func_parse)
 {
 	CREATE(session[fd], struct socket_data, 1);
-	CREATE(session[fd]->rdata, unsigned char, RFIFO_SIZE);
-	CREATE(session[fd]->wdata, unsigned char, WFIFO_SIZE);
+	CREATE(session[fd]->rdata, char, RFIFO_SIZE);
+	CREATE(session[fd]->wdata, char, WFIFO_SIZE);
 
 	session[fd]->max_rdata  = RFIFO_SIZE;
 	session[fd]->max_wdata  = WFIFO_SIZE;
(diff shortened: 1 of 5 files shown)


TeoTwawki:
I wish I were patient enough to explain my changes like this. You guys are lucky my every pull request doesn't just say "I did stuff for the thing"

helixhamin:
https://xkcd.com/936/

helixhamin:
Just realized you already posted that link.

PR #1342, opened 2015-04-22
Porter Moogle Implementation

MalRD:
Initial implementation of Porter Moogle functionality. The storage functionality of these moogles is implemented by storing toggled bits in the extra data of the "Storage Slip" items that are purchasable from the moogle. Each bit corresponds to if a particular item is or is not stored on the slip. 

Currently, extra data is not (easily) accessible within the Lua code, so functions were placed within lua_baseentity to perform the storage and retrieval of items. Extra data retrieval in Lua should be generalized at some point, perhaps returning the extra bytes as a table of numbers. 

Additionally, the Porter Moogle cannot store items that have started a Magian trial without completing it. Attempting to store one of these items should result in an error message, but currently does not. Since Magian Trials are not yet implemented, this is a low impact defect that can be resolved with the implementation of Magian Trials.


scripts/zones/Aht_Urhgan_Whitegate/npcs/Porter_Moogle.lua
```diff
@@ -0,0 +1,59 @@
...
+local e = 
+{
+    TALK_EVENT_ID       =   957,
+    STORE_EVENT_ID      =   958,
+    RETRIEVE_EVENT_ID   =   959,
+    ALREADY_STORED_ID   =   960,
+    MAGIAN_TRIAL_ID     =   963
+};
+
+RETRIEVE_DIALOG_ID  =   13462;
```
xdemolish: should probably move this to the zone's TextIDs.lua, and comment the message it displays (for compatibility with the batch is updates)

scripts/globals/porter_moogle_util.lua
```diff
@@ -0,0 +1,337 @@
...
+function buyStorageSlip(player, option, ITEM_CANNOT_BE_OBTAINED, ITEM_OBTAINED, NOT_HAVE_ENOUGH_GIL)
+    -- This is just because hilarious.
+    option = math.floor(option / 16) + (option % 16);
+    local hasItem = player:hasItem(slipIds[option]);
+    if (hasItem or player:getFreeSlotsCount() == 0) then
+        player:messageSpecial(ITEM_CANNOT_BE_OBTAINED, slipIds[option]);
+        return;
+    end
+
+    if (player:getGil() < 1000) then
```
xdemolish: i *think* this does pretty much the same as the if block below, and the message can probably be moved in an else there instead

scripts/zones/Aht_Urhgan_Whitegate/npcs/Porter_Moogle.lua
```diff
@@ -0,0 +1,59 @@
...
+
+-----------------------------------
+-- onEventFinish
+-----------------------------------
+function onEventFinish(player,csid,option)
+    -- Is there a default option value for cancelling selection windows?
+    if (csid == e.TALK_EVENT_ID and option < 1000) then
+        buyStorageSlip(player, option, ITEM_CANNOT_BE_OBTAINED, ITEM_OBTAINED, NOT_HAVE_ENOUGH_GIL);
+    else
+        player:setLocalVar('slipId', 0);
```
xdemolish: shouldn't the var be cleared anyways  (after buying the storage slip)

MalRD: SlipId is set whenever you trade a slip to the moogle for retrieval, so the only time it needs to be cleared is after you exit out of the retrieval menu. 

I'd actually prefer that the clear be inside porter_moogle_util, but while testing scenarios I moved it into Porter_Moogle and built the scripts without correctly it. I'll make a modification and move both handling of both onEventUpdate and onEventFinish back into the utility class.

scripts/globals/porter_moogle_util.lua
MalRD: The previous block was to validate that the event finishing was a retrieve event and not a store / talk. Per my below comment, I'm going to refactor a bit to clarify the intent.

teschnei:
FYI: DSP uses luaJIT, and thus has luaJIT's bitop (http://bitop.luajit.org/) - a whole lot easier to use those instead of reimplementing what they do

helixhamin:
Just wanted to mention, it seems like Mog Garden's Porter Moogle is not present.
http://wiki.ffxiclopedia.org/wiki/Porter_Moogle
Not complaining, just noting.

scripts/globals/porter_moogle_util.lua
```diff
@@ -0,0 +1,311 @@
+-----------------------------------
+-- Porter Moogle Utilities
+-- desc: Common functionality for Porter Moogles.
+-----------------------------------
+require("scripts/globals/common");
+local bit = require("bit");
```
teschnei: you can remove this (it's already a global)

scripts/globals/porter_moogle_util.lua
```diff
@@ -0,0 +1,311 @@
...
+
+----------------------------------------------------------------------
+-- desc : Checks if the supplied item is a Moogle Storage Slip.
+----------------------------------------------------------------------
+function isSlip(itemId)
+    if (itemId == nil) then
+        return false;
+    end
+
+    for i = 1, #slipIds, 1 do
```
teschnei: I think you can just check slipItems[slipId] not nil

and if you do, you can probably condense slipItems and slipIds into one table

scripts/globals/porter_moogle_util.lua
```diff
@@ -0,0 +1,311 @@
...
+    
+    player:setLocalVar('slipId', slipId);
+    player:startEvent(eventId, params[1], params[2], params[3], params[4], params[5], params[6], nil, slipIndex);   
+end
+
+----------------------------------------------------------------------
+-- desc : Begins the storage or retrieval process based on the items
+--        supplied in the trade.
+----------------------------------------------------------------------
+function doTrade(player, trade, e)
```
teschnei: since these are global, it's probably best to give them a name related to porter moogles directly (porterDoTrade, etc)

xdemolish: 'porterDoTrade' wat

porterMoogleTrade
cause there's enough broken engrish on dsp already

MalRD:
Alright, I made everything that isn't currently accessed outside of the utilities a local function, renamed the global functions, and corrected a small defect that didn't align with what actually happens on retail.

scripts/globals/porter_moogle_util.lua
```diff
@@ -0,0 +1,305 @@
...
+    
+    player:setLocalVar('slipId', slipId);
+    player:startEvent(eventId, params[1], params[2], params[3], params[4], params[5], params[6], nil, slipIndex);   
+end
+
+----------------------------------------------------------------------
+-- desc : Begins the storage or retrieval process based on the items
+--        supplied in the trade.
+----------------------------------------------------------------------
+function porterDoTrade(player, trade, e)
```
xdemolish: porterMoogleTrade (since that's what you went with in your npc scripts)

scripts/globals/porter_moogle_util.lua
```diff
@@ -0,0 +1,305 @@
...
+    [slipIds[21]] = { 26741, 26899, 27053, 27238, 27412, 26743, 26901, 27055, 27240, 27414, 26745, 26903, 27057, 27242, 27416, 26747, 26905, 27059, 27244, 27418, 26749, 26907, 27061, 27246, 27420, 26751, 26909, 27063, 27248, 27422, 26753, 26911, 27065, 27250, 27424, 26755, 26913, 27067, 27252, 27426, 26757, 26915, 27069, 27254, 27428, 26759, 26917, 27071, 27256, 27430, 26761, 26919, 27073, 27258, 27432, 26763, 26921, 27075, 27260, 27434, 26765, 26923, 27077, 27262, 27436, 26767, 26925, 27079, 27264, 27438, 26769, 26927, 27081, 27266, 27440, 26771, 26929, 27083, 27268, 27442, 26773, 26931, 27085, 27270, 27444, 26775, 26933, 27087, 27272, 27446, 26777, 26935, 27089, 27274, 27448, 26779, 26937, 27091, 27276, 27450, 26781, 26939, 27093, 27278, 27452, 26783, 26941, 27095, 27280, 27454 }
+};
+
+----------------------------------------------------------------------
+-- desc : Checks if the supplied item is a Moogle Storage Slip.
+----------------------------------------------------------------------
+local function isSlip(itemId)
+	if (slipItems[itemId] ~= nil) then
```
teschnei: i think this is literally the only check needed in this function - if its not nil, then it is a slip, if it is nil, it isn't

MalRD:
Got rid of the silly loop that was there for reasons unknown.

As for the earlier comment about the two table structure: it was done that way to enable an indexed lookup of the itemId for a particular index (slipIds[option]) instead of looping through the second table while keeping track of the iteration. I'm not married to the idea and will change it if you'd prefer a single table.

scripts/globals/porter_moogle_util.lua
```diff
@@ -0,0 +1,297 @@
...
+            player:messageSpecial(ITEM_CANNOT_BE_OBTAINED, slipIds[option]);
+            return;
+        end
+
+        if (player:getGil() < 1000) then
+            player:messageSpecial(NOT_HAVE_ENOUGH_GIL, slipIds[option]);
+            return;
+        end
+        
+        if (player:delGil(1000)) then
```
xdemolish: combine into one check (so player:delGil instead of getGil) cause i get to be picky as well
as far as i could tell, they do like exactly the same shit

if it doesn't work, i will take full responsibility and blame lautan

PR #1343, opened 2015-04-22
Finished the implementation of Magical Maps and added sanity checks

Racsr:
This pull completes the Magical Map vendor offerings, now all the offered maps can be purchased correctly. Map Vendors were also updated if they were missing the global magical_maps.
This also includes sanity checks in quests where the map may have already been purchased so that the player is not presented with a keyitem they already bought.


teschnei:
ohh, those enums are in such dire need of a refactor to use bit operators, but we can do that another time

PR #1344, opened 2015-04-23
Send_command

Turttle:
Updated the @send command to be able to send players to another zone
using auto translate, or another player.


scripts/commands/send.lua
```diff
@@ -10,27 +13,320 @@ cmdprops =
...
+---------------------------------------------------------------------------------------------------
+function onTrigger(player, p1, zoneId)
+    local word  = "";
+    local i     = 0;
+    local zone  = zoneId;
+	local targ1 = GetPlayerByName(p1);
+	
+	
+	-- Not enough info.. 
+	if(p1 == nil and (p2 == nil or zone == nil)) then
```
teschnei: how can p2 not be nil?

Turttle: Typo is fixed.

PR #1346, opened 2015-04-23
Fix rank San d'oria

Chronos77:
fix bug when finish BCNM he don't increase MissionStatus
fix wrong npc for mission 6.1


gedads:
Hi you're right for batallia, chasing quota is left from entrance and mission is right .
Could you separate qm2 and 3 letting chasing quotas in the right place and just moving sando 9-1 to the qm3

Chronos77:
Thx gedads ! i fix that :)

Chronos77:
Hmm.. i have a problem with commands @zone when i write Meriphataud Mountains oer Lower Delkfutts Tower he return zone id 119.

PR #1347, opened 2015-04-23
Roundel Earring

Scavenge:
Implemented Roundel Earring effects


teschnei:
Can you title your pull requests more appropriately? They are the main
identifier for what you've done

Scavenge:
good point i guess that's pretty ambiguous

PR #1358, opened 2015-04-24
"Murgleis sword augmenting convert" fixed.lua

b03605079:


b03605079:
yeah you're right, thanks!

nesstea:
Yay I helped.

teschnei:
should probably be a mod for this too if you feel like adding it while doing conqueror's (convert multiplier or something)

PR #1359, opened 2015-04-25
Implementing conqueror's augment to berserk (tested)

b03605079:


teschnei:
hopefully now you realise that checking for item ID is not very efficient and it'd be better to make it a modifier

b03605079:
i think you're right. thx!

teschnei:
makes it a lot easier in the future too if there are any other items that increase berserks potency

PR #1384, opened 2015-05-01
Send 2.0

Turttle:
Added the option to just move the player if both Target 1 and 2 are in
the same zone, instead of zoning them each time.

### DIFF
--- a/scripts/commands/send.lua
+++ b/scripts/commands/send.lua
@@ -322,7 +322,11 @@ function onTrigger(player, p1, zoneId)
 			if (zoneId ~= nil) then
 				local p2 = zoneId;
 				local targ2 = GetPlayerByName(p2);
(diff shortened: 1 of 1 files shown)


whasf:
Does this actually drag the player across the zone? I am wondering if that would cause unneeded load on the server due to monster actions having to be sent inbetween the two players as one moves. Zoning takes them out and plops them down where they're supposed to be

Turttle:
Yea, it just drags the player if in the same zone :/ But if it will create more load to the server might just leave it out.

xdemolish:
ya, dragging them across the zone would have the mobs aggro them forever too
closing cause it isn't really needed, but feel free to post the script (or any other custom/non-retail scripts you wish to share) on the forums

PR #1385, opened 2015-05-01
remove golden bat whitebone spawns

madwraith:
over 50% spawn locations were whitebone but I found no proof it spawns down there. Have 1 commented out spawn point left in for believers of rare whitebone spawn


whasf:
Did you intend to include Sekhmet spawn locations?

madwraith:
no im kind of confused as why its like that, those appear to be in the current version(same with 17199564s and 17404333 are not in the current file) , the part that has to do with golden bat is the 17199564s. im confused ;/

Issue #1361, opened 2015-04-25
/lockstyle command not working


Kthulupwns:
Not sure where to start on this one.

teschnei:
lockstyle changed completely in the most recent patch, which is why it's never a good idea to add new things into DSP as soon as they come out on retail :p

Kthulupwns:
Good call haha

PR #1392, opened 2015-05-04
Modified LockStyle Implementation

MalRD:
I'm not completely happy with this implementation, so I'm hoping for some feedback on potential improvements in addition to the normal. What is coded matches the retail behavior as best as I was able to ascertain. The following statements are true, from what I've gathered:

1. You can only set an item as a style if you currently own it (exists in any inventory).
2. You can only set an item as a style if you can equip that item normally (1+ job can equip).
3. A set weapon style only works if the skill type of the equipped weapon matches the styled weapon.
4. If you drop an item that is set as a style, it will be maintained until the next appearance update. When that occurs, the slot it was styled to will become empty.
5. If you drop an item that is set as a style but have another copy of the item, the style will remain.
6. If an item you have previously owned is used in a style you are attempting to lock, the currently equipped gear's appearance will be used instead.
7. If an equipment slot style is set to Remove, the styled appearance will be empty. 
8. If the style is locked without using an equipment set, the currently equipped gear is used as the style set.
9. Style remains locked through zoning and logging out/in.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -340,7 +340,8 @@ namespace charutils
             "campaign,"             // 22
             "playtime,"             // 23
             "isnewplayer,"          // 24
-            "campaign_allegiance "  // 25
+            "campaign_allegiance,"  // 25
+            "isstylelocked "        // 26
             "FROM chars "
             "WHERE charid = %u";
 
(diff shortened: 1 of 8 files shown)


src/map/utils/charutils.cpp
```diff
@@ -1753,6 +1805,126 @@ namespace charutils
...
+            && canEquipItemOnAnyJob(PChar, AItem));
+    }
+    
+    void SetStyleLock(CCharEntity* PChar, bool isStyleLocked) {
+        if (isStyleLocked) {
+            for (uint8 i = 0; i < SLOT_LINK1; i++) {
+                auto PItem = PChar->getEquip((SLOTTYPE)i);
+                PChar->styleItems[i] = (PItem == nullptr) ? 0 : PItem->getID();
+            }
+            memcpy(&PChar->mainlook, &PChar->look, sizeof(PChar->look));
```
teschnei: just change that bad boy to an assignment and it's good

MalRD:
Modified the memcpy to be an assignment. 

This should also resolve #1361.

PR #1395, opened 2015-05-05
Stone fix mp/castime/recastime values

madwraith:
Noticed in game the mp value for stone was not what was actually happening. Checked wikia and it matches whathe game says the mp value is. So matched stone values to what I found on wikia.


xdemolish:
it's intentional
retail changed spell cast times/formulas a while ago iirc but nobody's bothered to figure it out yet so old values are still being used

teschnei:
It's been figured out for a long time, just nobody has implemented it

xdemolish:
get implementing, figure-outer

PR #1397, opened 2015-05-05
updateClaim function added

Setz0r:
updateClaim LUA function added to make a monster claimed through script.
similar to updateEnmity but actually marks the monster as claimed.  good for scripted spawning of quest monsters or fished up NM

example in LUA:

SpawnMob(monsterid):updateClaim(player);

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -5918,6 +5918,30 @@ inline int32 CLuaBaseEntity::updateEnmity(lua_State *L)
 }
 
 /************************************************************************
+*                                                                       *
+*  updates enmity and claims monster									*
+*                                                                       *
+************************************************************************/
+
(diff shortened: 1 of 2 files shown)


maxtherabbit:
there are unrelated changes in both files that need to be cleaned up

src/map/lua/lua_baseentity.cpp
```diff
@@ -9111,10 +9135,10 @@ inline int32 CLuaBaseEntity::getPartyLeader(lua_State* L) // Todo: also add abil
             if (PLeaderName != nullptr)
             {
                 lua_pushstring(L, PLeaderName);
-                return 1;
-            }
+            return 1;
```
teschnei: ?

teschnei:
replace your tabs with spaces

Setz0r:
I updated the pull request to convert all tabs to spaces and remove unneeded code.

PR #1399, opened 2015-05-06
Kirin changes

dr01d3r:
Kirin now casts Astral Flow below 50-60% health.

Avatar summoned is chosen at random.

Modified Kirin's wind resistance so wind elemental nukes are resisted
less often.

Had to offset Kirin's pet entry in mob_pets so it summons Kirin's Avatar
instead of Genbu.
Kirin is now immune to sleep.


teschnei:
where does kirin actually use astral flow? i see the vars for it and the spawning of the pet, but no ability usage

scripts/zones/The_Shrine_of_RuAvitau/mobs/Kirin.lua
```diff
@@ -16,9 +17,21 @@ function onMobInitialize( mob )
...
+end
+
+-----------------------------------
 -- onMobFight Action
 -----------------------------------
 function onMobFight( mob, target )
+
+    if (mob:getHPP() < math.random(50,60) and mob:getLocalVar("astralFlow") == 0) then
+        -- Spawn Avatar
+        mob:spawnPet();
```
dr01d3r: Kirin spawns the pet.  The pet uses the appropriate ability based on the selected avatar model.

teschnei:
yes, but kirin has to do the 2hour animation ("kirin uses astral flow"), yes?

TeoTwawki:
does in retail

dr01d3r:
Okay.  I'll make the adjustment.

TeoTwawki:
need to make a mobskill script for the 2hr, have that script do the deed instead of onMobFight

teschnei:
no, I think having the astral flow script empty and doing it in the mobs script is fine

astral flow can vary greatly between the casters and having a generic script would be a big pain

plus the scripter has to script the pet to use the blood pact anyways, so it's not much extra work

TeoTwawki:
ic

xdemolish:
can probably give it chainspell with like 1 sec duration for insta-cast
or add anew lua binding to get/setCastTime

dr01d3r:
Is chainspell a reasonable solution?  What is involved in creating a new
lua binding for altering a skill's cast time?  Is either of these better
than creating a new family system for for Kirin_s_Avatar, so this can be
handled in the database?

teschnei:
I get home in 8 hours. If you could get on IRC to discuss this with me,
that'd be good. Otherwise, describe your issue and I can tell you the best
way to do it.  Also, is there a reason you made a new script for astral
flow instead of using the default?

xdemolish:
honestly, i'd prefer not to add the chainspell effect since you can just modify the spell cast time if you add a new lua binding (just threw it out there for the sake of it)
in fact, i'll add the lua bindings now :)

teschnei:
okay, here's the deal: don't defer the summoning to the script, and don't make a new script specifically for kirin: just make/use mobskills/astral_flow.lua, and have it not do anything except do the right message.  spawn your stuff in kirin's script.  as for the "cast time", I assume you mean the astral flow ability (which isn't a spell so chainspell won't help): it depends on what family ID you are using for the pet.  I don't think there's currently a family ID for astral flow instant summons, but there should be, and they should only have astral flow blood pacts, and they should have no cast time in the DB

dr01d3r:
Thanks Tyler.

It was suggested to me to do it this way but I can adjust it based on your
feedback (perhaps i misunderstood).  I will work through it again tonight
and get on IRC if I have further questions.

dr01d3r:
Thanks for being patient with me during this learning process.  I'm getting
much more familiar with how things work and how they should work within DSP.

dr01d3r:
The scope of this has changed slightly so closing this request to make a new one with a single commit.

PR #1402, opened 2015-05-07
SCRIPTS: replaced all updateEnmity with updateClaim

xdemolish:
i dont even know...


TeoTwawki:
@xdemolish I think not all those updateEmnity instances should actually be changed to claims. Like when an antlion nails me with pit ambush, I don't think it auto claims on retail. Pretty sure I remember someone outside my voking one off me anyway.

teschnei:
the only things that should be changed are the stuff that actually claims (spawning a mob from a ??? yeah, but a mob aggroing or a mob spawning minions, etc, all don't claim)

xdemolish:
https://github.com/DarkstarProject/darkstar/commit/92567ef6c810ba5c285077000d1b10f07068f056

PR #1403, opened 2015-05-07
CORE&SCRIPTS: Add Magic Damage mod

madwraith:
adding to help the new magic damage formula


teschnei:
TABS BE GONE!

madwraith:
think I killed them all

PR #1407, opened 2015-05-08
SCRIPTS: Case-Sensitive corrections

lichtl:
Correcting Hompoint name at Upper Delkfutts Tower since on Linux and other case-sensitive platforms the file was not being found.


lichtl:
Just realized that probably the better way to solve this is a fix in the npc_list. I have some other corrections so please close this I will reapply the SQL changes in a separate pull request

teschnei:
The case in the npc list is probably going to match what the client has in
the dats

lichtl:
Ok. what would be the preferred way? Just noticed even some more homepoints/npc might need to be fixed. Would you prefer to change the filename or do a correction in the database?

TeoTwawki:
@teschnei actually last I looked our homepoint names didn't match the dats in either casing or the usual spaces to underscores convention, or they'd all be like "Home_Point_#3.lua "

Nrrh:
^ That's what happens when they're manually added to npc_list. There's a lot of other records in the table with names that don't fall in like with the DATs. Standard naming convention (Can I have it?)

teschnei:
No preference then. Renaming a script is not a big deal at all, and neither
is changing the SQL name if it's already not the same as the DAT name

lichtl:
ok I've just reopend it. Gonna check later the other zones so at least Homepoint can be fixed within one pull request

teschnei:
As for a standard.. I'm okay with DAT format in entities and lowercase for
everything else (spells, mobskills, etc). They do need to be standardized
because iirc there are a few more places where case is not correct

Nrrh:
Got it. What about non-alphanumeric chars? Strip them? Replace with underscores?

teschnei:
Example?

Nrrh:
Hash tags, colons, apostrophes

Blade: Jin
HomePoint #01

Some names have apostrophes escaped, others have colons removed and not replaced with an underscore while others do replace them, etc...

Nrrh:
Just a quick look at npc_list shows more inconsistencies: blank vs no_name#. Looks like blank started being used instead of numbering no_names at some point

teschnei:
I like just replacing them with an underscore (except apostrophe s, which
should just be s).  In the future, I want to read both the DAT name and the
script name into core so we don't need to have script name like
selh\'teus.lua

teschnei:
Make a new issue for standardizing the names and we can talk there

gedads:
i added Blank  because packet is blank... nothing showed

lichtl:
closing this pull request and resubmit this as I now standardized the file names.

PR #1427, opened 2015-05-13
 SCRIPTS: added [dmg can be avoided by players]

b03605079:
the dmg outpout can be avoided by standing on/near the wyrm's two front feet.


teschnei:
It can be avoided because it's a cone, not because the damage dropoff is so extreme

xdemolish:
arent you supposed to be on vacation?
get vacating

b03605079:
ohhh i know what you mean now, thx :D

PR #1434, opened 2015-05-15
Added binding onWeaponskillHit to trigger mob script things when a player or pet uses TP on a mob, per discussion with Kjlotus…

TeoTwawki:
…. Tested with bst jug pets and smn avatars.

Modeled function on the existing onMagicHit, including some quirks that has. Which I intend to iron out later. This works though.

### DIFF
--- a/src/map/ai/ai_char_normal.cpp
+++ b/src/map/ai/ai_char_normal.cpp
@@ -2555,6 +2555,12 @@ void CAICharNormal::ActionWeaponSkillFinish()
         {
             Action.param = damage;
             Action.messageID = 185; //damage ws
+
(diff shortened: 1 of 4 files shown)


TeoTwawki:
May wanna leave this till Kj sees it, in case he wants me to change something. I know he's away for awhile and I can wait/rebase if needed.

teschnei:
can leave out the kukulkan script? i know it's commented out but there's like 3 things wrong with it and it's pretty much guaranteed to change once abyssea is actually looked at properly

TeoTwawki:
Sure but can you tell me what was wrong in IRC later? I just wanted to leave an example for when ppl actually try to use this.

TeoTwawki:
@teschnei Script removed as requested.

PR #1449, opened 2015-05-18
Update Sagheera.lua

eraffxi:
correction for adding ABC, was doubling amount added each time, spynix corrected but didn't feel like making the pull.


teschnei:
Those better not be tabs replacing spaces

Issue #1461, opened 2015-05-22
Perma-Buffs when Zoning


chipmungus:
When a player changes zone as a buff (or debuff?) is about to wear off (~1/2 seconds), the buff and its effects become permanent. Doesn't happen every time. This persists through other zone changes, and even changing job.

At first this happened with flee, so I added to:

scripts\globals\player.lua

the following:

if (zoning) then
        player:delStatusEffect(EFFECT_FLEE)
end

which worked, but I've since discovered that it can happen with *any* buff. 

Don't wanna just remove all buffs on zoning due to stuff that shouldn't be removed; Signet/Dedication/FoV & GoV buffs and probably more that I can't think of.

Desufire:
I had a player that had an Attack Down become permanent. Was a fairly potent one too. I told him to switch to a lvl 1 job and die. This cleared it up since death has a status erase.

chipmungus:
Tried adding a loop to delete any statuses with less than a set duration, but it seems delStatusEffect doesn't support variables in the same way that addStatusEffect does.

Removing all non-essential statuses on zoning still leaves the possibility of infinite debuffs.

I'll leave it up to someone with more experience.

TeoTwawki:
Durations between zero and one second seem to be erroneously saved as zero duration, which DSP uses for "infinite duration".

teschnei:
423ff9ec82ac724eaa70b44e4c1dd31c6668bc67

PR #1467, opened 2015-05-23
Perma-(De)Buff on zone Fix

chipmungus:
Added a function to status_effect_container, "delStatusEffectsByDuration"

Removes status effects which have a duration less than the time specified

Called the function in Scripts\Globals\Player.lua on zone, with a default time of 5 seconds.

Seems to have fixed issue #1461

### DIFF
--- a/src/map/status_effect_container.cpp
+++ b/src/map/status_effect_container.cpp
@@ -564,10 +564,8 @@ void CStatusEffectContainer::DelStatusEffectsByType(uint16 Type)
 }
 
 /************************************************************************
-*																		*
-*  Removes Status Effects by Duration   						*
-*																		*
-************************************************************************/
+		Removes Status Effects By Duration
+*************************************************************************/
 
 void CStatusEffectContainer::DelStatusEffectsByDuration(uint32 Duration)
 {
--- a/src/map/status_effect_container.cpp
+++ b/src/map/status_effect_container.cpp
@@ -564,6 +564,23 @@ void CStatusEffectContainer::DelStatusEffectsByType(uint16 Type)
 }
 
 /************************************************************************
+*																		*
+*  Removes Status Effects by Duration   						*
+*																		*
+************************************************************************/
+
+void CStatusEffectContainer::DelStatusEffectsByDuration(uint32 Duration)
+{
+	for (uint16 i = 0; i < m_StatusEffectList.size(); ++i)
+	{
+       if  (m_StatusEffectList.at(i)->GetDuration() <= Duration)
+		{
+			RemoveStatusEffect(i--, true);
+		}
+	}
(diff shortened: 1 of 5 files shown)


src/map/status_effect_container.cpp
```diff
@@ -564,6 +564,23 @@ void CStatusEffectContainer::DelStatusEffectsByType(uint16 Type)
...
+*																		*
+*  Removes Status Effects by Duration   						*
+*																		*
+************************************************************************/
+
+void CStatusEffectContainer::DelStatusEffectsByDuration(uint32 Duration)
+{
+	for (uint16 i = 0; i < m_StatusEffectList.size(); ++i)
+	{
+       if  (m_StatusEffectList.at(i)->GetDuration() <= Duration)
```
bendangelo: Can you fix this formatting.

bendangelo:
I'm kind of on the fence about this. I think the issue is status effects are not checked for removal while zoning. So on the next zone enter the status effect container should figure out the effects are past due. This leads me to believe this is an integer overflow bug.

xdemolish:
dont need to make a new commit for each file changed

there's probably a better way to do this, i might look into it eventually

teschnei:
Better to figure out why this is happening instead of band aiding it. Ie:
why are status effects becoming infinite duration? Are they being saved
with 0 duration? Is it the effect saving or effect loading that's doing it?

teschnei:
Best place to start is where they are saved: does the server tell the
difference between an effect with 0 duration (infinite effect) and an
effect with 0 remaining duration (start time + duration = current tick)? I
could see it accidentally saving remaining time as 0 (which gets loaded as
infinite) when it should not be saving at all

chipmungus:
Seems it could be dealt with by:

i, This "band-aid"

ii. Having the server save statuses with <1 duration as >1, bypassing the 0 = infinite thing when it loads after the zone.

iii. Remove "0 = infinite" as a rule. Not really sure why infinite duration status effects exist at all; there are none in retail that I'm aware of.

teschnei:
Iv. The server must calculate the remaining duration before saving it. If
the remaining duration is 0 but the buff duration is not, don't save it

TeoTwawki:
"Not really sure why infinite duration status effects exist at all; there are none in retail that I'm aware of."
Retail has multiple effects that do not wear off from time, but only by zoning/connection loss/death/whatever depending on which effect we are talking about. Farm yoursself up 4 days worth of Abyssea time, and go afk. Your ATMA will stay on the entire time.

teschnei:
there are tons of infinite duration effects, and monsters use them sometimes as well

423ff9ec82ac724eaa70b44e4c1dd31c6668bc67

PR #1470, opened 2015-05-23
CORE: Closed Position

Jupiter065:
Having closed position merits will dancers +3 acc and +3 evasion per
merit when facing the mob. I read that the angle is very small, so I set
it to ~20 degrees from dead on.


xdemolish:
commit labels please

Jupiter065:
Sorry, I don't know what that means.

teschnei:
https://github.com/DarkstarProject/darkstar/blob/master/README.md

Jupiter065:
Ah, thanks.

PR #1471, opened 2015-05-24
CORE / SCRIPTS: Fixing Pet Skill TP / Bloodpacts

Vadavim:
Prior to this change, the mobskill's TP value was assigned to the pet's TP value after this value had already been reduced to 0 (from preparing the skill). I first noticed this problem when tinkering around with avatar bloodpacts that use TP (like meteorite) and seeing that there was no effect with variance in TP.

In addition, many of the avatar bloodpacts had referenced pet:getTP(), but the pet's TP is reduced to 0 after preparing the skill, so I changed this to instead refer to skill:getTP(), which should now contain the correct TP value for the skills in question.

### DIFF
--- a/scripts/globals/abilities/pets/fire_ii.lua
+++ b/scripts/globals/abilities/pets/fire_ii.lua
@@ -14,7 +14,8 @@ function onAbilityCheck(player, target, ability)
 end;
 
 function onPetAbility(target, pet, skill)
-	local spell = getSpell(145);
+	
+    local spell = getSpell(145);
 	--calculate raw damage
 	local dmg = calculateMagicDamage(133,1,pet,spell,target,ELEMENTAL_MAGIC_SKILL,MOD_INT,false);
 	--get resist multiplier (1x if no resist)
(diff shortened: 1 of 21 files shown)


scripts/globals/abilities/pets/fire_ii.lua
```diff
@@ -24,11 +25,12 @@ function onPetAbility(target, pet, skill)
...
-	local tp = pet:getTP();
-	if tp < 100 then
-		tp = 100;
-	end
-	dmg = dmg * tp / 100;
+	local tp = skill:getTP();
+	--if tp < 100 then
+	--	tp = 100;
+	--end
+    print(tostring(tp));
```
bope12: left your debug print in

Vadavim: Whoops, sorry. Good catch. How can I fix that?

bendangelo: Make a new commit to this branch.

Vadavim:
Thanks! I didn't realize I could do quick edits on Github. Is this better, or did I not commit the fire II fix correctly?

bendangelo:
The fix is ok but please do not mix fixes together. This PR should only be about fixing tp in pet skills.

Vadavim:
I'm sorry, it was my first time doing something like this. I'll hopefully get the hang of it next time. :)

PR #1474, opened 2015-05-25
SCRIPTS windurst 8-1 to 9-1 additions/edits

eraffxi:


gedads:
hi, if(CurrentMission == VAIN and MissionStatus >= 1) then
player:startEvent(4);  that cs is for moon reading (windurst 9-2) 
also all dolls popup @ zone are missing (i have the scripts for all areas somewhere so that's not that important).

PR #1486, opened 2015-05-28
CORE: Allow monsters to still use spells with restricted expansion

madwraith:
Removed the check for restricted expansion in mob_spell_list. I figure if the player is fighting the monster than the monster should be allowed to use his spells. Also with what I see in spell.cpp it looks like there is want for monsters to have no spell restriction(at least with player spells I believe..). I do not know enough about blue spells so I did nothing to the way they are working. I let the spells get loaded and then do the check on the CanUseSpell.


teschnei:
ctrl+k,ctrl+d is a great thing

anyways, since you're changing the comments anyways, you could fix the typo too ("namespase")
and remove the russian one altogether, I don't even care anymore

teschnei:
as for blue spells, i assume it'll be fine because they subclass spell, but i also know it will be fine because you can't have any mobs casting blue magic without toau enabled (makes no sense when you consider what a blue magic spell is)

starlightknight:
For most scenarios the original code isn't an issue, but upon closer examination, this change in approach can help with same rare but important edge cases. 

Certain spells (Comet & Meteor in particular) were not available to players until Abyssea/SoA eras, and thus are tagged as such in the spell list sql. However, as early as RoZ or CoP, boss monsters or hnms, like KB & Promathia, could cast these. 

The change in approach will fix those cases (and similar). Sorry for missing those, it was an oversight :-(

However, you should also make sure CanLearnSpell is handled appropriately in this new version, and that the list of spells sent to the client is still filtered (very helpful when testing, and from a UX perspective)

teschnei:
a ton of cases, actually (tier 5/6 nukes, utsusemi: san, holy 2, dia 3), but yes, those were why this commit was authored (it was talked about extensively in irc for the past few days)

I still think you should at least idle in irc to keep track of what's goin on

madwraith:
godspeed

TeoTwawki:
Easier reading linky: https://github.com/DarkstarProject/darkstar/pull/1486/files?w=1  (hides the whitespace changes)

madwraith:
Sorry confused by the filtering of spells sent to client. From my testing players still can not cast or learn restricted spells. If the spells are not loaded monsters will not be able to use them. Need to understand more about filtering being spoken of and the benefits of it in testing.

starlightknight:
@madwraith In the original implementation, because i hijacked the spells as they were loaded, disabled spells were never sent to the client and added to the player's spell list. When switching back and forth to test a variety of configs, this makes it easily apparent as to which mode you're in, as any relevant spells just aren't there, as well as not leaving an confusion as to why the spell can't be used. If its by expansion, it just isn't on your list at all. This is also on par with what you would've seen in retail in that configuration as well (when the server did not have said expansion installed, because it didn't exist yet), as said spells would not have existed.

There are some nice options afforded by that behavior as well in the sense that battle ability wise, you can cleanly toggle between expansion levels with existing character data. On my server, which is only used with a small group of real life friends, we normally leave everything on, but occasionally we feel like playing old school style and flip some of the other expansions off for awhile. Filtering then auto adjusts what we the players see so that it isn't confusing. If you don't filter at load time it should be fairly easy to add an if check around your new function when building the player's spell list.

madwraith:
Moved the check to a better location

src/map/utils/charutils.cpp
```diff
@@ -2726,12 +2726,15 @@ namespace charutils
...
+    void filterEnabledSpells(CCharEntity* PChar)
+    {
+        for (int i = 0; i < MAX_SPELL_ID; i++)
+        {
+            if (spell::GetSpell(i) == nullptr || luautils::IsExpansionEnabled(spell::GetSpell(i)->getExpansionCode()) == false)
+            {
                 delBit(i, PChar->m_EnabledSpellList, sizeof(PChar->m_EnabledSpellList));
             }
-        }
+        }
```
teschnei: should get rid of those trailing whitespace before @TeoTwawki kills you

TeoTwawki: *sharpening my knives right now!*

madwraith:
whitespace death

PR #1489, opened 2015-05-28
Windurst 8-1, 8-2, 9-1

OzmaEra:


gedads:
Nice, just one small problem, the script for zones are still on the old var i used back then windurst81 but the mission var is now missionstatus

OzmaEra:
Thanks, but I'm unsure what these cutscenes even have to do with Vain. The wiki makes no mention of them. http://ffxiclopedia.wikia.com/wiki/Vain

Are you sure these are related to the mission?

xdemolish:
any of these involve the talking doll? cause if they do, you need to make sure that works too

gedads:
all zones are the talking dolls telling you to go north south east west etc. Worked after server message thing was fixed, should be still working.

gedads:
Hi, i'll have another look at it sunday night then i'll push it, i'm leaving for the week end and i'm in a hurry this morning. sorry for the delay

PR #1498, opened 2015-05-31
Core: Conquest influence fix

z16:
The current code for calculating influence has a few issues, mostly relating to integer underflow. It is also a bit redundant between nations and can be shortened. This pull request aims to address both these concerns.

The influence calculation itself presents a bit of a challenge, since it involves integer calculations with a limited pool of available resource, e.g. influence points.

##### Underflow and negative numbers

The current method uses unsigned integers, which underflowed as soon as the values went negative, resulting in absurdly high numbers in the database. The problem here is the values going negative in the first place.

This problem can be avoided by bounding the result of the calculation to a certain maximum. However, it is more elegantly solved by a proportional distribution, since that would guarantee no underflow by default. If a nation only had 12 points, and you based your calculations of the point loss on its ratio of the total available pool, it would never concede more than its maximum by design.

##### Proportional distribution

A nation with a large amount of influence should lose a larger chunk than a nation with a low amount of influence, when a third nation gains something.

Such a system is currently already in-place, but it calculates the ratio wrong, which results in more subtractions than the nation's pool of influence may allow. The points gained by one nation should be multiplied by the every other nation's ratio before they are subtracted.

##### Equality invariant

The number of total points should always add up to 5000. This is also not trivial to achieve, since just rounding or truncating numbers can lose information required to keep the invariant intact.

To guarantee this, two things need to be taken into account. First, the algorithm needs to guarantee that the amount gained by one nation never exceeds the sum of the amount lost by all other nations. Secondly, no points may be lost due to integer conversions.

#### Solution

The following algorithm addresses all three issues:
```c++
        int influences[4] =
        {
            Sql_GetIntData(SqlHandle, 0),
            Sql_GetIntData(SqlHandle, 1),
            Sql_GetIntData(SqlHandle, 2),
            Sql_GetIntData(SqlHandle, 3),
        };

        auto lost = 0;
        for (auto i = 0; i < 4; ++i)
        {
            if (i == nation)
            {
                continue;
            }

            auto loss = std::min<int>(points * influences[i] / static_cast<float>(5000 - influences[nation]), influences[i]);
            influences[i] -= loss;
            lost += loss;
        }

        influences[nation] += lost;
```

As mentioned above, it can never go negative for a nation, because of the proportional distribution. There is still a check in place, which is needed in case one nation gains more than 5000 points, which also can't be allowed.

The proportional distribution is guaranteed by the division by the available points between the three remaining nations. That way it always gets scaled to the maximum number of available points for each nation.

The reason it can guarantee the equality invariant is because the number added is not the number originally provided, but the number that was actually subtracted from the previous nations. In certain cases this may differ from the original number, such as when points are lost due to integral conversions. But this will, at most, differ by 3 points from the provided value.

A slightly more reliable number might be generated if the number was not truncated, but rounded. However, I did not consider this a big issue, so I did not implement it here. It can easily be added, if that is wanted.

### DIFF
--- a/src/map/conquest_system.cpp
+++ b/src/map/conquest_system.cpp
@@ -79,54 +79,25 @@ namespace conquest
 
         if (ret != SQL_ERROR && Sql_NextRow(SqlHandle) == SQL_SUCCESS)
         {
-            int san_inf = Sql_GetIntData(SqlHandle, 0);
-            int bas_inf = Sql_GetIntData(SqlHandle, 1);
-            int win_inf = Sql_GetIntData(SqlHandle, 2);
-            int bst_inf = Sql_GetIntData(SqlHandle, 3);
+            int influences[4] =
+            {
+                Sql_GetIntData(SqlHandle, 0),
+                Sql_GetIntData(SqlHandle, 1),
+                Sql_GetIntData(SqlHandle, 2),
+                Sql_GetIntData(SqlHandle, 3),
+            };
 
-            switch (PChar->profile.nation)
+            auto value = std::max<int>(5000 - influences[PChar->profile.nation], 0) / 3 + 1;
+            auto gain = 3 * value;
+            auto loss = -value;
+
+            for (auto i = 0; i < 4; ++i)
             {
-            case 0:
-            {
-                int total = dsp_max(5000 - san_inf, 0);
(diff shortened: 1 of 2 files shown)


xdemolish:
commit labels :D

teschnei:
commit labels on the commit messages

PR #1503, opened 2015-06-01
Scripts: made future use of EFFECT_ADDLE more intuitive after seeing some folks trying to implement mobskills using it get unexpected results.

TeoTwawki:
Will still function the same. Instead of setting negative power/subPower values we're just subtracting modifiers by that amount in addMod rather than adding a negative number to that value. Little bit closer to his original implementation, but without the swapped addMod/delMod that he was asked to change.

### DIFF
--- a/scripts/globals/effects/addle.lua
+++ b/scripts/globals/effects/addle.lua
@@ -9,8 +9,8 @@
 -----------------------------------
 
 function onEffectGain(target,effect)
-	target:addMod(MOD_FASTCAST, effect:getPower());
-	target:addMod(MOD_MACC, effect:getSubPower());
+	target:addMod(MOD_FASTCAST, -effect:getPower()); -- Yes we are subtracting in addMod()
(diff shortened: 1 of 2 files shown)


TeoTwawki:
https://github.com/DarkstarProject/darkstar/pull/1503/files?w=1

scripts/globals/effects/addle.lua
```diff
@@ -25,6 +25,6 @@ end;
 -----------------------------------
 
 function onEffectLose(target,effect)
-	target:delMod(MOD_FASTCAST, effect:getPower());
-	target:delMod(MOD_MACC, effect:getSubPower());
+	target:delModMod(MOD_FASTCAST, -effect:getPower());
+	target:delModMod(MOD_MACC, -effect:getSubPower());
```
xdemolish: delModMod

TeoTwawki:
Fixed.

PR #1505, opened 2015-06-02
SQL: item_weapon.sql - item shareback part 1

Loki79:


TeoTwawki:
Fixed

TeoTwawki:
Seems I spoke to soon, and there was an Indian burial ground under the graveyard bodies behind the room where the cult suicided! #1508

PR #1508, opened 2015-06-03
SQL: fixed syntax mistakes  /  SCRIPTS: added missing end

TeoTwawki:
Also filled in some missing IDs.


TeoTwawki:
Added missing "end" in AV script

Issue #1514, opened 2015-06-04
Kikoku relic damage proc not occuring on 75 cap, also paralyze working improperly


Kthulupwns:
pretty sure the item_mod sql needs to be updated for this specific item id 18312 (kikoku)

the paralyze proc on the weapon is proccing at the correct rate, but its not being overwritten by stronger versions of paralyze (I.E. Jubaku, Paralyze II, so on and so forth) which is a lua issue but I'm not sure how to correct it.) It is also overwriting stronger versions of paralyze which is parallel to the same issue.

Kthulupwns:
-- Kikoku (75)
-- -------------------------------------------------------
INSERT INTO `item_mods` VALUES (18312, 23, 20);
INSERT INTO `item_mods` VALUES (18312, 355, 137);
INSERT INTO `item_mods` VALUES (18312, 431, 1); -- Add. Effect
INSERT INTO `item_mods` VALUES (18312, 506, 160); -- 16% chance of extra dmg.
INSERT INTO `item_mods` VALUES (18312, 507, 300); -- Occ. 3x dmg

INSERT INTO `item_mods` VALUES (18317, 356, 153);


The isse was these 2 lines,

INSERT INTO `item_mods` VALUES (18312, 507, 160); -- 16% chance of extra dmg.
INSERT INTO `item_mods` VALUES (18312, 506, 300); -- Occ. 3x dmg

their 2nd variable was inversed, merely adjusted the value. Should work now ^^

TeoTwawki:
Regarding pralyze overwriting, is becaus enobody listened to me on how these are supposed to be written, so more thna dozen of teh existing item add.effect scripts are simply wrong.
```lua
target:delStatusEffect(EFFECT_PARALYSIS)
if (not target:hasStatusEffect(EFFECT_PARALYSIS)) then
    target:addStatusEffect(EFFECT_PARALYSIS, 15, 0, 30);
end
return SUBEFFECT_PARALYSIS, 160, EFFECT_PARALYSIS;
```
That is going to remove every paralyze on every proc and replace it. This is wrong behavior threes should be **no** delete of status effect happening and the paralyze being applied needs to be the correct power value. Then the core will decide if it should have no effect, overwrite, or be overwritten, based on the power level of the effect. A bunch of these were merged in this broken state and the author never adjusted his scripts after being told about it.

[...]

PR #1516, opened 2015-06-04
Update item_mods.sql

Kthulupwns:
The additional damage affect wasn't working because these values were incorrect, this should resolve the damage issue with #1514


Kthulupwns:
INSERT INTO item_mods VALUES (18312, 506, 160); -- 16% chance of extra dmg.
INSERT INTO item_mods VALUES (18312, 507, 300); -- Occ. 3x dmg

INSERT INTO item_mods VALUES (18317, 356, 153);

The isse was these 2 lines,

INSERT INTO item_mods VALUES (18312, 507, 160); -- 16% chance of extra dmg.
INSERT INTO item_mods VALUES (18312, 506, 300); -- Occ. 3x dmg

PR #1521, opened 2015-06-05
Scripts: Added script for Orcish Counterstance for those WotG Orcs.

TeoTwawki:


scripts/globals/mobskills/Orcish_Counterstance.lua
```diff
@@ -0,0 +1,30 @@
...
+-- Orcish Counterstance
+-- Used only by Orcs in Wings of the Goddess Areas.
+-- Certain NMs may have a higher power version of the effect.
+---------------------------------------------
+require("scripts/globals/status");
+require("scripts/globals/monstertpmoves");
+---------------------------------------------
+
+function onMobSkillCheck(target,mob,skill)
+    if (mob:getContinentID() == 3) then -- THE_SHADOWREIGN_ERA = 3
```
xdemolish: stick the continent ids in status.lua since it's required anyways

TeoTwawki:
Updated, but placed IDs in zone.lua instead of status.lua per discussion on IRC ("demo told me to"). :grinning:

scripts/globals/zone.lua
```diff
@@ -1,3 +1,80 @@
...
+------------------------------------
+
+    -- Zone Types
+    NONE        = 0
+    CITY        = 1
+    OUTDOORS    = 2
+    DUNGEON     = 3
+    BATTLEFIELD = 4
+    DYNAMIS     = 5
+    INSTANCED   = 6
```
xdemolish: ZONETYPE_CRAP

scripts/globals/zone.lua
```diff
@@ -1,3 +1,80 @@
...
+    ARAGONEAU_FRONT  = 38
+    FAUREGANDI_FRONT = 39
+    VALDEAUNIA_FRONT = 40
+    ABYSSEA          = 41
+    THE_THRESHOLD    = 42
+    ABDHALJS         = 43
+    ADOULIN_ISLANDS  = 44
+    EAST_ULBUKA      = 45
+
+    UNKNOWN          = 255
```
xdemolish: REGION_TYPE

scripts/globals/zone.lua
TeoTwawki: Its not required that they perfectly match whats in the source header file. Only that the use be constant in all lua script. I don't understand why the souce is usinging a prefix all those enums, when what they are is plain as day without any prefixing of the individual things
```cpp
enum ZONETYPE
{
    ZONETYPE_stuff = number,
}
```
is there an actual purpose in those prefixes? I normally like my lua stuff to match my core stuff...but these ones annoy me the same way a comment net to the enum telling me "enums zonetypes" would.

scripts/globals/zone.lua
TeoTwawki: F it, I'll make'em match. But it annoys me and I will have you all here of it! /ramble /gripe /complain

teschnei:
the purpose is making global namespace conflicts more difficult, since they are not encased in a table

TeoTwawki:
Its amended to include the prefixes, but juts sayin: those names are extremely likely to have said conflicts. But I guess consistency in formatting is important enough too and I'd be hypocrite if I argued it.

PR #1523, opened 2015-06-05
Update blade_metsu.lua

Kthulupwns:
Incorrect modifier value, should be set to 80% instead of 60. 

http://ffxiclopedia.wikia.com/wiki/Blade:_Metsu

I'll be working on reworking paralyze on the whole this week so kikoku should be good across the board for 75 cap


Kthulupwns:
Just realized theres a clause for seekers 80% mod.

PR #1524, opened 2015-06-06
SQL: 99 JSE Capes / Misc Item Model additions-corrections

Loki79:


Loki79:
Basic entries were actually in the last PR just needed to get the item_armor up.

PR #1528, opened 2015-06-06
CORE/SQL: expose latent effects to lua

xdemolish:
- moved item mod handling to lua

### DIFF
--- a/sql/item_latents.sql
+++ b/sql/item_latents.sql
@@ -1,2357 +1,713 @@
-
-SET SQL_MODE="NO_AUTO_VALUE_ON_ZERO";
-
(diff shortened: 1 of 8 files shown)


whasf:
Can't you merge this? Or are you waiting for kj to take a look at it

xdemolish:
waiting for minions to script the latents for each item

teschnei:
maybe we should put this on a different branch so we can open PRs to it

xdemolish:
as soon as i clean the query (cause fields r gone)

teschnei:
actually, think we'll do something else for the pet mods.. i started going through fixing everything but it's too much work for like 3 items

xdemolish:
guess the other way would involve adding a petmod field to item_latents

PR #1529, opened 2015-06-06
SCRIPT: Chaotic eye based off blue skill

Jupiter065:
Was based off of enfeebling magic skill


TeoTwawki:
Well that explains why it was almost always resisted.

PR #1545, opened 2015-06-08
CORE: Reprisal fix

Jupiter065:
Reprisal now gives the 50% bonus to block rates, the spike damage is
based off of damage blocked not damage taken, and it expires when it
reflects 2x the caster's max HP

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -565,14 +565,6 @@ uint16 CalculateSpikeDamage(CBattleEntity* PAttacker, CBattleEntity* PDefender, 
             // drain same as damage taken
             damage = damageTaken;
         break;
-        case SPIKE_REPRISAL:
-            damage += (float)damageTaken*0.3;
-            if(PDefender->StatusEffectContainer->HasStatusEffect(EFFECT_SENTINEL))
-            {
-                // bonus
-                damage *= WELL512::GetRandomNumber(1., 3.);
-            }
-        break;
         default:
         break;
     }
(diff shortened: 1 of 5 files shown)


src/map/utils/battleutils.cpp
```diff
@@ -667,6 +659,14 @@ bool HandleSpikesDamage(CBattleEntity* PAttacker, CBattleEntity* PDefender, apAc
                 if (Action->reaction == REACTION_BLOCK)
                 {
                     PAttacker->addHP(-Action->spikesParam);
+
+                    // Subpower is the remaining damage that can be reflected. When it reaches 0 the effect ends
+                    int remainingReflect = PDefender->StatusEffectContainer->GetStatusEffect(EFFECT_REPRISAL)->GetSubPower();
```
bendangelo: Can you use a temporary variable for the status effect.

src/map/utils/battleutils.cpp
```diff
@@ -1789,25 +1789,48 @@ int32 TakePhysicalDamage(CBattleEntity* PAttacker, CBattleEntity* PDefender, int
...
-					absorb = 100 - PDefender->m_Weapons[SLOT_SUB]->getShieldAbsorption();
-                    if((dsp_max(damage - (PDefender->getMod(MOD_PHALANX) + PDefender->getMod(MOD_STONESKIN)), 0) >0)
+        if(isBlocked)
+        {
+            uint8 absorb = 100;
+            if (PDefender->m_Weapons[SLOT_SUB]->IsShield())
+            {
+                if (PDefender->objtype == TYPE_PC)
+                {
+                    absorb = 100 - PDefender->m_Weapons[SLOT_SUB]->getShieldAbsorption();
```
bendangelo: Can you wrap this in a `dsp_cap` so there is no chance of overflow.

Jupiter065:
Sure

src/map/utils/battleutils.cpp
Jupiter065: What do you mean? like 
CStatusEffect reprisalEffect = PDefender->StatusEffectContainer->GetStatusEffect(EFFECT_REPRISAL);
?

Jupiter065:
Is this better?

src/map/utils/battleutils.cpp
xdemolish: ya, cept it's CStatusEffect*

Jupiter065: Right

bendangelo:
It looks ok. I'm not sure if the logic is correct though.

teschnei:
it is

Jupiter065:
Rebased it

PR #1548, opened 2015-06-08
SCRIPTS: Update Sobane.lua

Zenny89:
As was written, you were unable to trade a Kindred Seal to this NPC and complete Limit Break 4 if you had either "Signed in Blood" or "Flyers for Regine" quests active because of the "elseif" structure. Changed the statements to all be "if" so these trade actions will act independently.

### DIFF
--- a/scripts/zones/Southern_San_dOria/npcs/Sobane.lua
+++ b/scripts/zones/Southern_San_dOria/npcs/Sobane.lua
@@ -22,29 +22,28 @@ require("scripts/zones/Southern_San_dOria/TextIDs");
 ----------------------------------- 
 
 function onTrade(player,npc,trade)
-	local count = trade:getItemCount();
-	-- FLYERS FOR REGINE QUEST --
-	if(player:getQuestStatus(SANDORIA,FLYERS_FOR_REGINE) == QUEST_ACCEPTED) then
-		if(trade:hasItemQty(532,1) and count == 1) then -- Trade Magicmart Flyer
-			player:messageSpecial(FLYER_REFUSED);
-		end
+    local count = trade:getItemCount();
+    -- FLYERS FOR REGINE QUEST --
+    if (trade:hasItemQty(532,1) and count == 1) then -- Trade Magicmart Flyer
+        if (player:getQuestStatus(SANDORIA,FLYERS_FOR_REGINE) == QUEST_ACCEPTED) then
+            player:messageSpecial(FLYER_REFUSED);
+        end
 
-	-- SIGNED IN BLOOD QUEST --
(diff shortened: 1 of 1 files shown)


TeoTwawki:
Changing count was not at all needed, its just  variable storing the function you replaced it with. ```local count = trade:getItemCount();```

All you needed to do was move the limit break check to be the 1st one checked, or check the trade contents before the quest (depending one if retail allows them to work independently - sometimes on retail one quest **will** block another). Breaking it into multiple if/end blocks wasn't required to achieve that behavior.

```lua
	local count = trade:getItemCount();
	-- FLYERS FOR REGINE QUEST --
	if (trade:hasItemQty(532,1) and count == 1) then -- Trade Magicmart Flyer
		if (player:getQuestStatus(SANDORIA,FLYERS_FOR_REGINE) == QUEST_ACCEPTED) then
			player:messageSpecial(FLYER_REFUSED);
		end

	-- SIGNED IN BLOOD QUEST --
	elseif (trade:hasItemQty(1662,1) and count == 1) then
		if (player:getQuestStatus(SANDORIA,SIGNED_IN_BLOOD) == QUEST_ACCEPTED and player:getVar("SIGNED_IN_BLOOD_Prog") < 1) then
			player:startEvent(0x02DE,0,1662);
		end

	-- RIDING ON THE CLOUDS QUEST --
	elseif (trade:hasItemQty(1127,1) and count == 1) then -- Trade Kindred seal
		if (player:getQuestStatus(JEUNO,RIDING_ON_THE_CLOUDS) == QUEST_ACCEPTED and player:getVar("ridingOnTheClouds_1") == 2) then
			player:setVar("ridingOnTheClouds_1",0);
			player:tradeComplete();
			player:addKeyItem(SCOWLING_STONE);
			player:messageSpecial(KEYITEM_OBTAINED,SCOWLING_STONE);
		end
	end
end
```
Should make all 3 work independently

xdemolish:
commit labels please (see https://github.com/DarkstarProject/darkstar/blob/master/README.md)

Zenny89:
Okay - I realized the count thing after I changed and submitted it. I agree your change works; I was simply trying to make it look like the on trade action for Raminel.lua.

xdemolish:
gotta change the actual commit's message (not the same as the pull request title)

TeoTwawki:
When the DarkStar peeps talk about commit labels, they mean in the actual commit message, not the pull request title. To change it you'd have to use gits "amend" feature since this was your most recent commit o that branch (changing older commits gets slightly trickier).

xdemolish:
squash to one commit > amend last commit > change message > force push to remote branch > dun

TeoTwawki:
@xdemolish  I didn't even see he had 2 commits for this :x my bad

xdemolish:
problem is: if both conditions are met you'll get a message along the lines of "another event was called while the current one is running" (which probably isn't retail behaviour)

would be better to try and find a better way to restructure the elseifs

Zenny89:
Updated with suggested changes.

PR #1559, opened 2015-06-09
Scripts: blue magic head_butt is no longer 100% success rate stun, now has an actual resistance check.

TeoTwawki:
I did some testing, but the success rate is very high unless you find a mob more than 20 levels above you or with massive thunder resistance and not low lv. I'm unsure how retail accurate the success rate is but this is at least better than no resists check at all.


TeoTwawki:
https://github.com/DarkstarProject/darkstar/pull/1559/files?w=1

PR #1560, opened 2015-06-09
SCRIPTS: Jettatura Based on Blue Magic Skill

Zenny89:
Correct me if I'm wrong, but I think this should be based on Blue Magic instead of Dark Magic.


Hozu:
You are correct in that. I was working with the user who submitted the fixes for the other spells, and we forgot about this spell. Probably because it currently can't be learned.

PR #1568, opened 2015-06-10
PW Astral Flow

OzmaEra:


OzmaEra:
I could clean up the PW lua a little, but this is functioning. 

PW spawns without crashing the server, his pets change model when he does, and astral flow avatars are working (they all spawn at 75%, 50%, and 25%, use their AF abilities, and despawn after).

teschnei:
why make a specific lua for their astral flow?

OzmaEra:
not necessary i suppose

also, after looking at youtube videos it seems each of the surviving lamps all use astral flow, summoning a random avatar. so all of the sql stuff could be taken out as well and just use a math.random in lua to determine modelid which would then determine which mobskill is used.

youtube video showing astral flow: https://www.youtube.com/watch?v=SFsVuMQZ6ZY#t=35m44s

video showing form change from a melee perspective: https://www.youtube.com/watch?v=OBACND4dpe8#t=2m5s
it appears when PW changes form, players disengage him, suggesting that it is a full despawn and the 2nd mobid for PW spawns

teschnei:
the lamps don't summon avatars, there are 9 avatars and only 8 lamps.  (and
the avatar is not random, there's always one of each)

TeoTwawki:
> Unfortunately with the way PW is coded, it uses the astral flow in whatever form it's in at 75%, 50%, and 25% instead of final form only. However, this is still better than it not using AF at all until the whole thing is rewritten.

You could just use a set/getLocalVar to make sure its the final form before setting off the astral flow using ```mob:useMobAbility(MOBSKILL_ID_HERE)```

OzmaEra:
*"the lamps don't summon avatars, there are 9 avatars and only 8 lamps. (and the avatar is not random, there's always one of each)"*
Okay then it's coded just fine the way it is now, outside of the actual ability "Astral Flow" not being used, it just spawns the avatars at the specified HP%.

*"You could just use a set/getLocalVar to make sure its the final form before setting off the astral flow using mob:useMobAbility(MOBSKILL_ID_HERE)"*
The issue is the way PW is coded now it just changes forms based on what HP% it's at, when each form is supposed to be a separate fight from 100%. So if PW is rewritten to be like retail, then the code can stay the same, just make sure it only does it in final form with a local var, but the way it is now it would need to be coded to use it at 12%, 8%, 4% or around there.

TeoTwawki:
Which can be done the way I said, by tracking when to use/has already been used, using the localVar, even without recoding the entire thing (for now, as gedads appears to have been spot on in predicting PW was 2 mobIDs, it should eventually be redone)  :)

And quotes works like this:  ``` > text```
Github has a whole page on how to use the markdown in comments, the link to which is just above the comment input box on the right side. Other handy features include code highlighting, and the in line equivalent of most forum systems `code tags`.

PR #1579, opened 2015-06-11
SW Apollyon 3rd floor mimics

kitala1:
SCRIPTS: Apollyon 3rd floor now spawns armoury crates and mimics again. Also fixed some wonky references I found

### DIFF
--- a/scripts/globals/limbus.lua
+++ b/scripts/globals/limbus.lua
@@ -43,7 +43,7 @@ APPOLLYON_NW_SW_BCNM_LIST = {
                             };
 
 TEMENOS_LIST = {
-                             1298,{true,false,false,4,Temenos_Western_Tower}, -- 'Temenos_Western_Tower'	380 71 -184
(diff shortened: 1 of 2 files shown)


kitala1:
If someone has any insight as to why this broke in the first place, I'd gladly welcome it. I couldn't find any files  or evidence that broke the 3rd floor, it simply stopped popping armoury crates after the last server update we had. 

The issue I found was that onEventFinish was not being called when you select "Yes" when prompted to move to the next floor of Apollyon, but it was being called when you selected "No". All I did was move the call to spawn3fcoffers to onEventUpdate, which was being called every time I moved floors...

xdemolish:
commit labels please (see https://github.com/DarkstarProject/darkstar/blob/master/README.md)

kitala1:
Fixed, sorry about that

Hozu:
Oh hey this also fixes the chest drop from the elementals on 4f in SW. They weren't dropping either.

xdemolish:
squash > force push
(so it sticks all the changes in one commit since the old one without the label is still hanging around)

kitala1:
Done! :)

PR #1580, opened 2015-06-11
CORE:SQL:SCRIPTS: Remove Scavenge Recast Naming

madwraith:
Scavenge rate in settings.lua looks to no longer be relevant. The merit name needs updated since the scavenge merit no longer reduces the recastime. Thinking about it now I should rename it to MERIT_SCAVENGE_EFFECT. I will do this in my next commit. In packets/merit_points_categories.cpp I just updated the comment to avoid confusion.


madwraith:
godspeed!

PR #1583, opened 2015-06-12
Database/SQL - Promyvion NM drop adjustments

Hozu:
I adjusted the drop rates based on the data from ffxidb. For the
memospheres, since the database has only one rate yet there are three
slots, I divided the single rate by 3, then added 10 for the highest
rate, and subtracted 10 for the lowest rate, so it's more likely to drop
at least one memosphere.


xdemolish:
dumped on @teschnei cause lolmath

PR #1592, opened 2015-06-14
SCRIPTS: fix limbus exploit

eraffxi:
issue with the var not calling sometimes, insures that it will get reset on register + on entry to prevent abuse of entering without chips.


xdemolish:
squash/rebase > amend last commit > change commit message to include labels > force push

PR #1607, opened 2015-06-19
SCRIPTS/CORE: Add Scavenge Functionality

madwraith:
The following is my testing from retail and I did similar testing with the code I made for dsp.
The formula I am testing:
-floor(used * (jobLv / 200 + Scavenge Effect bonus))

Test formula:
-floor(3 arrow test * (66 / 200)) = .99 did not get an arrow
-floor(99 arrows used * (66 / 200)) = 32.67 got 32 arrows
-floor(304 arrows used * (66 / 200)) = 100.32 got 99 arrows
-floor(50 arrows used * (66 / 200 + Hunter's Socks(.05))) = 19 got 19 arrows

Use count variable tests:
-unlimited shot and recycle do not use count
-zoning resets use count
-using scavenge resets use count
-Using a different arrow resets use count
-switching to same arrow and using keeps count
-max use count is 99
-Do not have to have bonus gear on while using arrows
	Can put bonus gear on before using the skill to get bonus

Bonus from Gear/Merits:
-Found Hunter's Socks to be .05

~I did notestthese myself but am going with what wikia and bg have
http://ffxiclopedia.wikia.com/wiki/Scavenge_Effect
https://www.bg-wiki.com/bg/Scavenge

-Hunter's Socks +1 .05
-Orion Socks .07
- .05 per merit point up to .25

Least amount of arrows used to get stack back:
floor(122 arrows used * (99 / 200 + .07 + .25)) = 99.43

Heightist amount of arrows a lv 10 with no bonus would have to use to get 99 arrows 

-floor(1980 arrow test * (10 / 200)) = 99

### DIFF
--- a/scripts/globals/abilities/scavenge.lua
+++ b/scripts/globals/abilities/scavenge.lua
@@ -13,7 +13,7 @@ require("scripts/globals/status");
 -- onAbilityCheck
 -----------------------------------
 
-function onAbilityCheck(player,target,ability)
+function onAbilityCheck(player,target,ability,action)
     return 0,0;
 end;
 
--- a/scripts/globals/abilities/scavenge.lua
+++ b/scripts/globals/abilities/scavenge.lua
@@ -14,35 +14,59 @@ require("scripts/globals/status");
 -----------------------------------
 
 function onAbilityCheck(player,target,ability)
-	return 0,0;
+    return 0,0;
 end;
 
 -----------------------------------
 -- onUseAbility
 -----------------------------------
 
-function onUseAbility(user,target,ability)
+function onUseAbility(player, target, ability, action)
 
     -- RNG AF2 quest check
-    local FireAndBrimstoneCS = user:getVar("fireAndBrimstone");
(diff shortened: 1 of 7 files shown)


teschnei:
isn't gonna build on windows either without the vcproj files changed

madwraith:
Athe momenthere is an issue where on ability use the server crashes. DSGame says PANIC: unprotected error in call to Lua API (attempt to call a nil value). Then there are memory errors I get in VS while debugging: First-chance exception at 0x7567C41F in DSGame-server.exe: Microsoft C++ exception: std::out_of_range at memory location 0x002CE718.

teschnei:
the oor exceptions are normal, they are caught and handled (and have been there long before you made these commits)

teschnei:
find where it prints the panic, put a breakpoint there, and see it's doing when it does that

madwraith:
I found the cause of the error to be because of my messy test environment. I changed a lua_pcall param value to test something and never changed it back. Building this branch I found that my changes are working perfectly fine.

PR #1611, opened 2015-06-20
DATABASE: According to the wiki even though the aegis is a class 5 shield, it procs like a class 3 shield which should be about 50%.  This currently only handles the lvl 75 item.

dacrybabysuck:
DATABASE: According to the wiki even though the aegis is a class 5 shield, it procs like a class 3 shield which should be about 50%.  This currently only handles the lvl 75 item.


teschnei:
pretty sure aegis already has 50% block rate and this is adding 50% more

dacrybabysuck:
You're right. After looking through battleutils.cpp I see there's a special switch for it.  So we can ignore this pull request. Does it make sense to handle it from the db since there's a block rate mod, rather than having special cases in code?

dacrybabysuck:
Does the mod just adjust the current is or is it setting the base. I guess that would make the difference.

teschnei:
the block rate mod is for adjusting, for example reprisal

PR #1614, opened 2015-06-22
SQL/Scripts - Soulflayer & Yovra moves

Hozu:
Implemented Immortal Anathema, Mind Blast, Mind Purge, and Tribulation for Soulflayers, and Luminous Drape and Vitriolic Barrage for Yovra. For the Mind Blast BLU spell, all sources I checked did not say that bind er... paralysis overwrites itself/other sources so I removed it.

I did not implement Immortal Shield or Immortal Mind because I could not find a source for their power, nor do I know how to do the animation that exists while Immortal Shield buff is active.

### DIFF
--- a/scripts/globals/mobskills/Immortal_Anathema.lua
+++ b/scripts/globals/mobskills/Immortal_Anathema.lua
@@ -1,0 +1,27 @@
+---------------------------------------------
+--  Immortal Anathema
+--
+--  Description: Inflicts a curse on all targets in an area of effect.
+--  Type: Enfeebling
+--  Utsusemi/Blink absorb: Ignores shadows
+--  Range: AoE
+--  Notes: 
+---------------------------------------------
+require("scripts/globals/settings");
+require("scripts/globals/status");
+require("scripts/globals/monstertpmoves");
+
+---------------------------------------------
+function onMobSkillCheck(target,mob,skill)
+    return 0;
+end;
+
+function onMobWeaponSkill(target, mob, skill)
+    local typeEffect = EFFECT_CURSE_I;
+
+    local msg = MobStatusEffectMove(mob, target, typeEffect, 25, 0, 300);
+
+    skill:setMsg(msg);
+
+    return typeEffect;
+end;
--- a/scripts/globals/mobskills/Tribulation.lua
+++ b/scripts/globals/mobskills/Tribulation.lua
@@ -1,0 +1,41 @@
+---------------------------------------------
+--  Tribulation
+--
+--  Description: Inflicts Bio and blinds all targets in an area of effect.
+--  Type: Enfeebling
+--  Utsusemi/Blink absorb: Ignores shadows
+--  Range: AoE
+--  Notes: Bio effect can take away up to 39/tick.
+---------------------------------------------
+
+require("scripts/globals/settings");
+require("scripts/globals/status");
+require("scripts/globals/monstertpmoves");
+
+---------------------------------------------
+
+function onMobSkillCheck(target,mob,skill)
+    return 0;
+end;
+
+function onMobWeaponSkill(target, mob, skill)
+
+    local blinded = false;
+    local bio = false;
+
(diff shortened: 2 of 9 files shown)


gedads:
Hi if you're looking for the shields animations i think it's subanim 1 and 2

Hozu:
Ah, thank you. However, without knowing the strength of that magic stoneskin effect, I'd rather not implement it.

Hozu:
One sec I messed up here, didn't mean to put in that change here.

scripts/globals/mobskills/Immortal_Anathema.lua
```diff
@@ -0,0 +1,25 @@
...
+
+---------------------------------------------
+function onMobSkillCheck(target,mob,skill)
+    return 0;
+end;
+
+function onMobWeaponSkill(target, mob, skill)
+    local typeEffect = EFFECT_CURSE_I;
+
+    skill:setMsg(MobStatusEffectMove(mob, target, typeEffect, 25, 0, 300));
```
xdemolish: change this as well as the other skill:setMsg's to set it like so
```
local msg =  MobStatusEffectMove(mob, target, typeEffect, 25, 0, 300);

skill:setMsg(msg);
```

cause readability

scripts/globals/mobskills/Tribulation.lua
```diff
@@ -0,0 +1,29 @@
...
+function onMobSkillCheck(target,mob,skill)
+    return 0;
+end;
+
+function onMobWeaponSkill(target, mob, skill)
+
+    local typeEffect = EFFECT_BLINDNESS;
+
+    skill:setMsg(MobStatusEffectMove(mob, target, typeEffect, 20, 0, 120));
+    MobStatusEffectMove(mob, target, EFFECT_BIO, 39, 0, 120);
```
xdemolish: i summon @teschnei or @maxtherabbit (should probably push a separate message for the other effect resist/landing)

scripts/globals/mobskills/Vitriolic_Barrage.lua
```diff
@@ -0,0 +1,29 @@
...
+---------------------------------------------
+function onMobSkillCheck(target,mob,skill)
+    return 0;
+end;
+
+function onMobWeaponSkill(target, mob, skill)
+    local needles = 1000 / skill:getTotalTargets();
+    local typeEffect = EFFECT_POISON;
+
+    MobStatusEffectMove(mob, target, typeEffect, 20, 3, 60);
```
xdemolish: should prolly set the message here

Hozu:
Re: Tribulation - On the video I saw for retail, it says it inflicts blind, unless they're already blinded in which case it says it inflicts Bio (literally, [Target] is Bio.). However, I believe that when I had the message for Bio status, it returned that the player is KO. IIRC Tribulation is the only mobskill that inflicts Bio without doing any damage.

As for the rest, I'll take a look once I fix my database.

xdemolish:
```
if player:hasStatusEffect(EFFECT_ID) then
    -- do shit
end
```
though MobStatusEffectMove tests by adding the effect anyways

teschnei:
And it takes into account overwriting weaker versions etc

Hozu:
The issue I'm having is that I have to check to see if blind is inflicted by tribulation as if it's not, it returns the bio message. Though that does mean that merely checking for blind status isn't good enough...

teschnei:
Just check the value of msg after calling MobStatusMove

Hozu:
Here we go, made the requested changes and I fixed the message returned by Tribulation. I guess there was a UI update since the video I checked since it's now "[Target] is afflicted with Bio." so no more Engrish. While looking at Murk did allow me to solve the message issue, it was returning the wrong status messages so I fixed that as well.

Vitrolic Barrage returns the damage taken because it's damage + status, so it matches other moves of that type like Hurricane Wings.

PR #1615, opened 2015-06-22
Update Mob_pools.sql for Absolute Virtue Family Type

Kthulupwns:
Was oringally an incorrect type and was preventing him from using TP moves like Medusa Javelin and such.


xdemolish:
use commit labels please (see <https://github.com/DarkstarProject/darkstar/blob/master/README.md>)
amend last commit > change message to include label e.g. "SQL: updated absolute virtue pool type" (without quotes) > force push

Kthulupwns:
Ah okay, thanks demo

Hozu:
IMO keep AV as its own family because then other properties can be added to it but not polearm Aerns, like various status immunities. That means you'd simply have to enable the correct TP moves for AV in the mob_skills.sql file.

Kthulupwns:
In my opinion it seems easiest to put it in 440 because it uses the exact same tp moves as the other drg aern mobs, and can simply setMod  to modify his values as well. But that's up to you guys how you'd like to proceed

Hozu:
Hm, you're probably right. That's how its regen is handled anyway, IIRC.

Kthulupwns:
Well as well as Bracelets mode which can handled in Mob_pools_mod sql as well as lua triggering addMods.

teschnei:
put him in his own family, there's no guarantee that all his moves are the same (since he can use every 2hour, I think he needs them all explicitly defined for his family), or all his resists are the same as a regular aern

PR #1637, opened 2015-06-30
Scripts: Jailer of Fortitude pet check

Hozu:
Added a check to see if either Kf'Ghrah is still alive to see if it should mimic a spell casted on it.


Hozu:
Actually, please don't accept this quite yet - I have a couple of other changes I want to make to this first. I'll comment again once I force push the rest of the adjustments.

PR #1638, opened 2015-06-30
SQL/Scripts: Jailer of Fortitude pets + skills

Hozu:
-Added a check to see if either Kf'Ghrah is still alive to see if it should mimic a spell casted on it.
-Scripted the two Kf'Ghrah so they change form and have the correct jobs/resistances
-Set Kf'Ghrah HP to 9k based on pixel guesstimation (seriously - http://i.imgur.com/VgTv45Z.png the difference being 1058 damage taken - source video: https://youtu.be/YYAIJGrb-rQ?t=280).
-Added Vorpal Blade to Jailer of Fortitude family; added Sickle Slash and Damnation Dive to Kf'Ghrah family.
-Fixed the name for Kf'Ghrah BLM in mob_pools.sql


Hozu:
Unfortunately, due to lack of info regarding spells, I did not add them to the two Kf'Ghrah. The only spells I know of are Banish IV, Banishga III, and Flash from the Kf'Ghrah WHM. I have no confirmation about the Kf'Ghrah BLM so I didn't think it'd be right to add a partial list for one and nothing for the other.

Hozu:
I noticed that my form changing for the Kf'Ghrahs were incorrect as it was just from ball to one of spider/bird and back, and always either bird or spider. Fixed it so it's random between all three.

Hozu:
So I just found out that Jailer of Fortitude has a unique core - a prismatic core. That... is really cool so I got that going now.

I also found out that the BLM Kf'Ghrah has at least Blind and Bio 3 as castable spells. Still, since I've only seen it cast two spells total, I'm not comfortable adding those yet. I imagine that it gets Drain and Aspir as well, but who knows for sure?

whasf:
Sounds like you're trying to talk us out of merging this? I'm sure afraid to.. lol

Hozu:
Oh well it works much better than it currently does. The spells can always be added later.

scripts/zones/The_Garden_of_RuHmet/mobs/Kf_ghrah_blm.lua
```diff
@@ -0,0 +1,59 @@
...
+-- OnMobSpawn Action
+-- Set core Skin and mob elemental bonus
+-----------------------------------
+
+function onMobSpawn(mob)
+    mob:AnimationSub(0);
+    mob:setLocalVar("roamTime", os.time());
+    mob:setModelId(1168); -- Dark
+    mob:addMod(MOD_DARKRES, 50);
+    mob:addMod(MOD_LIGHTRES, -50);
```
xdemolish: https://www.youtube.com/watch?v=gvdf5n-zI14
never add or delete mods in onMobSpawn

add it in mob_spawn_mods instead

Hozu:
Wait, seriously? Oh man, that's... interesting to say the least. I'll take a look at that.

teschnei:
or onMobInitialize!

xdemolish:
didn't someone complain about onMobInitialize not setting the mods correctly at some point?

teschnei:
not that I know of.. should be fixable if that's the case though

xdemolish:
meh can't find it, onMobInitialize it is

Hozu:
By the way, setMod doesn't cause it to add every time, right? Only addMod? Fortitude itself uses setMod, though I'll end up moving his physical resistance to mob_spawn_mods.sql as well.

xdemolish:
setMod doesn't add

Hozu:
Ok moved the JoF and the two pet mods to mob_spawn_mods.sql. The reason I asked about addMod vs. setMod was for the regular Ghrah mobs, which were using addMod. I've changed them to use setMod instead.

scripts/zones/The_Garden_of_RuHmet/mobs/Jailer_of_Fortitude.lua
```diff
@@ -11,18 +11,11 @@ require("scripts/globals/magic");
...
-	-- Set the damage resists
-	-- According to https://www.bg-wiki.com/bg/Jailer_of_Fortitude
-	-- Has very high resistance to melee damage (approximately 95%).
-	-- Set damage resistance to match that.
-	mob:setMod(MOD_HTHRES,50);
-	mob:setMod(MOD_SLASHRES,50);
-	mob:setMod(MOD_PIERCERES,50);
-    mob:setMod(MOD_IMPACTRES,50);
+    -- Give it two hour
+    mob:setMod(MOBMOD_MAIN_2HOUR, 1);
```
teschnei: i see that this is from the script before, but you can't use MOBMOD in setMod, you must use setMobMod (currently this add 1 MOD_ICEDEF)

TeoTwawki:
> didn't someone complain about onMobInitialize not setting the mods correctly at some point?

I remember that but it got fixed.

Hozu:
Fixed the 2hr on Fortitude.

Hozu:
So I just realized that if say, a Ghrah spawned as light core, then was killed and respawns as fire, it retains the light res/dark weak from the light core. So I forced the elements that don't have the weakness/resistance to 0 on spawning.

I also moved the Core Meltdown 5% chance to the mobskill, because it was still blowing up too often, even when changed to only blow up in ball form.

Sorry about yet another change. Everything should be fine now.

teschnei:
is that what you observed happen or you think happens?

Hozu:
Erm, it was what I thought happens, but when I checked it with a much more obvious mod (element absorb) it turns out it wasn't the case. So I guess setting the resistance for the other elements to 0 isn't necessary, so I suppose I should remove that for the sake of efficiency. Sorry !

Hozu:
Ok removed the extra setMods. Didn't fully revert because of the Core Meltdown stuff and whitespace cleanup.

TeoTwawki:
@Hozu little request should you happen to do any more editing: my eyes suck, so `if(` and other stuff squished up against parenthesis is hard for me to see. I wind up looking at it in 150% zoom just to tell it isn't `i[` in both notepad++ and in my browser. `if (` looks so much better. 

Any of DSP's files you happen to clean up with stuff like tabs to spaces, if you also would add the space between the `if (` and the `) then` 's along the way, I'd personally appreciate that extra care. Don't need to make another change solely for that though unless one of the devs asks you to.

Hozu:
Understandable, I'll keep that in mind.

PR #1648, opened 2015-07-03
Scripts: Healing Breath and Spirit Link updates

Enignite:
Corrected Deep Breathing values to those for healing breath not elemental breath

Updated TP to skill:getTP, corrected values based on retail data and to only increase for every 20% TP

Corrected main formula so that gear is properly calculated

Updated Spirit Link so that player no longer loses HP if wyvern is at full HP with Empathy merits.

### DIFF
--- a/scripts/globals/abilities/pets/healing_breath_i.lua
+++ b/scripts/globals/abilities/pets/healing_breath_i.lua
@@ -13,26 +13,30 @@ function onAbilityCheck(player, target, ability)
 end;
 
 function onPetAbility(target, pet, skill, master)
-	-- TODO: Correct base value (45/256).  See Healing Breath III for details.
+
+   -- TODO:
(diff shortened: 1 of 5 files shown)


xdemolish:
squash the commits > double check the commit is labeled e.g. "SCRIPTS: updated healing breath and spirit link" > force push > leave a comment when done so we know to check back :)

Enignite:
Sorry, first time doing this. Is it correct now?

xdemolish:
yep
leaving for @teschnei to look at

teschnei:
Looked fine assuming it was tested

Enignite:
Made these a while ago but retested it on most recent build before commiting and it was working, other than Healing Breath IV as I don't think that's coded yet (just made the lua for when it is implemented).

Issue #1452, opened 2015-05-19
When a player is charmed by a Corse mob, the player does not move (the player should move and attack people on the mob's enmity list)


dazusu:

ashdam:
I think charm AI is not implemented yet.

teschnei:
They will move and attack other people, but it won't display on their own
client yet

dazusu:
The other people around them are not seeing it either (ie, the people 
getting attacked).

teschnei:
Okay, I'll keep it in mind next week when I nuke the AI classes

xdemolish:
yus, been waiting for AI rewrite for long tiem

PR #1653, opened 2015-07-04
Scripts/SQL: Gurfurlur, Tiamat, and Charm

Hozu:
-Added Incessant Fists, and scripted Arcane Stomp. Since it applies Magic Shield status, I handled the absorption there.
-Added an HP check on Pleiades Ray.
-Made Gurfurlur's guards respawn indefinitely while engaged.
-Added his two hour and the ability to use it multiple times.
-Changed the power of Charm moves to 0, as that enables charmed players to be controlled by AI. (looks like this fixes part of issue #1452 - still missing the ability to act against charmed players though.
-Moved Tiamat's ranged defence mod to Mob Pools from scripts. Deleted her duplicate magic defence mod. Magic damage taken is now about as much as people who have experience fighting Tiamat on retail expect.

### DIFF
--- a/scripts/globals/mobskills/Arcane_Stomp.lua
+++ b/scripts/globals/mobskills/Arcane_Stomp.lua
@@ -19,19 +19,11 @@ function onMobSkillCheck(target,mob,skill)
 end;
 
 function onMobWeaponSkill(target, mob, skill)
---[[
-    power = 1;
-    tic = 0;
-    duration = 60;
-
-    typeEffect = EFFECT_NAME;
-    skill:setMsg(MSG_BUFF);
-    if(mob:hasStatusEffect(typeEffect) == true) then
-        oldEffect = mob:getStatusEffect(typeEffect);
-        oldEffect:setPower(power);
-        oldEffect:setDuration(duration);
-    else
-        mob:addStatusEffect(typeEffect,power,tic,duration);
+    
+    local duration = 300;
+    local typeEffect = EFFECT_MAGIC_SHIELD;
+    local msg = MobBuffMove(mob,typeEffect,3,0,duration);
+	
+    skill:setMsg(msg);
     return typeEffect;
-	]]
 end;
--- a/scripts/globals/mobskills/Pleiades_Ray.lua
+++ b/scripts/globals/mobskills/Pleiades_Ray.lua
@@ -15,23 +15,30 @@ require("scripts/globals/monstertpmoves");
 ---------------------------------------------
 
 function onMobSkillCheck(target,mob,skill)
-	return 0;
+    local result = 1;
+    local mobhp = mob:getHPP();
+
+    if (mobhp <= 20) then
+        result = 0;
+    end;
+
+    return result;
 end;
 
 function onMobWeaponSkill(target, mob, skill)
-	local duration = 120;
+    local duration = 120;
 
-	MobStatusEffectMove(mob, target, EFFECT_PARALYSIS, 40, 3, duration);
-	MobStatusEffectMove(mob, target, EFFECT_BLINDNESS, 40, 3, duration);
-	MobStatusEffectMove(mob, target, EFFECT_POISON, 10, 3, duration);
-	MobStatusEffectMove(mob, target, EFFECT_PLAGUE, 5, 3, duration);
-	MobStatusEffectMove(mob, target, EFFECT_BIND, 1, 0, duration);
-	MobStatusEffectMove(mob, target, EFFECT_SILENCE, 1, 0, duration);
-	MobStatusEffectMove(mob, target, EFFECT_SLOW, 128, 0, duration);
+    MobStatusEffectMove(mob, target, EFFECT_PARALYSIS, 40, 3, duration);
+    MobStatusEffectMove(mob, target, EFFECT_BLINDNESS, 40, 3, duration);
+    MobStatusEffectMove(mob, target, EFFECT_POISON, 10, 3, duration);
+    MobStatusEffectMove(mob, target, EFFECT_PLAGUE, 5, 3, duration);
+    MobStatusEffectMove(mob, target, EFFECT_BIND, 1, 0, duration);
+    MobStatusEffectMove(mob, target, EFFECT_SILENCE, 1, 0, duration);
+    MobStatusEffectMove(mob, target, EFFECT_SLOW, 128, 0, duration);
 
-	local dmgmod = 1;
-	local info = MobMagicalMove(mob,target,skill,mob:getWeaponDmg()*7,ELE_FIRE,dmgmod,TP_NO_EFFECT);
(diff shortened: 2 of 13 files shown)


TheMuffinPirate:
Excellent! Not being able to nuke tiamat was proving quite annoying.

scripts/globals/mobskills/Arcane_Stomp.lua
```diff
@@ -19,19 +19,9 @@ function onMobSkillCheck(target,mob,skill)
...
-    if(mob:hasStatusEffect(typeEffect) == true) then
-        oldEffect = mob:getStatusEffect(typeEffect);
-        oldEffect:setPower(power);
-        oldEffect:setDuration(duration);
-    else
-        mob:addStatusEffect(typeEffect,power,tic,duration);
+    
+    local duration = 300;
+    local typeEffect = EFFECT_MAGIC_SHIELD;
+    skill:setMsg(MobBuffMove(mob,typeEffect,3,0,duration));
```
xdemolish: change this (and the others) to use a var when setting the message, for readability e.g
```
local msg = MobBuffMove(mob,typeEffect,3,0,duration);

skill:setMsg(msg)
```

Hozu:
Done, sorry about that.

scripts/globals/mobskills/Pleiades_Ray.lua
```diff
@@ -15,23 +15,28 @@ require("scripts/globals/monstertpmoves");
 ---------------------------------------------
 
 function onMobSkillCheck(target,mob,skill)
-	return 0;
+    local mobhp = mob:getHPP();
+    if (mobhp <= 20) then
+        return 0;
+    end;
+
+    return 1;
```
TeoTwawki: Should use an else to put that in or variablize the return value.

Hozu:
Good point to have some sort of failsafe. I like the second solution so I went with that. Also did the same with Fulmination because that's where I copied the HP check from, and cleaned up the whitespace on it.

TeoTwawki:
@Hozu keep up the good work :)

Hozu:
I noticed that Vrtra has the same thing as Tiamat's duplicated mdb buff, but it's regen. 125 from the db, 30 from scripts. Not sure how much it's supposed to have though so I didn't touch it, though I can assure you 155/tic makes it pretty much impossible to kill on a 75 era server.

TeoTwawki:
@Hozu Just my opinion.: Sounds like someone didn't understand that dsp *usually* does mob's modifiers in a db table, does their own in script (I tend to do my non-dsp thinsg in the scripts instead of the tables) and accidentally doubled it up during some pull request after merging DSP updates to their own fork somewhere in the past. If both the db and the script use the same value, we can just delete the scripted line for it and if the db value for it is wrong, someone can adjust that.

TeoTwawki:
The db one is prolly right are closer to correct than the script, because dsp normally does all the modifiers for mobs in the db while people customizing their fork tend to do them in scripts.

Hozu:
Ok I removed the extra 30/tic regen from the scripts since the DB has 125/tic already. Wish I could get retail data for its regen because I still think 125/tic is excessive.

TeoTwawki:
Shouldn't be to hard for somebody to get that number from retail later and adjust the db entry as needed. They just need to use a plugin/addon for either windower or ashita that tells them the mobs exact HP and see how much of their damage they just did gets healed per tick.

teschnei:
I wrote the vrtra script, and I am still debating removing the mod db
tables entirely. As for finding out the regen, the server does not send
exact hp, you can only go by %

Hozu:
Ah, I see. Should I undo the Vrtra change and move Tiamat mods to the script instead?

TeoTwawki:
> I am still debating removing the mod db

@teschnei I thought the only reason those tables still existed was so you didn't have to individually script multiple monsters that otherwise would not need any script at all.

teschnei:
well, I can see merit to the family mod table for the exact reason you described, and probably pools I guess since a pool is across zones and a script isn't, it's just annoying to have 2 places that can do the same thing (the reason fairly obvious, lol).  I haven't done anything about it yet because I'm not sure how I wanted to handle it, but I do think that mods for specific things (NMs, or a mob that only appears in one zone) should be in the script only, leaving the db ones specifically for modifiers to multiple monsters across a family or pool

TeoTwawki:
Another advantage of based mods over db to consider: not having to either remember wtf this line is or maintain sql comments.

Hozu:
What should I do with this then? Move the mods for Tiamat and Vrtra into their scripts instead of the DB? Or is it fine as it currently is?

Deadwing888:
To try to get some info on this I just shot some footage of Vrtra. The HP number appears to be roundabout 100k vs dsp's 40k. I didn't get a perfect reading on his regen due to some... blood aggro, but what I did get suggested his regen was in the neighborhood of 50 hp a tick.

Hozu:
100k!? Oh man... ffxiclopedia states Vrtra to have 40k HP, Jormundgand to have 80k HP, and Tiamat to have 120k HP. Did they get buffed at some point?

Deadwing888:
No I think the 40k number is pure fantasy someone made up. 40k can be zerged with 2 kclub drks and it took at least 6 way back when

Hozu:
After looking at both Tiamat and Jormundgand on bgwiki, it says both have 100k HP. Perhaps all three of them have 100k HP and the same amount of auto regen?

Deadwing888:
In terms of tiamat HP I parsed a roughly hour-long tiamat that had roughly 120k damage done to him. This lead me to believe he had about 100k hp with about a 25 hp per tick regen. That's some 6 year old memories combined with guesswork. Jorm we're camping right now to get hp and regen numbers I think 80k is a smidge low

korietsu:
Glad to see Hozu's putting in the hotfixes from our testing.

Deadwing888:
Jormungand 100k hp CONFIRMED

teschnei:
next time, please do not combine unrelated changesets into a single pull request - it makes it much harder to review

scripts/zones/Halvung/mobs/Gurfurlur_the_Menacing.lua
```diff
@@ -18,11 +18,41 @@ end;
 
 function onMobEngaged(mob,target)
 
-	SpawnMob(17031593,180):updateEnmity(target);
-	SpawnMob(17031594,180):updateEnmity(target);
-	SpawnMob(17031595,180):updateEnmity(target);
-	SpawnMob(17031596,180):updateEnmity(target);
+    mob:setLocalVar("Gurfurlur", mob:getID());
```
teschnei: why set a local var as his ID? why not just use getID when you need the ID?

Hozu:
That... is a very good question. What the hell was I thinking? Changed it to not use a local var.

And yeah, sorry about making changes to many different things in one submission. Someone else compared it to shoving a gigantic bill through Congress.

PR #1658, opened 2015-07-05
CORE: Cure cast time reductions

Jupiter065:
White mage merit cure cast reduction
item bonus cure cast reduction

### DIFF
--- a/src/map/ai/states/magic_state.cpp
+++ b/src/map/ai/states/magic_state.cpp
@@ -254,8 +254,17 @@ uint32 CMagicState::CalculateCastTime(CSpell* PSpell)
         cast = cast * (1.0f - ((songcasting > 50 ? 50 : songcasting) / 100.0f));
     }
 
-    int16 fastCast = dsp_cap(m_PEntity->getMod(MOD_FASTCAST),-100,50);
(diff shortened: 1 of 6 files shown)


src/map/ai/states/magic_state.cpp
```diff
@@ -253,6 +253,16 @@ uint32 CMagicState::CalculateCastTime(CSpell* PSpell)
...
     }
+    if (PSpell->isCure())
+    {
+        int32 cureCast = m_PEntity->getMod(MOD_CURE_CAST_TIME);
+        if (m_PEntity->objtype == TYPE_PC)
+        {
+            cureCast += ((CCharEntity*)m_PEntity)->PMeritPoints->GetMeritValue(MERIT_CURE_CAST_TIME, (CCharEntity*)m_PEntity);
+        }
+        cureCast = dsp_cap(cureCast, 0, 100);
+        cast = cast * ((100.0f - cureCast) / 100.0f);
```
teschnei: you're going to have to verify that cure cast is multiplicative with fast cast and not additive

Jupiter065: Ok, what would the cap be then? 80% like in that thread?

Jupiter065:
Ok, I've made it additive with fast cast and capped it at 80%.

teschnei:
yep, 80% is right

PR #1672, opened 2015-07-06
SCRIPTS: Added BLM Potency Merits Checks

Zenny89:
This is my attempt at implementing BLM Potency Merits. Please let me know if I stuck this code in the wrong function or I made any mistakes.

SCRIPTS: Added BLM Potency Merits Checks


TeoTwawki:
@Zenny89 is that the only file to change at all?
does anything in your local copy have a conflict mark?
if you're a tortoise user should be a yellow ! mark

PR #1682, opened 2015-07-08
SCRIPTS: Checks for Avatar Magical Attack Merits

Zenny89:


teschnei:
I don't think this is the right way to do this. You can't handle physical
attack or accuracy in this fashion, thus it makes more sense to handle this
in the same place you'd handle those

Zenny89:
I agree, but I can't figure out where to handle those.

teschnei:
Easiest spot would be to just add the stats when their stats are initially
calculated. Petutils I think has a stat calculation function for avatars

PR #1686, opened 2015-07-09
CORE: Added SMN T1 Merits

Zenny89:


src/map/utils/petutils.cpp
```diff
@@ -1306,7 +1306,11 @@ namespace petutils
                     }
                 }
             }
-
+	     CCharEntity* PChar = (CCharEntity*)PMaster;
```
teschnei: you need to check if the PMaster is a character: C-style casts don't check this for you, it'll just crash or do other horrible things

if (PMaster->objtype == TYPE_PC)

Zenny89:
Made correction

src/map/utils/petutils.cpp
teschnei: now move the cast inside the if and indent the if properly and it's good to go

TeoTwawki:
oh nope fooled me because you have a tab and some spaces in the same leading space and github renders that stupidly..Don't do that. *ever*.

Zenny89:
Closed because I changed my repository - see other pull request.

PR #1697, opened 2015-07-11
Scripts/SQL: Enhances Rice Ball effect + Hachiman TP latent fix

Hozu:
I've added the bonus effects of each rice ball to each piece of gear as food-activated latents using food ID as the parameter. I removed the old system for the rice ball as it was granting the boost when the gear was not equipped - that was not retail behaviour. Fix for issue #1963.


Hozu:
I noticed that the Hachiman TP latent was using TP > 100 so I corrected that as well.

PR #1705, opened 2015-07-12
SCRIPTS: Globals title variable fix

The-Halcyon-Days:
Intends to fix global variable issues with quest that give a title that has the same name as it.
This would change title global variables to have the 'TITLE_'.
This request should change all instances in which these variables are used to the new variable name.


teschnei:
Probably easier to put them into a global table names titles. Then you'd do
titles.title_name

The-Halcyon-Days:
Noticed some issues with the method/Regex I used and I think this will take more work s just closing for now.

Issue #1425, opened 2015-05-13
Zilart: BCNM Temple of Ugga causing hangup


Desufire:
There are no errors that I can find, otherwise I'd upload a dump, but players are hanging up after win in BCNM Temple of Uggalepih. The cs plays and Grav'iton says:

"It all started with a mere crystal and the mysterious power that it held..." and then the game hangs on a black screen. 
Commands release and posfix do nothing to help. bring and homepoint get the player away from black screen, but mission flag does not update doing this and they must do the fight again.

I've looked into the scripts and couldn't see anything wrong with them.

gedads:
remind me a fix from long ago... https://www.youtube.com/watch?v=5r4zcFHcLeU
where are you stuck, @ first cs, second cs or exit?

Desufire:
First cs transitioning into the second.

Reported to me that they waited a few minutes in case of long load time. I didn't know I could add the KI to access the fight, so I'll be doing that now to see for myself. I hope I didn't just report this for no reason.

Desufire:
Edit: After testing it, I was able to recreate the problem on 2 other GMs. Each allowed 5 minutes plus for CS to load and nothing happened. The only thing being shown is chat box. But again, there are no errors being shown on server logs, so I don't know what's going on.

gedads:
ok i'll have a look.

AlexandreCA:
the same problem.

gedads:
same issue https://github.com/DarkstarProject/darkstar/issues/1411

teschnei:
7b76bdaaa7993ed9a9e1fefb84b8b6de9d584993

AlexandreCA:
I think there is still a small problem with the 2nd CS or we are told the story of Zillart . A black screen appears . The video in 4:03 min show good the situation.  

 https://www.youtube.com/watch?v=5r4zcFHcLeU

teschnei:
3fe203c

AlexandreCA:
Thank you , but no. The problem has not changed :(

TeoTwawki:
@teschnei its stalling at the chang over from cs 0x0007 to 0x0008, 1 cs plays fien (unless they use the eternity plugin, which usually breaks it) and the 2nd never starts. The zone.lua script has onEventFinish for the 1st start the second.

if I (at)cs them, these both play just fine even causing one to trigger the other, because of that weirdness where the zone lua gets read for (at)cs.

teschnei:
[...]

PR #1727, opened 2015-07-15
Scripts: ZM4 CS Hanging Fix

The-Halcyon-Days:
Should fix issue #1425
Adjusts Event parameters to get desired result
Mapped out event parameters for this BCNM


The-Halcyon-Days:
Does not fix hanging after 3rd CS in the chain and unable to find a solution for now after some testing so just closing the pull request for now.

PR #1735, opened 2015-07-16
SCRIPTS: San d'Oria mission 2-3

b03605079:
[Before] can't proceed after getting mission from gatehouse.
[After] fixed.

### DIFF
--- a/scripts/zones/Chateau_dOraguille/npcs/Halver.lua
+++ b/scripts/zones/Chateau_dOraguille/npcs/Halver.lua
@@ -110,7 +110,7 @@ function onTrigger(player,npc)
 		-- Mission San D'Oria 2-3 Journey Abroad 
 		elseif(currentMission == JOURNEY_ABROAD and MissionStatus == 11) then
 			player:startEvent(0x01fb);
-		elseif(currentMission == JOURNEY_ABROAD and MissionStatus == 0) then
+		elseif(currentMission == JOURNEY_ABROAD and MissionStatus == 1) then
 			player:startEvent(0x01f9);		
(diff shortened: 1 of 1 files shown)


scripts/zones/Chateau_dOraguille/npcs/Halver.lua
```diff
@@ -110,7 +110,7 @@ function onTrigger(player,npc)
 		-- Mission San D'Oria 2-3 Journey Abroad 
 		elseif(currentMission == JOURNEY_ABROAD and MissionStatus == 11) then
 			player:startEvent(0x01fb);
-		elseif(currentMission == JOURNEY_ABROAD and MissionStatus == 0) then
+		elseif(currentMission == JOURNEY_ABROAD and MissionStatus == 1) then
```
TeoTwawki: Where is this set to the one value of 1 that it is checking for?

TeoTwawki: If at the very start of a mission your players have a var of 1 instead of zero, that's caused by another bug someplace else, because all missions start with zero. The gate guard should be simply adding the mission without setting a var so 1st NPC after that should be looking for zero.

b03605079: I think i found out what the problem is. The door npc of papal chamber screw up everything.
Thanks homie!

TeoTwawki: At the end of every mission, the status variable should be reset to zero. if you found the one that isn't being reset and fixed that, awesome and thank you.

PR #1751, opened 2015-07-18
SQL: Weather latents + Crafting torques

Hozu:
Implemented weather latents for whatever items I could find on the wiki.
Implemented the crafting torques that weren't already in the database.


Hozu:
A typo, oops. Fixed.

Hozu:
So I just noticed that the Almah Torque did not have the Store TP mod. I thought it had no mods set but nope, someone gave it the Subtle Blow mod but not the Store TP one. Fixed now.

PR #1754, opened 2015-07-19
Update merits to match retail

some1here:


TeoTwawki:
Much better :thumbsup:

PR #1758, opened 2015-07-19
CORE: Added SMN T1 Merits

Zenny89:
CORE: Added SMN T1 Merits

### DIFF
--- a/src/map/utils/petutils.cpp
+++ b/src/map/utils/petutils.cpp
@@ -1307,6 +1307,16 @@ namespace petutils
                 }
             }
 
+            
+            if (PMaster->objtype == TYPE_PC)
+            {
+                CCharEntity* PChar = (CCharEntity*)PMaster;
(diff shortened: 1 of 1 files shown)


src/map/utils/petutils.cpp
```diff
@@ -1307,6 +1307,15 @@ namespace petutils
                 }
             }
 
+            CCharEntity* PChar = (CCharEntity*)PMaster;
```
teschnei: still need to move this into the if

TeoTwawki:
@Zenny89 you can use the "amend" feature to edit your last commit without adding a new one (so you won't have to squash again) and then force push to update this pull request with the change teschnei pointed out.

Zenny89:
I deleted my last repository while playing around with GitHub last week, so that's why it's a new request. I don't quite understand why/how I can move that into the if without the "If" statement making sense.

Zenny89:
Corrected

PR #1764, opened 2015-07-20
Sql Item mod update

some1here:


some1here:
There is some reason that the system has errors with the descriptions of the mods put past the ";" will error the system and not allow changes.. I am not sure as to why.

"@@ -43,59 +43,43 @@ CREATE TABLE IF NOT EXISTS item_mods (" reverted too after changes

PR #1765, opened 2015-07-20
SQL: CoP Wyrm adjustment

Hozu:
Changed the HP of Tiamat, Vrtra, and Jormungand to 100k, and their auto-regen to 25/tic. This was after a discussion with Deadwing888, who has collected a lot of HNM data from retail. I'm certain that Vrtra was previously given a high rate of auto-regen because the wiki value of 40k HP was used, yet actual fights on retail seemed to be long for it to have only 40k HP.
Changed the name prefix on Tiamat (no more "the Tiamat").
Changed the behaviour on Cynoprosopi (NW Apollyon wyrm) to match Fafnir/Nidhogg (no turn).


Deadwing888:
The KS99 "Wyrm" needs noturn flagged as well

Hozu:
Good point, I never checked that one.

PR #1767, opened 2015-07-20
SQL\SCRIPTS: Camisado/Noctoshield

Zenny89:
Added Camisado and Noctoshield as well as framework for other Diabolos BPs.

### DIFF
--- a/scripts/globals/abilities/pets/camisado.lua
+++ b/scripts/globals/abilities/pets/camisado.lua
@@ -1,0 +1,28 @@
+---------------------------------------------------
+-- Camisado M=3.5
+---------------------------------------------------
+
+require("scripts/globals/settings");
+require("scripts/globals/status");
+require("scripts/globals/monstertpmoves");
+require("scripts/globals/summon");
+
+---------------------------------------------------
+
+function onAbilityCheck(player, target, ability)
+    return 0,0;
+end;
+
+function onPetAbility(target, pet, skill)
+	local numhits = 1;
+	local accmod = 1;
+	local dmgmod = 3.5;
+
+	local totaldamage = 0;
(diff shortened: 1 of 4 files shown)


Hozu:
Yo, you forgot to make it so noctoshield lasts longer with master's summoning magic skill over the cap. It should be like how Hastega does it IIRC: https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/abilities/pets/hastega.lua

Zenny89:
Not according to the wiki. Compare http://ffxiclopedia.wikia.com/wiki/Noctoshield with http://ffxiclopedia.wikia.com/wiki/Frost_Armor or http://ffxiclopedia.wikia.com/wiki/Hastega_%28Blood_Pact%29

Hozu:
http://forum.square-enix.com/ffxi/threads/20767-Blood-Pact-Ward-testing calc the duration from here

Hozu:
hm looks like most BPs will have to be revamped because mod_summoning just returns the skill directly? From modifier.h: MOD_SUMMONING = 0x75, // Summoning Magic Skill (modId = 117)

So it looks like it should be if the master has more than 300 skill, duration is 180 + (skill - 300).

Hozu:
Oh no... it turns out that the way the skill amount is being called upon, it's not doing it right. You have to do the following to get the summoner's skill: 
summoner:getSkillLevel(38)
where 38 is the number for summoning skill.

So this means all BPs don't have any sort of extended time on them because it's calling 0. Time to fix that. Thank you for this submission because now I discovered that issue.

Do duration = 180 + bonusTime, where bonusTime is summoner:getSkillLevel(38) - 300, making sure to cap it at +200 max and 0 min. Then you'll be good for Noctoshield.

Hozu:
Actually, see my latest pull request, since I've had to make some changes.

Zenny89:
Updated with Hozu's changes.

PR #1768, opened 2015-07-20
Adds all items Lv76-99 currently in AH.

Dalthine:
Edits item_mods.sql to add all items currently showing on the Auction House for Lv76-99.
Includes nonfunctional or incorrectly categorized items (ref: Ranged category showing a dagger and greatswords, and Lv.99 items that appear at the bottom of categories).
Made a few small fixes here and there for issues that I encountered while working.

### DIFF
--- a/sql/item_mods.sql
+++ b/sql/item_mods.sql
@@ -1024,6 +1024,16 @@ INSERT INTO `item_mods` VALUES (10451, 421, 5); -- Increases critical hit damage
 INSERT INTO `item_mods` VALUES (10452, 369, 1);
 
 -- -------------------------------------------------------
+-- Praeco Doublet
+-- -------------------------------------------------------
+INSERT INTO `item_mods` VALUES (10453, 1, 50); -- Defense + 50
+INSERT INTO `item_mods` VALUES (10453, 12, 15); -- INT + 15
+INSERT INTO `item_mods` VALUES (10453, 13, 15); -- MND + 15
+INSERT INTO `item_mods` VALUES (10453, 14, 15); -- CHR + 15
+INSERT INTO `item_mods` VALUES (10453, 30, 5); -- Magic Accuracy + 5
+INSERT INTO `item_mods` VALUES (10453, 455, 10); -- Song Spellcasting Time + 10
+
+-- -------------------------------------------------------
 -- Asura Samue
 -- -------------------------------------------------------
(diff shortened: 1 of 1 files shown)


TeoTwawki:
Gave it a look in TortoiseGitMerge since github won't show everything, and looks pretty good!

You didn't use the commit labels though, which the devs want people doing now. :(
So you'll need to squash and use the SQL label `SQL: blah blah blah` in the commit message.

teschnei:
no i don't

TeoTwawki:
mmmK

xdemolish:
it only really bothers me when the commit message is ambiguous to be honest (aka all my own commit messages lol)

teschnei:
as noted by Loki, you need to remove the skill bonuses from item level from all the weapons.  they are already accounted for in item_armor/item_weapon

you can use this query to see the offending rows: SELECT * FROM item_mods where modId >= 0x50 and modId <= 0x7A and value > 30;

teschnei:
oh, except shields I think

Dalthine:
Strange.  Does it not display the iLvl skill additions on your stats, in that case?  Because I was testing some of the items out and the skill seemed correct when keeping it in mods.  If it's a hidden stat though I'll remove those skills from the mods file.

Dalthine:
As an example, I equipped the Ohrmazd (iLvl119 hand-to-hand), and my skill goes from 424 to 666, which represents the skill addition on the weapon.  And it has the iLvl skill on the item_weapon table.  It seems like if it was doubled I should have 908 skill.

teschnei:
you don't get the bonus skill on your skill page (except for shield, parry, and guard) - check your attack and accuracy (or check parry/guard since those should be doubled)

Dalthine:
Guarding went from 417 to 659, which is correct based on the mods changes.  I think Attack is showing the jump, though, going up the 484 (double) points.  So, as I understand it I should remove all of the iLvl stats except shield?  I've been in the process of doing so in the meantime.

teschnei:
Yes, that's right

Dalthine:
Okay.  Just created pull request #1776 to fix.

TeoTwawki:
Sorry, I didn't see that at all when I looked. Was looking for bad modID's and sloppy syntax etc.

Dalthine:
It was my fault.  I apologize for the trouble.

PR #1769, opened 2015-07-21
Scripts: BP Ward duration bonus

Hozu:
The way the skill was being obtained previously as always returning a value of 0, so I changed it. Now it grabs the master's summoning magic skill. Each BP ward has a base duration, and a bonus of one second for every point of skill above 300, capping out at 500 skill.

Source: http://forum.square-enix.com/ffxi/threads/20767-Blood-Pact-Ward-testing

### DIFF
--- a/scripts/globals/abilities/pets/crimson_howl.lua
+++ b/scripts/globals/abilities/pets/crimson_howl.lua
@@ -5,6 +5,7 @@
 require("scripts/globals/settings");
 require("scripts/globals/status");
 require("scripts/globals/monstertpmoves");
+require("scripts/globals/utils");
 
 ---------------------------------------------------
 
@@ -13,8 +14,10 @@ function onAbilityCheck(player, target, ability)
 end;
 
 function onPetAbility(target, pet, skill, summoner)
-	local duration = 30 + summoner:getMod(MOD_SUMMONING);
-	target:addStatusEffect(EFFECT_WARCRY,9,0,duration);
-	skill:setMsg(MSG_BUFF);
-	return EFFECT_WARCRY;
+    local bonusTime = utils.clamp(summoner:getSkillLevel(SKILL_SUM) - 300, 0, 200);
+    local duration = 60 + bonusTime;
+
+    target:addStatusEffect(EFFECT_WARCRY,9,0,duration);
+    skill:setMsg(MSG_BUFF);
+    return EFFECT_WARCRY;
 end
--- a/scripts/globals/abilities/pets/ecliptic_growl.lua
+++ b/scripts/globals/abilities/pets/ecliptic_growl.lua
@@ -5,6 +5,7 @@
 require("scripts/globals/settings");
 require("scripts/globals/status");
 require("scripts/globals/monstertpmoves");
+require("scripts/globals/utils");
 
 ---------------------------------------------------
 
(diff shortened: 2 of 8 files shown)


scripts/globals/abilities/pets/crimson_howl.lua
```diff
@@ -13,8 +13,15 @@ function onAbilityCheck(player, target, ability)
...
 
 function onPetAbility(target, pet, skill, summoner)
-	local duration = 30 + summoner:getMod(MOD_SUMMONING);
-	target:addStatusEffect(EFFECT_WARCRY,9,0,duration);
-	skill:setMsg(MSG_BUFF);
-	return EFFECT_WARCRY;
+    local bonusTime = summoner:getSkillLevel(38) - 300; -- summoning skill
+    if (bonusTime < 0) then
+        bonusTime = 0;
+    elseif (bonusTime > 200) then
```
TeoTwawki: There is utils.clamp() for minimum/maximum restricting values letting you write shorter code. :smiley: 

`local bonusTime = utils.cliamp(summoner:getSkillLevel(38) - 300, 0, 200);`
2 lines instead of 6.  (you'd need to require the utils.lua global)
`require("scripts/globals/utils");`

Hozu:
Oh hey, thanks. Changed my submission to use that function.

Hozu:
Fixed comment, because it's not Blood Pact doesn't shorten to BW.

scripts/globals/abilities/pets/crimson_howl.lua
```diff
@@ -13,8 +14,10 @@ function onAbilityCheck(player, target, ability)
 end;
 
 function onPetAbility(target, pet, skill, summoner)
-	local duration = 30 + summoner:getMod(MOD_SUMMONING);
-	target:addStatusEffect(EFFECT_WARCRY,9,0,duration);
-	skill:setMsg(MSG_BUFF);
-	return EFFECT_WARCRY;
+    local bonusTime = utils.clamp(summoner:getSkillLevel(38) - 300, 0, 200);
```
teschnei: don't use a magic number, there are skill enumerations in status.lua

Hozu:
Yeah, that wasn't such a good idea when I look back at why BLU enfeebles weren't working. Changed it and tested and it yielded the desired results.

scripts/globals/abilities/pets/crimson_howl.lua
```diff
@@ -13,8 +14,10 @@ function onAbilityCheck(player, target, ability)
 end;
 
 function onPetAbility(target, pet, skill, summoner)
-	local duration = 30 + summoner:getMod(MOD_SUMMONING);
-	target:addStatusEffect(EFFECT_WARCRY,9,0,duration);
-	skill:setMsg(MSG_BUFF);
-	return EFFECT_WARCRY;
+    local bonusTime = utils.clamp(summoner:getSkillLevel(SUMMONING_SKILL) - 300, 0, 200);
```
TeoTwawki: I suggest using the ones in status.lua for now: https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/status.lua#L1889 since that global is already required here anyway.. For some derptarded reason different editors variable-lized some skill ID's multiple times in different globals without noticing each other's work. I plan to eventually make a pull request standardizing them all into a single place and single naming convention and I'll adjust all scripts using the existing ones at that time.

TeoTwawki: SUMMONING_SKILL would only work here if you added a require for magic.lua

TeoTwawki: it likely worked when you tested because some other thing had already loaded magic.lua just coincidently

Hozu:
Surprisingly, SUMMONING_SKILL did work without including it. But I should probably add it anyway. Using MOD_SUMMONING only returns the skill granted by gear, not the total skill, so it's not useful.

TeoTwawki:
I'd just go with SKILL_SUM rather than add another require, because its gonna get changed anyway.

Hozu:
Fixed with SKILL_SUM.

Issue #1772, opened 2015-07-21
Jettatura crashes server


Dalthine:
Was just experimenting a bit last night to find out what blue magic does and doesn't work, and discovered that trying to cast Jettatura will cause the server to crash. It might need to be disabled similar to head butt and voracious trunk until it can be fixed.

Dalthine:
Pinecone bomb, not head butt.

Hozu:
Whoops I forgot to remove the use of the var "power" for it when I was doing copypasta, my bad.

Hozu:
My pull request has been accepted - this issue is now resolved.

PR #1773, opened 2015-07-21
Scripts: Gaze check for BLU enfeebles + fix for #1772

Hozu:
Awful Eye, Blank Gaze, Chaotic Eye, Jettatura, and Light of Penance check if both caster is facing target and if target is facing caster.
Yawn is different - it only needs to check if the target is facing the caster.
Set the power for the terror on Jettatura to 1 instead of an undefined variable.


Hozu:
I did not know that BLU gaze spells only checked to see if the target was facing the caster. Removed the additional check of caster facing the target.

PR #1774, opened 2015-07-21
SCRIPTS: Mission 4-1 Bastok citizen check & event ID

b03605079:
Adding Bastok citizen check & fixing some incorrect event ID
[Before] Sandy & Windy Rank4 citizen could get Bastok m4-1 that makes "MissionStatus" set to '1'
[After] fixed

### DIFF
--- a/scripts/zones/RuLude_Gardens/npcs/Goggehn.lua
+++ b/scripts/zones/RuLude_Gardens/npcs/Goggehn.lua
@@ -53,7 +53,7 @@ function onTrigger(player,npc)
 	         	player:startEvent(0x0023);
 	        elseif(player:hasKeyItem(MESSAGE_TO_JEUNO_BASTOK)) then
 		        player:startEvent(0x0037);
-		elseif(player:getRank() <= 4) then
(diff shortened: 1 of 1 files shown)


TeoTwawki:
cs 0x0004 never gets used for anything?

b03605079:
that's a good point. in fact there're same problems for sandy & windy's npc.
let me have a look in retail version.

b03605079:
tested in retail: 
[before m3-3] cs 0x0004
[after m3-3] cs 0x0065

TeoTwawki:
https://github.com/DarkstarProject/darkstar/pull/1774/commits
In the future, please make commit messages descriptive of the changes. "updated x" tells the reader of your pull request absolutely nothing and should someone need to look back in the history makes this change a pain in the rear to locate.

If you squash and force push you can still change the commit message here. Something like what you have as the pull request title would be appropriate.

If you are using either github for windows or the github website to do your edits..Those suck. git bash or tortoise git (or both!) are generally the best way to go if on windows. I dunno about linux, only ever command lined things there.

xdemolish:
if you could do what teo suggested i can blind merge and hope nothing breaks

PR #1789, opened 2015-07-23
Core/SQL: Implement AGI boost version of "Enhances Trick Attack"

Hozu:
Mod give % AGI bonus to Trick attack, and if the mod is on a piece of equip, it has to be equipped on hit.
Added said mod to Rogue Armlet's +1 (15% bonus AGI).
Gave a proper description for my "Enhances Drain/Aspir" mod from earlier.


Joeyjohjjo:
Thanks for doing this. However, shouldn't Vajra (ID 18996) also have a modifier of 10?

Hozu:
Vajra "enhances Sneak Attack" is actually different - it increases the critical damage by 10%. That's why I included AGI when I named the modifier. https://www.bg-wiki.com/bg/Vajra_%28Level_75%29

teschnei:
It's just flat damage bonus actually - it stacks differently than with
actual critical hit damage

Hozu:
Oh? Interesting. And that is why I didn't add the mod for Vajra. I did THF AF+1 hands by request.

teschnei:
note that since SA/TA are always crits, they work out to be the same except in regards to stacking with other crit damage, so it's not really a big difference

either way, it does not multiply DEX/AGI like rogues/raiders gloves do

PR #1794, opened 2015-07-24
Spirit Within

Turttle:
Adjusted the ws to match new TP values. Corrected calculations using
correct old values from:
BGwiki
https://www.bg-wiki.com/index.php?title=Spirits_Within&oldid=104829 and
FFXIclopedia
http://ffxiclopedia.wikia.com/wiki/Spirits_Within?oldid=306648

Used updated formula for "new - Adoulin" Values
https://www.bg-wiki.com/bg/Spirits_Within

### DIFF
--- a/scripts/globals/weaponskills/spirits_within.lua
+++ b/scripts/globals/weaponskills/spirits_within.lua
@@ -7,9 +7,11 @@
 -- Delivers an unavoidable attack. Damage varies with HP and TP.
 -- Not aligned with any "elemental gorgets" or "elemental belts" due to it's absence of Skillchain properties.
 -- Element: None
+-- Modifiers: HP (Damage is calculated by TP Multiplier times current HP) 
 -- Modifiers: HP:
--- 100%TP    200%TP    300%TP
--- 12.5%       50%      100%
+-- 1000%TP    2000%TP    3000%TP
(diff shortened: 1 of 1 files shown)


scripts/globals/weaponskills/spirits_within.lua
```diff
@@ -22,11 +23,11 @@ function onUseWeaponSkill(player, target, wsID)
 	local TP = player:getTP();
 	local WSC = 0;
 	if (TP == 300) then
-		WSC = HP * 0.46875;
+		WSC = HP * (15/32);
```
teschnei: these need to be scaled linearly - spirits within does not use breakpoints

scripts/globals/weaponskills/spirits_within.lua
```diff
@@ -34,9 +35,9 @@ function onUseWeaponSkill(player, target, wsID)
 		if (TP == 300) then
 		WSC = HP;
 		elseif (TP >= 200) then
-		WSC = HP * .5
+		WSC = floor(HP * ( floor(0.72*TP) - 96 ) / 256 )
```
teschnei: these numbers are the old formula: ((300*0.72) - 96) / 256 = 15/32

PR #1797, opened 2015-07-25
Scripts/SQL: Implement Hpemde family TP moves

Hozu:
Corrected the animations that were used, as only Sinuate Rush was using the correct one. For some reason, they're separated by other animations like player WS ones. Strange...
Swapped the mob skill ID of Sinuate Rush and Temporal Shift - this fixes the issue where one of those moves would call the script of the other. This meant I also had to change the ID in blue_spell_list.sql so Temporal Shift would be learned from the correct move.

### DIFF
--- a/scripts/globals/mobskills/Ichor_stream.lua
+++ b/scripts/globals/mobskills/Ichor_stream.lua
@@ -1,0 +1,25 @@
+---------------------------------------------
+--  Ichor Stream
+--  Family: Hpemde
+--  Description: Spews venomous ichor at targets in a fan-shaped area of effect.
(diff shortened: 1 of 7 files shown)


Leelink1:
Hey. Added these to my test server. And it was crashing on Tail_thrust. on CanGainStatusEffect,  GetStatusEffect, and addStatusEffect. Changed local typeEffect = EFFECT_PARALYZE; to local typeEffect = EFFECT_PARALYSIS; and that seems to make it work with no crashes

Hozu:
...and that's what I get for testing with godmode on, hahaha. Fixed.

PR #1799, opened 2015-07-25
SCRIPTS: Added Repair

Zenny89:


scripts/globals/abilities/repair.lua
```diff
@@ -0,0 +1,131 @@
...
+-----------------------------------
+-- onAbilityCheck
+-----------------------------------
+
+function onAbilityCheck(player,target,ability)
+    if (player:getPet() == nil) then
+        return MSGBASIC_REQUIRES_A_PET,0;
+    else
+        local id = player:getEquipID(SLOT_AMMO);
+        if(id >= 18731 and id <= 18733 or id == 19185) then
```
TeoTwawki:
```lua
        if ((id >= 18731 and id <= 18733) or id == 19185) then
```

Makes it clearer what you are doing and avoids future probs when the next editor adds any new item SE creates that would get added here..

PR #1803, opened 2015-07-25
Scripts/SQL: Implement Elemental Siphon

Hozu:
Changed the message in abilities.sql to match the retail message ([Player] regains [mp] MP)
Elemental Siphon has a pet check on use to make sure it's a spirit, then on ability use it checks the element of the spirit (bit of shuffling here), the day/weather, spirit's MP, and the difference between player's max and current MP.

### DIFF
--- a/scripts/globals/abilities/elemental_siphon.lua
+++ b/scripts/globals/abilities/elemental_siphon.lua
@@ -1,0 +1,78 @@
+-----------------------------------
+-- Ability: Elemental Siphon
+-- Drains MP from your summoned spirit.
+-- Obtained: Summoner level 50
+-- Recast Time: 5:00
+-- Duration: Instant
+-----------------------------------
+require("scripts/globals/settings");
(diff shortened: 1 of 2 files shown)


scripts/globals/abilities/elemental_siphon.lua
```diff
@@ -0,0 +1,85 @@
...
+        spiritEle = 4
+    elseif (spiritEle == 4) then
+        spiritEle = 2
+    elseif (spiritEle == 5) then
+        spiritEle = 6
+    elseif (spiritEle == 6) then
+        spiritEle = 3
+    end;
+
+    local basePower = player:getSkillLevel(SKILL_SUM) - 50;
```
teschnei: make sure it doesn't go negative!

Hozu:
Whoops, forgot that some people just don't like to skill their magic. Added that check to make basePower be min 0.

PR #1807, opened 2015-07-26
Scripts: Fix Qu'Hau Spring

Hozu:
Fixed it so the CS for both Divine Might and the DM request actually plays.
Extended the hour check by one hour because it's 0:00-3:00, not 0:00-2:00.

### DIFF
--- a/scripts/zones/RoMaeve/npcs/Qu_Hau_Spring.lua
+++ b/scripts/zones/RoMaeve/npcs/Qu_Hau_Spring.lua
@@ -14,18 +14,17 @@ require("scripts/globals/keyitems");
 -----------------------------------
 
 function onTrade(player,npc,trade)
-   local DMRepeat = player:getQuestStatus(OUTLANDS,DIVINE_MIGHT_REPEAT);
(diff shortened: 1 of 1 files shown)


scripts/zones/RoMaeve/npcs/Qu_Hau_Spring.lua
```diff
@@ -34,13 +33,13 @@ end;
 
 function onTrigger(player,npc)
     local CurrentMission = player:getCurrentMission(WINDURST);
-	local MissionStatus = player:getVar("MissionStatus");
+    local MissionStatus = player:getVar("MissionStatus");
 
-	if(CurrentMission == VAIN and MissionStatus >= 1) then
-		player:startEvent(2);
+    if(CurrentMission == VAIN and MissionStatus >= 1) then
```
TeoTwawki: Small request both for my OCD and my bad eyesight: could you make all if `if(` consistently `if (` please? this file has both styles in it.

Hozu:
Sure. Closing this for a sec while I re-implement the full moon check.

PR #1808, opened 2015-07-26
Spirit Within Revised

Turttle:
Corrected calculations using
correct old values from:
BGwiki
https://www.bg-wiki.com/index.php?title=Spirits_Within&oldid=104829 and
FFXIclopedia
http://ffxiclopedia.wikia.com/wiki/Spirits_Within?oldid=306648

Used updated formula for "new - Adoulin" Values
https://www.bg-wiki.com/bg/Spirits_Within

### DIFF
--- a/scripts/globals/weaponskills/spirits_within.lua
+++ b/scripts/globals/weaponskills/spirits_within.lua
@@ -41,14 +41,8 @@ function onUseWeaponSkill(player, target, wsID)
 		params.ftp100 = 0.125; params.ftp200 = 0.5; params.ftp300 = 1;
 	end
 		
-	if (TP == 300) then
-		hp_param = Hitpoint * params;-- 0.46875 @300 TP -- ADOULIN 1.000 @300 TP
-	elseif (TP >= 200) then
-		hp_param = Hitpoint * params;-- 0.18750 @200 TP -- ADOULIN 0.500 @200 TP
-	elseif (TP >= 100) then
-		hp_param = Hitpoint * params;-- 0.12500 @100 TP -- ADOULIN 0.125 @100 TP
-	end
-
+	hp_param = Hitpoint * fTP(TP, params.ftp100,params.ftp200,params.ftp300); 
+	
 	local damage = target:breathDmgTaken(hp_param);
 	damage = damage * WEAPON_SKILL_POWER
 	return 1, 0, false, damage;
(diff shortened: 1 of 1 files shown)


scripts/globals/weaponskills/spirits_within.lua
```diff
@@ -6,42 +6,51 @@
...
+	--attack multiplier (only some WSes use this, this varies the actual ratio value, see Tachi: Kasha) 1 is default.
+	params.atkmulti = 1;
+	
+	if (USE_ADOULIN_WEAPON_SKILL_CHANGES == true) then
+		params.ftp100 = 0.125; params.ftp200 = 0.5; params.ftp300 = 1;
+	end
+		
 	if (TP == 300) then
-		WSC = HP * 0.46875;
+		hp_param = Hitpoint * params;-- 0.46875 @300 TP -- ADOULIN 1.000 @300 TP
```
teschnei: you want to do hp_param = Hitpoint * fTP(TP, params.ftp100,params.ftp200,params.ftp300) instead of the whole if/else chain - it calculates the proper coefficient for you

Turttle: That make things even simpler.

teschnei: you still need the local TP = player:getTP() (the TP in the function call refers to the local)

scripts/globals/weaponskills/spirits_within.lua
```diff
@@ -6,42 +6,45 @@
...
-		WSC = HP * .5
-		elseif (TP >= 100) then
-		WSC = HP * .125;
-		end
+		params.ftp100 = 0.125; params.ftp200 = 0.5; params.ftp300 = 1;
 	end
-
-	local damage = target:breathDmgTaken(WSC);
+		
+	hp_param = Hitpoint * fTP(TP, params.ftp100,params.ftp200,params.ftp300);
```
Turttle: @teschnei Would it be better to just call "ftp" from the "scripts/globals/weaponskills" since fTP(TP, params.ftp100,params.ftp200,params.ftp300); is already defined in line 77 of the ws.lua. 
"hp_param = Hitpoint * ftp;" Tho it has "...+ bonusfTP;" at the end of that function to which I do not know how it will off set the actual weapon skill.

TeoTwawki:
whadja close fer ?

Issue #1139, opened 2015-03-23
Namas Arrow SC Properties


Kthulupwns:
Namas arrow WS is Opening and closing Darkness, when it's only SC properties are 

Skillchain Properties: Light/Distortion

It's also doing the incorrect of damage as it's hitting much harder then sidewinder, when it's damage should be marginally less then Sidewinder (as it was on retail anyway). That's easily fixed in lua though. Thanks for your time.

Fr0zt:
I just checked both sidewinder and namas arrow's scripts and they are correct. Are you testing on the same mob lvl and type for consistent results?

PR #1812, opened 2015-07-27
Weapon Skill Properties Update

Turttle:
Weapon Skill properties update. Now should do correct skillchain and get correct
items buffs acc (gorgets/obi). #1139


Hozu:
erm, I put in a print into the gorget script to tell me when the gorget bonus happens, and what DSP currently has is correct... same with the skillchain properties. I don't know why the numbers in the .sql file don't match the numbers in status.lua though.

teschnei:
SC properties are enumerated in battleentity.h

the backwards ones you are looking at is probably the SUBEFFECT ones, which are the actual value used in the packet for the effect
like Hozu, I don't think they were wrong to begin with, unless you can provide proof that they were

Turttle:
Yea I was looking at the wrong reference table >.>

PR #1814, opened 2015-07-27
Scripts: Fix for Switchstix eating 10k currency pieces.

Hozu:
Since you do not actually trade a 10k currency piece to Switchstix for your relic weapon (you trade it to the corresponding blank spot instead), I simply made a check to block the trade of each of the 10k pieces.

It's necessary for the 10k piece to be in each relic's subarray because Switchstix calls that data when you talk after having seen the stage 4 CS. However, it existing in the subarray is why it would be accepted (and subsequently eaten) prior to this fix.


teschnei:
As discussed in IRC, the final CS is wrong, but I'll merge this to stop people from losing their 10000-pieces by accident for now

PR #1816, opened 2015-07-28
Scripts/SQL/Core: NIN nuke bonuses

Hozu:
Implement T1 magic attack bonus + elemental resistance reduction duration bonus
Implement T2 San nuke magic attack bonus + magic accuracy bonus. Modified the power of these merits from 1 to 5 in merits.sql.
Implement "enhances Ninjutsu damage" mod (magic attack bonus for NIN nukes); added said mod to Koga Hatsuburi (+1). Was mistakenly a direct damage bonus rather than magic attack bonus.
Changed the tic length for Innin and Yonin from 10 seconds to 15 seconds - the effect is not supposed to fully decay at the end of the 5 minutes.
Implemented the bonus to Ninjutsu nukes from behind while Innin is active.

Sources: https://www.bg-wiki.com/bg/Category:Ninjutsu#Job_Abilities_and_Merits https://www.bg-wiki.com/bg/Innin https://www.bg-wiki.com/bg/Koga_Hatsuburi


teschnei:
i am now realizing that stuff like this is the reason all the magic functions have the "bonus" parameter - lets use it instead of checking spellids in the main nuke functions

Hozu:
Hm? Ok I'll close this for now, and take another look when I have time.

teschnei:
basically: applyResistance has an optional param "bonus", and addBonuses also has an optional param for bonus MAB, we just need to be able to pass those through doNuke, and then you can add all ninjutsu specific bonuses in doNinjutsuNuke, and all specific spell bonuses in the spell scripts themselves (as a param to doNinjutsuNuke).  for example: katon bonus in katon script calls doNinjutsuNuke with param of katon MAB/MACC/whatever bonus - doNinjutsuNuke adds all ninjutsu specific bonuses to them (innin, which i assume is actually also just a MAB bonus), and then that passes the bonusmab/macc to doNuke which does the rest of the calculations

PR #1823, opened 2015-07-30
SCRIPTS: Tateeya fix

Zenny89:


teschnei:
i highly doubt that fixes it - i'm 99% sure if i named it subid I wanted the subid

Zenny89:
I traded her a few attachments and she converted them all, and I was able to equip them all to the automaton.

teschnei:
Yes, but where did you get the attachments? If you used additem, you
probably added the wrong one (there are two itemids for every attachment,
only one is auction housable). The auction house one uses subid to point to
the real one

Zenny89:
Yes, I used additem: For Attuner, I used the ID found here: http://www.ffxiah.com/item/8453/attuner so 8543. Where can I find the correct one?

teschnei:
Search it on ffxiah. One of them will be stackable, I think it's that one

Zenny89:
Got it - closing this pull request and will check to see if this NPC works with the existing code later.

PR #1844, opened 2015-08-02
remove level sync while entering bcnm, consume ammo during reward, update speed of sandworm

Scavenge:
people were apparently able to sync before entering a bcnm, and then unsync once inside in order to remove all level restrictions and be 75 for any bcnm.  this will prevent level sync from being active prior to applying the restriction for the bcnm.  there may be a better way to do this, considering that it probably shouldn't let you initiate the bcnm until you do not have level sync active at all, or just change your level restriction from the level sync's to the battlefield's.


xdemolish:
holding off until someone with retail access confirms how it's handled

teschnei:
do not mix unrelated things into a single pull request

Scavenge:
I take it there is no way to separate changes that are all on the same branch and I'll probably have to create separate branches for them?

teschnei:
That is indeed what you are supposed to do

PR #1846, opened 2015-08-02
Update Chumimi.lua

AaronDemoncia:
Originally worked as Intended
BLM AF hands: http://ffxi.allakhazam.com/db/quests.html?fquest=411


Hozu:
I'll give you the AF hands quest, but not this one. It makes no sense from a story perspective to have solved the problem by completing AF3, then having the people try to figure out how to solve said problem in AF2 afterwards.

Also, you linked the hands quest there. http://ffxi.allakhazam.com/db/quests.html?fquest=276

PR #1848, opened 2015-08-02
Update Guslam.lua

AaronDemoncia:
As per BGwiki and Allahkazam for all Borghertz Hands quests, the previous AF quest simply needs to have been accepted.

### DIFF
--- a/scripts/zones/Upper_Jeuno/npcs/Guslam.lua
+++ b/scripts/zones/Upper_Jeuno/npcs/Guslam.lua
@@ -46,63 +46,63 @@ function onTrigger(player,npc)
 	
 	if(player:getMainLvl() >= 50 and player:getVar("BorghertzAlreadyActiveWithJob") == 0) then
 		if(player:getMainJob() == 1 and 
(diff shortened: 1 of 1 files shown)


Hozu:
If you make it check for just accepted, the quest won't be able to begin if it's completed. I think you should check if it's completed or accepted. Or rather, the original file before my change was correct and it should be reverted to that.

AaronDemoncia:
Id accept that, as it was perfectly functional as was.

AaronDemoncia:
As you mentioned, thats a valid addition and it would (hopefully) prevent issues with people whom had already completed the quests being locked out, Coded in "or QUEST_COMPLETED" flags

Hozu:
er, adding an or like that, while makes sense logically, doesn't work with code. That's how Qu Hau Spring was broken for a while. Doing ~= QUEST_AVAILABLE is the way to go, and that's what it originally had.

AaronDemoncia:
*sigh* way to go LUA ill just go ahead and do that then.

PR #1851, opened 2015-08-02
Scripts/SQL: Qiqurn TP moves, minor fixes for other mobskills

Hozu:
Scripted Faze, Kibosh, and Sandspray for Qiqurn; also scripted and enabled Deadeye with the three aforementioned moves for Cheese Hoarder Gigiroon.
Changed the accuracy down effect on Actinic Burst to Flash. Power modified to suit the change in status.
Fixed Charm mobskills - properly this time.
Enabled Digest on Claret, because it was disabled for some reason. It is supposed to use the move.

### DIFF
--- a/scripts/globals/mobskills/Deadeye.lua
+++ b/scripts/globals/mobskills/Deadeye.lua
@@ -1,0 +1,39 @@
+---------------------------------------------
+--  Deadeye
+--  Family: Qiqurn
+--  Description: Lowers the defense and magical defense of enemies within range.
+--  Type: Magical
+--  Utsusemi/Blink absorb: Ignores shadows
+--  Range: Unknown
+--  Notes: Used only by certain Notorious Monsters. Strong effect.
+---------------------------------------------
+require("scripts/globals/settings");
+require("scripts/globals/status");
+require("scripts/globals/monstertpmoves");
+---------------------------------------------
+
+function onMobSkillCheck(target,mob,skill)
+    return 0;
+end;
+
+function onMobWeaponSkill(target, mob, skill)
+
+    local defDown = false;
+    local mDefDown = false;
+
+    defDown = MobStatusEffectMove(mob, target, EFFECT_DEFENSE_DOWN, 50, 0, 120);
+    mDefDown = MobStatusEffectMove(mob, target, EFFECT_MAGIC_DEF_DOWN, 50, 0, 120);
+
+    skill:setMsg(MSG_ENFEEB_IS);
+
+    -- display defense down first, else magic defense down
+    if (defDown == MSG_ENFEEB_IS) then
+        typeEffect = EFFECT_DEFENSE_DOWN;
+    elseif (mDefDown == MSG_ENFEEB_IS) then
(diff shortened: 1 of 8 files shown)


scripts/globals/mobskills/Deadeye.lua
```diff
@@ -0,0 +1,39 @@
...
+
+    defDown = MobStatusEffectMove(mob, target, EFFECT_DEFENSE_DOWN, 50, 0, 120);
+    mDefDown = MobStatusEffectMove(mob, target, EFFECT_MAGIC_DEF_DOWN, 50, 0, 120);
+
+    skill:setMsg(MSG_ENFEEB_IS);
+
+    -- display defense down first, else magic defense down
+    if (defDown == MSG_ENFEEB_IS) then
+        typeEffect = EFFECT_DEFENSE_DOWN;
+    elseif (mDefDown == MSG_ENFEEB_IS) then
```
TeoTwawki: Did you test this out in game and see both the scenarios for messages you intended actually display in game?

Hozu:
I did a while back when I went to figure out how Murk worked from when I added some other mobskill a while back, and that skill worked (returned the message for the second status if it couldn't inflict the first). Though, since both statuses overwrite themselves if their power is the same or greater than what the target already has, it always ends up returning defense down. Not likely to get more than -50% defense.

scripts/globals/mobskills/Deadeye.lua
TeoTwawki: And retail behavior is to display only 1 of the messages if both effects land?

TeoTwawki: As long as thats retail and workin as you intended :+1:

Hozu: Surprisingly, retail only shows one status when two are inflicted at once, and no damage is inflicted. It's pretty dumb, but that's how it is.

And I remember the move now, it was Tribulation.

TeoTwawki: There's a couple on retail that show multiple messages, but most moves only do 1. And several that are "damage and this status" only show the damage part, leaving players to later notice they have this here bad effect and go "where'd that come from?!" :laughing:

teschnei: there are none that show multiple messages (it's not even possible i don't
think)

TeoTwawki: @teschnei If I ever remember which the hell one if it is that does it in retail, I'll try and get you a screen cap. And multiple messages in DSP is pretty near impossible (requires some fakery).

TeoTwawki: @teschnei I mainly just wanted to be sure that if/else had been tested, looked weird.

Hozu: Yeah that if/else does look weird. I don't really understand it either. I saw that it worked for Murk so I copied it.

TeoTwawki: Good 'nuff fer me Hozu.

PR #1852, opened 2015-08-02
Scripts: Added "enhances Trick Attack" bonus (AGI version) to WS

Hozu:
Added the AGI bonus version of "enhances Trick Attack" (THF AF+1 hands) to Weaponskill damage calculation.
Thanks to ciradan for informing me that this was necessary.


Hozu:
Fixed this conflict.

PR #1860, opened 2015-08-03
Update mountain_buster.lua

AaronDemoncia:
Adding Bind to Mountain Buster (player Blood Pact) as Monster Variant still has it but this doesn't.

### DIFF
--- a/scripts/globals/abilities/pets/mountain_buster.lua
+++ b/scripts/globals/abilities/pets/mountain_buster.lua
@@ -23,12 +23,10 @@ function onPetAbility(target, pet, skill)
 	target:delHP(totaldamage);
 	target:updateEnmityFromDamage(pet,totaldamage);
 
-	return totaldamage;
-	
-	if damage > 0 then
-
-		if(target:hasStatusEffect(EFFECT_BIND) == false) then
-			target:addStatusEffect(EFFECT_BIND, 1, 0, duration);
-		end
+	if (damage > 0) then
+	    if (target:hasStatusEffect(EFFECT_BIND) == false) then
+	    target:addStatusEffect(EFFECT_BIND, 1, 0, duration);
+	    end
 	end
+	return totaldamage;
 end
--- a/scripts/globals/abilities/pets/mountain_buster.lua
+++ b/scripts/globals/abilities/pets/mountain_buster.lua
@@ -22,8 +22,13 @@ function onPetAbility(target, pet, skill)
 	totaldamage = AvatarFinalAdjustments(damage.dmg,pet,skill,target,MOBSKILL_PHYSICAL,MOBPARAM_SLASH,numhits);
 	target:delHP(totaldamage);
 	target:updateEnmityFromDamage(pet,totaldamage);
-	
-	local typeEffect = EFFECT_BIND;
-	
+
 	return totaldamage;
+	
+	if damage > 0 then
+
+		if(target:hasStatusEffect(EFFECT_BIND) == false) then
+			target:addStatusEffect(EFFECT_BIND, 1, 0, duration);
+		end
+	end
 end
--- a/scripts/globals/abilities/pets/mountain_buster.lua
+++ b/scripts/globals/abilities/pets/mountain_buster.lua
@@ -22,6 +22,8 @@ function onPetAbility(target, pet, skill)
 	totaldamage = AvatarFinalAdjustments(damage.dmg,pet,skill,target,MOBSKILL_PHYSICAL,MOBPARAM_SLASH,numhits);
 	target:delHP(totaldamage);
 	target:updateEnmityFromDamage(pet,totaldamage);
-
+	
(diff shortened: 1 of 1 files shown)


scripts/globals/abilities/pets/mountain_buster.lua
```diff
@@ -22,6 +22,8 @@ function onPetAbility(target, pet, skill)
 	totaldamage = AvatarFinalAdjustments(damage.dmg,pet,skill,target,MOBSKILL_PHYSICAL,MOBPARAM_SLASH,numhits);
 	target:delHP(totaldamage);
 	target:updateEnmityFromDamage(pet,totaldamage);
-
+	
+	local typeEffect = EFFECT_BIND;
```
TeoTwawki: that doesn't do what you think it does.

TeoTwawki: All you did there was create  local variable named "typeEffect" and store the global variable EFFECT_BIND in it. You have no code making use of typeEffect to actually apply EFFECT_BIND to anything.

TeoTwawki: Take a peek in some other moves that both do damage and apply a status effect.

AaronDemoncia:
Lets try this again, Code taken and modified from Shadowstich, removed player TP check for duration adjustment and implamented, unsure if its best before or after the Return totaldamage but i assumed the dmg check for 0 dmg would require it to be after the check

scripts/globals/abilities/pets/mountain_buster.lua
```diff
@@ -24,4 +24,11 @@ function onPetAbility(target, pet, skill)
 	target:updateEnmityFromDamage(pet,totaldamage);
 
 	return totaldamage;
```
TeoTwawki: Nope, you still missed something important while reading the other script(s): returns end the scripts execution, it has to be last, move it right before the last "end".
Also, don't use tabs. Format with 4 spaces and () your conditions for readability.
```lua
    if (damage > 0) then
        if(target:hasStatusEffect(EFFECT_BIND) == false) then
            target:addStatusEffect(EFFECT_BIND, 1, 0, duration);
        end
    end
    return totaldamage;
end
```

TeoTwawki: Nope, you still missed something important while reading the other script(s): returns end the scripts execution, it has to be last, move it right before the last "end".
Also, don't use tabs. Format with 4 spaces and () your conditions for readability.
```lua
    if (damage > 0) then
        if (target:hasStatusEffect(EFFECT_BIND) == false) then
            target:addStatusEffect(EFFECT_BIND, 1, 0, duration);
        end
    end
    return totaldamage;
end
```

scripts/globals/abilities/pets/mountain_buster.lua
```diff
@@ -23,5 +23,10 @@ function onPetAbility(target, pet, skill)
 	target:delHP(totaldamage);
 	target:updateEnmityFromDamage(pet,totaldamage);
 
+	if (damage > 0) then
+	    if (target:hasStatusEffect(EFFECT_BIND) == false) then
+	    target:addStatusEffect(EFFECT_BIND, 1, 0, duration);
```
Hozu: You need to define duration somewhere before this line, or just use a number directly. If duration isn't holding a number, this would crash a server upon use.

TeoTwawki: @Hozu good catch

TeoTwawki: @AaronDemoncia After fixing that nil value Hozu pointed out, please squash the commits:

(In the future you can use the "amend" and "force push" features to avoid making extra commits when changing things, if you prefer).

AaronDemoncia:
Doing most of this work in browser and finallizing in github for windows, still learning lua so i dont want to just force push things without at least a second opinion crashing is no fun for anyone.

Hozu:
Amend + force push would merely replace the file in the commit with a new one, as if the old changes were never there. It wouldn't be merged into DSP. You would have to comment here afterwards so people could see it, however.

TeoTwawki:
> Doing most of this work in browser and finallizing in github for windows

Eww eww eww oh god eww. Doing work via the website or the terribad github for windows is extremely not recommended. Literally the worst way to do this they giv eyou so many problems and make learning hard. I recomend 1 or both of git for windows using git bash, or git for windows + tortoise git (which gives you a handy right click shell extension menu).

https://msysgit.github.io/
https://tortoisegit.org/

> Amend + force push would merely replace the file in the commit with a new one, as if the old changes were never there

Git Amend modifies your last commit made to the branch. Force push causes local copy to 100% overwrite the remote. Thus if you amend instead of making new commits, you can force push to replace that last commit on the remote. Being several commits in already this doesn't work for this pull request hence my saying squash.

PR #1862, opened 2015-08-03
Scripts/SQL: Jailer of Hope

Hozu:
Enabled the ability to spawn it via trade.
Gave it the ability to cast Burst II and Thundaga III instantly for up to a minute after using Plasma Charge.
Bumped up Phuabo movespeed to 70 as they are supposed to have enhanced movement speed.

### DIFF
--- a/scripts/zones/AlTaieu/mobs/Jailer_of_Hope.lua
+++ b/scripts/zones/AlTaieu/mobs/Jailer_of_Hope.lua
@@ -10,8 +10,7 @@ require("scripts/globals/status");
 -----------------------------------
 
 function onMobInitialize(mob)
-    mob:setMobMod(MOBMOD_MAIN_2HOUR, 1);
-    mob:setMobMod(MOBMOD_2HOUR_MULTI, 1);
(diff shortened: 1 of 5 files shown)


Hozu:
Fixed this conflict.

Scavenge:
Does it allow instant cast only from the time it first uses it, or does it refresh the timer while spikes are active? Seems this method could cause it to lose instant cast even if it still has spikes or reuses the move.

Hozu:
Every time Plasma Charge is used, it sets the one var to current server time + 60. The effect is supposed to remain even if spikes gets removed, which is why I wrote it the way I did. If it uses Plasma Charge while it's already able to cast spells, it just resets it to 60 seconds again.

Whenever the 60 seconds has passed, the spelllist gets set to an empty one and the var is set to 0 so it doesn't keep setting the list repeatedly (I think onMobFight happens every second or something).

Turttle:
onMobFight is called every 3 seconds

teschnei:
onMobFight has been every second for a long time... probably over a year

PR #1863, opened 2015-08-03
SCRIPTS: ZM4 CS hanging fix

Enignite:
Moved csid checks out of Zone.lua into temple_of_uggalepih.lua, tested and allowed CSs to progress and provides KI and next mission after they have finished.

### DIFF
--- a/scripts/zones/Sacrificial_Chamber/Zone.lua
+++ b/scripts/zones/Sacrificial_Chamber/Zone.lua
@@ -65,16 +65,4 @@ end;
 function onEventFinish(player,csid,option)
 	-- printf("CSID: %u",csid);
 	-- printf("RESULT: %u",option);
-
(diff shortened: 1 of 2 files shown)


Hozu:
Did you test it by winning the fight instead of using the command to play cutscenes directly? I was testing for someone else who tried to fix ZM4 and they couldn't get the third CS to play after the second one when you won the fight, but if you used the command it worked.

Enignite:
I set my current mission to Temple of Ugg, entered the BCNM and killed the tonberry, it played the CSs and then left the BCNM/gave KI/next mission (and then after it worked I tested it with 2 players, the one who had just progressed and a second I had just set mission on to make sure it didn't mess up the first players progress).

Hozu:
Excellent. I'm glad someone was able to figure this out because I sure couldn't.

PR #1866, opened 2015-08-03
Scripts/SQL: NIN T1/T2 nuke merits + Innin/Yonin adjustment

Hozu:
Innin/Yonin: Changed the tic duration from 10 to 15 - Innin and Yonin bonuses don't fully decay. They go from 30 to 10.
Added a subpower of 20 to each, which is reduced by 1 every second tic. This is because the enmity+/- effect is to start at 20 instead of 30, and they decay to 10 by the end.

Nukes: Added bonusMab param to doNinjutsuNuke and doNuke so they can pass on the a mag atk bonus to addBonus.
Added T1 mag atk bonus and resist down duration bonus to all 18 Ninjutsu nukes.
Added Innin mag atk bonus from behind (uses power of Innin effect).
Added "enhances Ninjutsu damage" mod for Koga Hatsuburi (+1) - it's mag atk bonus for Ninjutsu nukes. Removed the incorrect implementation of it from magic.lua.
Added T2 mag atk bonus and mag acc bonus to the 6 San nukes. This required an update to merits.sql (changed power from 1 to 5).


Hozu:
Fixed conflicts.

teschnei:
still has conflicts

Hozu:
Oh I see why, the absorb mod on liberator was accepted first. One moment.

Hozu:
Re-fixed conflicts by changing the mod number in modifer.h, status.lua, and in the entries in item_mods.sql.

Hozu:
srsly travis, it compiled just fine for me.

Moved the NIN_NUKE_BONUS mod in modifier.h to be with the Ninja mods since I did that with the TRICK_ATK_AGI mod.

scripts/globals/effects/innin.lua
```diff
@@ -33,9 +36,9 @@ end;
 -----------------------------------
 
 function onEffectLose(target,effect)
-	--remove the remaining power
-	target:delMod(MOD_ACC,1*effect:getPower());
-	target:delMod(MOD_EVA,-1*effect:getPower());
-	target:delMod(MOD_ENMITY,-1*effect:getPower());
-	target:delMod(MOD_CRITHITRATE,effect:getPower());
+    --remove the remaining power
+    target:delMod(MOD_ACC,1*effect:getPower());
```
xdemolish: this too (and any others which are dumb)

Hozu:
Whoops, missed that one. Also took care of yet another new conflict (blocked by whitespace changes...).

PR #1881, opened 2015-08-06
Bcnm level sync exploit fix

Scavenge:


teschnei:
battlefield.cpp:234

do the check in there instead

xdemolish:
reopen when requested changes have been made

PR #1885, opened 2015-08-06
Scripts/SQL: Gargouille TP moves

Hozu:
Added all five TP moves. Note that since they don't have form changing scripted (I don't know the conditions for that), Dark Mist and Dark Orb will end up not being used by them under normal circumstances.

Changed the family in mob_skill.sql from 117 to 118 because that's what all of the Gargouille in mob_pools.sql are set to, and nothing appears to be set to family 117.

### DIFF
--- a/scripts/globals/mobskills/Bloody_claw.lua
+++ b/scripts/globals/mobskills/Bloody_claw.lua
@@ -1,0 +1,36 @@
+---------------------------------------------
+--  Terror Eye
+--  Family: Gargouille
+--  Description: Steals an enemy's HP. Additional effect: Reduces a random stat.
+--  Type: Physical
+--  Utsusemi/Blink absorb: Ignores shadows 
+--  Range: Melee
+--  Notes: Despite the attack ignoring Utsusemi, it is physical, and therefore capable of missing entirely. 
+---------------------------------------------
+require("scripts/globals/settings");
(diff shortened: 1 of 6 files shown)


teschnei:
aren't they only used while flying?

Hozu:
Dark Orb and Dark Mist are the two flying only moves. Triumphant Roar and Terror Eye are the standing moves. Bloody Claw has no restriction.

teschnei:
yeah - so why can't you implement them?

Hozu:
I suppose I could, they just won't be used.

After some further testing, I noticed that forcing Dark Orb and Dark Mist when it's standing, it will go into the air to use it. The other three moves will have it land to use them if it's flying. Hm...

Hozu:
Ok I scripted Dark Mist and Dark Orb. After some testing, only Bloody Claw looks like it should be used in either phase - when it uses Terror Eye in flying phase, it kind of just appears as standing, uses the move, then goes back into the air quickly (vice versa for Dark Mist). On the other hand, Bloody Claw has it move to stand for the attack, then take off, with both landing/taking off happening smoothly (ie. looks natural and intended).

Because of that, only Bloody Claw is in either phase, Dark Mist/Orb is flying only, Triumphant Roar/Terror Eye is standing only.

PR #1893, opened 2015-08-07
SCRIPTS: fixing spells with capital letters in the filename for linux

dacrybabysuck:
Yawn and several etudes were committed with capital letters.  This was causing linux builds to not be able to find them when using them in game.


TeoTwawki:
Wouldn't it be great if people stuck to existing conventions so you didn't have to do fixes like these? Even better, wouldn't it be great if DSP had a ***single*** convention for all file names instead of most of them **Mixed_Case** and some of them **lower_case** so people would be less likely to make these mistakes? I think they should match what the client displays.

dacrybabysuck:
Agreed.  It's interesting that the mobskills are capped, but everything else appears to be lower case.

TeoTwawki:
And a good portion of the mob skills were cased wrong in the DB when added, so when the missing scripts get added you'll have the same issue as these spells on Linux if the person adding them doesn't notice it :frowning:

xdemolish:
apparently i've been waiting to confirm merge for the past 9 minutes
http://prntscr.com/821vt0

PR #1902, opened 2015-08-09
Core/SQL: Jailer weapon additional swings

Hozu:
Added an extra swing in AddAttackSwing, with it checking if a Virtue Stone is equipped and if the weapon has the modifier for it. This swing stacks with Double/Triple/Quad Attack traits, and decrements Virtue Stones for the swing. Faith Baghnakhs can swing two times per attack round (one per hand), consuming two Virtue Stones.

Jump does not proc the extra swing as per http://www.bluegartr.com/threads/94701-Random-Question-Thread-XI-You-re-Wrong?p=3879961&highlight=jailer#post3879961

Also corrected the model for Justice Sword. http://i.imgur.com/613VXpw.png

### DIFF
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -1230,6 +1230,7 @@ MOD_SHARPSHOT       = 0x13A -- Sharpshot accuracy bonus (modId = 314)
 MOD_ENH_DRAIN_ASPIR = 0x13B -- % damage boost to Drain and Aspir(modId = 315)
 MOD_TRICK_ATK_AGI   = 0x208 -- % AGI boost to Trick Attack (if gear mod, needs to be equipped on hit) (modId = 520)
 MOD_NIN_NUKE_BONUS  = 0x20A -- magic attack bonus for NIN nukes (modId = 522)
+MOD_AMMO_SWING      = 0x20B -- Extra swing rate w/ ammo (ie. Jailer weapons). Use gearsets, and does nothing for non-players. (modId = 523)
 
 -- Mythic Weapon Mods
 MOD_AUGMENTS_ABSORB    = 0x209  -- Direct Absorb spell increase while Liberator is equipped (percentage based) (modId = 521)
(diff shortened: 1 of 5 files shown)


scripts/globals/status.lua
```diff
@@ -1230,6 +1230,7 @@ MOD_SHARPSHOT       = 0x13A -- Sharpshot accuracy bonus (modId = 314)
 MOD_ENH_DRAIN_ASPIR = 0x13B -- % damage boost to Drain and Aspir(modId = 315)
 MOD_TRICK_ATK_AGI   = 0x208 -- % AGI boost to Trick Attack (if gear mod, needs to be equipped on hit) (modId = 520)
```
TeoTwawki: 0x208 already in use bud

TeoTwawki:
Teo derp comment deleted! Nothing to see here!

src/map/attackround.cpp
```diff
@@ -193,54 +194,80 @@ void CAttackRound::CreateAttacks(CItemWeapon* PWeapon, PHYSICAL_ATTACK_DIRECTION
...
+    if (m_attacker->objtype == TYPE_PC)
+    {
+        // Check for ammo
+        CCharEntity* PChar = (CCharEntity*)m_attacker;
+        CItemArmor* PAmmo = PChar->getEquip(SLOT_AMMO);
+        uint8 slot = PChar->equip[SLOT_AMMO];
+        uint8 loc = PChar->equipLoc[SLOT_AMMO];
+        if (slot != 0) // slot is 0 if nothing is equipped in the ammo slot. Checking the ID of the item in the ammo slot without this check crashes if nothing is in the ammo slot.
+        {
+            if (PAmmo->getID() == 18244 && dsprand::GetRandomNumber(100) < m_attacker->getMod(MOD_VIRTUE_SWING))
```
xdemolish: dont ever hardcode item ids, use a mod

Hozu:
I'm going to look into getting this to work via gearsets, makes more sense that way. Easier to handle the Virtue Stone.

Hozu:
Ok done via gearsets. Actually this is much more neat because now I don't need nearly as many checks in the core. Also when digging though BG I found that the swing rate for all of the weapons is most like 50% instead of 45%.

PR #1915, opened 2015-08-12
Partially cleanup some Dynamis nonsense.

TeoTwawki:
Only thing I really corrected was Dynamis Lord's instant self depop and the timing on Ying and Yang repopping each other. Its 30 seconds, not 21. I killed the timer on Dynamis Lord because it : 

**A)** wasn't actually working as you can't setLocalVar on a mob that is not even up yet resulting on comparing os.time() with zero meaning always depop Dynamis Lord every tick of onMobFight... 

and **B)** I couldn't even find any source for it (only time I ever saw a Dyna lord fight, it was over well before this timer would have been up). I almost corrected his spawn requirements but decided I just don't care about *non* Neo Dynamis.

I didn't correct his spawn conditions but I commented in [Dynamis_Lord.lua](https://github.com/DarkstarProject/darkstar/commit/7a78eaa4f6598aa8e99f40e4fc40f2785f9c6d3f#diff-6f1961dac1835fbfc64de0a537dba6cdR5) if anyone else wants to. I just can't bring myself to care enough to do all that for *non* Neo Dynamis.

If someone else wants to re-implement the timer and or correct the spawn conditions, have at it!

### DIFF
--- a/scripts/zones/Dynamis-Xarcabard/mobs/Dynamis_Lord.lua
+++ b/scripts/zones/Dynamis-Xarcabard/mobs/Dynamis_Lord.lua
@@ -1,33 +1,40 @@
 -----------------------------------
 -- Area: Dynamis Xarcabard
--- NPC:  Dynamis_Lord
+--  NM:  Dynamis_Lord
+-- 
+-- In OLD Dynamis, he is spawned by killing 15 Kindred NMs..
+-- ..NOT by killing Ying and Yang.
+--
+-- In Neo Dynamis, he is spawned by trading
+-- a Shrouded Bijou to the ??? in front of Castle Zvahl. 
+--
+-- Suggested method for old school Dyna: mask bits
+-- Duke Berith        bit 0
+-- Marquis Decarabia  bit 1
+-- Count Zaebos       bit 2
+-- Duke Gomory        bit 3
+-- Marquis Andras     bit 4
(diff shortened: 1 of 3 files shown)


TeoTwawki:
Oh I totally forgot to mention: Ying and Yang are supposed to each give Dynamis Lord an Immunity, one physical one magical. I didn't correct that either!

teschnei:
but that's the easy part

TeoTwawki:
Yeah easy enough that I'd even be willing to set aside my old dyna hate for it. It totally slipped my mind and I hit push. I'd go tack it onto this right now but I have some other stuff I really should get done. Maybe later in the week if nobody else cares to.

gedads:
waiting for either KJlotus or whasf to tell me to add neo dynamis instead of old one. files (i mean db mobs)are ready but i think it will give nothing with actual system (a lot of PH popping mobs).

TeoTwawki:
@gedads I was going to checkout the branch and see what else needed done yet to make it ready-to-use and get it done, but have been busy with other things so never got around to it. My RL is being a pain lately :(

gedads:
neo dynamis branch is outdated , db is 1 year old at least now. i updated it locally 1 month ago but i need to check it again since i messed groups with august update. (replaced neo dynamis groups by escha ruaun groups)

PR #1923, opened 2015-08-14
SQL: Corrected the mobskillIDs that should be checked for some spells.

Hozu:
Jettatura and Corrosive Ooze check the wrong mobskillID when it comes to learning spells. The latter one is due to my fixing of the swapped moves.

### DIFF
--- a/sql/blue_spell_list.sql
+++ b/sql/blue_spell_list.sql
@@ -70,7 +70,7 @@ INSERT INTO `blue_spell_list` VALUES ('570','138','2','0','1','0','0'); -- Blood
 INSERT INTO `blue_spell_list` VALUES ('572','154','1','6','1','0','0'); -- Sound Blast
 INSERT INTO `blue_spell_list` VALUES ('573','1445','3','4','1','0','0'); -- Feather Tickle
 INSERT INTO `blue_spell_list` VALUES ('574','146','2','19','1','0','0'); -- Feather Barrier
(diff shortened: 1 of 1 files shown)


TeoTwawki:
@Hozu challenge: figure out how the F to untangle Barbed Crescent and Self-Destruct!

Hozu:
I thought nobody knows the Self-Destruct formula? Without that I'm not touching it with a 10 foot pole.

I was unaware that Barbed Crescent is a BLU spell. Huh.

TeoTwawki:
Well when people try to learn Barbed Crescent, they obtain Self-Destruct instead. Because both have the same ID in mobskills table but the packet Fomors use for their skills is a little different because they are actually WS'ing.

TeoTwawki:
So you prolly can't fix that atm, I was making a small (and lame) joke.

Hozu:
Oh well that's interesting. Didn't know that. And of course the one spell from a Fomor just happens to share the mobskillID of another BLU spell.

xdemolish:
if canMerge then
    GetEntityByName('Hozu'):leaveComment();
ens

Hozu:
Yeah this is good to go. The self-destruct/barbed crescent issue is something that has to be handled some other way, I'm afraid.

TeoTwawki:
@xdemolish sorry i confused you by bringing that up, I was only messin with Hozu a bit.

xdemolish:
ya you better be, i had to hold off clicking a green button because of you

PR #1924, opened 2015-08-14
Updated San d'Oria Mission 1-3 BCNM

kamajii7:
Updated Spawn points for mobs in the BCNM. Also updated enmity for those mobs so they will aggro players when attacked.

### DIFF
--- a/scripts/zones/Ghelsba_Outpost/mobs/Fodderchief_Vokdek.lua
+++ b/scripts/zones/Ghelsba_Outpost/mobs/Fodderchief_Vokdek.lua
@@ -12,6 +12,16 @@ function onMobSpawn(mob)
 end;
 
 -----------------------------------
+-- onMobEngaged
+-----------------------------------
+
+function onMobEngaged(mob,target)
+	--GetMobByID(17350662):updateEnmity(target); -- Fodderchief_Vokdek
+	GetMobByID(17350663):updateEnmity(target); -- Sureshot_Snatgat
+	GetMobByID(17350664):updateEnmity(target); -- Strongarm_Zodvad
+end;
+
+-----------------------------------
 -- onMobDeath Action
 -----------------------------------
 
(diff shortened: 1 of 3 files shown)


teschnei:
They shouldn't need script changes: BCNM mobs should superlink by default

kamajii7:
They did not in my tests. The mobs just stood there even if I was already attacking one.

TeoTwawki:
In my checking these 3 guys do not link, but they agro (behave like regular orcs). If you are high enough level the other orcs just stare at you, like they are hoping you'll forget to kill them next.

Maybe something just wasn't flagged right in their db info or this being Ghelsba Uutpost which is also a regular zone had something to do with it?

xdemolish:
there's a mobmod for superlink so you should prolly use that instead
https://github.com/DarkstarProject/darkstar/blob/master/src/map/mob_modifier.h#L58

xdemolish:
ignore what i said earlier you email readers

Hozu:
"Maybe something just wasn't flagged right in their db info or this being Ghelsba Uutpost which is also a regular zone had something to do with it?"

I'm inclined to believe this is the case because the CoP 8-3 fight does not have the pots linking.

kamajii7:
From my understand all BCNM's should have True Sight and True Hearing and should aggro regardless of level. Does anyone recall if that is true? I have not check the BCNM code yet I can do that next.

Hozu:
In the core, actual BCNMs have superlink, though there are a couple of fights that have specific exceptions (Diabolos (Prime) and Promy Vazhl). The fights that aren't actual BCNMs appear to lack this, which is why the 8-3 pots don't link.

TeoTwawki:
Maybe we should just have a BCNM specific behavior field and the default zero value be "everything superlink" so those can be "actual" BCNM as well then.

xdemolish:
no, no touching bcnms unless full rewrite cause i hate them

xdemolish:
as in: no touching core side of it

TeoTwawki:
@xdemolish I know you were trying to keep the existing scripts unchanged when you took a stab at it..and if I do it I'm prolly changing the effin scripts bits partly too. That chunk in core that basically does a setPos() after a static magic number check made me /headshake /sigh

xdemolish:
i dont even know what we're doing with this so closing

PR #1932, opened 2015-08-16
Scripts: Esuna fix, Sacrifice target

Hozu:
Made it so Esuna can only be cast on self, and Sacrifice on others only.
Redid the stored statuses as non-locals so they'll actually delete the statuses from others hit by the spell.
Made the effects deleted actually random instead of going in a set order.

### DIFF
--- a/scripts/globals/spells/esuna.lua
+++ b/scripts/globals/spells/esuna.lua
@@ -10,44 +10,69 @@ require("scripts/globals/magic");
 -----------------------------------------
 
 function onMagicCastingCheck(caster,target,spell)
-	return 0;
+    if (caster:getID() ~= target:getID()) then
+        return MSGBASIC_CANNOT_PERFORM_TARG;
+    else
+        return 0;
+    end;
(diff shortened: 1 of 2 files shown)


teschnei:
No lua globals, ever

Hozu:
Really? Fine, I'll close this for now since it will take forever for me to figure it out for the core.

teschnei:
Just use a player local var

teschnei:
It's pretty much exactly what they are for

Hozu:
Oh... I used them for mobs before, why didn't I think about using them for players?

...and seriously? I can't reopen this now because I forced pushed already?

teschnei:
They are useful for players too as a temporary var, since they never go to
the db they don't wreck performance, and they are wiped every time you
zone. So they're useful for a lot of things and won't be leaving memory
hanging around over the runtime of the server

Hozu:
Ah, I see. Thanks for the advice.

PR #1940, opened 2015-08-17
Implemented Vajra SA and TA Augments, added Vajra 75~99-4 to SQL

IxAalexI:
Also corrected some improper values and added MOD_AUGMENTS_CONVERT to status.lua (missed that last commit)


teschnei:
it's not a crit damage buff, it's just another multiplier to SA/TA damage

IxAalexI:
https://www.bg-wiki.com/bg/Vajra_%28Level_75%29 states "Testing seems to indicate +10% Critical Attack Bonus to Sneak Attack and Trick Attack." 

Couldn't see a different way of adding it without messing with source or adding a mod that would apply even if SA or TA was missed.

teschnei:
Correct, but that research was done before crit damage from atma/gear was really a thing (so there wasn't a good way to be sure at 75 cap).  take a look at the 80 vajra page's references, near the bottom of that thread's first page they talk about it being independent of the crit damage boost cap - and every vajra page after the 75 one has it listed as just SA/TA damage bonus

IxAalexI:
Ok, where do I look for directly affecting SA and TA? attack.cpp?

teschnei:
Yep. If it's not directly in there, then it'll be in a battleutils function called from there.  Will most likely require a few lines in weaponskill.lua too since we don't have that all centralized in one place, unfortunately.  And maybe wherever the blue magic does their SA calc too (THF/BLU isn't popular but still..)

IxAalexI:
Something like:
    {
		m_trickAttackDamage = m_attacker->DEX() * ((100+(m_attacker->getMod(MOD_AUGMENTS_ SA)))/100) ;
    }

?

teschnei:
that would only multiply the bonus damage from dex/agi, rather than the
whole attack

IxAalexI:
/************************************************************************
*																		*
*  Processes the damage for this swing.									*
*																		*
void CAttack::ProcessDamage()
{
	// Sneak attack.
	if (m_attacker->GetMJob() == JOB_THF &&
		m_isFirstSwing &&
		m_attacker->StatusEffectContainer->HasStatusEffect(EFFECT_SNEAK_ATTACK) &&
		((abs(m_victim->loc.p.rotation - m_attacker->loc.p.rotation) < 23) ||
        m_attacker->StatusEffectContainer->HasStatusEffect(EFFECT_HIDE)))
	{
		m_trickAttackDamage = m_attacker->DEX();
	}

Hmm, that's all there is for processing SA damage in attack.cpp, I'll dig around some more. Closing this for now.

teschnei:
the damage is added and the actual swing damage is right after that - you
could just do it right after the isCritical check

IxAalexI:
Like this?

    // Apply Sneak Attack Augment Mod
    if (m_attacker->getMod(MOD_AUGMENTS_SA) > 0)
    {
        m_damage += (m_damage * ((100+(m_attacker->getMod(MOD_AUGMENTS_
SA)))/100));
    }

teschnei:
and check for m_trickAttackDamage > 0, or else it applies to all attacks

PR #1943, opened 2015-08-18
Implemented Vajra SA and TA Augments, added Vajra 75~99-4 to SQL

IxAalexI:
Also corrected some improper values.


Hozu:
Just a heads up - your mod currently only affects melee attacks - not WSs. I made the same mistake when I implemented THF AF+1 hands.

IxAalexI:
Thanks for the heads up, looking at it now to figure out where to stick the mods in.

src/map/attack.cpp
```diff
@@ -363,9 +363,21 @@ void CAttack::ProcessDamage()
...
 	}
 
+    // Apply Sneak Attack Augment Mod
+    if (m_attacker->getMod(MOD_AUGMENTS_SA) > 0 && m_trickAttackDamage > 0 && m_attacker->StatusEffectContainer->HasStatusEffect(EFFECT_SNEAK_ATTACK))
+    {
+        m_damage += (m_damage * ((100+(m_attacker->getMod(MOD_AUGMENTS_SA)))/100));
+    }
+
+    // Apply Trick Attack Augment Mod
+    if (m_attacker->getMod(MOD_AUGMENTS_SA) > 0 && m_trickAttackDamage > 0 && m_attacker->StatusEffectContainer->HasStatusEffect(EFFECT_TRICK_ATTACK))
```
TeoTwawki: I think you meant to use MOD_AUGMENTS_TA on that line.

IxAalexI: Nah that's what I meant to do, both SA and TA return damage under m_trickAttackDamage. Looked odd to me too but that's what's already there.

TeoTwawki: lol. I was like "but the comment says..." and then got to "nvm".

IxAalexI:
Fixed

PR #1945, opened 2015-08-18
SQL/Scripts: Sheep Cap (+1)

Hozu:
Because reasons.

### DIFF
--- a/scripts/globals/items/sheep_cap_+1.lua
+++ b/scripts/globals/items/sheep_cap_+1.lua
@@ -1,0 +1,29 @@
+-----------------------------------------
+--  ID: 26720
+--  Sheep Cap +1
+--  When used, you will obtain one of eleven random items. They are:
+--  Faerie Apple, Sunflower Seeds, Red Moko Grass, La Theine Cabbage, Boyahda Moss,
+--  Pine Nuts, Beaugreens, Acorn, Dung, Batagreens, and Moko Grass
+-----------------------------------------
+
+-----------------------------------------
+-- OnItemCheck
+-----------------------------------------
+
+function onItemCheck(target)
+    local result = 0;
+    if (target:getFreeSlotsCount() == 0) then
+        result = 308;
+    end
+    return result;
+end;
+
+-----------------------------------------
+-- OnItemUse
+-----------------------------------------
+
+function onItemUse(target)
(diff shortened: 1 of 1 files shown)


TeoTwawki:
:sheep:

scripts/globals/items/sheep_cap_+1.lua
```diff
@@ -0,0 +1,29 @@
...
+    return result;
+end;
+
+-----------------------------------------
+-- OnItemUse
+-----------------------------------------
+
+function onItemUse(target)
+    local giftList = {4363,4505,1845,4366,919,2213,4571,4504,8929,4367,833}
+    local gift = math.random(0,10);
```
xdemolish: math.random(1. 11)
lua indexes start at 1

TeoTwawki: wrong in this case, math.random can start at zero without issue - we have literately 50 odd cases of it doing this.

xdemolish: he's using it to access an element in the table, meaning the first element has an index of 1 and last 11

TeoTwawki: I should not allow myself to speak after 2am passes.

Hozu: Wait are you serious? I did a 0 - X array element check for Esuna and I never had an issue with it when I had one status. If it has one status then it would do random 0 - 0 and it removed the status so... I'll do a print to see if it returns the correct item.

Hozu: Ok I confirmed with a print that that giftList[0] didn't print, and nothing after it happened. But that leaves me confused as hell as to why my Esuna script is able to do anything when there's only one removable status.

Hozu: Oh @TeoTwawki explained to me my Esuna script works with 0 because I specifically set element 0 in the loop. That makes a lot of sense.

teschnei: You definitely shouldn't do that either

teschnei: Oh, never mind, its not as bad as I thought. Still convention to start at 1
though

PR #1952, opened 2015-08-19
SQL: Khimaira Immunity

Zenny89:
Based on here: http://ffxi.allakhazam.com/db/bestiary.html?fmob=4576 and a post I read by @Deadwing888, Khimaira should be immune to Bind, Sleep, and Gravity, not just Sleep.


Deadwing888:
I thought he was immune to bind because nobody had enough macc back in the day to land it. Turns out if you stack macc with AV items and pimpest 75 gear you can actually land it for a few seconds 20% of the time or so. I think if you gave him (MOD_ICERES, 120) or so it'd be about right.

PR #1954, opened 2015-08-19
Core/Scripts/SQL: Implement increased roll range mod.

Hozu:
To do this, I added a binding in the core for setRange for ability. So in the scripts it's just ability:setRange(value).
Added a modifier that would be used for the increased range, and added it to Luzaf's Ring.
Modified all roll scripts + Double-Up so it will set the range to 8 (retail value before any mods) + the mod value.


xdemolish:
if the retail value is 8 (and the ability's range in dsp is also 8), shouldn't you just getRange() (cause whoever looks at it later will be loike dafuq is 8)

Hozu:
I could, but that's work! Ok I'll go make another binding to get the range...

xdemolish:
always keep in mind what d00d who looks at it later will have to go through to understand whatever monstrosity is concocted in the name of dsp

teschnei:
Need to make sure setrange isn't acting on the global ability list, or else
everyone on the server gets the range boost

xdemolish:
it's not https://github.com/DarkstarProject/darkstar/blob/master/src/map/ai/ai_general.h#L183
(if it is then it's your fault for not learningme2c++)

teschnei:
I checked ai_general.cpp, it looks good

Hozu:
Updated with the binding for ability:getRange()

PR #1956, opened 2015-08-19
SQL: Quick Draw range

Hozu:
"Quick Draw can be used from a max distance of 22'." http://ffxiclopedia.wikia.com/wiki/Quick_Draw


Hozu:
Closing due to ffxiclopedia being unreliable for this range.

PR #1957, opened 2015-08-20
Added Heka's Kalasiris, Nefer khat, and Nefer Kalasiris NQ and +1 gear set bonuses along with fixed a repeat ID.

Strafeous:
Added Heka's Kalasiris + head combinations for 3 tick refresh along with the Nefer Khat/Nefer Kalasiris NQ and HQ combos for 2 tick refresh.


Strafeous:
The indents come out odd on Github for me. I use 4 space indents on NPP but they're set to 8 at times, is that an option in my Settings?

xdemolish:
settings > preferences > tab settings > tab size 4 > tick replace by space

Strafeous:
Thank you

PR #1960, opened 2015-08-21
De-Hex ModID lists in modifier.h and status.lua because

TeoTwawki:
**Because:**
1) These are decimal everywhere else..
2) These are neither offsets nor masks, just arbitrary numbered lists..
3) Mistakes have been made repeatedly by the "base 16 challenged" folk among us..

No other changes in this commit, so nothing busted or incorrect that wasn't already.


TeoTwawki:
I have a 2nd PR coming that requires #1960 to be merged first

Also whoever started trying to order parts of the enum in the header file by job instead of ID, that was a bad idea. The mess in status.lua is bad enough from that only being partly and halfheartedly done (multiple competing organization methods = chaos, boys and girls..so please cut that out).

PR #1962, opened 2015-08-21
SQL: Updated AH category for multiple items

rpetit3:
Please remove.


TeoTwawki:
@rpetit3 It is still possible to make it 1 commit by doing a rebase. Prior to the replacing you did, could have just rolled back the history instead (on tortoisegit, this is as easy as right click > showlog > reset > mixed > ok). 

Hope you aren't using that vile github for windows program - this would be only the tip of the iceberg in terms of difficulty doing basic things in git using that horrible thing.

TeoTwawki:
> please remove.

Why? Just updating your branch updates the pull request. And a force push would overwrite anything you want to remove from github's side.

PR #1964, opened 2015-08-21
Added a good amount of new items and fixed a few older ones.

Strafeous:
Will update item_mod whenever I get off work since it's deciding to be very odd.


Strafeous:
I'll pull when I get back from work, there's problems and I don't have time to figure them out D:

PR #1966, opened 2015-08-21
SCRIPTS: Rush Modifier

Zenny89:
According to here: http://ffxiclopedia.wikia.com/wiki/Calculating_Blood_Pact_Damage
and here: http://ffxi.allakhazam.com/forum.html?forum=250&mid=1254863299247793724
Rush should have a subsequent damage modifier of 2 similar to the other multi-hit Blood Pacts.

### DIFF
--- a/scripts/globals/abilities/pets/rush.lua
+++ b/scripts/globals/abilities/pets/rush.lua
@@ -1,5 +1,5 @@
 ---------------------------------------------------
--- Double Slap M=6, 2 (still guessing here)
+-- Rush M=5, 2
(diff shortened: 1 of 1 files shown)


scripts/globals/abilities/pets/rush.lua
```diff
@@ -1,5 +1,5 @@
 ---------------------------------------------------
--- Double Slap M=6, 2 (still guessing here)
+-- Rush M=5, 2
```
xdemolish: m=5wat

PR #1969, opened 2015-08-21
SCRIPTS: Added Summon Skill P.Acc. Bonus

Zenny89:
Summoning magic should add 1 Accuracy per level over the cap: http://ffxiclopedia.wikia.com/wiki/Category:Summoning_Magic


teschnei:
that's a weird place to put it - why don't you add it to the avatars base acc when they are summoned (ie. where the merit acc is)?  you can't realistically expect the summoning magic skill level to change dramatically while an avatar is out, right?

Zenny89:
Sorry, it only applies to BPs - left that part out.

Hozu:
Various equips give skill, so it can change whenever the player feels like it.

teschnei:
oh duh, never mind me

PR #1978, opened 2015-08-24
SQL/Core/Scripts: MobMod for scripted 2hrs

Hozu:
This modifier will allow mobs to have 2hrs in their particular moveset in mob_skill.sql without being able to use it as a random tp move. If the mod is set to 1, each 2hr script will return 1 in the onMonSkillCheck function. This will allow all mobs with 2hrs that are unscripted to use them as normal.

When an ability is forced to be used by the script, it bypasses the onMobSkillCheck function, and just uses the skill. This is why I didn't need to add an additional check. Also, since the skill parameter is used to pass along the duration for some of these scripted 2hrs, I modified the check to use the param value if the mobmod is set and the param value is not 0. Otherwise, it defaults to the standard duration.

I looked through mob_skill.sql to see what mob families had 2hrs in their skillset, and determined whether they needed the new mobmod, or simply to have the standard enable 2hr mobmod set, and added them accordingly to mob_spawn_mods.sql. Cardians and Tyger had their 2hr skills removed as they weren't necessary - Cardians got them as normal mobmods for the specific NMs, while Tyger has no mention of having Mighty Strikes on any wiki, so I commented it out for now.

This also fixes the issue of mobs being unable to use specific 2hr abilities, such as Benediction and Hundred Fists.


Hozu:
Sorry about that Prishe. Fixed.

TeoTwawki:
Is the goal to just have less near duplicate lines in the sql table?

I don't understand the need for a MobMod since there is already a flag that does this. I've used it extensively on custom craps. If so flagged, mob will not use the move without being explicitly told to via the script command `mob:useMobAbility(ID)` This even includes non 2 hrs (the core doesn't know a 2hr from any other skill, its just that flag).

tl;dr 
Teo doesn't get what problem you are solving.

TeoTwawki:
p.s. I'd wager that Tyger mighty strikes thing was someones custom that slipped into their pull request and went unnoticed.

Hozu:
Well the main purpose was to fix certain 2hrs so they can actually be used by mobs (literally no mob could use Hundred Fists/Benediction/Mighty Strikes and some others unless they were scripted to). For example, the current check for Mighty Strikes is https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/mobskills/Mighty_Strikes.lua#L13 while Blood Weapon has https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/mobskills/Blood_Weapon.lua#L14 (ie. inconsistent).

Basically I don't want scripted 2hr use to interfere with non-scripted 2hr use. The check for the mod in each 2hr script also blocks it from being called as a regular tp move if it's added as a skill in mob_skills.sql, rather than using the param as a check.

TeoTwawki:
Those checks are all wonky...nothing blocks them if forced in script: the skill will go off regardless of what is in `onMobSkillCheck()` I don't think it even gets called. Scripted 2 hours should be using `mob:useMobAbility()` in that mobs `onMobFight()` and **if its flagged they can't use it unless forced, so thsoe checks wer eonly for non scripted usage and never belonged** crap like checking the mobs HP should have been in `onMobFight()`. The param, thats the duration when used with 2hrs so that one returning 1 whenever the param wasn't zero is just massive derp.

I guess my opinion is the checks should be removed rather than worked around, and those old usages corrected to fit a more generic instead of mob specific usage. I just got up and saw my IRC log though looks like new mobMob was already ok'd there?

...But lookin at the old code sure makes one /shakehead /sigh, dudnit?

TeoTwawki:
Actually...I kinda think all 2hrs should be scripted 2hrs, instead of letting the c++ core randomly decided when a mob will 2hr, could do it all in lua. Not that many mobs 2hr and setting a random hp lv to go off at is pretty simple for the ones that need it random (don't try and make that change lol, I'm just saying).

Hozu:
Forgot about Ix'Aern MNK and its helpers.

xdemolish:
i dont even know what im supposed to do with this
dumping on @teschnei

PR #1989, opened 2015-08-26
SQL/SCRIPTS: Death Scythe NQ/HQ Drain Additional Effect

waytim:
Set at 10%~ activation as per wiki discussion.


scripts/globals/items/death_scythe.lua
```diff
@@ -0,0 +1,37 @@
...
+-- Additional Effect: Drains HP
+-----------------------------------------
+require("scripts/globals/status");
+
+-----------------------------------
+-- onAdditionalEffect Action
+-----------------------------------
+function onAdditionalEffect(player,target,damage)
+    local chance = 10;
+    if (target:getMainLvl() > player:getMainLvl()) then
```
TeoTwawki: NO. I am sick of explaining these are NOT level corrected and that even if it were this method of attempting level correction isn't even anywhere near accurate.

TeoTwawki: See every add effect pull request ever.

TeoTwawki: On melee weapons with added effects:
* on status effect: fixed chance + chance to resist 
(in which case there is no message the effect just never happens)

* on element damage: fixed chance, and fixed damage that is modified ONLY by elemental resistance 
(fire dmg vs water elemental = you see the log say you hit for zero)

Drain effects work as if they were dark element dmg.

TeoTwawki: I apologize for the tone-in-text I just gave you..Got annoyed at seeing it yet again.

waytim:
Amended as discussed with Teo.

PR #1995, opened 2015-08-26
Scripts/SQL/Core: Elemental Obi bonus as mods, Hachirin-no-Obi

Hozu:
I changed the hardcoded IDs used to check for elemental obis as mods. This made it easier to implement the Hachirin-no-Obi, as well as cleaning up hardcoding.

Added Hachirin-no-Obi as an item and equipable item, as well as adding all of the new obi mods to it. The old obis received the one appropriate mod they should have.

Added a modifier for no MP depleted on cast as a % chance to activate, and gave it to the Hachirin-no-Obi (1%).

Fixed the bug where spells that were weak to the current day were getting a damage boost instead of penalty.


Hozu:
Note that the opposite day/weather bonus thing (ie. casting Drain in Dynamis on Lightsday) would have the bonus stack properly for the day/weather bonus (would be 1.15 in that instance).

PR #1997, opened 2015-08-27
SQL: Add CanSendPOL flag to various items.

Hozu:
I noticed some items were missing this flag because the delivery box was rejecting them even though the UI said they should go through. I checked in POLutils to make sure. They are: all coffer/chest keys, most abjurations, and the purple/brown belt quest items. The black belt quest items do not have this flag, however.


Hozu:
Eh? Why did travis fail when I had no error putting the sql file into my db?

teschnei:
read the log, looks like it powered off for some reason

you should be able to restart it

Hozu:
Strange. Let's see if it works this time.

PR #2007, opened 2015-08-30
SCRIPTS: Removed respawn time for certain force spawned NM crawlers

waytim:
Guardian Crawler, Drone Crawler, Queen Crawler, Matron Crawler, and Awd Goggie. According to retail, these ???s have no 15 min respawn; they disappear when the NM is successfully popped, and reappear after its death, able to be force popped immediately afterward. A player can continue to pop these as long as they have enough of the required rolanberries to do so.

### DIFF
--- a/scripts/zones/Crawlers_Nest/mobs/Drone_Crawler.lua
+++ b/scripts/zones/Crawlers_Nest/mobs/Drone_Crawler.lua
@@ -14,21 +14,20 @@ function onMobSpawn(mob)
 end;
 
 -----------------------------------
+-- onMobDespawn
+-----------------------------------
+
+function onMobDespawn(mob)
+    if (mob:getID() == 17584131) then
+        GetNPCByID(17584457):setStatus(STATUS_NORMAL); -- qm3
+    elseif (mob:getID() == 17584132) then
+        GetNPCByID(17584458):setStatus(STATUS_NORMAL); -- qm4
+    end
+end;
+
+-----------------------------------
 -- onMobDeath
 -----------------------------------
 
-
 function onMobDeath(mob,killer)
-
-	mob = mob:getID();
-
-	if (mob == 17584131) then
-		GetNPCByID(17584457):hideNPC(900); -- qm3
-	elseif (mob == 17584132) then
-		GetNPCByID(17584458):hideNPC(900);-- qm4
-	end
-
 end;
-
-
-
--- a/scripts/zones/Crawlers_Nest/mobs/Guardian_Crawler.lua
+++ b/scripts/zones/Crawlers_Nest/mobs/Guardian_Crawler.lua
@@ -14,17 +14,20 @@ function onMobSpawn(mob)
 end;
 
 -----------------------------------
+-- onMobDespawn
+-----------------------------------
+
+function onMobDespawn(mob)
+    if (mob:getID() == 17584129) then
+        GetNPCByID(17584455):setStatus(STATUS_NORMAL); -- qm1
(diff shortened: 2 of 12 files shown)


TeoTwawki:
I suggest having the NPC also re-pop on the mobs despawn as well as its death because if a player/party dies/wipes on one, it should despawn after a bit (I don't know if DSP has them set to do so, but they do on retail).
```lua
function onMobDespawn(mob)
    <code here>
end;
```

waytim:
Thanks Teo, updated so mob despawns to also set ???s visible.

scripts/zones/Crawlers_Nest/mobs/Drone_Crawler.lua
```diff
@@ -20,15 +20,28 @@ end;
...
 function onMobDeath(mob,killer)
 
-	mob = mob:getID();
-
-	if (mob == 17584131) then
-		GetNPCByID(17584457):hideNPC(900); -- qm3
-	elseif (mob == 17584132) then
-		GetNPCByID(17584458):hideNPC(900);-- qm4
-	end
+    mob = mob:getID();
```
TeoTwawki: Thats actually a bad practice, and also wasn't local'd. Better to just say that bit of code twice than to var it as "mob" when mob is already the argument name of the function if only so future editors/readers don't get confused. 

General rule of thumb: 
Variablize stuff that is either changing farther down the line in your code, or that is annoying to read (like really long like like this: `GetMobByID(12345678):getCurrentAction() == ACTION_ANNOY_TEOS_EYES` because right/left scrolling or word wrapping code is "TEH DEBIL!"  I tell you) and *good* code rarely gets very "wide" anyway.

This is only used twice in the same script, and its pretty short. it's ok to say `Mob:getID()` twice here.

scripts/zones/Crawlers_Nest/mobs/Guardian_Crawler.lua
```diff
@@ -19,12 +19,26 @@ end;
 
 function onMobDeath(mob,killer)
 
-	local mob = mob:getID();
+    local mob = mob:getID();
 
-	if (mob == 17584129) then
-		GetNPCByID(17584455):hideNPC(900); -- qm1
-	elseif (mob == 17584130) then
```
TeoTwawki: remind me to help previous author face-palm themselves. :wink:

TeoTwawki:
The rest looks good.

waytim:
Updated again.. thanks for everything, i want it to be purrrfect!

waytim:
Everything is updated and pretty, let me know if there's anything else i need to change.

TeoTwawki:
Someone said something in IRC about not needing both of onMobDespawn and onMobDeath..If onMobDespawn runs both on normal depops **and** the fadeout from death, they are correct and the onMobDeath bits can be removed. Otherwise, I don't have anything else to add to this.

(I dunno if that is how it works, and did not think to ask before pointing out the despawning mob possibility in his previous revision)

waytim:
Updated and done. For future reference: hiding the ??? onMobDespawn is enough; not needed on both that and onMobDeath. If a player kills the mob first, a few seconds later it will run the despawn function anyway when its body disappears.

TeoTwawki:
Good to know, thanks for updating.

xdemolish:
i win

TeoTwawki:
@xdemolish Was there a a prize? I didn't know there was a contest! Someone needs to tell us these things! If there's a decoder ring, you better watch your back! :cry:

Issue #1646, opened 2015-07-02
King Behemoth adjustments


Deadwing888:
I noticed King Behemoth cast meteor for the first time today, really cool. 

There are a few things still off about the fight though and I thought I'd bullet point them here in case someone wants to take a swing at it. 

~~1) The meteor needs to do a lot more than it currently is. I clocked retail KB meteor doing 1112 damage to my naked, unbuffed, rdm on a neutral day.~~
~~2) KB should be immune to silence~~
~~3) KB should be immune to stun~~
~~4) KB should have no regen~~
~~5) Thunderbolt does about 3x too much.~~
6) KB should have single target draw in, and only use it on the person who has hate when they leave the main behemoth area (past the stone columns)
7) KB has enstun on his melee hits, I'd eyeball it at 20% proc with about a 2 sec duration.
8) Kick out is doing ~3x too much damage and needs to be made an aoe.
9) Wild horn is doing ~3x too much damage
10) In addition to his double attack (which feels like an accurate rate) he has a ~10% triple attack.

I'll eventually muddle my way through the changes myself and do a pull request, but if someone more experienced can knock it out before then that'd be great. ~Wing

TheMuffinPirate:
This all makes sense but he does use shockwave currently

Deadwing888:
Great, I'll remove that one.

Zenny89:
@Deadwing888 you say that Double Attack % looks right, but when I look in the code I don't see anywhere that he has Double Attack. Are you sure?

TeoTwawki:
I can confirm DSP's King Behemoth has no double attack at all (not in in script nor mob_spawn_mods.sql nor mob_pool_mods.sql now in mob_family_mods.sql, and there is no place else for it to have been put, so it ain't there boys).

Deadwing888:
Idk what the code says, but when you actually go out and actually pop it he does actually double attack at about a 10% rate. Could be on account of his job being set to war?

I'm also noticing that his delay is too high, I'd estimate retail KB has a delay of about 200 (though I never frame counted to be sure).

It's great that stuff on this list is getting knocked out, but it's important that the nerfs on the list happen as well as the buffs (especially thunderbolt), or he will be too hard for 75 era servers when it turns into a kite fight. Also if his draw-in can't be configured to be single target+ only used when someone walks past the columns it should probably be turned off all together. 75s will need to be able to kite properly.

TeoTwawki:
[...]

PR #2008, opened 2015-08-30
SCRIPTS/SQL: King Behemoth Adjustments

Zenny89:
See Issue #1646 Items 5,7, and 10.


scripts/zones/Behemoths_Dominion/mobs/King_Behemoth.lua
```diff
@@ -15,6 +15,18 @@ function onMobInitialize(mob)
     mob:setMobMod(MOBMOD_MAGIC_COOL, 60);
 end;
 
+-----------------------------------
+-- onAdditionalEffect Action
+-----------------------------------
+function onAdditionalEffect(mob,target,damage)
+    if ((math.random(1,100) >= 81) or (target:hasStatusEffect(EFFECT_STUN) == true)) then
```
TeoTwawki: 80% chance and ignores both stun resits and thunder resist, is that retail?

scripts/zones/Behemoths_Dominion/mobs/King_Behemoth.lua
```diff
@@ -1,6 +1,6 @@
 -----------------------------------
 -- Area: Behemoth's Dominion
---  HNM: King Behemoth
```
TeoTwawki: That line didn't need changed at all.

scripts/zones/Behemoths_Dominion/mobs/King_Behemoth.lua
TeoTwawki: `math.random(0,99) >= 20` Because you are returning zero's if this condition is true (alternatively you can flip your >= to <= but this is more intuitive: 20% chance is what you intended, right?)

`applyResistanceAddEffect(player,target,ELE_LIGHTNING, EFFECT_STUN)` will check the resistance and returns the resistance value with 1.0 being fully unresisted. I'd assume a player has to full resist the effect for it to not land, and odds are the estimated proc rate was while wearing no stun resist gear.

TeoTwawki: Update and new comment post so your aware: 

1. After posting I remembered that the existing trunk examples for mobs with melee **status effects** very much suck. I don't think even one is retail accurate. The element dmg and drain ones are alright but the status ones look to be just random like you were doing. So I for one am not gonna expect perfection here - **if a core dev doesn't come along and insist on a resist check, I'm not going to either, and it can stay just a fully random thing till how those should be done is sorted out.**

Also:

2. I maybe should have pointed you to one of the other resist functions (there some in magic.lua and at least 1 in monstertpmove.lua as well) instead of `applyResistanceAddEffect(mob,target,ELE_LIGHTNING, EFFECT_STUN)`

3. I just looked and that function for the resist I pointed you at hasn't been updated yet, but thats what @maxtherabbit said to do (tag, you're it!) for the status effects in addEffect scripts when we were doing items. No idea if mobs should use same thing or not and I think I used more then 1 resist function on those items anyways...  >.>

maxtherabbit:
all status effects should call applyResistanceEffect(player,spell,target,diff,skill,bonus,effect)

Zenny89:
That's fine but I literally copied the stun code from a commit that occurred less than a month ago: https://github.com/DarkstarProject/darkstar/pull/1862/files

maxtherabbit:
well I can't blame you, or @Hozu even, since the codebase is wrought with "if ((math.random(1,100) >= 81) or (target:hasStatusEffect(EFFECT_SUCK) == true)) then" type stuff - but it still needs to go away

Zenny89:
So I need to get rid of the "random" function completely and simply call applyPlayerResistance(mob,effect,target,diff,bonus,element)?

maxtherabbit:
yeah I think that's probably the best way for us to handle mob additional effects

maxtherabbit:
but there still needs to be some kind of random function, to determine the proc rate of the effect

like there's a random check to see if it procs, and then the resistance function checks if it sticks

Zenny89:
Can anyone explain why my merge failed and how I can fix it?

xdemolish:
try a rebase

Zenny89:
That's what called this problem. I did fetch,etc. and then committed the merge.

sql/mob_pool_mods.sql
```diff
@@ -215,6 +212,15 @@ INSERT INTO `mob_pool_mods` VALUES (459, 23, 50, 0); -- Attack
 INSERT INTO `mob_pool_mods` VALUES (459, 73, 25, 0); -- Store TP  
 
 -- -------------------------
+<<<<<<< HEAD
```
TeoTwawki: Signs of incorrectly handled edit conflicts.

Zenny89: Yes, I'm aware it's incorrect.

bendangelo: You can safely delete lines 223 - 234

PR #2016, opened 2015-09-01
SQL/SCRIPTS: Adding Ramponneau NM

waytim:
http://ffxiclopedia.wikia.com/wiki/Ramponneau

NM has undispellable shock spikes and an enblizzard effect.  Set lottery at 20% chance.


scripts/zones/West_Sarutabaruta_[S]/mobs/Toad.lua
```diff
@@ -0,0 +1,25 @@
...
+require("scripts/zones/West_Sarutabaruta_[S]/MobIDs");
+
+-----------------------------------
+-- onMobDeath
+-----------------------------------
+
+function onMobDeath(mob,killer)
+
+    local mob = mob:getID();
+    if mob == Ramponneau_PH then
```
Comment (TeoTwawki) [inline scripts/zones/West_Sarutabaruta_[S]/mobs/Toad.lua]: Don't name variable the same as the function arg.

Comment (TeoTwawki) [inline scripts/zones/West_Sarutabaruta_[S]/mobs/Toad.lua]: Also:
`if (stuff) then`
 not
 `if stuff then`
for styling consistencies sake

Comment (TeoTwawki) [inline scripts/zones/West_Sarutabaruta_[S]/mobs/Toad.lua]: Nothing else stands out to me atm.

sql/mob_family_system.sql
```diff
@@ -537,6 +537,7 @@ INSERT INTO `mob_family_system` VALUES (493,'Ashrakk',9,'Demon',1,40,90,140,6,3,
 INSERT INTO `mob_family_system` VALUES (494,'Sekhmet',6,'Beast',1,60,94,90,1,1,4,2,6,6,2,3,3,3,3,1,1,1,1,1,1,1,1.25,0.875,1,1,1,5);
 INSERT INTO `mob_family_system` VALUES (495,'Astral_Flow_Pet',5,'Avatar',0,40,100,120,3,3,3,3,3,3,3,3,3,3,3,1,1,1,1,1,1,1,1,1,1,1,1,0);
 INSERT INTO `mob_family_system` VALUES (496,'Mandragora',17,'Plantoid',0,40,107,120,3,3,4,5,4,4,2,3,3,3,3,1,1.25,1,1,1.25,1.25,1.25,1,1.25,1,0.875,1.25,4); -- for chaneque unique TP move
+INSERT INTO `mob_family_system` VALUES (497,'Poroggo',7,'Beastmen',0,40,70,140,6,3,6,3,1,5,4,3,3,3,3,1,1,1,1,1,1.125,1,1,1.125,0.5,0.5,1,6);
```
TeoTwawki: Oh, just read on wiki he does no tp moves at all and casts no spells, if that's the only reason for the new family we can actually just set this in his script:
`mob:SetMobAbilityEnabled(false);` And with no spell list he won't try to cast.

Otherwise, ignore this post.

There was an NM I put in Upper Deklfutts Tower I think, when I was new to the project..I need to go back and do the same thing to him. He error's the console log about trying to use TP when he has no skills for his FamilyID whenever someone fights him :frowning:

waytim:
Updated Teo, should be good now.

scripts/zones/West_Sarutabaruta_[S]/mobs/Ramponneau.lua
```diff
@@ -0,0 +1,68 @@
...
+-----------------------------------
+
+function onAdditionalEffect(mob,target,damage)
+    
+    local power = math.random(4,15);
+    local params = {};
+    params.bonusmab = 0;
+    params.includemab = false;
+    power = addBonusesAbility(mob, ELE_ICE, target, power, params);
+    power = power * applyResistanceAddEffect(mob,target,ELE_ICE,0);
```
Comment (TeoTwawki) [inline scripts/zones/West_Sarutabaruta_[S]/mobs/Ramponneau.lua]: Shucks, sorry @waytim i forgot to tell you mob additional effect isn't the same stuff as item additional effects.

And I don't see the scripting for spikes here, you used the mobmod to enable scripted spikes. I'll PM you examples for both in IRC.

waytim:
Leaving spikes at 10 dmg. This guys ready to rock!

TeoTwawki:
*Note for devs*: After trying to get the damage calculation right with him for several hours, I told @waytim to not bother scripting the spikes and just let the status effect power handle it for now while I try and rework how we handle mobs with auto-spikes that are supposed to vary. Will talk on IRC about the problem later.

TeoTwawki:
The only semi-serious issue is beyond his control, see issue #2017 For now as a work around he's setting its TP to zero. If he gives it its own family ID but with no skills, it spams the server log every time someone fights it.

waytim:
Tested with https://github.com/DarkstarProject/darkstar/commit/88edd3724125772add28879b20df8f768b2c7a68 and is working. Removed mob:setTP(0) and only using mob:SetMobAbilityEnabled(false). Thank you @teschnei and @TeoTwawki

PR #2019, opened 2015-09-02
Dear past me, please pay better attention to what we copy from who and where

TeoTwawki:
IMaybe I haven't been to Iceland because I'm busy dealing with YOUR crummy code, past self.

### DIFF
--- a/scripts/zones/Attohwa_Chasm/mobs/Sargas.lua
+++ b/scripts/zones/Attohwa_Chasm/mobs/Sargas.lua
@@ -3,8 +3,7 @@
 --  NM:  Sargas
 -----------------------------------
 require("scripts/globals/status");
------------------------------------
-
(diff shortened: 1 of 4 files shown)


TeoTwawki:
Todo: actually figure out EXACTLY how retail calculates this crap, instead of continuing to replicate very fishy looking stuff that I think was just ripped from melee add effects at the time I copy pasta'd it (over 6 months old code = I don't remember things any more). I do remember testing it and deciding that the actual damage was in the right ballpark, at least.

TeoTwawki:
Oh my bad, I forgot to actually tell you what I was fixing.. >. > I missed a require in all these scripts, so players fighting them could wind up being one shitted because the dmg calcs would go insane if they hadn't just happend to cause that global to already be loaded by doing something else..

PR #2020, opened 2015-09-02
SQL: Composure and Velocity Shot effect flag

Hozu:
Added the flag that enables the status to wear upon zoning, as they're both supposed to do that.
http://ffxiclopedia.wikia.com/wiki/Velocity_Shot
http://ffxiclopedia.wikia.com/wiki/Composure


TeoTwawki:
Goodbye  favorite rdm bug I was hoping nobody would ever notice.

PR #2029, opened 2015-09-05
SCRIPTS: Empty Vessel Quest answer

waytim:
http://ffxiclopedia.wikia.com/wiki/An_Empty_Vessel

Changed answer needed from "Read the scroll?" to "Burn the scroll?" as stated on wiki. This is one of the biggest problems encountered on my server as people followed the wiki, I tested and it works. Any input?

### DIFF
--- a/scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
+++ b/scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
@@ -75,7 +75,7 @@ function onEventUpdate(player,csid,option)
 			player:setVar("SuccessfullyAnswered",success+1);
 		elseif (option == 6) then
 			player:setVar("SuccessfullyAnswered",success+1);
-		elseif (option == 7) then
+		elseif (option == 7 or option == 9) then
 			player:setVar("SuccessfullyAnswered",success+1);
 		elseif (option == 12) then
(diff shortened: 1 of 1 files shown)


TeoTwawki:
which wiki? All answers in DSP are what BG wiki claims, and not what ffxiclopedia claims. Changing it to match a different wiki doesn't exactly make it correct. 

..And I swear to god I remember giving the same answers on retail 3 diff days and getting different results for the exact same answers. :confused:

TeoTwawki:
@ everybody:
Honetsly, if nobody can confirm either one as being retail perfect, I'd rather we just go with what ffxiclopedia says because the majority of players looking at a wiki for it are going to look at that one (in which case, I am fer merging this).

waytim:
@TeoTwawki I see, the two are different on that question only. And i agree 99% are going to look at ffxiclopedia,  The fact is this quest as is far from retail, it appears to use some kind of points system since success can come from different combos of answers, but appears random.  Or... we COULD say if option == 7 or option == 9 couldn't we? End the problem right there!

TeoTwawki:
@waytim  That would be fine with me as well. My thinking was: as long as we're not being retail perfect may as well minimize people reporting it as "broken".

waytim:
Updated to match both wiki's.

PR #2031, opened 2015-09-06
Removed code that resets monster TP to zero when a monster has no skills to use

TeoTwawki:
@teschnei I went ahead and removed the message too. Discovered that a few elemental type mobs from newer content the project does have implemented yet actually have a few TP moves, so checking the ecosystem had to go, and I didn't wanna see that message in log every time somebody fought an elemental nor go hunt down every script-less elemental first.

### DIFF
--- a/src/map/ai/ai_mob_dummy.cpp
+++ b/src/map/ai/ai_mob_dummy.cpp
@@ -715,20 +715,10 @@ void CAIMobDummy::ActionAbilityStart()
 
     std::vector<CMobSkill*> MobSkills = battleutils::GetMobSkillsByFamily(m_PMob->getMobMod(MOBMOD_SKILLS));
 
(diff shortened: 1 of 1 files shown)


TeoTwawki:
@teschnei I know you told me it'd be fine, but I went ahead and fed some skill-less mobs to anyway just to be sure there were no adverse effects.

bendangelo:
https://github.com/DarkstarProject/darkstar/pull/2031/files#diff-f9ce0a6c732916d22a79166e350d583dR836 Line 836. Mob tp still get reset.

teschnei:
you need to keep the MobSkills.size() check and return immediately if it's 0

TeoTwawki:
Amended, thanks.

TeoTwawki:
..screwed up my amend..re-amaneding..

TeoTwawki:
There should be correct now.

Issue #2028, opened 2015-09-05
Magic damage not breaking bind effect (with details)


Deadwing888:
In day-to-day play on a darkstar server, bind holding through magic damage is the number one thing obviously off from retail.

After a few hours of testing I've nailed down these specifics:

Magic damage from nukes have a chance to break bind.
DOT spells that do initial damage have a chance to break bind. (yes dia, yes bio, no poison)
Being hit and having spikes deal damage has a chance to break bind.
The chance to break bind does not vary based on how much damage is done.
The chance to break bind does not vary based on macc or player int vs mob int.
The chance to break bind does vary based on your level checked against the mob's level.

My testing indicated the chance for magic damage to break bind based on level go like this:

Too weak- 0% chance to break
Easy Prey- 15% chance
Decent Challenge- 15% chance
Even Match- 30% chance 
Tough- 75% chance
Very Tough- 100% chance
Incredibly Touch- 100% chance

Now, I believe in the past I've seen bind hold on an IT vs a nuke, but in two hours of testing I couldn't recreate that. If it were capped at 5% chance to hold and 95% chance to break on each end I don't think that would be the end of the world. I also don't think it would be the end of the world if it functioned based on an int check of player vs mob, but it would need to ramp up to 95% very quickly once mob int was higher than player int. If anyone has any questions I'd be happy to go back in and do more testing.

TeoTwawki:
> Now, I believe in the past I've seen bind hold on an IT vs a nuke, but in two hours of testing I couldn't recreate that. 

Retail has it being a stupidly rare occurrence. DSP does have a bind break chance, its just a status flat % for now instead of being calculated though.

> The chance to break bind does not vary based on how much damage is done.

I am very certain it **does**, in a very small and meaningless way (good old SE..). Its trumped by the level correction involved. Nothing matters nearly as much as that level difference when it comes to "will the next nuke/melee attack break it?" (the answer is going to be yes anyway, on anything that matters, barring crazy luck - as you saw on your VT and IT attempts even a 1dmg dia is going to break that 199 or 200 tries).

Deadwing888:
I was not aware DSP had a bind break chance coded for magic. I admit my test server build is a month or so old, but I found it impossible to break bind with nukes there, ditto on nasomi where I regularly play.

If you're certain it does, you're certain it does, but I eyeballed absolutely no difference between bio for 20 stone for 200 and thunder 3 for 800 in an hour of playing with the same mob.

[...]

PR #2033, opened 2015-09-06
Eliminated situations where bind never breaks, added new function for handling the check for bind breakage.

TeoTwawki:
This is a temp quick fix to get rid of bind never breaking at all. #2028

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1870,9 +1870,8 @@ namespace battleutils
         {
             PDefender->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_DAMAGE);
 
-            //40% chance to break bind when dmg received
-            if (PDefender->StatusEffectContainer->HasStatusEffect(EFFECT_BIND) && dsprand::GetRandomNumber(100) < 40)
-                PDefender->StatusEffectContainer->DelStatusEffect(EFFECT_BIND);
+            // Check for bind breaking
+            BindBreakCheck(PAttacker, PDefender);
 
             switch (PDefender->objtype)
             {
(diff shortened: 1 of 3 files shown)


TeoTwawki:
Added new function for it, as requested.

src/map/utils/battleutils.cpp
```diff
@@ -4363,6 +4362,62 @@ namespace battleutils
...
+    {
+        if (PDefender->StatusEffectContainer->HasStatusEffect(EFFECT_BIND))
+        {
+            uint16 BindBreakChance = 0; // 0-1000 (100.0%) scale. Maybe change to a float later..
+            uint16 LvDiffByExp = charutils::GetRealExp(PAttacker->GetMLevel(), PDefender->GetMLevel()); // This is temp.
+
+            // Todo: replace with an actual calculated value based on level difference. Not it, Bro!
+            // This entire block of conditionals should not exist. Lv diff really shouldn't be handled at the exp check level either.
+            // It might not even be in sync with the check values the entire way from lv 1 to lv 99 for all we know.
+            if ((PAttacker->objtype == TYPE_PC || PAttacker->objtype == TYPE_PET) && PDefender->objtype == TYPE_MOB)
```
bendangelo: What about PvP? I think a player hitting a binded player has a chance to break bind.

xdemolish: pvp was taken out like a year ago

TeoTwawki: I'd edit to account for the day pvp returns, but that whole conditional branch will likely be changed to something that doesn't need object types anyway (used the /check because we have real data)

TeoTwawki: I'd edit to account for the day pvp returns, but that whole conditional branch will likely be changed to something that doesn't need object types anyway (used the /check because we have  no real data)

TeoTwawki: ignore the first of my double post, you email readers (typos)

teschnei: why is the object check there? we have no data for bind on players, sure, but is there any reason to assume that 40% static is more accurate than the level check?

TeoTwawki: the the type check is because I don't want to call those GetRealExp function for non players. as for the rate on players, if you have a formula you'd like me to use, pls share

TeoTwawki: remember *none* of this is to my liking to start with.

TeoTwawki: will read any comments and amend as needed when I wake back up

teschnei: you called GetRealExp before the check so it gets called either way

at any rate, the only thing GetRealExp checks is levels, which is why the parameters are both CBattleEntity

TeoTwawki: I derped 2 ways then. Thought that only worked when a player was involved, if its safe I'll just use it without any object check.

TeoTwawki:
Amended

PR #2035, opened 2015-09-06
Scripts/SQL: Cerberus family mobskill direction check

Hozu:
-All skills check if the target is behind the mob - if it is, Scorching Lash is the only eligible move.
-Added no turn behaviour to Cerberus and Sarameya,
-Made the Stoneskin effect of Magma Hoplon unable to be removed by dispel effects.
-Added Burn spell to Sarameya's spell list (saw it in a video).
-Increased the prepare time for Gates of Hades to 3 seconds.


Deadwing888:
He can still use gates, undulation, and magma hoplon when people are attacking from behind.

Hozu:
Really? Well shit.

Issue #2034, opened 2015-09-06
COP 5-3 Tenzen's Path KI Bug


towbes:
Currently if a player forgets to check the ??? in Battalia Downs twice in order to get the KI Delkfutt Recognition Device, then they go and kill the Disaster Idol, their mission var gets set to 7 and they are unable to go back to obtain the KI in order to progress the mission.  According to the wiki http://ffxiclopedia.wikia.com/wiki/The_Pursuit_of_Paradise_(Tenzen%27s_Path) a player should be able to obtain the KI even after they kill the idol.

In the script darkstar/scripts/zones/Batallia_Downs/npcs/qm4.lua , the if statement should be

 player:getVar("COP_Tenzen_s_Path") == 6 or 7    

in order to allow the player to get the KI even after defeating the idol.

Another solution might be to not update the mission var if the player didn't have the KI yet, though I think the first example is more aligned with what the wiki says.

For example adding the KI check in
scripts/zones/Lower_Delkfutts_Tower/mobs/Disaster_Idol.lua

if (killer:getCurrentMission(COP) == THREE_PATHS and killer:getVar("COP_Tenzen_s_Path") == 6 and player:hasKeyItem(DELKFUTT_RECOGNITION_DEVICE) == true))  then 
      killer:setVar("COP_Tenzen_s_Path",7);

Hozu:
They shouldn't even be able to spawn the Disaster Idol NM without the KI, and if someone else spawns it, it probably should not advance the mission progress. I'll have to double check that last one though.

Hozu:
Oh right, this issue is now resolved.

PR #2036, opened 2015-09-06
Scripts: Tenzen's Path KI bug fix + Sea requirement check

Hozu:
-Made the ??? in Batallia Downs check for the progress before and after the defeat of the Disaster Idol. This is because you don't have to defeat it again if you forgot to get the KI. Currently, if you defeat the NM, you are stuck unable to get the KI.
-Changed the requirements to use various Sea stuff from KI Light of Al Taieu to CoP current mission is past The Warrior's Path. The KI is not Sea access, as you are supposed to lose it at some point, and only one race is to get it back. However, I'm not sure when the loss is supposed to happen, so I left that out.

### DIFF
--- a/scripts/zones/Batallia_Downs/npcs/qm4.lua
+++ b/scripts/zones/Batallia_Downs/npcs/qm4.lua
@@ -14,12 +14,13 @@ require("scripts/globals/missions");
 -----------------------------------
 
 function onTrigger(player,npc)
-	if (player:getCurrentMission(COP) == THREE_PATHS  and  player:getVar("COP_Tenzen_s_Path") == 5) then	
-	 player:startEvent(0x0000);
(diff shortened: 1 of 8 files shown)


Hozu:
This fixes #2034.

xdemolish:
is it supposed to give the keyitem to anyone who's completed warrior's path? cause the way you've got it now it does

Hozu:
Hm? The only thing I changed about being given the KI for the one Tenzen's path thing is allowing for the progress var being 6 or 7 instead of just 6. It's set to 6 from the first CS at the ???, and set to 7 when the NM is killed if it's at 6.

The other changes merely replaced checking for the Light of Al'Taieu to checking for CoP mission past The Warrior's Path.

xdemolish:
whatever

PR #2037, opened 2015-09-06
Scripts: Removed the rear facing block that I just put in for certain…

Hozu:
… Cerberus moves.

Gates of Hades, Magma Hoplon, and Ululation. Turns out I was wrong in assuming it was completely like Spike Flail.


Hozu:
See the latest comments in #2035.

Deadwing888:
Fafhogg can use hurricane wing from behind in addition to Spike Flail

PR #2039, opened 2015-09-06
SCRIPTS: Breath Mantle and High Breath Mantle

Vavehl:
These two scripts should be retail-like with the following caveat:

1) The enchantments persist when the item is removed (which is not supposed to happen).  In short, you can equip the item, use it to gain the enchantment for 30 minutes, then remove it and still keep the enchantment for the remaining duration.


teschnei:
If you set the script type of those items to SCRIPT_EQUIP (1), it will run onItemCheck whenever the item is equipped or unequipped, and you can check for the status effect there

Vavehl:
Thanks, teschnei -- that was a lot easier than I was expecting it to be.  The Breath Mantle (#15486) and High Breath Mantle (#15487) now have their script-types set to SCRIPT_EQUIP in item_armor.sql.

teschnei:
note that setting the scripttype only makes it call onItemCheck on equip/unequip - you still have to check the effect in the script (your current scripts do nothing in onItemCheck)

Vavehl:
Okay, the enchantments should now expire when the items are unequipped.  My worry is that there may be different EFFECT_ENCHANTMENT effects that get wiped when removing (or equipping) the Breath Mantle or High Breath Mantle.

Is there a function that would allow me to check the subID (located in the char_effects table) of a character's EFFECT_ENCHANTMENT, to ensure that I can specifically remove the enchantment for either items #15486 (Breath Mantle) or #15487 (High Breath Mantle)? Or maybe I'm approaching this whole thing incorrectly?  When you get a chance, teschnei, would you mind nudging me in the right direction one more time?

Thanks for your help; it is really appreciated.

teschnei:
instead of using hasStatusEffect, use getStatusEffect to get the effect if it exists.  then you can check if its nil or not (no effect will return you nil) and if it's not nil, you can do effect:getSubID() to get the subid

Vavehl:
Thanks again, teschnei.  I'm still having a bit of a problem calling getSubID(), and I'm not familiar enough with Lua to track down my error.  When I run the following code snippet, I get the following error: attempt to call method 'getSubID' (a nil value).  Is this happening because I'm not including a required library for the getSubID() method?  I'll keep tinkering away at it until I get it right, heh.

```

function onItemCheck(target)
	if (target:getStatusEffect(EFFECT_ENCHANTMENT) ~= nil) then
		local effect = target:getStatusEffect(EFFECT_ENCHANTMENT);
		if (effect:getSubID() == 15486) then
			target:delStatusEffect(EFFECT_ENCHANTMENT);
		end;
	end;
	return 0;
end;
```

teschnei:
oh, I got the name wrong, sorry
basically, check the src/map/lua folder for the lua objects.  You're
getting a status effect here, so you'd want to check lua_statuseffect.h.
All the functions defined in there should be the ones callable from Lua.
In this case, you're looking for getSubType

TeoTwawki:
Does retail let you have multiple enchantment effects? I thought we were limited to just one but its been a long time since I had to know that so I don't remember.

Vavehl:
Thanks again, @teschnei -- you were a huge help.  I'll remember to look in the src/map/lua folder for all the objects and functions from now on.  I need to do some more testing on these two scripts, but (as they currently stand) the Breath Mantle and High Breath Mantle should no longer overwrite an existing enchantment when they are equipped.

@TeoTwawki: Heh, I can't remember either whether or not retail allows multiple enchantments, either.  Until evidence is found that they're allowed, I'll err on the side of caution and keep it at one.  These two scripts should adhere to that "one enchantment" rule -- and they shouldn't overwrite a different enchantment (such as from a High Mana Wand or a Talisman Obi) when they're equipped or unequipped.

I'll try to do a bit more testing tomorrow to make sure that they're functioning as intended.

TeoTwawki:
@Vavehl  :+1:

Vavehl:
I've made a few more changes over the past week, but I think these two items are scripted in a retail-like manner.  Any further suggestions/ideas to edit either the Breath Mantle or High Breath Mantle?

scripts/globals/items/high_breath_mantle.lua
```diff
@@ -0,0 +1,53 @@
...
+require("scripts/globals/settings");
+require("scripts/globals/status");
+
+-----------------------------------------
+-- OnItemCheck
+-----------------------------------------
+
+function onItemCheck(target)
+	if (target:getStatusEffect(EFFECT_ENCHANTMENT) ~= nil) then
+		local effect = target:getStatusEffect(EFFECT_ENCHANTMENT);
```
teschnei: put this before the if, then just have the if check "if (effect ~= nil)", that way you don't iterate the whole buff list twice

Vavehl:
Great suggestion -- I didn't even pay attention that I was calling the function twice.

xdemolish:
tabs to 4 spaces

Vavehl:
Right on, changed the indentation from tabs to 4-spaces.

PR #2044, opened 2015-09-07
SCRIPTS: Updated Angon to be able claim monsters on use

Enignite:
Angon will claim an unclaimed monster.


Hozu:
Oh so that's how to do it. Well then.

PR #2050, opened 2015-09-08
ZM8 bcnm blackscreen fix

Sacshop:
Input the proper event ID for winning the bcnm.


teschnei:
hm, I doubt that this is the only bcnm in the game that doesn't use a cs in the 32000 range, is this from retail?

Sacshop:
Yes - this would make it 100% retail accurate and no longer just leave you at a black screen upon winning.

teschnei:
so just to confirm - you went on retail, capped the packet, and that is what the csid was

Sacshop:
No - let me do that now then ~

PR #2056, opened 2015-09-10
No more chat crash.

TeoTwawki:
@teschnei I haven't been able to trigger a chat message client crash since changing to to this.

### DIFF
--- a/src/map/packets/chat_message.cpp
+++ b/src/map/packets/chat_message.cpp
@@ -28,18 +28,19 @@ This file is part of DarkStar-server source code.
 
 CChatMessagePacket::CChatMessagePacket(CCharEntity* PChar, CHAT_MESSAGE_TYPE MessageType, int8* buff)
 {
-     // Determine the return message length..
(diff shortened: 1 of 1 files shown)


TeoTwawki:
Fixed the indent too. https://github.com/DarkstarProject/darkstar/pull/2056/files?w=1

teschnei:
This looks like what we used to have, which would cause buffer overflows in zmq (note that packet->size should never ever be bigger than 0x104/2, because the buffer is 0x104 bytes)

it also does not check buffSize, which is another problem

TeoTwawki:
Damn. Knew there was a reason we "broke" this..

teschnei:
git blame would show you the commit and hopefully the reason :)

TeoTwawki:
Nope, or I'd have not PR'd it. :cry: back to drawing board!

teschnei:
git log, sorry: https://github.com/TeoTwawki/darkstar/commit/0f2bb5e1c4a3647b760a9e91c21079b9ba8f5043

PR #2060, opened 2015-09-12
Refactor magic resist

bendangelo:
Consolidated the common code between magic resist functions.

I broke the code down into down methods:

getMagicHitRate - returns p which is basically a percentage of the chance
getMagicResist - returns the magic resistance

### DIFF
--- a/scripts/globals/magic.lua
+++ b/scripts/globals/magic.lua
@@ -320,7 +320,6 @@ end;
 
 function applyResistance(player,spell,target,diff,skill,bonus)
 
-    local resist = 1.0;
     local magicaccbonus = 0;
     local element = spell:getElement();
     local castersWeather = player:getWeather();
@@ -335,8 +334,7 @@ function applyResistance(player,spell,target,diff,skill,bonus)
         end
     end
 
-    --get the base acc (just skill plus magic acc mod)
-    local magicacc = player:getSkillLevel(skill) + player:getMod(79 + skill) + player:getMod(MOD_MACC) + player:getILvlMacc();
+    local magicacc = 0;
 
     if player:hasStatusEffect(EFFECT_ALTRUISM) and spell:getSpellGroup() == SPELLGROUP_WHITE then
         magicacc = magicacc + player:getStatusEffect(EFFECT_ALTRUISM):getPower();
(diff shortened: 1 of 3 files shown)


scripts/globals/magic.lua
```diff
@@ -336,7 +336,7 @@ function applyResistance(player,spell,target,diff,skill,bonus)
     end
 
     --get the base acc (just skill plus magic acc mod)
-    local magicacc = player:getSkillLevel(skill) + player:getMod(79 + skill) + player:getMod(MOD_MACC) + player:getILvlMacc();
+    local magicacc = player:getILvlMacc();
```
teschnei: Item level magic acc should be in the same place as the skill level macc (it is basically the same thing)

teschnei:
You should be using 4 spaces per indent everywhere - currently I see 2 spaces and 4 spaces in the lua script, and a tab character in the cpp/h files

scripts/globals/magic.lua
```diff
@@ -618,211 +467,124 @@ function applyResistanceEffect(player,spell,target,diff,skill,bonus,effect)
...
+    minorBonus = minorBonus + (affinityBonus-1) * 200;
+  end
+
+  --add acc for RDM group 1 merits
+  if (element > 0 and element <= 6) then
+    minorBonus = minorBonus + caster:getMerit(rdmMerit[element]);
+  end
+
+  -- BLU mag acc merits - nuke acc is handled in bluemagic.lua
+  if (skill == BLUE_SKILL) then
```
teschnei: as the comment says, this should just be in bluemagic.lua passed in as bonus macc

teschnei:
you made a binding for getSkillType but never used it (every script already defines the skill type manually in their calls), not sure if you were intending to keep that for something

other than those it's good, so push those fixes here and then I'll merge it

scripts/globals/magic.lua
```diff
@@ -618,211 +467,124 @@ function applyResistanceEffect(player,spell,target,diff,skill,bonus,effect)
...
+  local p = 0;
+
+  if(minorBonus == nil) then
+    minorBonus = 0;
+  end
+
+  -- Get the base acc (just skill + skill mod (79 + skillID = ModID) + magic acc mod)
+  local magicacc = caster:getMod(MOD_MACC);
+
+  if(skillType ~= 0) then
```
xdemolish: stick a space after between if/elseif's and ('s cause readability
my second monitor sucks :(

scripts/globals/magic.lua
bendangelo: I moved it because skill is not always defined. Check the ability method.

scripts/globals/magic.lua
bendangelo: I just copied this from the original method.

teschnei: I know, but if we're here to refactor it, might as well...

scripts/globals/magic.lua
teschnei: hmm, add it to "local magicacc = caster:getMod(MOD_MACC)" then in getMagicHitRate

scripts/globals/magic.lua
bendangelo: Just checked, can't move it because blue spells use `applyResistanceEffect` for checking resistance and that doesn't pass through bluemagic.lua.

Anyway yeah, I don't want to touch too many things. I'll refactor more in smaller pieces.

bendangelo:
I guess I'll just change my editor's space indentation to 4. I'm keeping skilltype for later.

src/map/lua/lua_spell.cpp
```diff
@@ -176,6 +176,13 @@ inline int32 CLuaSpell::getID(lua_State *L)
 	return 1;
 }
 
+inline int32 CLuaSpell::getSkillType(lua_State *L)
+{
+	DSP_DEBUG_BREAK_IF(m_PLuaSpell == nullptr);
```
teschnei: still tabs here

PR #2066, opened 2015-09-13
Fix crash

TeoTwawki:
Fixes a crash triggered when mob has nothing to do but spam alternate ACTION_ATTACK and ACTION_MOBABILITY_START...They cycle so quickly that other things break.

I caused this in ceb079ae714f71fa8d3fd78b1033aceffa02fbec but it was not revealed during testing. I was testing on elemental, which occasionally break up their melee by casting. I triggered this crash during normal gameplay by silencing an elemental that had full TP..Made a fix after seeing a crash dump that had:
```
 	DSGame-server.exe!CAIMobDummy::ActionAttack() Line 1313	C++
 	DSGame-server.exe!CAIMobDummy::TransitionBack(bool skipWait) Line 2592	C++
 	DSGame-server.exe!CAIMobDummy::ActionAbilityStart() Line 722	C++
 	DSGame-server.exe!CAIMobDummy::ActionAttack() Line 1431	C++
 	DSGame-server.exe!CAIMobDummy::TransitionBack(bool skipWait) Line 2592	C++
```
Repeated for about 5 thousand lines in my call stack..

### DIFF
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -27,6 +27,7 @@
 #include <string.h>
 
 #include "mobentity.h"
+#include "../utils/battleutils.h"
 
 CMobEntity::CMobEntity()
 {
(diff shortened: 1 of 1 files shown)


src/map/ai/ai_mob_dummy.cpp
```diff
@@ -719,6 +719,7 @@ void CAIMobDummy::ActionAbilityStart()
     // We don't have any skills we can use, so let's go back to attacking
     if (MobSkills.size() == 0)
     {
+        m_PMob->PBattleAI->SetMobAbilityEnabled(false);
         TransitionBack(true);
```
bendangelo: This would break mob's that don't have any skills and then later on get skills. I think the best solution is to prevent the mob from going into this method if it doesn't have any skills. Add the check to line 1427 and 1905.

TeoTwawki: > This would break mob's that don't have any skills and then later on get skills.

What are you smoking?

If you add skills to mobs that don't have any you are restarting so it can see the SQL change anyways, at which point this no longer effects that mob.

TeoTwawki: If you meant mobs starts with no skills then skill get unlocked during the fight..please show me any one mob that A) actually does that and B) is not capable of setting that bool again when it does so.

PR #2069, opened 2015-09-14
TOAU Mission 1

Turttle:
Implemented last zone check. To match retail, Land of Sacred Serpent can only be completed by traveling from Mhaura to Aht Urhgan Whitegate using the Open sea route to Al Zahbi(Ferry).


teschnei:
Nah. I've completed the mission many times without ever taking the boat on
retail

PR #2071, opened 2015-09-14
Prevent some server variables related to the old Land Kings from incrementing multiple items on a single kill.

TeoTwawki:
Today @Hozu mentioned to me that the were increment per-party member because of the nature of onMobDeath.

Since these never despawn except after being killed, and onMobDespawn will run after onMobDeath has finished, this should be a safe change. The only way for it to increment more than intended here would require abuse of GM commands.


TeoTwawki:
Amended to fix the broken Knight Crabs / King Arthro as well...

hooksta4:
Ix drk has the same issue with Serv var adding however many people are in the party...

TeoTwawki:
@hooksta4 I just checked and that deos not appear to be the case at all. There is no var being repeatedly incremented there.

PR #2073, opened 2015-09-14
Land of Sacred Serpent Last Zone check

Turttle:
Implemented last zone check. To match retail, Land of Sacred Serpent can
only be completed by traveling from Mhaura to Aht Urhgan Whitegate using
the Open sea route to Al Zahbi(Ferry).


Turttle:
Moving the check to the NPC:  Salaheem's Sentinels. Removing Zone check. Adding KI Boarding_Pass check.

PR #2074, opened 2015-09-14
Boarding Permit Check

Turttle:
Implemented Boarding Permit KI check to be able to progress on the TOAU
story line. Boarding Permit is obtained on TOAU mission 1.


teschnei:
see my comment on https://github.com/DarkstarProject/darkstar/pull/2069

PR #2080, opened 2015-09-16
Refining Magic Resist Formula

bendangelo:
The magic resist formula really sucks, and is not even close to retail, so lets make this a team effort to improve it. I have gone ahead and wrote a simple function that outputs magic hit rate per level as we currently have it.

When you run dsgame, you'll see something like this:
```
[15/Sep] [21:09:39][LUA Script] -------- CasterLvl: 61
[15/Sep] [21:09:39][LUA Script] Lvl: 61 vs 56, 95%, MA: 211, ME: 153
[15/Sep] [21:09:39][LUA Script] Lvl: 61 vs 57, 95%, MA: 210, ME: 157
[15/Sep] [21:09:39][LUA Script] Lvl: 61 vs 58, 95%, MA: 209, ME: 162
[15/Sep] [21:09:39][LUA Script] Lvl: 61 vs 59, 95%, MA: 208, ME: 166
[15/Sep] [21:09:39][LUA Script] Lvl: 61 vs 60, 95%, MA: 207, ME: 171
```

Let's discuss how we can improve the formula hopefully with some research to back it up.

We should first agree on a baseline, what should the magic hit rate be for these mobs? Assuming general stats with maxed skills.

Lvl 10 vs lvl 5 mob?
Lvl 10 vs lvl 6 mob?
Lvl 10 vs lvl 7 mob?
Lvl 10 vs lvl 8 mob?
Lvl 10 vs lvl 9 mob?
Lvl 10 vs lvl 10 mob?
Lvl 10 vs lvl 11 mob?
Lvl 10 vs lvl 12 mob?
Lvl 10 vs lvl 13 mob?
Lvl 10 vs lvl 14 mob?
Lvl 10 vs lvl 15 mob?

### DIFF
--- a/scripts/globals/magic.lua
+++ b/scripts/globals/magic.lua
@@ -334,21 +334,21 @@ function applyResistance(player,spell,target,diff,skill,bonus)
         end
     end
 
-    local magicacc = 0;
+    local percentBonus = 0;
 
     if player:hasStatusEffect(EFFECT_ALTRUISM) and spell:getSpellGroup() == SPELLGROUP_WHITE then
-        magicacc = magicacc + player:getStatusEffect(EFFECT_ALTRUISM):getPower();
+        magicaccbonus = magicaccbonus + player:getStatusEffect(EFFECT_ALTRUISM):getPower();
     end
 	
     if player:hasStatusEffect(EFFECT_FOCALIZATION) and spell:getSpellGroup() == SPELLGROUP_BLACK then
-        magicacc = magicacc + player:getStatusEffect(EFFECT_FOCALIZATION):getPower();
+        magicaccbonus = magicaccbonus + player:getStatusEffect(EFFECT_FOCALIZATION):getPower();
     end
     --difference in int/mnd
 	
     if (diff > 10) then
-        magicacc = magicacc + 10 + (diff - 10)/2;
+        magicaccbonus = magicaccbonus + 10 + (diff - 10)/2;
     else
-        magicacc = magicacc + diff;
+        magicaccbonus = magicaccbonus + diff;
     end
 	
     --Add acc for dark seal
@@ -376,7 +376,7 @@ function applyResistance(player,spell,target,diff,skill,bonus)
         magicaccbonus = magicaccbonus + 25;
     end
 
-    local p = getMagicHitRate(player, target, skill, element, magicacc, magicaccbonus);
+    local p = getMagicHitRate(player, target, skill, element, percentBonus, magicaccbonus);
 
     return getMagicResist(p);
 end;
@@ -395,7 +395,7 @@ function applyResistanceEffect(player,spell,target,diff,skill,bonus,effect)
 
     local magicaccbonus = 0;
     local element = spell:getElement();
-    local magicacc = 0;
+    local percentBonus = 0;
 
     if (bonus ~= nil) then
         magicaccbonus = magicaccbonus + bonus;
@@ -408,17 +408,17 @@ function applyResistanceEffect(player,spell,target,diff,skill,bonus,effect)
     end
 
     if player:hasStatusEffect(EFFECT_ALTRUISM) and spell:getSpellGroup() == SPELLGROUP_WHITE then
-      magicacc = magicacc + player:getStatusEffect(EFFECT_ALTRUISM):getPower();
+      magicaccbonus = magicaccbonus + player:getStatusEffect(EFFECT_ALTRUISM):getPower();
     end
     if player:hasStatusEffect(EFFECT_FOCALIZATION) and spell:getSpellGroup() == SPELLGROUP_BLACK then
-      magicacc = magicacc + player:getStatusEffect(EFFECT_FOCALIZATION):getPower();
+      magicaccbonus = magicaccbonus + player:getStatusEffect(EFFECT_FOCALIZATION):getPower();
     end
 
     --difference in int/mnd
     if (diff > 10) then
-        magicacc = magicacc + 10 + (diff - 10)/2;
+        magicaccbonus = magicaccbonus + 10 + (diff - 10)/2;
     else
-        magicacc = magicacc + diff;
+        magicaccbonus = magicaccbonus + diff;
     end
     --add acc for ele/dark seal
     if (player:getStatusEffect(EFFECT_DARK_SEAL) ~= nil and skill == DARK_MAGIC_SKILL) then
@@ -465,7 +465,7 @@ function applyResistanceEffect(player,spell,target,diff,skill,bonus,effect)
         end
 
         if (effectres > 0) then
-            magicacc = magicacc - target:getMod(effectres);
+            percentBonus = percentBonus - target:getMod(effectres);
         end
 
         if(effect == EFFECT_LULLABY) then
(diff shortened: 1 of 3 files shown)


teschnei:
you're better off doing this matching gear/stats of some empirical data to
some monster, there isn't going to be any good data of a non-75 BLM wearing
no gear with capped skills

bendangelo:
Sure but this is to first set a general baseline, then once it seems right we can further refine it with realistic scenarios.

Just looking at the initial info `Lvl: 61 vs 56, 95%` is definitely wrong.

bendangelo:
Just pushed a new formula. I been playing around with it and seems ok so far. I'm concerned it'll make HNMs too resistant to magic. Does anyone know the chances of resist of a 75 vs 95 HNM? Less than 50% non-resist?

teschnei:
i know my 99 sch non-resist rate against a VD HKCNM boss, but that's all

(it's 5% without buffs)

teschnei:
I can probably write a bot to just cast spells and record the damage for a few hours, though.  Think I only have 99 casters, but it's still good info - try to get some info on level penalties etc

bendangelo:
I think that'll be useful. I need info on chances of resist vs same level and vs much higher levels.

Deadwing888:
The resist rate of a level 75 vs a 95 HNM is going to vary wildly based on what element you're nuking vs which hnm. For instance it's completely pointless to nuke anything but water or thunder vs Tiamat.

It's also going to vary wildly based on how much elemental magic gear and macc you're sacking. The general rule of thumb in the era was you wanted to be north of 340 elemental magic skill on Tiamat before you started focusing on MAB. Though we also used threnody and I'm super hazy on exactly how that lowered meva. 

But then if you look at Hydra (even though it's only level 85) it's completely pointless to nuke it period as it's got just too much meva. I think on the classic HNM front it's going to be a real game of fine tuning each one's magic evasion for each element; you won't be able to make one formula fit all.

bendangelo:
That's good info thanks Deadwing. Just tested that case with the current formula.

75 vs 95 with 340 magic acc has a 46% chance to not get resisted.
75 vs 95 with 340 magic acc casting a spell it's weak against has a 58% chance to not get resisted.

I noticed at all levels, casting a spell on a mob of the same level (40 vs 40, 75 vs 75) will have around a 50-60% chance to not get resisted at all. This is without any gear and buffs. With good gear it can easily be boosted to 90%, vs EM and even vs VT.

I think the formula is pretty good the way it is.

Deadwing888:
I read through the formulas in magic.lua till my head hurt (I'm just not code oriented enough to grasp it all on the first read). There are some things that are odd to me, but I'm going to wait and look at it with fresh eyes tomorrow, and then maybe parse retail to see how it matches.

Now when I actually play on Nasomi I seem to notice debuffs only ever getting half resists? I see a lot of full duration and half duration binds and sleeps, but never ever see quarter eighth or sixteenth resists. The exception to this may be paralyze which I think I have seen get smaller resists.

I've also noticed elemental seal doesn't seem to be boosting magic accuracy by enough. Wiki doesn't really lend any magic accuracy specifics; however the conventional wisdom in the level 75 days was that unless the mob was immune to a debuff or a specific damage type then the spell would always land full strength. On darkstar it's a boost but it just doesn't feel like enough.

In terms of the Tiamat test, I think it shows the divide between elements HNM are strong to and element's they're weak to is not wide enough. For instance just last night in sky we had a black mage do a 1k quake vs Kirin (who is of course an earth-based mob). In retail he would have been resisted for sure. 

In terms of fixing this sort of thing I know in sql there is a place to adjust elemental strengths and weaknesses for particular elements; however when I did some tests with that it seemed to affect MDB and Meva at the same time. I got the most true-to-retail results by boosting MOD_THUNDERRES etc as a line in the HNM's .lua file but this generated some complications because if I boosted it to the correct amount then it became impossible to stun properly. Is there a better way for me to be going about this?

In my mind if MOD_STUNRES etc could be configured to accept a particular value for "stun meva" then it would be possible to side-by-side tests with each HNM using an identical character in retail as a control variable (I already have this sort of data for some hnms). Doing this with every mob would obviously be insane; however I do think in terms of the classic game there are fifty or so NMs for which they do need this personal touch to get that retail feel.

Sorry for typing such a long comment.

bendangelo:
>  I see a lot of full duration and half duration binds and sleeps, but never ever see quarter eighth or sixteenth resists.

Yeah right now it's either full, half or resisted. I believe the duration is slightly random on retail but it isn't here.

> I've also noticed elemental seal doesn't seem to be boosting magic accuracy by enough.

Just checked this adds 256 magic acc which is quite a lot. I will test this in game though.

> if I boosted it to the correct amount then it became impossible to stun properly. Is there a better way for me to be going about this?

There isn't currently. What this tells me is elemental resistances has less affect on status effects of the same type. So all of thunder resistance shouldn't carry over to stun resistance, it should maybe be 50%.

teschnei:
Enfeeble duration is not random on resists on retail

bendangelo:
@teschnei Reason I mention that is because I found this info from ffxi wiki. http://ffxiclopedia.wikia.com/wiki/Magic_Hit_Rate Look at the bottom of the page

teschnei:
hm.. i'll rephrase then: not all enfeebles have randomization on duration (slow, for example)

bendangelo:
So the issues brought up by @Deadwing888 I will resolve in an other PR. So I want to get this thing merge in, I tested these changes in realistic scenarios and the new resist rates seem reasonable.

If there are no other concerns I'll rebase and merge this in today.

teschnei:
No, I haven't looked at it yet

TeoTwawki:
> Reason I mention that is because I found this info from ffxi wiki. http://ffxiclopedia.wikia.com/wiki/Magic_Hit_Rate Look at the bottom of the page

keep in mind any of the wikis can be wrong at times. It happens.

Even people who have played the game for years will occasionally say something so crazy you'll be like "What? Are we playing the same game??" while they are in the same party with you..And that guy could be editing the wiki and the guys who know better not caring enough to correct it. Though in general I trust Bg folks when math is involved and trust 'clopedia when its guide-like.

Not saying whatever that page says is wrong, just saying its always a possibility, and the one on wikia especially always did have a lot of conjecture (oh man, gardening info took years to clean up..).

bendangelo:
Yeah I agree, but right now I'm leaving that issue to an other PR.

I just started doing some testing on HNMs and the numbers just don't add up. In mobutils mobs over lvl 83 get a huge bonus to magic evasion and it's impossible to get a decent hit rate, even with the best gear. I have to rework the equation a little bit, thinking about removing the `scale` variable and just giving mobs a better magic evasion ranking as they level up, it would mathematically make more sense.

More testing tomorrow.

Deadwing888:
I don't know if it's retail or not, but removing the scale variable and doing a more conventional meva rank based on level makes loads of sense to me. The certain HNM with unconventionally high meva could just have it sprinkled in via their .lua

I know it may be a different issue/pull request but I was thinking the other night about the Elemental Seal thing, and I'm wondering if it's actually got plenty of magic accuracy, but what's happening is it's hitting some sort of 95% magic hit rate cap? I guess I have two questions:
 
1) Does darkstar have a magic hit rate cap?
2) Should it? I know melee hit rate caps at 95% but I sort of thought magic hit rate went all the way to 100% and I know for a fact Elemental Seal should take it there unless it's hitting trait proc or immunity on a debuff.

teschnei:
1) it does
2) both wikis say 95%, and that seems in line with what I've personally
experienced, but I have no reason to believe elemental seal goes past this
(it's certainly more than 256 flat macc though, as I just slept a 130
rabbit as a naked sch99 with only 282 enfeeble)

PR #2140, opened 2015-09-28
Scripts: Fix Cura's future Misery support

Hozu:
-Made the spells store the power of the Misery mod as a local var upon casting if the caster's ID == the target's ID. This enables all who are affected by the spell to get the boosted heal.
-Made the spell only able to be cast on the caster.

Please note, that since the Misery mod itself is not fully implemented (doesn't store the damage from the last hit), the boost won't actually be applied - I tested by manually setting the mod power.

### DIFF
--- a/scripts/globals/spells/cura.lua
+++ b/scripts/globals/spells/cura.lua
@@ -16,105 +16,113 @@ require("scripts/globals/magic");
 -----------------------------------------
 
 function onMagicCastingCheck(caster,target,spell)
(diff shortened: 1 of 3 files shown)


xdemolish:
i demand misery
make me miserable

xdemolish:
or just rename the commit so the message is clear that you added support for when misery does eventually get implemented

Hozu:
Changed commit message + title of pull.

PR #2143, opened 2015-09-28
SCRIPTS: Even More Guillibles Travels fix and cleanup

waytim:
In my original PR I forgot to add the .sql changes, but discovered now I could use one npc lua for all 3 without needing to change names in SQL. Also, for Personal Hygiene I removed an unnecessary var and changed another to local var.


scripts/zones/Yuhtunga_Jungle/npcs/Blue_Rafflesia.lua
```diff
@@ -0,0 +1,132 @@
...
+            end
+        else
+            player:setVar("FirstBlueRafflesiaCS", 1);
+            player:addItem(1144);        -- Rafflesia Nectar
+            player:messageSpecial(ITEM_OBTAINED,1144);
+        end
+    elseif (csid == 0x0016 and option == 1) then
+        if (player:getVar("FirstBlueRafflesiaCS") == 1 and player:getVar("ThirdBlueRafflesiaCS") == 1) then
+            player:setVar("FirstBlueRafflesiaCS", 0);
+            player:setVar("ThirdBlueRafflesiaCS", 0);
```
xdemolish: instead of using 3 separate vars, use a bitmask

e.g. in the first one, set it to
```
    local rafflesiaCS = player:getVar("RafflesiaCS");
    bit.bor(rafflesiaCS, 0x01); -- set the bit for first cs
```

in the second one, do the same but set it to 0x02, and in the third do the same but set it to 0x04 (start at 0x01 for flagged, then multiply each subsequent value by 2 for each additional mask e.g. 0x01, 0x02, 0x04, 0x08)

once you've set each flag, you can check by adding the bits you want to check and comparing using `bit.band(mask, bits to check)` e.g. 
```
if bit.band(rafflesiaCS, 0x01) then -- first cs only
    -- do shit
elseif bit.band(rafflesiaCS, 0x05) then -- we want to check both the first and third have been triggered, so add the bits for first (0x01) + third (0x04) = 0x05
    -- do shit
```

waytim:
Thanks for the explanation demo, but it's a bit over my head even with how clearly you laid it out.  I need to check that they have clicked on that npc before or else they could keep getting the item over and over. I'm sorry I ever PR'd these quests, i did not realize the amount of work and learning required,  but I don't want to leave this broken now that it's been merged, so I need someone to literally walk me through this on IRC sometime if they could.

xdemolish:
well you can either bit.bor or set the var to var+bits you wanna set e.g.
```
player:setVar("RafflesiaCS, player: getVar("RafflesiaCS") + 0x01);
```
in the cs thing crap

bendangelo:
I think the PR is ok without using bitwise operators, I am kinda against using them because it breaks clarity and there isn't a good reason to use them in this case. Are you trying to save database space? Why not just use more than one var?

xdemolish:
less vars to track/clear if it does break at some point, and i dislike player vars in general cause errybody uses forgets to clear them

oh and i only got as far as setting three separate vars flagged with a value of 1 before making my previous comment (didn't actually take a proper look at the time) so if there's no point in bit'ing here, might as well merge

TeoTwawki:
@waytim @xdemolish 
We actually have a function that handles bitmasked vars as well, and its much easier for new people to grasp. See NPC's involved in bastok's Stamp Hunt Quest, or any of the 4 Lure of the Wildcat quests, or the summoner unlock quest, for examples. I believe that function actually handles the bits in C++ instead of lua, but its wrapped in a fairly easy to use function. So far as I know it is specifically for player vars only, which means its for this exact use case.

I hope that helps future cases.

Also:
@bendangelo I detest player vars for same reason plus that table is already bloaty and massive even with a small number of players. The fewer additional we use the better.

TeoTwawki:
For clarification: I am not talking about changing an already merged PR, I am pointing out a thing for future PR's :smile:

PR #2145, opened 2015-09-29
SQL: Goblin NM family/skill list ID fix for Dynamis - Beaucedine

Hozu:
Changed their family to 373 so they can use the various Goblin Dice moves.


Hozu:
Updated the skill list ID as well.

PR #2156, opened 2015-09-30
SQL: Dimensional Death/Barbed Crescent ID check fix

Hozu:
Another broken spell with the mobskill/family separation, again due to fomor stuff. Though this separation means that Barbed Crescent should finally be learnable.


teschnei:
FYI: I can basically guarantee it will break again when I merge the ai
rewrite

Hozu:
Really? Well it can be easily fixed at that point at least.

teschnei:
Yes, just that I probably won't remember to do it :p

Hozu:
Rather have it working now then break later, and be easily fixed after someone catches it over waiting for the ai rewrite to be finished and fix it then, honestly.

teschnei:
I'm not saying this shouldn't be merged, just something I thought of since
this will essentially have to be reverted then

bendangelo:
I'm going to merge this because it does fix the issue.

PR #2162, opened 2015-10-01
Scripts: Couple of extra Spirit Link effects

Hozu:
Added the removal of Blind, level/3 tic regen of 90 seconds, and only wake from sleep if pet is healed as per https://www.bg-wiki.com/bg/Spirit_Link

### DIFF
--- a/scripts/globals/abilities/spirit_link.lua
+++ b/scripts/globals/abilities/spirit_link.lua
@@ -60,21 +60,25 @@ function onUseAbility(player,target,ability)
     local pet = player:getPet();    
     local healPet = drainamount * 2;
     local petTP = pet:getTP();
+    local regenAmount = player:getMainLvl()/3; -- level/3 tic regen
     
     if (player:getEquipID(SLOT_HEAD)==15238) then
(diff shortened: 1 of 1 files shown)


scripts/globals/abilities/spirit_link.lua
```diff
@@ -60,21 +60,25 @@ function onUseAbility(player,target,ability)
...
     if (math.random(1,2) == 1) then
         pet:delStatusEffect(EFFECT_DOOM);
     end
-    
+    if (pet:getHP() < pet:getMaxHP()) then -- sleep is only removed if it heals the wyvern
+        removeSleepEffects(pet);
+    end
+
+    pet:addHP(healPet); --add the hp to pet
+    pet:addStatusEffect(EFFECT_REGEN,regenAmount,3,90,0,0,0) -- 90 seconds of regen
```
bendangelo: Missing semi colon

xdemolish: lua doesn't care about semi-colons

but ya, semi-colon :p

PR #2165, opened 2015-10-03
Making several cpp functions use proper weaponSlot. Adding SLOT_KICKS. 

GerrardOfTitan:
This is my first time issuing a pull request in github. I'm sorry if I missed any required memos on github in general, or for the darkstar project. 

I'm also not sure that this change is complete. I'm particularly concerned i missed some corner cases involving kick attacks. I did a quick test of kick attacks and they seem to be working (not crashing the zone server), but that's the extent of my testing of kick attacks. 

This proposed change: 

Adding new SLOTTYPE named SLOT_KICKS. 

Changing CBattleEntity::m_Weapons to have a new weapon for SLOT_KICKS. It's a default weapon, except with SKILL_H2H. 

Changing battleutils::GetDamageRatio to take additional argument weaponSlot. 

For kick attacks, making CAttack::GetWeaponSlot() return SLOT_KICKS. 

Changing CBattleEntity::ATT to take arguments weaponSlot and attackOffset. 

Refactoring implementation of CBattleEntity::ATT and CBattleEntity::ACC. 

Removing MOD_ATT as a valid option for CLuaBaseEntity::getStat. Replacing with new Lua binding CLuaBaseEntity::getATT. 

Fixes an issue where there are two files in the repo with the same path case insensitive (searing_light.lua). This change removes one of the paths.

Making other necessary follow-up changes.


teschnei:
firstly... why? what is the problem being fixed?

secondly, I'm not really a fan of SLOT_KICKS. There's no equipment slot for "kicks", so it doesn't really make any logical sense and is kind of just confusing

GerrardOfTitan:
There's several problems I'm trying to fix. 

First, the current code always calculates attack according to the main-hand weapon, including when making dual-wield attacks with the off-hand weapon. This doesn't require SLOT_KICKS. 

Second, kick attacks should not use the base weapon damage of the main-hand weapon (AFAIK). This is what SLOT_KICKS is for. I agree it looks hacky. I wonder if there's a better way. Previously, I coded it up where instead of SLOT_KICKS, I had SLOT_NONE, and I did a bunch of checks for SLOT_NONE and/or nullptr CWeapon pointers, but in the end I think that was worse. I'm definitely open to comments here.

GerrardOfTitan:
Also, concerning proper etiquette, what is the best place to discuss changes? A forum? Here in pull requests? Open an issue first?

teschnei:
here is fine, but IRC works too if there will be lots of back and forth that doesnt require commenting directly on lines (mostly because each comment emails everyone)

I don't think you're right on either points though - the HORRILE HORRIBLE code in CAttack does check both the main/sub weapon different, and the kick attacks

although looking at this it's using the main weapon's damage rating for kick attacks when it should just be damage from hand to hand skill plus kick mod..

GerrardOfTitan:
Yea, I'm new to the code. Sorry for the complex change first. I have a couple easier changes that I'll do a pull request for, such as Cyranuce_M_Cutauleon (the drg unlock BCNM mob). 

Let me find the IRC channel and see if I can log in. Also more-or-less completely new to IRC.

PR #2166, opened 2015-10-03
Fixing drg-unlock BCNM and Chaos_Breath. 

GerrardOfTitan:
Fixing mob in BCNM that unlocks drg. Making it use mobtpskills. 

Also enabling Chaos_Breath, one of the tp skills that the mob uses.


bendangelo:
thanks

PR #2167, opened 2015-10-03
Fixing CBattleEntity::ATT and battleutils::GetDamageRatio to take into account main vs sub weapon

GerrardOfTitan:
Changing battleutils::GetDamageRatio to have additional argument weaponSlot in order to properly calculate the attack value.
- Made necessary follow-on changes.

Changing CBattleEntity::ATT to have additional argument attackNumber in order to properly calculate the attack value.
- Made necessary follow-on changes.

Refactor of CBattleEntity::ATT, ACC, RATT, and RACC.
- While doing the refactor, found and fixed a bug in CBattleEntity::RATT. It was using MOD_ATTP instead of MOD_RATTP for PUP pets.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -2067,7 +2067,7 @@ inline int32 CLuaBaseEntity::getStat(lua_State *L)
         case MOD_INT:  lua_pushinteger(L, PEntity->INT()); break;
         case MOD_MND:  lua_pushinteger(L, PEntity->MND()); break;
         case MOD_CHR:  lua_pushinteger(L, PEntity->CHR()); break;
-        case MOD_ATT:  lua_pushinteger(L, PEntity->ATT()); break;
+        case MOD_ATT:  lua_pushinteger(L, PEntity->ATT(SLOT_MAIN, 0)); break;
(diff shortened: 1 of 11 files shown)


GerrardOfTitan:
I noticed a couple other things during my refactor. For two-hander weapons: the code adds 75% of dex as a bonus to acc for PCs only, not monsters, but the code adds 75% of str as a bonus to att for PCs and monsters. Is that intended? Is this moot because monsters cannot use two-hander weapons? I don't know offhand.

src/map/lua/lua_baseentity.cpp
```diff
@@ -2067,7 +2067,7 @@ inline int32 CLuaBaseEntity::getStat(lua_State *L)
         case MOD_INT:  lua_pushinteger(L, PEntity->INT()); break;
         case MOD_MND:  lua_pushinteger(L, PEntity->MND()); break;
         case MOD_CHR:  lua_pushinteger(L, PEntity->CHR()); break;
-        case MOD_ATT:  lua_pushinteger(L, PEntity->ATT()); break;
+        case MOD_ATT:  lua_pushinteger(L, PEntity->ATT(SLOT_MAIN, 0)); break;
```
teschnei: we are definitely going to have to expose the slot to Lua and have the scripts check it

PR #2168, opened 2015-10-03
Elemental Affinity revamp + misc. corrections

Hozu:
Looking at this http://ffxiclopedia.wikia.com/wiki/Magic_Affinity page on the wiki, it's clear that the affinity boosts need to be separated for when Magian stuff is implemented. I've also added an all affinity mod to be used by the Iridal/Chatoyant Staff.

The affinity damage mods gives +/-5% damage per level, affinity acc mods give +/-10 acc per level, and affinity perp mods give -/+1 perp per level. Because of this, each item will have one more level of affinity than listed on various wikis (example Fire Staff has 2 in each of fire affinity dmg/acc/perp).

I've corrected some items mods in the database to use elemental resistance instead of affinity, and corrected the bug where using the staff strong against the element of your avatar (ie. Ice Staff with Garuda) reduces perpetuation. I've also (re?)implemented the elemental accuracy mods used by the elemental grips.


Hozu:
Corrected a minor mistake - I was grabbing the all affinity damage mod in the affinity accuracy function instead of the all affinity accuracy mod.

bendangelo:
Thanks.

PR #2171, opened 2015-10-04
fixing PC DRG pet wyvern breath abilities

GerrardOfTitan:
I tested the fix for healing breaths for defensive-type wyvern, hybrid-type wyvern. Tested that offensive breaths are used for sub war and not used for sub whm. 

The person who is maintaining this code might think another way is cleaner, but this works. One thing that I might want to do personally is create a constant in a header for this skill list id to avoid a magic hardcoded number.

### DIFF
--- a/src/map/utils/petutils.cpp
+++ b/src/map/utils/petutils.cpp
@@ -1400,6 +1400,10 @@ namespace petutils
 		PPet->SetMJob(JOB_DRG);
 		PPet->SetMLevel(PMaster->GetMLevel());
 
+        if (PMaster->objtype == TYPE_PC) {
+            PPet->m_MobSkillList = 193; //set the mob skill list id for player dragoon wyvern pets
+        }
+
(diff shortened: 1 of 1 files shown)


teschnei:
it's probably better for player wyverns to have their own mob pool (IIRC the pool used for player summons is in pet_list)

src/map/utils/petutils.cpp
```diff
@@ -1400,6 +1400,10 @@ namespace petutils
 		PPet->SetMJob(JOB_DRG);
 		PPet->SetMLevel(PMaster->GetMLevel());
 
+        if (PMaster->objtype == TYPE_PC) {
```
bendangelo: You actually don't even need this condition. This method is only used by players.

bendangelo:
Should probably add a new mob pool and grab the skill list from the db. Don't think it's currently being loaded.

GerrardOfTitan:
Redoing change as per above suggestions. Will open new pull request.

PR #2195, opened 2015-10-07
[SCRIPT] ZNM Tinnin

nesstea:
Coded the AI for the T4 ZNM. Adjusted mobskills to reflect the wiki.


scripts/globals/mobskills/Serpentine_Tail.lua
```diff
@@ -13,21 +13,21 @@ require("scripts/globals/monstertpmoves");
...
 
-    if (mobSkin == 1796) then
+        if (mobSkin == 1796) then
+            return 0;
+        else
+            return 1;
+        end
+    end
+
+    if (mob:AnimationSub() < 2 and target:isBehind(mob, 48) == true and mob:getFamily() == 313) then -- Tinnin
```
bendangelo: To be honest I don't like all these conditions. I feel like it's better to make multiple mob skill lists and change the mobs skill list when it changes states. You can do it from lua like so `mob:setMobMod(MOB_SKILL_LIST, {listId})`

scripts/zones/Wajaom_Woodlands/mobs/Tinnin.lua
```diff
@@ -0,0 +1,161 @@
...
+function onCriticalHit(mob)
+    local critNum = mob:getLocalVar("crits");
+    
+    if ((critNum+1) > mob:getLocalVar("CritToTheFace")) then  -- Lose a head
+        if (mob:AnimationSub() == 0) then
+            mob:AnimationSub(1);
+            mob:setMod(MOD_REGEN, 60);
+            mob:setLocalVar("headTimer", os.time() + math.random(60,190));
+        elseif (mob:AnimationSub() == 1) then
+            mob:AnimationSub(2);
```
bendangelo: This is where you would change it's skill list. `mob:setMobMod(MOBMOD_SKILL_LIST, {listid})`

nesstea: I would need 3 skill lists in the db and it always has access to TP moves. It loses breath attacks when the heads poof.

PR #2196, opened 2015-10-07
[SCRIPT] Imp defense and horn-breaking AI

nesstea:
You can now break horns! This isn't all the imps in the game, but should be plenty enough until the others need coding/AI.


nesstea:
I missed the non-versioned files.

bendangelo:
There are two imp families but you are only checking for one. Should you check for both families?

Hozu:
Aren't Imps able to pull out another horn after some period of time?

nesstea:
not that I recall.

bendangelo:
Thanks

PR #2199, opened 2015-10-07
CORE: fix

nasomi:


teschnei:
please don't merge core changes before travis is done building, that's why we have it

PR #2200, opened 2015-10-07
Fixed deleting spells/missions/ki's/quests on multi-server platforms

nasomi:
Information was deleted on changing servers is reloaded to prevent writing bad information to the blobs. Has no negative effects on single servers but is critical for multi-servers.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -1296,7 +1296,7 @@ inline int32 CLuaBaseEntity::setRankPoints(lua_State *L)
     DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
 
     CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
-
+	charutils::LoadMissions(PChar);
     PChar->profile.rankpoints = (int32)lua_tointeger(L, 1);
     charutils::SaveMissionsList(PChar);
     return 0;
@@ -1312,7 +1312,7 @@ inline int32 CLuaBaseEntity::addRankPoints(lua_State *L)
     DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
 
     CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
-
+	charutils::LoadMissions(PChar);
     PChar->profile.rankpoints += (int32)lua_tointeger(L, 1);;
     charutils::SaveMissionsList(PChar);
     return 0;
@@ -1341,7 +1341,7 @@ inline int32 CLuaBaseEntity::setRank(lua_State *L)
     DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
 
     CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
-
+	charutils::LoadMissions(PChar);
     PChar->profile.rank[PChar->profile.nation] = (int32)lua_tointeger(L, 1);;
     charutils::SaveMissionsList(PChar);
     return 0;
@@ -1358,7 +1358,7 @@ inline int32 CLuaBaseEntity::addQuest(lua_State *L)
     DSP_DEBUG_BREAK_IF(lua_isnil(L, -2) || !lua_isnumber(L, -2));
 
     CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
-
+	charutils::LoadQuests(PChar);
     uint8 questID = (uint8)lua_tointeger(L, -1);
     uint8 logID = (uint8)lua_tointeger(L, -2);
 
@@ -1393,7 +1393,7 @@ inline int32 CLuaBaseEntity::delQuest(lua_State *L)
     DSP_DEBUG_BREAK_IF(lua_isnil(L, -2) || !lua_isnumber(L, -2));
 
     CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
-
+	charutils::LoadQuests(PChar);
     uint8 questID = (uint8)lua_tointeger(L, -1);
     uint8 logID = (uint8)lua_tointeger(L, -2);
 
@@ -1460,7 +1460,7 @@ inline int32 CLuaBaseEntity::completeQuest(lua_State *L)
     DSP_DEBUG_BREAK_IF(lua_isnil(L, -2) || !lua_isnumber(L, -2));
 
     CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
-
+	charutils::LoadQuests(PChar);
     uint8 questID = (uint8)lua_tointeger(L, -1);
     uint8 logID = (uint8)lua_tointeger(L, -2);
 
@@ -1535,7 +1535,7 @@ inline int32 CLuaBaseEntity::addMission(lua_State *L)
     if (LogID < MAX_MISSIONAREA && MissionID < MAX_MISSIONID)
     {
         CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
-
+		charutils::LoadMissions(PChar);
         if (PChar->m_missionLog[LogID].current != LogID > 2 ? 0 : -1)
         {
             ShowWarning(CL_YELLOW"Lua::addMission: player has a current mission\n" CL_RESET, LogID);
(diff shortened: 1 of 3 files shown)


teschnei:
screw, meet hammer

xdemolish:
taaaaaaaaaabs

teschnei:
you used tabs in a file filled with spaces, and spaces in a file filled with tabs

asdfhjklasdfasd

xdemolish:
not that i bothered to check, but you should prolly have these called in charutils::LoadChar since you wrote functions for them anyways (them being loading keyitems/missions/quests etc)

nasomi:
When i made it they were all on tabs.

teschnei:
either way, i don't really like the idea of loading the entire thing over and over - maybe a better idea is to never save the entire blob, and just flip the bit individually in the db when you learn/obtain it (if you can flip bits individually in an sql statement, that is)
if this fixes the problem though, then i suppose the problem was not that ki/spells/whatever blobs were being saved with uninitialized characters (since this wouldn't change that), but some normal function overwriting their blob with nothing? maybe there's an out of bounds array write, or the code to initialize the blob (with nothing) is being called when it shouldnt? I think, if you are okay with testing something different for me nasomi, I'd like to change those blobs to a container with bounds checking and automatic initialization (if possible) before doing something like this

nasomi:
It only loads it before it's updated. It's a safety check, and without hte functions data is lost constantly. I've been fighting iwth this data loss for a while. It may not be a great forever fix but it's critical to not lose data every single time a player zones right now. The functions are only called when the information is changed, not when reading, so it should not call the functions to often.

nasomi:
I've been talking with lautan about non-blob containers. The blobs themselves aren't the issue, it's how the data is being accessed and stored in memory that is creating the problems, and the blobs being difficult for humans to read. I've been running these fixes for 8 months and haven't had a problem with them yet.

nasomi:
if you want to hcange it up and move to table like merits, i'm all for it though

teschnei:
I am aware of how your change works, but a) this is difficult to test because I don't think it's "every single time a player zones" as you claim, and b) implies that there is a problem somewhere else, and this is a bandaid. Thus, I want to fix the actual problem (which appears to be that the arrays are reset in some fashion when they should not) by making that not possible.  It's silly to have the blobs loaded in memory if we can't even trust them to be right!
However, if it really is every single time a player zones, then I can test it myself.
Thinking about it a bit more, I may have an idea of a similar problem (though I still don't see how even that would reset everything).  I should have some time after work tonight to look into it - I really want to get the multiple server bugs all fixed ASAP

bendangelo:
It's possible to set individual bits with mysql `UPDATE users SET permission = permission | 16;`. But how about we just create a new table to store these attained items. Trust me performance will not be an issue.

nasomi:
Sorry, not every single time, but every time something changers, then the person transfers from one zone to another, where the two zones are on different servers, when the previous zone is also a different server, as best I can determine.

If it helps, the charutils::savekeyitems(pchar); in packet_system.cpp also wipes players KI's upon checking KI's. i have removed that line for the time being.

teschnei:
No

teschnei:
if the problem is core side memory overwriting (which is most likely, evidenced by the fact that these changes work), then changing the storage scheme in the database will do nothing, as it will still overwrite it all with nothing

xdemolish:
^ and you'd need a table for keyitems, another for missions, another for quests


unless i have no clue what even is being discussed here in which case this comment will be edited out of existence

nasomi:
Changing the storage could work, because you can't overwrite something with nothing when there's multiple fields. I'm not concerned about performance, but there are over two thousand ki's, not to mention hundreds of spells, etc, per character, could become a fairly large table, it just seems wasteful to do it like merits is done.

The issue is memory overwriting, the memory reserved is being written by something else, because the blob doesn't become null all the time, sometimes there is data in it, it's just bad data.

bendangelo:
So one solution is to do a bitwise set like so:
```
UPDATE chars SET spells = spells | %s;
``
But the spell list would still have to be reloaded because the data in memory might be different from the database. No matter which approach we should still be reloading the spells / whatever after saving.

The long term solution is to just create a new table called `char_attainables` or whatever. It will store all key items, spells, titles. When a new key item is gained, insert a new row. When a key item is removed, delete the row. This is a fool proof system and is how the database was designed to be used. With the proper indexes performance is not an issue.

We could keep using blobs but there will always be potential for these problems in the future.

teschnei:
doesn't fix the issue at hand, though (the memory in the game server is still corrupt! and we cannot skip saving it into memory and loading it from the db when we need it, because that is a huge performance loss!)

why should we reload the spells after saving? what is the use case of such a function?  when will the spell list ever change between saving and reloading?

Don't know if a bitwise set on the entire blob would work (what about when you unlearn a spell? memory is corrupted -> unmerit a spell -> deleted all your spells still).  I am for setting/unsetting the specific bit when you learn or unlearn a spell permanently (and never saving the whole blob), but that is just going to fix the loss of data being permanent (you will still have them "lost" until you relog)

nasomi:
My functions reload the spells *before* updating and saving, to make sure it has correct data before it adds or deletes a spell, then saves to the database.

bendangelo:
Ok so you're right we have to first reproduce the issue before we can make a proper fix for it. I don't know how to reproduce the problem but there are some steps we can take to prevent data loss and safe guard against the issue. We can even throw in some assertions to help catch the error in action.

Here's what I'm suggesting:
1. use std::bitset instead of uint8 arrays to hold the data. No reason to roll our own.
2. Create a new `char_attainables` table. It'll look like this: charId, typeId, value
3. Move all the blobs to this new table (Or whichever ones make sense)
4. addSpell / delSpell type methods will just insert / delete rows
5. I'll throw in some assertions for sanity checking

On your point of not needing to reload spells. Sure there is no reason to reload if the system is working correctly. It's more of a safe-guard to always ensure you have the latest data.

So with these changes worst case a player would just have to relog to get back their key items / whatever. In the mean team we can dig into the root cause.

I took a look at some of the examples of corrupted data and I don't think the bitwise operations will help us. Because it'll end up writing bad data and potentially giving the player random key items / spells.

teschnei:
1. yes
2. no
3. no
4. yes, but toggling the bit individually in the blob instead of adding rows in a new table
5. yes

bendangelo:
You have to explain why. The performance isn't any different.

teschnei:
it's functionally equivalent to doing a single bitset on the blob, so it's a waste of time to change it all, and potentially introduce new bugs, just to have it do the same thing

(this is assuming that we no longer save the blob all at once like we do now, and only "save" single bits at a time)

bendangelo:
I understand

nasomi:
As long as the age old issue doesn't persist, previously the blobs would change length with new spells, missions, KI's, etc. When that happened, whoever updated teh blob inserted the item with it's group, instead of appending it to the end. This makes sense, however it would ruin pre-existing data, which was an issue before. However this fix addresses the memory being overwritten. This may be due to more than one game server being on the same physical machine sharing memory with other game servers? Shot in the dark there. However all this does is reload data before updating data, so it's performance difference should be next to nil, being that the information is not changed all that often. Perhaps a handful of times for a character in a given session.

whasf:
So I've been following this thread all day since it was making my phone go all crazy..

I appreciate that you guys are trying to make DSP better, but I don't think this PR is the best way. There is something wrong in the core that is corrupting the blob. I agree with Kjlotus that it needs to be diagnosed and fixed, not worked around.

My suggestion is to create a PR with some debugging and we can go from there. Or I can do that, whatever.

nasomi:
I agree a workaround isn't an ideal long term solution, but there's no way for me to run my server without this code. It would destroy every players data. Because of that, anyone who tried to run their server in the fashion that I run mine would end up with corrupt data. That is why I made the PR.

bendangelo:
So it's not possible to do the bitwise save like I suggest. It does not work on blobs over 64bits.

bendangelo:
I think we should just create a table for each column. There is no way to group them without being confusing and the only benefit is a slight speed boost which might not even be necessary.

teschnei:
i got a neat POC for a bitset that we can use instead of std::bitset that is directly usable in the packet buffers, so most likely hold off on changing the data types core side til I can finish that off (so change the blobs but keep the c arrays)

also closing this since once the c arrays are converted, we will have bounds checking on each bitset, and we can figure out where to go from there

PR #2203, opened 2015-10-07
scripts/zones/abyssea_la_theine/npcs 

vrtra:
Heya, We have been trying to get abby up and running on our server. We are trying to make it as close to retail as possible. If these nm's are ok then will gladly release more. :)

-Wadski


nesstea:
funny commit names :)

Issue #2204, opened 2015-10-07
Client Crash - Quest Black Market 


RunAwayDSP:
https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Norg/npcs/Muzaffar.lua

Client will crash after completing Black Market Quest, but not every time maybe 25%-50% of the time.

TeoTwawki:
Incoming fix.

TeoTwawki:
#2207

PR #2207, opened 2015-10-07
Fix for issue #2204

TeoTwawki:
Explanation: the reason we null TextIDs.lua is because with every zone loading the same file name, you can wind up with null or invalid text ID from the previous zone being loaded instead the correct TextID from the zone a player is currently in.

### DIFF
--- a/scripts/zones/Norg/npcs/Muzaffar.lua
+++ b/scripts/zones/Norg/npcs/Muzaffar.lua
@@ -1,36 +1,37 @@
 -----------------------------------
 -- Area: Norg
--- NPC: Muzaffar
+--  NPC: Muzaffar
 -- Standard Info NPC
 -- Quests: Black Market
--- @zone 252
--- @pos 16.678, -2.044, -14.600
+-- @pos 16.678, -2.044, -14.600 252
 -----------------------------------
-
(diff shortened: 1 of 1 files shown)


scripts/zones/Norg/npcs/Muzaffar.lua
```diff
@@ -1,36 +1,37 @@
...
 -- Area: Norg
--- NPC: Muzaffar
+--  NPC: Muzaffar
 -- Standard Info NPC
 -- Quests: Black Market
--- @zone 252
--- @pos 16.678, -2.044, -14.600
+-- @pos 16.678, -2.044, -14.600 252
+-----------------------------------
+package.loaded["scripts/zones/Castle_Zvahl_Baileys/TextIDs"] = nil;
```
bendangelo: Isn't this supposed to be Norg?

TeoTwawki: woops, failboat, amending

TeoTwawki: I was distracted by stupid

TeoTwawki:
@bendangelo fixed my copy paste derp

bendangelo:
thanks

PR #2208, opened 2015-10-08
Scripts/SQL: Snake Eye force 11 chance update

Hozu:
Client says each merit past the first gives a 10% chance to hit 11 if roll is 5 or higher.


bendangelo:
The chance for this happening increases by 5% for each point in Snake Eye NOT including the first, up to a maximum of 20% (the in-game text stating 10% per upgrade is an error)

http://ffxiclopedia.wikia.com/wiki/Snake_Eye

Hozu:
Ugh.... you've got to be kidding me SE why.

PR #2213, opened 2015-10-08
Scripts: Animsub reset for respawning imps.

Hozu:
Animsubs don't reset automatically so it needs to be reset on spawn so each mob doesn't have permanently broken horns.

### DIFF
--- a/scripts/zones/Abyssea-Konschtat/mobs/Bakka.lua
+++ b/scripts/zones/Abyssea-Konschtat/mobs/Bakka.lua
@@ -18,6 +18,7 @@ end;
 -----------------------------------
 
 function onMobSpawn(mob)
+    mob:AnimationSub(0);
 end;
 
 -----------------------------------
--- a/scripts/zones/Abyssea-Konschtat/mobs/Dapifer_Imp.lua
+++ b/scripts/zones/Abyssea-Konschtat/mobs/Dapifer_Imp.lua
@@ -17,6 +17,7 @@ end;
 -----------------------------------
 
 function onMobSpawn(mob)
+    mob:AnimationSub(0);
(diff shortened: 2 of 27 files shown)


bendangelo:
Shouldn't the animation sub reset for every mob respawning?

Hozu:
Every mob should respawn with a specific subanim, generally 0, yeah. Some mobs don't ever use 0 though - Aerns start in 1.

I take it you're thinking of a better way to do this?

nesstea:
Good test would be that tinnin i just did. ill go check it.

Hozu:
Good point I don't think Tinnin would spawn in 0 - I'm assuming all three heads is 0, though I didn't look at the specific subanims.

nesstea:
All 3 is 0 yes.

nesstea:
Confirmed. Animation sub does not reset on spawn (for any mob it seems). DB for Tinnin is set at 2 and i spawned it a second time with 3 heads.

Hozu:
Yeah they don't reset by default - I made a recent pull that handled the CoP wyrms so they're fine, but if there's a DB value for default subanim and it's not working/being used, uh...

bendangelo:
I saw a line in ai mob dummy which is supposed to reset the animation sub but it was commented out for some reason. I'll look into it later.

teschnei:
It was commented out because it set the animation to some hardcoded number,
if we're thinking of the same thing (ezekyel code for sea mobs)

nesstea:
Anyway, this commit is legit for the imps. Just have to remember to reset the AnimationSub on spawn.

Hozu:
Fix in core makes this no longer required so closing.

PR #2215, opened 2015-10-08
[SCRIPT][SEA] Goldfish AI

nesstea:
They now function like they should! I am not sure if the numbers are correct, but they are no where near game breaking.


Hozu:
Shouldn't they have a % damage taken increase in open mouth form instead of -def?  I believe they take double damage from all sources.

teschnei:
Also, they only go in swim animation if they're in the water areas. Don't
think we can handle that yet though

nesstea:
"Hpemdes that have been damaged and start to attack will occasionally open their mouths for approximately 30 seconds; this doubles their attack output, and decreases their defense by half, much like Berserk."

Ok I have the Melee ATTP and DEFP modified instead of the actual DEF and ATT mods. I am not sure how to modify the magic side of things to consider a 50% damage increase. MDEF is set at 0. I assume it can be a negative, but that is just adding flat damage.

Any suggestions @teschnei? I see the mod MOD_DMGMAGIC, but is a positive 12 equivalent to 12% more damage taken or 12% less damage taken?

teschnei:
Do the opposite of what shell (the spell) does

scripts/zones/AlTaieu/mobs/Ul_hpemde.lua
```diff
@@ -24,6 +28,55 @@ end;
...
+    end
+end;
+
+-----------------------------------
+-- onMobFight
+-----------------------------------
+
+function onMobFight(mob, target)
+
+    if (mob:getHPP() < 100) then -- it will attack once it has been damaged.
```
xdemolish: not that i checked whether getHPP returns float or int, but its prolly best to check hp < max hp instead of hp% (cause 99.99% would still come out as 100% - probably)

PR #2216, opened 2015-10-08
ATMA effect script.

TeoTwawki:
It nothing but the effect script. I was never able to get the NPC to behave retail.

I have concluded that I have started work on a number of things that I will never actually finish, and that I should just share back what I can from those things so that others can make use of it going forward.

I know of several people trying to do abyssea stuff, and I hope this saves them some time.


teschnei:
Rise, monster script!

TeoTwawki:
Lol

teschnei:
Wonder if it'd be better to have those all in a global. The thing is that
this 800 line script has to be loaded and parsed each time the game server
needs a function from it (including on tick), while a global is just once
when required. Not saying you really need to spend the time changing it,
but just make a global, copy the existing thing in line for line and just
call it straight up from the effect script. "To-do:  turn this into a table"

TeoTwawki:
I was saying something along these lines when I talked to about it before and you were like naw it can't be that bad. Now that you see, you know I really meant it was F'ing huge! heh.

And it'll only get worse when someone adds the atmacite from voidwatch (same effect in client).

I'll amend this in a minute or too when I get off the phone.

teschnei:
Current me apologises on behalf of past me

TeoTwawki:
Like so (for now) ?

nesstea:
Looks nice Teosan

TeoTwawki:
@nesstea I kinda hate it. I am hoping someone gets of of all the if/elseif trash as they do abyssea craps. Legion dudes had it as a ton of separate effect scripts and omg poor status.lua stuffed full of things. And Omni had it calling modifiers that didn't exist crashing left and right. So I hurriedly squished it together and nuked bad mods.

xdemolish:
drop a comment once it's all done and table'd
script made me cry

TeoTwawki:
@teschnei -nudge- did I get what you wanted? 

Demo thinks ima redo it into that table you said I didn't have to right away. I don't wanna do it before all the required parts exist. Right now they don't..Besides missing modifiers, I never did finish that stuff for putting latents in status effects, and ATMA needs that badly (a perfect example of not remembering wtf I was doing from starting to many things..Disappointed :disappointed: with myself).

teschnei:
Sorry, out of town all weekend. Canadian thanksgiving, not home til tonight

TeoTwawki:
@teschnei  Oh. See you when you get back then. Have a good holiday!

PR #2217, opened 2015-10-08
Scripts/SQL: Qutrub weapon breaking + mobskills

Hozu:
-Implemented the ability to break Qutrub weapons. On a critical hit, there's a chance to break the first weapon. From that point onwards, it swaps between fists and second weapon once a minute. The second weapon can be broken when it's being wielded as well.
-Scripted Fortifying Wail, Hex Palm, and Unblest Jambiya. Added animsub checks to each of the mobskills as appropriate.
-Corrected the animations of Animating Wail and the formerly unimplemented mobskills, as well as the AoE type/range.

### DIFF
--- a/scripts/globals/mobskills/Animating_Wail.lua
+++ b/scripts/globals/mobskills/Animating_Wail.lua
@@ -1,6 +1,12 @@
 ---------------------------------------------------
--- Animating Wail
--- Increases attack speed.
+--  Animating Wail
+--  Family: Qutrub
+--  Description: Let's out a wail that applies Haste to itself and nearby allies.
+--  Type: Enhancing
+--  Can be dispelled: Yes
+--  Utsusemi/Blink absorb: N/A
+--  Range: Self
+--  Notes:
 ---------------------------------------------------
 
 require("scripts/globals/settings");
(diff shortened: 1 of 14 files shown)


Hozu:
Note that I didn't reset the subanims on respawn - @bendangelo said he'd take a look to see if there's a better way if it can be handled. I will add the subanim reset if it's determined that that is the best way.

xdemolish:
why the hell does travis refuse to build on your pull requests

Hozu:
Dunno. I just did a force push with no changes, let's see if it will build this time.

teschnei:
Can you not restart it yourself since it's your branch?

xdemolish:
ya it's gud
@teschnei shouldn't subanim be reset on spawning anyways (since localvars are)

teschnei:
Is there a db column or mob mod for it? Otherwise we don't know which
animation to change to. Shinryu spawns in 1 or 2, for example (0 is not
used)

Hozu:
mob_pools.sql has 
`animationsub` tinyint(1) NOT NULL DEFAULT '0',
I don't think it's used because Tinnin has that value set to 2 yet it always spawns with three heads. Well not anymore due to the script.

teschnei:
Easy fix probably, then

xdemolish:
https://github.com/DarkstarProject/darkstar/blob/master/src/map/utils/zoneutils.cpp#L475

Hozu:
Well that's a thing, too bad it's only in the initialization and not spawn...

Hozu:
Animsub is reset on spawn in the core now so that issue is fixed. There are no outstanding issues for this so it's ready for merging if nobody has any problems with it.

PR #2218, opened 2015-10-08
Moved player spells from a blob to a table.

bendangelo:
Added a new folder called migrations which will convert spell blobs into the new format.

It's a simple lua script, I'll write up how to run it later.

EDIT: I'm considering moving the migration scripts to python because it's a much easier installation for everyone. I only used lua because we were already using it.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -333,17 +333,16 @@ namespace charutils
             "nation,"               // 13
             "quests,"               // 14
             "keyitems,"             // 15
-            "spells,"               // 16
-            "abilities,"            // 17
-            "titles,"               // 18
-            "zones,"                // 19
-            "missions,"             // 20
-            "assault,"              // 21
-            "campaign,"             // 22
-            "playtime,"             // 23
-            "isnewplayer,"          // 24
-            "campaign_allegiance,"  // 25
-            "isstylelocked "        // 26
+            "abilities,"            // 16
+            "titles,"               // 17
+            "zones,"                // 18
+            "missions,"             // 19
+            "assault,"              // 20
+            "campaign,"             // 21
+            "playtime,"             // 22
+            "isnewplayer,"          // 23
+            "campaign_allegiance,"  // 24
+            "isstylelocked "        // 25
             "FROM chars "
             "WHERE charid = %u";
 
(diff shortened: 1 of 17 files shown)


teschnei:
we can't *really* use lua like that because the server doesn't require an interpreter (just the dev headers/library), so people probably wont have one (I do think python is a better idea, since everyone who has that will have the interpreter)

bendangelo:
Redid the scripts in python and added a readme. Need someone to test the script on a large dataset and we should be good to go

src/map/utils/charutils.cpp
```diff
@@ -766,6 +759,37 @@ namespace charutils
         luautils::OnGameIn(PChar, zoning == 1);
     }
 
+    void LoadSpells(CCharEntity* PChar)
+    {
+        // disable all spells
+        size_t length = sizeof(PChar->m_SpellList.data());
+        memset(PChar->m_SpellList.data(), 0, length);
```
teschnei: do we need this memset? if the data isn't initialized to 0s, then thats something the bitset should fix

bendangelo: I put this in because I thought charentities were not recreated on zoning.

teschnei: they are recreated on zoning

bendangelo: Actually I would prefer to keep this just so `LoadSpells` can be called any time and the spells are loaded on a "clean slate"

teschnei: okay, I will make a bitset method to do that then instead of memsetting it

bendangelo:
Tested the migration on real data and it works fine. So next steps is to merge this in or find a guinea pig.

src/map/utils/charutils.cpp
teschnei: PChar->m_SpellList.clear()

bendangelo:
Rebased into a single commit.

src/map/utils/charutils.cpp
```diff
@@ -766,6 +759,61 @@ namespace charutils
...
+        // disable all spells
+        PChar->m_SpellList.reset();
+        PChar->m_EnabledSpellList.reset();
+
+        // Compile a string of all enabled expansions
+        std::vector<const char*> expansions {"COP", "TOAU", "WOTG", "ACP", "AMK", "ASA", "ABYSSEA", "SOA"};
+        std::string enabledExpansions;
+
+        uint8 count = 0;
+        for (const char* expan : expansions)
```
teschnei: for (auto&& expan : {"COP", "TOAU", "WOTG", "ACP", "AMK", "ASA", "ABYSSEA", "SOA"})

src/map/utils/charutils.cpp
```diff
@@ -766,6 +759,61 @@ namespace charutils
...
+            Sql_NumRows(SqlHandle) != 0)
+        {
+            while(Sql_NextRow(SqlHandle) == SQL_SUCCESS)
+            {
+                uint16 spellId = Sql_GetUIntData(SqlHandle, 0);
+
+                if (spell::GetSpell(spellId) != nullptr)
+                {
+                    PChar->m_SpellList[spellId] = true;
+                    PChar->m_EnabledSpellList[spellId] = true;
```
teschnei: get rid of EnabledSpellList and just replace everywhere it was used with SpellList

bendangelo:
Done

PR #2227, opened 2015-10-09
Create scroll_of_blizzaja.lua

some1here:
Making scroll work


TeoTwawki:
Thank you :heart:

nesstea:
Can you do one commit for all the scrolls @some1here  instead of one commit for each of them?

some1here:
Sorry for the second batch but this was to patch(xx) originally.  Then Teo  wanted it closer...... less space

TeoTwawki:
@some1here its ok. If you get more you can combine when they are small things. I wasn't gonna ask you for anything more than what you already did.

PR #2231, opened 2015-10-10
Scripts/SQL: Corrected valid targets for Auspice/Esuna/Sacrifice

Hozu:
Disabled the ability to cast Auspice on others via command - it's a self only spell that's AoE like Esuna.


teschnei:
why would you do that in the script?

Hozu:
It's already done in the database but that only affects what you can target via menu cast or <stpc> or <stal> or whatever. Doing /ma Auspice name would still work on others.

teschnei:
i think you are mixing "the client" with "the database"

Hozu:
...You are correct, it was wrong in the DB. Changed it there instead.

PR #2232, opened 2015-10-10
[SCRIPT][SEA] Jailer of Prudence

nesstea:
There is a binding that is not coded yet. Kj said the binding will be there in the future, so I left it commented.
Tested and is working as intended otherwise.

### DIFF
--- a/scripts/zones/AlTaieu/mobs/Jailer_of_Prudence.lua
+++ b/scripts/zones/AlTaieu/mobs/Jailer_of_Prudence.lua
@@ -1,9 +1,13 @@
 -----------------------------------
 -- Area: Al'Taieu
---  NM:  Jailer of Prudence
+-- NM:  Jailer of Prudence
+-- IDs: 16912846, 16912847
+-- AnimationSubs: 0 - Normal, 3 - Mouth Open
+-- Wiki: http://ffxiclopedia.wikia.com/wiki/Jailer_of_Prudence
 -----------------------------------
 
 require("scripts/globals/status");
+require("scripts/zones/AlTaieu/mobIDs");
 
 -----------------------------------
 -- onMobInitialize Action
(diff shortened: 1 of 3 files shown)


scripts/zones/AlTaieu/mobs/Jailer_of_Prudence.lua
```diff
@@ -19,6 +24,17 @@ end;
 -----------------------------------
 
 function onMobSpawn(mob)
+    mob:AnimationSub(0); -- Mouth closed
+    mob:addStatusEffect(EFFECT_FLEE,100,0,30);
```
Hozu: Wasn't there a way to make it so when this status wears there wouldn't be a message?

nesstea: Wiki says it gains "Flee"

xdemolish: >Wasn't there a way to make it so when this status wears there wouldn't be a message?
addStatusEffectEx

Hozu: That's the one. It gains the effect, but I believe it doesn't give a message about it wearing.

Deadwing888:
His initial flee is actually longer than 30 seconds. I believe 60.

He also gets a brief flee when he uses perfect dodge.

scripts/zones/AlTaieu/mobs/Jailer_of_Prudence.lua
nesstea: You guys know how to do that? I do not.

Hozu: Use addStatusEffectEx instead of addStatusEffect. Same parameters, just a different function name.

xdemolish: https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_baseentity.cpp#L4403

just takes icon as a param after id (instead of power being the next param), the rest are the same as addStatusEffect

PR #2240, opened 2015-10-11
Scripts: Frenetic Rip mobsub check

Hozu:
Only is available when horn is broken, not the other way around...


Hozu:
0 being horn is there, 1 being horn broken. Want it to be valid only in 1, not 0. Right now imps have no valid moves in 1.

nesstea:
My bad :o

Hozu:
Modified comment in the script for clarity.

PR #2244, opened 2015-10-11
Scripts: Make Astral Flow BPs consume all MP.

Hozu:
Because they're not consuming any right now.


TeoTwawki:
LOL. Thanks for catching that!

PR #2248, opened 2015-10-12
Update Nomad_Moogle.lua

some1here:
some updates


AlexandreCA:
simply copy and paste one of my topics on the forum darkstar . Really.. I confirm the switch to lvl 90 but it would be nice to continue thereafter.

PR #2252, opened 2015-10-12
Update Nomad_Moogle.lua

some1here:
update


bendangelo:
There should probably be a check for `MAX_LEVEL` that disables this quest unless cap is higher than 75.

xdemolish:
https://github.com/DarkstarProject/darkstar/blob/master/README.md#pull-requests

PR #2254, opened 2015-10-12
Update missions.lua

some1here:
Added RoV


TeoTwawki:
In the future please avoid random unintentional line ending changes. (if you intend to do it, say so - you need to start having *real* commit messages! "update x" is not descriptive.)

bendangelo:
Thanks for contributing, but please have more descriptive commits.

PR #2260, opened 2015-10-13
Correct all misuses of utils.clamp throughout trunk. This fixes a bug with Phalanx.

TeoTwawki:


TeoTwawki:
I blame copy pasta!

TeoTwawki:
I was absorbing mobs ice spikes when I had ~~failanx~~ phalanx up.

TeoTwawki:
Side note: seem to be a lot of clamping happening before more math is done to things, not sure that's right but didn't touch it. Hopefully its not possible for those effect durations to go negative from resistance.. >.>

bendangelo:
Duration is clamped to 1 for most status effects

PR #2261, opened 2015-10-13
Scripts: Correct weather check for Qu'Hau spring

Hozu:
After taking a look in zone_weather.sql, I noticed that Ro'Maeve can't have clear weather, at all. I'm assuming the data is correct, particularly because the image on the wiki shows the filled spring and the reflection of the sky is partly cloudy, which is what happens during the sunshine weather.

Also I did bad math in my previous fix to this spring and gave it an extra game hour, so I'm bring it back down.


Hozu:
The picture, for reference: http://vignette1.wikia.nocookie.net/ffxi/images/e/ee/QuHauSpring.png/revision/latest?cb=20081126092915

PR #2271, opened 2015-10-16
Update settings.lua to add ENABLE_ROV

some1here:
added ENABLE_ROV


xdemolish:
commit messages as in: actual commit title, not just pull request title

PR #2276, opened 2015-10-16
Core/Scripts: Dread Spikes max drain cap

Hozu:
Also fixed underflow bug with Reprisal subpower so it can actually wear off before the duration expires.

### DIFF
--- a/scripts/globals/spells/dread_spikes.lua
+++ b/scripts/globals/spells/dread_spikes.lua
@@ -16,11 +16,13 @@ function onSpellCast(caster,target,spell)
     local duration = 180;
     local power = 0;
     local typeEffect = EFFECT_DREAD_SPIKES;
+    local drainAmount = target:getMaxHP() / 2;
 
     if (caster:hasStatusEffect(EFFECT_COMPOSURE) == true and caster:getID() == target:getID()) then
-      duration = duration * 3;
+        duration = duration * 3;
     end
-    if (target:addStatusEffect(typeEffect,power,0,duration)) then
+
+    if (target:addStatusEffect(typeEffect, power, 0, duration, 0, drainAmount, 1)) then
         spell:setMsg(230);
     else
         spell:setMsg(75);
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -730,6 +730,16 @@ namespace battleutils
 
                         if (PDefender->isAlive())
                         {
+                            // Subpower is the remaining damage that can be drained. When it reaches 0 the effect ends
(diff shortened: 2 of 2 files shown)


scripts/globals/spells/dread_spikes.lua
```diff
@@ -16,12 +16,18 @@ function onSpellCast(caster,target,spell)
     local duration = 180;
     local power = 0;
     local typeEffect = EFFECT_DREAD_SPIKES;
+    local drainAmount = target:getHP() / 2;
```
TeoTwawki: "Dread Spikes effect lasts one minute, or until the caster has drained up to half of their *maximum* HP. "

From wiki, emphasis mine. You are using current HP not max, at time of cast.

Hozu: Whoops, fixed. I guess I spaced out when I read half of current max HP, since max HP usually doesn't change... though DRK does have Drain 2.

src/map/utils/battleutils.cpp
```diff
@@ -730,6 +730,16 @@ namespace battleutils
 
                         if (PDefender->isAlive())
                         {
+                            // Subpower is the remaining damage that can be drained. When it reaches 0 the effect ends
+                            int remainingDrain = PDefender->StatusEffectContainer->GetStatusEffect(EFFECT_DREAD_SPIKES)->GetSubPower();
```
bendangelo: Would you might storing the status effect in a local variable. That'll make this easier to read. It can then be used in both places.
`CStatusEffect* PStatus = ..`

Hozu: Wait is there a way to make it grab whatever spikes effect you have? Otherwise I'd have to grab the effect in each case individually which... doesn't make it usable in both places and would be confusing.

TeoTwawki: The request already confuses me. We'd either be calling just dread spikes PStatus when none of the others are written that way, or we'd need to redefine PStatus for each one. This is more confusing, not less.

bendangelo:
Well it looks fine to me. Thanks

src/map/utils/battleutils.cpp
teschnei: GetStatusEffect can return nullptr, check it

there is no guarantee that the effect is the only thing that set action.spikesParam to SPIKE_DREAD or SPIKE_REPRISAL

auto PEffect = PDefender->StatusEffectContainer->GetStatusEffect(EFFECT_DREAD_SPIKES)
if (PEffect) {

Hozu: I'm not sure what you're getting at with spikes being set by other things because wouldn't a mob having auto-spikes (if there's one with auto Dread Spikes, pretty sure there's no mob on retail with Reprisal in any form...) not have a limited amount of draining capacity?

So I should change something with this?

PR #2278, opened 2015-10-16
Core: Hasso/Seigan (re)cast penalty

### LINKED ISSUE #1833: Hasso/Seigan Not doubling recast timers.
Kind of keep forgetting to post this but the issue is that SAM/NIN is just completely broken at the moment in terms of survability, not to mention that Third Eye and Utsusemi Stack, with utsusemi taking priority no less. Assuming this is just an oversight but it's a relevant issue that effects game balance.

Hozu:
Implemented the 1.5x (re)cast time for any spell while under the effect of either ability. Removed the negative fastcast mod on the two effects because that was improper. Fixes #1833, though the additional part of Utsusemi and Third Eye being mutually exclusive remains broken.

### DIFF
--- a/scripts/globals/effects/hasso.lua
+++ b/scripts/globals/effects/hasso.lua
@@ -14,7 +14,6 @@ function onEffectGain(target,effect)
 	target:addMod(MOD_STR,effect:getPower());
 	target:addMod(MOD_HASTE_ABILITY,102);
 	target:addMod(MOD_ACC,10);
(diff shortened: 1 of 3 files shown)


Hozu:
Yes I realize my additions are inconsistent with each other, but they're consistent with the rest of the things in their respective function.

bendangelo:
Looks good thanks a lot for contributing.

PR #2292, opened 2015-10-19
SQL: Correct Dynamis Bastok mega boss mobskill list

Hozu:
Totally shouldn't have the quadav NM moves.


bendangelo:
Actually, mob skill list 705 is correct. The skills in the list are just wrong, the skill list should contain all Adamantking Effigy skills plus wrath of gu`dha.

Hozu:
I haven't come across any reference as to the mega boss having that move. Nothing about it on dynamisbums from what I was told.

bendangelo:
http://ffxiclopedia.wikia.com/wiki/Dynamis_-_Bastok. Actually it's all Quadav NMs have this skill. So the skill list was set on the wrong mobs.

Hozu:
So the statue shouldn't have the move, yes? Just the statue moves?

bendangelo:
ok it's correct then

PR #2294, opened 2015-10-20
[SCRIPT] Medusa TP moves and 2Hour

nesstea:
This is all I am going to work on for Medusa for now. This allows her to use TP moves and 2hour like the wiki says.

She still needs work with the ads and the respawning AI, etc.


nesstea:
Remaking branch because I hate rebasing.

PR #2295, opened 2015-10-20
[SCRIPT] Medusa progress

nesstea:
Scripted 2hour
Modified Gorgon Dance
Modified mobskills
Still needs work, but this is a step in the right direction. I won't be working on her for a while.


Hozu:
I still think you should have left the PW file out of this. Your change is allowing it to use the move regardless of the model. <_<

nesstea:
PW changing forms = new mobskill lists. Don't matter.

scripts/zones/Arrapago_Reef/mobs/Medusa.lua
```diff
@@ -1,28 +1,45 @@
...
-	SpawnMob(16998866,180):updateEnmity(target);
+-----------------------------------
+-- onMobFight Action
+-----------------------------------
 
+function onMobFight(mob, target)
+    local HPP = mob:getHPP();
+    if (mob:getLocalVar("usedees") == 0) then
+        if (HPP <= mob:getLocalVar("eeshpp")) then
+            mob:useMobAbility(479); -- Eagle Eye Shot
```
bendangelo: 479 is eagle eye shot for goblins. This is a bad way of doing two hours. It's better to use the mobmod `MOBMOD_SCRIPTED_2HOUR`

TeoTwawki: I don't see whats bad other than he called a a specific ID that doesn't work for the mob he's using. For most other things like mighty strikes that method works quite well, and had he used a version that wasn't gob specific (I'm assuming the animation is wrong with that one he used) it'd be fine as well. He's setting the level that triggers the 2hr within a random range instead of a static 60 (random is closer behavior to retail, btw) and I don't see why we needed a mobmod for everything under the sun anyway while making the core handle it - I'd rather we just did all the 2hr via script than have 5 ways of doing a 2hr, since I can make whatever I need to emulate retail behavior happen via script.

scripts/globals/mobskills/PW_Gorgon_Dance.lua
```diff
@@ -1,31 +1,32 @@
 ---------------------------------------------
--- Gorgon Dance
+--  Gorgon Dance
 --
--- Description: Petrifies all targets in an area of effect.
--- Type: Enfeebling
--- Utsusemi/Blink absorb: Ignores shadows
--- Range: 10' radial?
--- Notes: Used only by Medusa. Starts using it at 25%.
+--  Description: Petrifies all targets in an area of effect.
```
TeoTwawki: @nesstea  Why'd  you  bother  with  changing  the  spacing ? 
Single space was correct formatting anyway, not that anybody cares whats in the headers or we wouldn't have 40 diff versions among the skill scripts.

scripts/zones/Arrapago_Reef/MobIDs.lua
```diff
@@ -0,0 +1,6 @@
+-- Medusa
+Medusa    = 16998862;
+ExonOne   = 16998863;
```
TeoTwawki: @nesstea No need for this. Can use relative IDs since they are all next to each other.
Instead of:
```lua
function onMobEngaged(mob, target)
    target:showText(mob, MEDUSA_ENGAGE);
    SpawnMob(ExonOne,   180):updateEnmity(target);
    SpawnMob(ExonTwo,   180):updateEnmity(target);
    SpawnMob(ExonThree, 180):updateEnmity(target);
    SpawnMob(ExonFour,  180):updateEnmity(target);
end;
```

you'd use:
```lua
function onMobEngaged(mob, target)
    target:showText(mob, MEDUSA_ENGAGE);
    SpawnMob(mob:getID()+1, 180):updateEnmity(target);
    SpawnMob(mob:getID()+2, 180):updateEnmity(target);
    SpawnMob(mob:getID()+3, 180):updateEnmity(target);
    SpawnMob(mob:getID()+4, 180):updateEnmity(target);
end;
```

Groupings like this are stable through client ID shifts: if medusa gets moved by one her helpers get moved by the same amount and stay the same relative to each other, every single time.

scripts/zones/Arrapago_Reef/mobs/Medusa.lua
bendangelo: Sure a new lua binding like `useMobTwoHour` is the best solution. Hard coding values like this is never a good solution.

TeoTwawki: -*blinks*-

You wanna make a binding and think the existing one, for a value that never really changes, instead of the existing binding that works correctly? All that needs done is call that with the correct ID, or variablize the 2hr names in a global if you hate seeing raw IDs for them. 

(Even EES is not an example of changing - there are literally multiple different ones that just happen to have the same name and run the same script)

TeoTwawki: -*blinks*-

You wanna make a binding and think the existing one, for a value that never really changes, instead of the existing binding that works correctly, is hard coding? All that needs done is call that with the correct ID, or variablize the 2hr names in a global if you hate seeing raw IDs for them. 

(Even EES is not an example of changing - there are literally multiple different ones that just happen to have the same name and run the same script)

TeoTwawki: I hope I don't sound hostile, I'm simply having a hard time seeing your pov there.

bendangelo: The value does change, for Maat they are totally different values, it could be different for other mobs too. The AI also needs to remember if it already used it's two hour, so it won't use it again. This `useMobAbility` doesn't do that.

All of this is already handled in the core, and as a long term solution (what if we discover, some animations were wrong? Now we have to fix multiple scripts instead of fixing one method) I feel adding a new method to initiate a mob two hour is just better. It's a no-brainer method call, you want the mob to two hour now? Ok call this method, done.

TeoTwawki: No, no they do NOT change. The skill ID never, ever gets changes - the different ID is a totally different skill. The localVar is handling tracking, and gets wiped out when no longer needed on its own. It being handle din core is precisely what I have an issue with. You'll wind up with a  bunch of redundant hard coded checks in core to see which version of Eagle Eye Shot to use, instead of just calling the correct one form that specific mobs script. How is that better?

teschnei: pros and cons.  On one hand, you can have it automatically deal with "don't use this again", and automatically find the right 2hour (maybe)

on the other hand, it makes things like mobs with 3 or more different 2hours a little more annoying to do (ark angels have 4 each..).  Where would you draw the line, though? use a special binding for ranged attacks? another one for natural meditation when being hit by a weaponskill? etc etc... the mobskill IDs probably aren't going to change ever, so it being a hardcoded number isn't really much of a problem

TeoTwawki: If it makes everyone happy I can whip up a lua function to auto handle each 2hr including not using it twice, and the multiple version can be handled as (maybe with better naming, but I'm psuedoCoding here):
`oneTimeSpecial(EagleEyeShot)`
`oneTimeSpecial(EagleEyeShotGoblin)`
`oneTimeSpecial(EagleEyeShotMaat)`
etc.

..But I don't personally see the need to mess with it.

bendangelo: I'm not saying mob skill id's could change from the client side. I'm saying we don't yet have all the two hour mob skills, for example there is an EES animation for each family, so we'll have to add a new mob skill for each. I just don't like the idea of hard coding the skill id when we don't have to.

We can push the two hour logic to scripts for sure, but we need better control for two hours, something to of course remember if it was already used (no need to manage yourself). We'll need something like: `mob:useTwoHour(JOB_THF)`, (Use two hour regardless if already used) `mob:hasUsedTwoHour(JOB_THF)`, `mob:resetTwoHour(JOB_THF)`.  (with no args it could default to current main job). 

We don't need new bindings for this (but I would still be for adding them to the mob, it's just cleaner). Could add a new `mob` or `mobutils` namespace for these types of methods. 

```
if(mobutils.hasUsedTwoHour(mob, JOB_THF) == false) then
mobutils.useTwoHour(mob, JOB_THF)
end
``` etc..

File: scripts/zones/Arrapago_Reef/MobIDs.lua (hunk 3 above)
Comment (nesstea) [inline scripts/zones/Arrapago_Reef/MobIDs.lua]: I know I can do that. I wanted to do it this way to make it easier for anyone else coding it that may not understand that.

Comment (teschnei) [inline scripts/zones/Arrapago_Reef/MobIDs.lua]: Do it the way Teo said. Readability is good, but not if you have to
sacrifice maintainability for it (especially since id shifts happen so
often)

File: scripts/zones/Arrapago_Reef/mobs/Medusa.lua (hunk 1 above)
Comment (nesstea) [inline scripts/zones/Arrapago_Reef/mobs/Medusa.lua]: @bendangelo  I could have done that scripted 2hour method, but I was not confident at all that it would use it. Plus, there is little/no documentation regarding this in the function header.  You asked me to submit the code I have. This method works just fine. The animation for EES works. I agree with @TeoTwawki with this being more accurate to retail 2hours, except that a mob should use their 2hour way before they hit 5% life. :)

File: scripts/globals/mobskills/PW_Gorgon_Dance.lua (hunk 2 above)
Comment (nesstea) [inline scripts/globals/mobskills/PW_Gorgon_Dance.lua]: Uh I usually do single space. Confused why there is two spaces. O.o

File: scripts/zones/Arrapago_Reef/mobs/Medusa.lua (hunk 1 above)
Comment (TeoTwawki) [inline scripts/zones/Arrapago_Reef/mobs/Medusa.lua]: @bendangelo 

> if(mobutils.hasUsedTwoHour(mob, JOB_THF) == false) then
mobutils.useTwoHour(mob, JOB_THF)
end

You just repeated my suggestion and called it cleaner. 

Except I think it saner to say the 2hr name, not the job abbreviation - I do **not** want a bunch of conditionals figuring out which EES to call, lets just call the correct one in the first place its not that hard for scripters to test and see if it animates correct. You are saying the using the input is hard coding, which is precisely the opposite..By setting the ID base don a non input criteria you are hard coding the result. Doing so is not cleaner, its using hard coding to solve what you see as hard coding. 

For the record, I detest whats in core - why would you tie what 2hr they can use to their job at all? Many mobs use a 2hr that doesn't match their job with the most notable being AV and the ark angels. Its the worst possible way to handle them! 

>what if we discover, some animations were wrong? Now we have to fix multiple scripts instead of fixing one method

The animations aren't even part of the script. What we would do is add the alternates to the table with a different animation ID, and call that ID instead. Not exactly an earth shattering bug there.


Btw we should probably avoid saying TwoHour or TwoHr etc btw, since they aren't even called that in retail anymore. Not on 2 hour timers any more (and in DSP mobs generally get them once per spawn which is very likely retail behavior) so calling it that may cause some noob confusion in some distant future despite our familiarity with the term right now.

@nesstea I would hazard a guess that what retail does for these is the chance to "go off" increases as the HP decreases. rather than just being a random number. This weights it toward lower HP values while still letting it happens at a wide range of values. However, me trying to prove that would make my head asplode.

Comment (bendangelo) [inline scripts/zones/Arrapago_Reef/mobs/Medusa.lua]: Well I disagree using an id is not better. This will force every scripter to look up the correct mobskill and has the potential to use the wrong one, in fact in this case `479` is the wrong mobskill.

All this two hour stuff won't get solved in this PR, so let's figure it out later.

@nesstea I'm about 100% certain `479` is not the correct mob skill. Run `@injectaction 11 19` with Medusa selected, it will most likely run the wrong animation.

Comment (teschnei) [inline scripts/zones/Arrapago_Reef/mobs/Medusa.lua]: it's probably 1675 or 1676.  Animation ID is prooobably in the 1300s somewhere (with the rest of her animations)

**Comment by bendangelo** [comment]
Feels like this has been abandoned, so I'll do you a favor and find the EES animation for you when I get home.

**Comment by nesstea** [comment]
Just been busy. I'll do it today.

**Comment by nesstea** [comment]
The animation is 1338 for her 2hour. Am I creating a new mobskill or am i modifying a different one?

**Comment by bendangelo** [comment]
I'll take a look at it later. When you find her animation you'll have to create a new mob skill for it.

**Comment by nesstea** [comment]
Could you show me how to do that when you do? I can learn something this way!

**Comment by nesstea** [comment]
Not to rush, but I'd like to edit some more mob_skill_lists. This has been here for a long time.

File: sql/mob_skills.sql [hunk 4]
```diff
@@ -1597,9 +1597,9 @@ INSERT INTO `mob_skills` VALUES (1548,1353,'Haymaker',4,12,2000,0,4,0,0,3);
 INSERT INTO `mob_skills` VALUES (1549,1354,'Incessant_Fists',0,7,2000,0,4,0,0,3);
 INSERT INTO `mob_skills` VALUES (1550,1355,'Arcane_Stomp',1,25,2000,0,1,0,0,3); -- corected ? to 1
 INSERT INTO `mob_skills` VALUES (1551,1356,'Pleiades_Ray',1,10,2000,0,4,0,0,3);
--- INSERT INTO `mob_skills` VALUES (1552,1552,'Petrifaction',0,7.0,2000,1500,4,0,0,0);
+INSERT INTO `mob_skills` VALUES (1552,1552,'Petrifaction',0,7.0,2000,1500,4,0,0,0);
 INSERT INTO `mob_skills` VALUES (1553,1335,'PW_Shadow_Thrust',0,10,2000,2000,4,0,0,0);
--- INSERT INTO `mob_skills` VALUES (1554,1554,'Tail_Slap',0,7.0,2000,1500,4,0,0,0);
+INSERT INTO `mob_skills` VALUES (1554,1554,'Tail_Slap',0,7.0,2000,1500,4,0,0,0);
```
bendangelo: Just tested, this is not the right animation id. Same for petrifaction

nesstea: Medusa does some goofy animations with that new EES btw.

nesstea:
I made the changes with Teo's commit expertise :)

scripts/globals/mobskills/PW_Gorgon_Dance.lua
```diff
@@ -7,25 +7,26 @@
...
+
 ---------------------------------------------
 
 function onMobSkillCheck(target,mob,skill)
-    local mobSkin = mob:getModelId();
-
-    if (mobSkin == 1865) then
+    
+    local mobhp = mob:getHPP();
+    if (mobhp <= 25) then -- She's under 25%, it's okay to use this.
```
Hozu: stahp fucking with the PW check pls

TeoTwawki: I think that script doesn't get used anymore anyway since the mobskill refactor?

nesstea: correct

Hozu: If that's what will happen then there would be no need for a check in the onMobskill function.

nesstea: Which is why I removed it :o

scripts/zones/Arrapago_Reef/mobs/Medusa.lua
```diff
@@ -1,28 +1,46 @@
...
-	SpawnMob(16998866,180):updateEnmity(target);
+-----------------------------------
+-- onMobFight Action
+-----------------------------------
 
+function onMobFight(mob, target)
+    local HPP = mob:getHPP();
+    if (mob:getLocalVar("usedees") == 0) then
+        if (HPP <= mob:getLocalVar("eeshpp")) then
+            mob:useMobAbility(1296); -- Eagle Eye Shot
```
bendangelo: This should be 1675. 1296 is just the animation.

sql/mob_skills.sql
bendangelo: The animation for tail slap and petrifaction is still wrong.

scripts/globals/mobskills/PW_Gorgon_Dance.lua
Hozu: I mean there shouldn't even be a HP check either.

scripts/zones/Arrapago_Reef/mobs/Medusa.lua
nesstea: I meant to do that. Changing now.

scripts/globals/mobskills/PW_Gorgon_Dance.lua
Hozu: Like there's no guarantee that it be in Medusa form below 25%, might be above that and then change to something else before it reaches 25%.

nesstea:
Updated ben.

bendangelo:
Ok I like it. Thanks.

PR #2296, opened 2015-10-20
Scripts/SQL: Sky god enspell/element resist fixes

Hozu:
-Changed enspell damage to be independent of stats (fixes zombie glitch as well).
-Changed the family elemental resistance to be more in line with what it's supposed to be

Note: Missing Kirin's enspell damage.


bendangelo:
It's better than what was there before but some form of resist is needed.

Hozu:
Hm? I thought resist was handled by it already?

bendangelo:
Yeah it's there. I didn't expand it to see the lines below.

PR #2305, opened 2015-10-21
Item Mods to Claidheamh Soluis

some1here:
Claidheamh Soluis added 
Sword skill +242 Parrying skill +242 Magic Accuracy skill +18
http://www.ffxiah.com/item/20718/claidheamh-soluis


teschnei:
Item level mods (skill, macc) on item level weapons (99+) don't go in item_mods. they go in item_weapon under ilvl_xxx.

some1here:
The mods are not working(and no show) threw item_weapon and this makes it the same as retail so you can see the differences in stats

teschnei:
You don't play retail.  I play retail.  You don't see the difference in
stats on retail.

You may use /checkparam <me> to verify that they are working properly (not
the skills page)

PR #2310, opened 2015-10-23
Added Jar.lua to scripts\zones\Davoi\npcs fixes #2264

### LINKED ISSUE #2264: Missing lua for the jar in Davoi. Involved in Test My Mettle quest.
The quest itself is no great shakes, but people do like to steal the boots for fire resist set.

BRFord88:
Script allows players to obtain Power Sandals from the Jar in Davoi.


bendangelo:
Do you need the quest active to get the power sandals? I think you might

BRFord88:
I don't think so because the Jar is supposed to be a timed pop (somewhere around 21-24 hours I believe).  The quest is just to turn the boots in.

PR #2316, opened 2015-10-23
Core/Scripts: Rework forced spawn NPC disappearance

Hozu:
I created a new lua binding that will set a hidden NPC to unhide after the specified period of time - the existing function only works on non-hidden NPCs. I've also created a variable in settings.lua that holds a default time for the force spawning NPC hiding time (defaults to five minutes).

Upon force spawning the mob, the NPC will be hidden indefinitely, until the despawn (NOT death action) of the mob it spawned - at that time, the hiding time is updated to the time specified with the new setting.

Other misc. fixes that I've done while going through the files:
-Fixed the NPCIDs various ???s
-Made the ??? for Jailer of Fortitude not move around as it's untouched - it only moves after its death.
-Unswapped the comment labels in the zone.lua file of The Garden of Ru'Hmet for the ??? of Ix'DRK and Jailer of Faith;
-Used stored ID for the ??? for Jailer of Faith because it's there and was being set twice in the same function for some reason. Same with a couple mobs.
-Changed the spawn point of Jailer of Faith/Fortitude to be at the ???'s pos rather than the player's pos.


bendangelo:
Looks good, thanks

PR #2328, opened 2015-10-25
Core/SQL: Damage type for specific weapons + change check from dmgtype to skilltype

Hozu:
Added the special damage type properties that were missing for the specific weapons on this page: http://ffxiclopedia.wikia.com/wiki/Damage_Types Also gave Lex Talonis and Griffinclaw slashing damage instead of none (yes, the former weapon is slashing despite being a rapier according to http://ffxiclopedia.wikia.com/wiki/Lex_Talionis). Iridial/Chatoyant staff got blunt damage as well.

In the core the two things I changed were checking for h2h damage type when it would make more sense to check for weapon being of h2h skill. Making that change allows for h2h weapons to be of different damage types while being able to continue working for those particular checks.


Hozu:
Changed the few checks that was looking for h2h damage type to h2h skill so h2h weapons of different damage types can continue working.

PR #2330, opened 2015-10-25
[SCRIPT]: Add/Update commands

santssoft:
* addeffect: add subid and subpower to command
 * animatenpc: add name of npc after id
 * animatesubnpc: change animatesub of a npc
 * deleffect: delete effect of a pc


bendangelo:
thanks

PR #2332, opened 2015-10-25
Scripts: Castle Oztroja top door opening fix

Hozu:
The door opens after lighting only *one* torch, not all four. The old script already lights all four torches after lighting one anyway.


bendangelo:
thanks

PR #2336, opened 2015-10-27
Fix for alliance exp.

jwonesh:
Iterate over all available parties.


teschnei:
4,706 additions, 4,693 deletions not shown

probably you changed line endings, so I will ask you to commit unix line endings and then amend your pull request so that I can read the diff on github instead of having to check it out
(plus the repository should have unix line endings anyways)

PR #2337, opened 2015-10-27
Fix for alliance xp.

jwonesh:
Iterate over all available parties. Moved previous loop to inner loop (that is why you see a bunch of indents).

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -3081,7 +3081,7 @@ namespace charutils
 				}
 				for (uint8 j = 0; j < numAllianceParties; j++)
 				{
-					for (uint8 i = 0; i < PChar->PParty->members.size(); ++i)
+					for (uint8 i = 0; i < PParty->members.size(); ++i)
 					{
 						exp = 0;
 						monsterbonus = 1.0f;
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -3081,12 +3081,13 @@ namespace charutils
 				}
 				for (uint8 j = 0; j < numAllianceParties; j++)
 				{
-					for (uint8 i = 0; i < PParty->members.size(); ++i)
+					CParty* innerPParty = (CParty*)PChar->PParty->m_PAlliance->partyList[j];
+					for (uint8 i = 0; i < innerPParty->members.size(); ++i)
 					{
 						exp = 0;
 						monsterbonus = 1.0f;
 						chainactive = false;
-						CCharEntity* PMember = (CCharEntity*)PChar->PParty->m_PAlliance->partyList[j]->members[i];
+						CCharEntity* PMember = (CCharEntity*)innerPParty->members[i];
 						if (PMember->getZone() == PMob->getZone())
 						{
 							if (map_config.exp_party_gap_penalties == 1)
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -3075,233 +3075,240 @@ namespace charutils
             baseexp = GetRealExp(maxlevel, PMob->GetMLevel());
             if (baseexp != 0)
             {
-                for (uint8 i = 0; i < PChar->PParty->members.size(); ++i)
-                {
-                    exp = 0;
-                    monsterbonus = 1.0f;
-                    chainactive = false;
-                    CCharEntity* PMember = (CCharEntity*)PChar->PParty->members[i];
-                    if (PMember->getZone() == PMob->getZone())
-                    {
-                        if (map_config.exp_party_gap_penalties == 1)
-                        {
-                            if (maxlevel > 50 || maxlevel > (PMember->GetMLevel() + 7))
-                            {
-                                exp = (float)baseexp*(float)((float)(PMember->GetMLevel()) / (float)(maxlevel));
-                            }
-                            else
-                            {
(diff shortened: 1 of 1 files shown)


jwonesh:
This is not ready, I see something I missed. Will fix tomorrow as it is late already.

teschnei:
tabs -> spaces

jwonesh:
Fixed. Github doesn't seem to be smart enough to ignore total whitespace changes (even if they are just spaces), so you can add ?w=1 to PR link to ignore whitespaces when diffing.

jwonesh:
Also, I just had a thought that I need to make sure that MLevel is being calculated correctly. If we are in alliance, MLevel should be the highest person's level, yeah? I need to double check that is what it is doing when I get home form work. If it is only grabbing party's highest level, then we can have 2 alliances, one full of 75s and the other of low levels, where the 75 party can steamroll mobs and the low level party would get xp relative to their level.

teschnei:
Yeah, but I don't want tabs in the repository :)
also, when using w=1 I can't comment on specific lines

I am pretty sure the main level is okay.  Whenever someone acts on a mob, it sets the level to their level if it is lower than their level.  So it should be alliance agnostic

jwonesh:
Ok that is what I was thinking too.

I did realize that I will need to add a check if there is an alliance just before setting the PMember variable, as it will be null if I try to do so when no alliance exists. Didn't catch that when testing as I only made an alliance and saw that it worked. I will fix that when I get home from work.

src/map/utils/charutils.cpp
```diff
@@ -3075,233 +3075,241 @@ namespace charutils
...
+			                {
+				                if (PMember->StatusEffectContainer->HasStatusEffect(EFFECT_SIGNET) && PMob->m_Element > 0 && dsprand::GetRandomNumber(100) < 20 &&
+					                PMember->loc.zone == PMob->loc.zone) // Need to move to SIGNET_CHANCE constant
+				                {
+					                PMember->PTreasurePool->AddItem(4095 + PMob->m_Element, PMob);
+				                }
+			                }
+			                if (PChar->PParty != nullptr && PChar->PParty->m_PAlliance != nullptr && PMob->m_Type == MOBTYPE_NORMAL)
+			                {
+				                if ((Pzone > 38 && Pzone < 43) || (Pzone > 133 && Pzone < 136) || (Pzone > 184 && Pzone < 189)) AddExperiencePoints(false, PMember, PMob, exp, 1, false);
```
xdemolish: Pzone->getType() != ZONETYPE_DYNAMIS

teschnei:
seems your tabs -> spaces missed a bunch of tabs still

jwonesh:
Hmm I set visual studio to use 4 spaces instead of tab and redid all of the tabs. Let me take another look.

teschnei:
CTRL+K, CTRL+F is auto-format selection (which should be good)
or CTRL+K, CTRL+D is auto-format entire document (which will be good if the document has been auto-formatted before, no guarantees on that one)

jwonesh:
I reformatted per your suggestion @teschnei

teschnei:
Great! I will get to the actual review tonight then

jwonesh:
I would still recommend (after verifying that I have not introduced tabs into the repo) using the ignore-whitespace flag when viewing the diff, as Git still seems to think a lot more has changed than what really has.

teschnei:
Yep, I saw the actual changes :)

src/map/utils/charutils.cpp
```diff
@@ -3075,30 +3075,37 @@ namespace charutils
...
-                {
-                    exp = 0;
-                    monsterbonus = 1.0f;
-                    chainactive = false;
-                    CCharEntity* PMember = (CCharEntity*)PChar->PParty->members[i];
-                    if (PMember->getZone() == PMob->getZone())
-                    {
-                        if (map_config.exp_party_gap_penalties == 1)
+                uint8 numAllianceParties = 1;
+                if (PChar->PParty->m_PAlliance) numAllianceParties = PChar->PParty->m_PAlliance->partyList.size();
```
teschnei: is it possible to use PChar->ForAlliance and use a lambda?  I want to hide the partylist/memberlist eventually so if you could get it to use that instead (if possible) I'd like that

jwonesh:
@teshnei I'll see what I can do tonight

jwonesh:
@teschnei I have rewritten to use lambda but I now get a compilation error which I cannot seem to fix, no matter how many stack overflow posts I read. Admittedly, my c++ skills are very basic (as I work with Java, normally), and my functional programming skills are even less developed. Here is the error that I get:

Error	C2893	Failed to specialize function template 'unknown-type std::invoke(_Callable &&,_Types &&...)'	DSGame-server	c:\microsoft visual studio 14.0\vc\include\type_traits	1494

It seems to have something do with using parameterized classes inside of the Lambda, but I cannot figure anything further out.

teschnei:
commit your code and I can look at it

jwonesh:
The solution became so clear once I took a step back and committed the code lol. I am recommitting the code for you to look at.

teschnei:
I added the alliance penalty values and optimized it a tad, thanks for contributing :)

jwonesh:
No problem, sorry my changes couldn't be of more use.

teschnei:
your changes were fine, all I did was remove some duplicated code because
of them :)

PR #2344, opened 2015-10-28
Scripts: Added command to set the equipment model for players.

Hozu:
Independent of gear, this will help people find the correct model for gear without having to use an ugly @exec command or worse, modify in the database and reboot. Or play around with packet injection.


Hozu:
I forget who sent me an early variant of this command but this is what I used when I had to find the model for implementing Angon.

PR #2350, opened 2015-10-30
Scripts: Fix soft cap application for Pollen and Healing Breeze

Hozu:
It doesn't help when the soft cap on power happens after power is used, so I moved it to before just like the rest of the spells. Tests show that they're in line with the data from the wiki.


bendangelo:
thanks

PR #2355, opened 2015-10-31
SCRIPTS: Fixed some weapon skill params for both 75 and 99 eras.

Fr0zt:
Changed some default params to be retail accurate for 75 era. Updated Adoulin params to current values. Converted tabs to spaces in all weapon skill files.


teschnei:
would have preferred having the tabs to spaces in a separate commit so I can actually read the weaponskill changes

teschnei:
nevermind, w=1 works fine with that many files
blade: yu has a typo

Fr0zt:
Typo in Blade: Yu has been corrected.

Hozu:
I'm curious as to what your source is for these, as I don't see any numbers for the accuracy penalty on either wiki for Sidewinder and Slug Shot. BG Wiki has "Has a substantial Ranged Accuracy penalty at 100TP." though 20% is already fairly significant isn't it? In a full ranged accuracy set in the 75 era you can get more than 400 ranged accuracy, so -20% is already -80, so at -50% for -200 I question the ability to land with any reliability whatsoever.

Fr0zt:
Slug Shot and Sidewinder: You are correct that there is no good data on the accuracy values for these weapon skills. I looked around and couldn't find anything solid. -50% is probably too much, however, I don't think that it should be at -20%. If you look at other weapon skills in DSP that "accuracy varies with TP" they are all set to -20% and they should have a lesser penalty than Slug Shot and Sidewinder. Perhaps -30% is more to your liking?

Coronach: I used data from the following page:
https://www.bg-wiki.com/index.php?title=Coronach&oldid=116386
This is from June 2009. If @teschnei wants to keep Coronach at it's current value then I will revert this change.

Hozu:
Uh... probably -30%/-15%/0% is closer to what it should be, yeah. Even -80 acc (before sushi) isn't THAT much for the 5.0 ftp. But I'm just guessing here.

teschnei:
Well, to be fair, it's like 95% chance likely that the "penalty" is a flat
amount and not a percent, because of the way accuracy works (-20 accuracy
is -10% hit rate, etc)

Deadwing888:
As a 75 cor in retail for many years, I'll say -40 acc sounds about right to me for slugshot at 100 tp. Although thinking back it did seem to miss more than 5% even when racc should have been capped...

It's certainly not -50% accuracy.

Fr0zt:
Is there is a good way to add flat acc penalties? The only acc params for weapon skills (as far as I am aware) are the percent based ones.

teschnei:
i'd just change the acc param one to be flat

pretty sure none of them are %

Fr0zt:
I was under the impression that params.acc was a percent accuracy modifier.
Here is a comment from line 25 in sickle_moon.lua:
--accuracy mods (ONLY USE FOR ACCURACY VARIES WITH TP) , should be the acc at those %s NOT the penalty values. Leave 0 if acc doesnt vary with tp.

Is this not correct?

teschnei:
like I said, I very much doubt that they are actually percents.  Even
things that are normally percents (corsair buff, geomancer buff, bard song)
are all flat values for accuracy, since it's already a percent and doesn't
scale on levels

for example, a -30% acc penalty at 100 acc (level 20 ish) is 15% hit rate
loss, while -30% acc at 1000 acc (level 119) is 150% hit rate loss (ie. it
will never hit something of your level)

Fr0zt:
I just did some testing with Sidewinder while setting params.acc100 to -40 and it has a 20% hit rate. Unless that is not what you meant by setting it to a flat rate then it seems like it does need to be a percent value.

teschnei:
the weaponskills.lua code (currently and incorrectly) assumes it is a
percent reduction, which is what I'm saying should be changed

Fr0zt:
I see, however fixing that is beyond my ability. Should I just change the acc back to what it was until that gets corrected?

Hozu:
Yeah it's probably for the best to revert those particular changes if you can't figure out how to get it to be a flat acc penalty. Someone else may be able to do it another time.

bendangelo:
ci failed

PR #2358, opened 2015-11-01
Core: DRG sub High Jump enmity loss reduction + more h2h skill check …

Hozu:
…fixes

Because searching DAMAGE_HTH via github didn't reveal all the things that are there. Hell, even the search I just did didn't reveal the check I fixed in the Jump function, which is the first one I found while looking for something else. https://github.com/DarkstarProject/darkstar/search?utf8=%E2%9C%93&q=DAMAGE_HTH&type=Code

As for the thing I was really looking to do, here's the source: https://www.bg-wiki.com/bg/High_Jump

### DIFF
--- a/src/map/ai/ai_mob_dummy.cpp
+++ b/src/map/ai/ai_mob_dummy.cpp
@@ -1841,7 +1841,7 @@ void CAIMobDummy::ActionAttack()
                                 else
                                 {
                                     int16 naturalh2hDMG = 0;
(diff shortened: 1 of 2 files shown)


teschnei:
the github search only shows the first 2 occurrences of a string per file,
so you shouldn't be using that

Hozu:
That explains that. I did a search through both files and found one more check that had to be changed, so there it is.

Hozu:
If the checks would stop failing that'd be great because it compiled and ran fine for me...

PR #2364, opened 2015-11-04
SQL Job Points Player menu

some1here:
Right below merits is job points
some info is on https://www.bg-wiki.com/bg/Job_Points#Acquiring_Capacity_and_Job_Points
as well as 
https://github.com/Windower/Resources/blob/master/lua/job_points.lua


teschnei:
what is the purpose of "upgrade", "upgradeid", and "catagoryid"? why are they there if they are not different?

some1here:
I was using the merits as the way this was constructed. Please just tell me if a different build was better.

teschnei:
you can remove those 3 columns since they make no sense in the context of job points

teschnei:
also, the file failed to import - you should probably test it before opening a pull request

sql/job_points.sql
```diff
@@ -0,0 +1,248 @@
+/*
+MySQL Data Transfer
+Source Host: localhost
+Source Database: dspdb
+Target Host: localhost
+Target Database: dspdb
+Date: ???
+*/
+
+SET FOREIGN_KEY_CHECKS=0;
```
bendangelo: Shouldn't need this line

some1here: Fails check if SET FOREIGN_KEY_CHECKS=0; is removed

teschnei: looks like appveyor failed because of github giving a 500 error - nothing to do with the sql file (appveyor doesnt even run the sql files, actually)

some1here:
"This usually happens when 'git.cgi' is running with the wrong permissions."
please add if I am not allowed darkstar/sql/job_points.sql
https://community.webfaction.com/questions/3019/git-http-error-500

links to back up data were in initial comment

teschnei:
it has nothing to do with you, just restart the check if it does that

sql/job_points.sql
```diff
@@ -0,0 +1,242 @@
...
+INSERT INTO `job_points` VALUES ('460', 'divine_emblem_effect', '2', '7');
+INSERT INTO `job_points` VALUES ('462', 'sepulcher_duration', '1', '7');
+INSERT INTO `job_points` VALUES ('464', 'palisade_effect', '1', '7');
+INSERT INTO `job_points` VALUES ('466', 'enlight_effect', '1', '7');
+INSERT INTO `job_points` VALUES ('512', 'blood_weapon_effect', '2', '8');
+INSERT INTO `job_points` VALUES ('514', 'arcane_circle_effect', '1', '8');
+INSERT INTO `job_points` VALUES ('516', 'soul_enslavement_effect', '1', '8');
+INSERT INTO `job_points` VALUES ('518', 'last_resort_effect', '2', '8');
+INSERT INTO `job_points` VALUES ('520', 'souleater_duration', '1', '8');
+INSERT INTO `job_points` VALUES ('522', 'weapon_bash_effect, '1', '8');
```
lichtl: Hi, I think this field needs to be closed.

PR #2365, opened 2015-11-04
Update ZoneIDs added Escha - Ru'Aun 

some1here:
Escha - Ru'Aun 289 as per
https://github.com/Windower/Resources/blob/master/lua/zones.lua


bendangelo:
Why'd you close this?

some1here:
I did not close it on purpose if I did.... I will reopen either way

some1here:
wrong file this just keeps failing adding a name and it is not important

PR #2368, opened 2015-11-04
Update job_traits.lua three traits added

some1here:
Added as per widower reference https://github.com/Windower/Resources/blob/master/lua/job_traits.lua#L54

### DIFF
--- a/src/map/trait.h
+++ b/src/map/trait.h
@@ -140,6 +140,9 @@ enum TRAITTYPE
     TRAIT_INSPIRATION         = 121,
     TRAIT_SLEIGHT_OF_SWORD    = 122,
     TRAIT_DAKEN               = 123
+    TRAIT_SUPERIOR            = 124 
+    TRAIT_MAGIC_ACC_BONUS     = 125
+    TRAIT_MAGIC_EVA_BONUS     = 126
 };
 
 #define MAX_TRAIT_ID    124
--- a/src/map/trait.h
+++ b/src/map/trait.h
@@ -143,8 +143,7 @@ enum TRAITTYPE
     TRAIT_SUPERIOR            = 124, 
     TRAIT_MAGIC_ACC_BONUS     = 125,
     TRAIT_MAGIC_EVA_BONUS     = 126
-    
-};
+  };
 
 #define MAX_TRAIT_ID    124
 
--- a/src/map/trait.h
+++ b/src/map/trait.h
@@ -143,6 +143,7 @@ enum TRAITTYPE
     TRAIT_SUPERIOR            = 124, 
     TRAIT_MAGIC_ACC_BONUS     = 125,
     TRAIT_MAGIC_EVA_BONUS     = 126,
+    
 };
 
 #define MAX_TRAIT_ID    124
--- a/src/map/trait.h
+++ b/src/map/trait.h
@@ -145,7 +145,7 @@ enum TRAITTYPE
     TRAIT_MAGIC_EVA_BONUS     = 126
   };
 
-#define MAX_TRAIT_ID    124
+#define MAX_TRAIT_ID    127
 
 /************************************************************************
 *                                                                       *
--- a/src/map/trait.h
+++ b/src/map/trait.h
@@ -142,7 +142,7 @@ enum TRAITTYPE
     TRAIT_DAKEN               = 123,
     TRAIT_SUPERIOR            = 124, 
     TRAIT_MAGIC_ACC_BONUS     = 125,
-    TRAIT_MAGIC_EVA_BONUS     = 126,
+    TRAIT_MAGIC_EVA_BONUS     = 126
     
 };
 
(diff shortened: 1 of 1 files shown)


src/map/trait.h
```diff
@@ -139,7 +139,10 @@ enum TRAITTYPE
     TRAIT_PRIMEVAL_ZEAL       = 120,
     TRAIT_INSPIRATION         = 121,
     TRAIT_SLEIGHT_OF_SWORD    = 122,
-    TRAIT_DAKEN               = 123
+    TRAIT_DAKEN               = 123,
+    TRAIT_SUPERIOR            = 124, 
+    TRAIT_MAGIC_ACC_BONUS     = 125,
+    TRAIT_MAGIC_EVA_BONUS     = 126,
```
bendangelo: Extra comma

src/map/trait.h
```diff
@@ -139,7 +139,10 @@ enum TRAITTYPE
...
     TRAIT_INSPIRATION         = 121,
     TRAIT_SLEIGHT_OF_SWORD    = 122,
-    TRAIT_DAKEN               = 123
+    TRAIT_DAKEN               = 123,
+    TRAIT_SUPERIOR            = 124, 
+    TRAIT_MAGIC_ACC_BONUS     = 125,
+    TRAIT_MAGIC_EVA_BONUS     = 126,
 };
 
 #define MAX_TRAIT_ID    124
```
xdemolish: increase so it's greater than the highest defined id by 1

src/map/trait.h
```diff
@@ -139,7 +139,11 @@ enum TRAITTYPE
     TRAIT_PRIMEVAL_ZEAL       = 120,
     TRAIT_INSPIRATION         = 121,
     TRAIT_SLEIGHT_OF_SWORD    = 122,
-    TRAIT_DAKEN               = 123
+    TRAIT_DAKEN               = 123,
+    TRAIT_SUPERIOR            = 124,
```
TeoTwawki: What job(s) actually have these 3 new traits?

teschnei: Every job has superior. I'm guessing that the magic accuracy traits are
also from gifts

some1here: "Superior" Equipment like Blurred Knife +1 http://ffxiclopedia.wikia.com/wiki/Blurred_Knife_%2B1 it to be equiped It is in the gift section

TeoTwawki: These should probably not even be handled via traits list then.

teschnei: Traits are a client side thing. The trait IDs are from the dats. We don't
get to choose what's there and what isn't

some1here: Pleas check the Widower scanned values https://github.com/Windower/Resources/blob/master/lua/job_traits.lua#L54

TeoTwawki: Kj..its a list of strings in that dat and we used an offset to make the first trait "one" and has ZERO TO DO with choosing what is a trait. Likewise I don't give 2 squirts about windowers resource list, they are reading the same strings and deciding anything past a certain offset must be traits. We could easily says traits ended at this value and "gift" started at this next value, internally. Very simply: do they show in game as a trait? Then I am comfy calling it a trait. But please do not make arguments that because the string was in a client file that means something special.

TeoTwawki: The weaponskills are in the same dat, as are job abilities. All same file. So are Fomor ws and monstrosity crap.

teschnei: The next one is magic accuracy bonus. I mean... I can go ask a friend with
superior unlocked, but I don't see why they'd randomly have superior
between other traits. Besides, having superior as a trait makes total
sense. It's the exact same thing as dual wield

TeoTwawki: Thats fine, if its a trait its a trait. If it makes sens internally to code is **as** as trait, gerat. Just don't expect being in a file to mean that. Theres a ton of non traits, same file.

TeoTwawki: The description didn't *sound* like a trait though.

src/map/trait.h
```diff
@@ -139,8 +139,11 @@ enum TRAITTYPE
...
     TRAIT_SLEIGHT_OF_SWORD    = 122,
-    TRAIT_DAKEN               = 123
-};
+    TRAIT_DAKEN               = 123,
+    TRAIT_SUPERIOR            = 124, 
+    TRAIT_MAGIC_ACC_BONUS     = 125,
+    TRAIT_MAGIC_EVA_BONUS     = 126
+  };
 
 #define MAX_TRAIT_ID    124
```
bendangelo: This has to be 127

nesstea: When you modify any traits, spells, etc., there is a constant you need the change too. 
like ben said, "#define MAX_TRAIT_ID    124" ---> 127

bendangelo:
looks good, thanks

PR #2377, opened 2015-11-06
Dmgtaken

m241dan:
Fixed the Borked MagicDmg Taken Formula. It's pretty clear from the wiki that this formula applies to both physical and magical, so it made sense to me that it likely applied to all "damage takens." Using the wiki accurate formula, I updated all the items in the database and found that only 80% of them were using the "borked" numbers and augments certainly were not. This should streamline the whole damage taken system and give us clearly legable item mods in the database.


src/map/utils/battleutils.cpp
```diff
@@ -4360,13 +4362,14 @@ namespace battleutils
...
+        float resist = 1.0f + floor( 256.0f * ( PDefender->getMod(MOD_UDMGPHYS) / 100.0f )  ) / 256.0f;
         resist = dsp_max(resist, 0);
         damage *= resist;
 
-        resist = 1.0f + (PDefender->getMod(MOD_DMGPHYS) / 100.0f) + (PDefender->getMod(MOD_DMG) / 100.0f);
-        resist = dsp_max(resist, 0.5f);
-        damage = damage * resist;
+        resist = 1.0f + ( floor( 256.0f * ( PDefender->getMod(MOD_DMGPHYS) / 100.0f ) ) / 256.0f )
+                      + ( floor( 256.0f * ( PDefender->getMod(MOD_DMG)     / 100.0f ) ) / 256.0f );
+        resist = dsp_cap(resist, 0.5f, 1.5f); //assuming if its floored at .5f its capped at 1.5f but who's stacking +dmgtaken equip anyway???
```
bendangelo: This will definitely break invincible. It has a -100% damage taken mod. Maybe the `dmgphys` should be calculated and capped first, then apply the uncapped dmg physical.

m241dan:
invincible just needs to be changed to MOD_UDMGPHYS 100 then.

src/map/utils/battleutils.cpp
m241dan: invincible just needs to be changed to MOD_UDMGPHYS 100 then.

m241dan: It already is.

bendangelo: Oh I didn't see you're multiplying resist to damage twice. It should work fine.

PR #2378, opened 2015-11-06
Fixed InnoDB deprecated command

CodeCorrupt:
I was receiving an error saying a command I was trying to use was deprecated when importing the .sql files to a new DB. After some playing around I discovered that removing the ROW_FORMAT=FIXED option fixed the issue.


CodeCorrupt:
This error was found using Windows server 2012-R2 and MySQL57

bendangelo:
thanks

PR #2397, opened 2015-11-11
Added some missing map keyitems + missing zone.h changes

TeoTwawki:


TeoTwawki:
Tossed in the zone.h that changes that somehow didn't get in #2395

PR #2400, opened 2015-11-12
SQL/Scripts: Dragon Fruit au Lait/Kitron Juice usable/recipe

Hozu:
Made those two items usable and gave them synth recipes. Also corrected the Regen effect on the Dragon Fruit au Lait.


Hozu:
By the way the estimated value is from both ffxiclopedia and bgwiki, but it does have the verification needed thing on the former.

bendangelo:
Thanks

PR #2401, opened 2015-11-13
Scripts/SQL: Fix the spawning of TEs in Dynamis - Xarcabard

Hozu:
The checks were being thrown out of whack because of onMobDeath running for every player in the group, so I borrowed code from the Maat's Cap check to fix it. Each trigger mob will set a specific bit on the appropriate variable if it's not already set, and when the variable has reached the correct value (ie. all triggers are dead), the TE will spawn and variable is cleared.

I also corrected the mob group of two TEs, as they were spawning as eyes instead of statues.

### DIFF
--- a/scripts/zones/Dynamis-Xarcabard/mobs/Vanguard_Eye.lua
+++ b/scripts/zones/Dynamis-Xarcabard/mobs/Vanguard_Eye.lua
@@ -19,7 +19,7 @@ end;
 -----------------------------------
 
 function onMobSpawn(mob)
-	mob:setMobMod(MOBMOD_SUPERLINK, mob:getShortID());
+    mob:setMobMod(MOBMOD_SUPERLINK, mob:getShortID());
 end;
 
 -----------------------------------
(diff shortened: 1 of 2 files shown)


Hozu:
Also there must have been mob ID shifts at some point because the specific mobs being referred to had different mobIDs... thankfully they were labeled in the mob_spawn_points.sql table, otherwise I'd have never found the correct ones.

whasf:
This has been tested and it works?

Hozu:
I tested it out myself with a party of 2, but it's always nice if someone else can verify it for me.

Hozu:
Re-tested to be sure, is fine as-is.

scripts/zones/Dynamis-Xarcabard/mobs/Vanguard_Eye.lua
```diff
@@ -35,27 +35,47 @@ end;
...
+    
+    -- 035 039: spawn 043 when defeated
+    if (mobID == 17330718) then
+        local southTE = GetServerVariable("[DynaXarcabard]TE43_Trigger");
+        if (bit.band(southTE, bit.lshift(1,0)) ~= 1) then
+            SetServerVariable("[DynaXarcabard]TE43_Trigger",bit.bor(southTE, bit.lshift(1,0)))
+        end
+    elseif (mobID == 17330756)  then
+        local southTE = GetServerVariable("[DynaXarcabard]TE43_Trigger");
+        if (bit.band(southTE, bit.lshift(1,1)) ~= 1) then
```
teschnei: this statement can't ever be false - the band is going to either be 2 or 0.  If you're checking whether a bit is set or net, compare it with 0 (you either have compare with 0 or with the shifted number again, and 0 is easier)

scripts/zones/Dynamis-Xarcabard/mobs/Vanguard_Eye.lua
```diff
@@ -35,27 +35,47 @@ end;
...
+        SpawnMob(17326086);
+    -- 144-149: spawn 150 when defeated
+    elseif (mobID >= 17330913 and mobID <= 17330918) then
+        local wallTE = GetServerVariable("[DynaXarcabard]TE150_Trigger");
+        if (bit.band(wallTE, bit.lshift(1, (mobID - 17330913))) ~= 1) then
+            SetServerVariable("[DynaXarcabard]TE150_Trigger",bit.bor(wallTE, bit.lshift(1, (mobID - 17330913))))
+        end
+    end
+    
+    if (GetServerVariable("[DynaXarcabard]TE150_Trigger") == 63) then
```
teschnei: don't you think it's a bit silly to get the server variable, flip a bit, save it, and then load it again to check to spawn? and why check the other two server variables if you know only one can ever be checked once (the one that got its bit flipped)?

scripts/zones/Dynamis-Xarcabard/mobs/Vanguard_Eye.lua
Hozu: Check for == 0? Guess someone messed it up with the maat's cap check then heh.

scripts/zones/Dynamis-Xarcabard/mobs/Vanguard_Eye.lua
Hozu: hm.. good point, I'll check the appropriate var only when the bit is set.

Hozu:
Oh crap I got this branched messed up because I had to revert to before the client update stuff. I'm going to have to close this and make a new pull.

PR #2404, opened 2015-11-13
Scripts: Make mobs despawn on teleport effect

Hozu:
For future handling of the handful of mobs that cast warp (ie. animated weapons, at least in old school, do they still do it on retail?). Player warps work fine.

Also corrected a minor typo in the despawnmob command.


teschnei:
They don't, but there are others that do

bendangelo:
thanks

PR #2408, opened 2015-11-14
Scripts/SQL: Fix the spawning of TEs in Dynamis - Xarcabard

Hozu:
The checks were being thrown out of whack because of onMobDeath running for every player in the group, so I borrowed code from the Maat's Cap check to fix it. Each trigger mob will set a specific bit on the appropriate variable if it's not already set, and when the variable has reached the correct value (ie. all triggers are dead), the TE will spawn and variable is cleared.

I also corrected the mob group of two TEs, as they were spawning as eyes instead of statues.

### DIFF
--- a/scripts/zones/Dynamis-Xarcabard/mobs/Vanguard_Eye.lua
+++ b/scripts/zones/Dynamis-Xarcabard/mobs/Vanguard_Eye.lua
@@ -19,7 +19,7 @@ end;
 -----------------------------------
 
 function onMobSpawn(mob)
-	mob:setMobMod(MOBMOD_SUPERLINK, mob:getShortID());
+    mob:setMobMod(MOBMOD_SUPERLINK, mob:getShortID());
 end;
 
 -----------------------------------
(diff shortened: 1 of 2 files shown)


Hozu:
I screwed up my branch so I had to make a new one here. This is after all of the changes that @teschnei  suggested.

scripts/zones/Dynamis-Xarcabard/mobs/Vanguard_Eye.lua
```diff
@@ -35,27 +35,48 @@ end;
...
+            northTE = 0;
+        end
+        SetServerVariable("[DynaXarcabard]TE60_Trigger", northTE);
+    -- 114: spawn 112 when defeated
+    elseif (mobID == 17326790) then
+        SpawnMob(17326086);
+    -- 144-149: spawn 150 when defeated
+    elseif (mobID >= 17330913 and mobID <= 17330918) then
+        local wallTE = GetServerVariable("[DynaXarcabard]TE150_Trigger");
+        if (bit.band(wallTE, bit.lshift(1, (mobID - 17330913))) == 0) then
```
xdemolish: might be better to set 17330913 as a var for easier editing if/when ids shift

scripts/zones/Dynamis-Xarcabard/mobs/Vanguard_Eye.lua
```diff
@@ -35,27 +35,48 @@ end;
...
+    if (mobID == 17330718 or mobID == 17330756) then
+        local southTE = GetServerVariable("[DynaXarcabard]TE43_Trigger");
+        local mobOffset = 0;
+        if (mobID == 17330756) then -- since the mobIDs aren't sequential there isn't an easy way to handle them outside of ID checks...
+            mobOffset = 1;
+        end
+        if (bit.band(southTE, bit.lshift(1, mobOffset)) == 0) then
+            southTE = bit.bor(southTE, bit.lshift(1, mobOffset));
+        end
+        if (southTE == 3) then
```
teschnei: probably better to move this into the previous if, right? it should only check to spawn it if the var actually changed?

Hozu: I was about to say that since each mob can only be killed once it shouldn't matter, but then I realized this whole commit is because onMobDeath runs for every person in the group. Changes made, re-tested and it still works.

PR #2412, opened 2015-11-14
Scripts/SQL: Achamoth mobskill list + Exuviation base amount

Hozu:
Exuviation heals for a base amount in addition to the extra amount from removing statuses.


Hozu:
Cancelled what..

bendangelo:
I like. thanks

PR #2413, opened 2015-11-14
tabled fov/gov cause why the hell would you even if/elseif tree that …

xdemolish:
…hard


m241dan:
Why would you commit this??? While I appreciate the work done, and it's much better to use lua tables, if you only do FoV, GoV side breaks. You may want to revert this in the project or update the GoV side.

xdemolish:
apparently groundsofvalor didn't get committed so its still sitting on my pc

i'll commit it in a few hours

PR #2416, opened 2015-11-15
changed the ForgeYourDestiny_timer to a correct value of 10800

whitevamp:
changed the ForgeYourDestiny_timer to a correct value of 10800 witch is 3 hrs / 3 in-game days, instead of the 10368 witch is 2.885 hrs /2.885 in-game days.


teschnei:
is it supposed to be 3 hours or 3 game days? they're not the same

whitevamp:
I saw in 1 place that it was the same, just looked around a bit more and 2 places say there not the same, and the original number was correct, will close this pull.

PR #2422, opened 2015-11-16
Pankration related NPC work..

TeoTwawki:
Tested and as retail accurate as can reasonably be without copying the client hang bug retail has.

(If you have one hundred million Jettons, retail trips a glitch.)


TeoTwawki:
@Fiocitrine **IT IS DONE.**

Fiocitrine:
lol

PR #2437, opened 2015-11-24
CORE: Mobs no longer path through walls...

m241dan:
Because Bad Friggin Dyslexic Algorythm


bendangelo:
I don't think this changes anything?

m241dan:
Sure fixed it on my server.

Fr0zt:
Mobs sure as hell path through walls on my server, as well as walk on water/air. I know they are supposed to wallhack when they don't have a path to you, but mobs wallhack all the time for me.

bendangelo:
I see what this fixes. The mob won't look at the point if it is already on the target point. I think the `LookAt` method should be slightly refactored but this is a good fix.

PR #2442, opened 2015-11-27
MOBMOD_STANDBACK_TIME no longer exists

TeoTwawki:
Did not test (and not going to) If the behavior was changed as well as the name, I'll push it off onto the guy that broke this and close the PR.

### DIFF
--- a/scripts/zones/The_Celestial_Nexus/mobs/Eald_narche.lua
+++ b/scripts/zones/The_Celestial_Nexus/mobs/Eald_narche.lua
@@ -1,12 +1,16 @@
 -----------------------------------
 -- Area: The Celestial Nexus
--- NPC:  Eald'Narche - Phase 1
+--  NPC: Eald'Narche - Phase 1
 -- Zilart Mission 16 BCNM Fight
 -----------------------------------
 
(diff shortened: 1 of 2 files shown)


bendangelo:
Wait, the new mod `mobmod_standback_cool` is different from the old one. This mod will make the mob cast magic more often when it is not within melee range. I think the original intent was to prevent this mob from standing back. So maybe it should just be removed?

TeoTwawki:
I can amend to comment it out if you think thats closer to the original behavior
I didn't look at wtf the replacement mod does relative to the original one and the descriptions of the old and new are ambiguously similar (not digging through old source to compare them, ya know?)

TeoTwawki:
@bendangelo amened, and thats as far as I go with this.

teschnei:
FYI: yes the original intent was he never stands back.

TeoTwawki:
I should have mentioned at the start: the reason I made this quickfix is it was crashing servers by calling nil modifier, and the guy in IRC didn't know why so I looked.

teschnei:
I think that much was evident in your original email :)

bendangelo:
Looks good thanks

PR #2446, opened 2015-12-01
Scripts: No standback for Bahamut/Eald'narche

Hozu:
Proper mobmod for Bahamut, added proper one for Eald'narche after the old one was removed due to refactoring of the mobmod.


TeoTwawki:
@Hozu thanks for fixing Eald'nards. I wasn't aware we could -1 the HP based standback mod till now.

Hozu:
Yeah -1 makes them never stand back, while 100 makes them stand back forever (Qiqirn RNGs).

Issue #2450, opened 2015-12-04
Ultima's Equalizer is... something else


Hozu:
In the mobskill table and in the monster tp moves doc, Equalizer is listed as 1005. However, when the mob uses it, it calls upon #749 instead.

teschnei:
1005 - 749 = 256

does that help?

Hozu:
I vaguely remember reading something about a 256 offset for CoP era and beyond mobskills... I'll have to look into it and see why specifically Equalizer is messed up.

PR #2451, opened 2015-12-04
Scripts/SQL: Add/update mobskills for (proto) Omega/Ultima.lua

Hozu:
-Corrected the animations for Omega's moves
-Did some scripting for all four mobs, mostly Proto Omega and some Proto Ultima.
-Added Particle Shield, Ion Efflux, and Discharger.
-Updated the checks on the other mobskills due to the scripting done for Proto Omega/Ultima.
-Separated the mob skill lists for the four mobs, as they all have different moves.
-Removed Equalizer due to #2450

Citadel Buster was not added because I couldn't think of a good way to have it triggered. I would think that SE uses a dummy mobskill to trigger the countdown start, but I have no idea what they do for sure. Nuclear Waste doesn't have a good way to be added either, so the six elemental moves will remain usable at will during the time they're supposed to be usable.


scripts/globals/mobskills/Colossal_Blow.lua
```diff
@@ -11,21 +11,30 @@ require("scripts/globals/monstertpmoves");
...
-	local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,MOBSKILL_PHYSICAL,MOBPARAM_SLASH,info.hitslanded);
-	target:delHP(dmg);
-	return dmg;
-end;
+    local currentHP = target:getHP();
+    -- remove all by 5%
+    local damage = 0;
+
+    -- if have more hp then 30%, then reduce to 5%
+    if (currentHP / target:getMaxHP() > 0.2) then
```
xdemolish: mob:getHPP()

Hozu: That was a copy and paste from Throat Stab since the old version of the script simply had it doing damage. I'll correct that in a moment.

Hozu: Done.

bendangelo:
:+1: Looks good

nesstea:
GJ Hozu. U be a sexy BIOtch

PR #2463, opened 2015-12-09
SQL\DATABASE: Mob Skill Updates for AV Family ID

Kthulupwns:


bendangelo:
Do all these skills work?

Hozu:
They should, they're the moves from polearm Aerns, which are already scripted and are used in the polearm Aern list.

bendangelo:
Ok thanks for contributing

PR #2471, opened 2015-12-11
Update Bubble Shower for King Arthro

Musashi-:


Musashi-:
Updated bubble shower for King Arthro fight from Sandworm tele

scripts/globals/mobskills/Bubble_Shower.lua
```diff
@@ -10,6 +10,23 @@ require("scripts/globals/monstertpmoves");
 ---------------------------------------------------
 
 function onMobSkillCheck(target,mob,skill)
+	mob = mob:getID();
+	if (mob == 17129519) then
```
TeoTwawki: That is incorrect. Mob isn't an ID, its an entity. To get a mobs ID from its entity, use `mob:getID()`. *However* we should **not** be using mob IDs as the condition for these things because mobIDs can change over time form client updates. Better to check some other condition associated with that mob, like its family ID just as an example..Actually. I think i know the best way of all:check the skill list itself because that will always be unique to the mob in question. `mob:getMobMod(MOBMOD_SKILL_LIST) == LIST_ID_HERE`

@bendangelo

TeoTwawki: @bendangelo didn't mean to tag you, was about to ask a question when I realized the answer: you made a mobMod for it (which was awesome, thanks)

TeoTwawki: @Musashi- just wanted to let you know that checking `mob:getPool()` will soon be an option as well, in the event checking the skill list won't do the job.

Musashi-:
Thanks man I appreciate all the in put. I hope this works out for everyone
and somebody can make it even better. Thanks again.

PR #2481, opened 2015-12-11
SCRIPTS/LUA: Medusa Javelin Update for AV

Kthulupwns:
Changed to deal the correct amount of damage specifically for AV, seeing as he is the only mob in the family ID that uses medusa javelin to require this modifier thought this would be the most straightforward approach.

### DIFF
--- a/scripts/globals/mobskills/Medusa_Javelin.lua
+++ b/scripts/globals/mobskills/Medusa_Javelin.lua
@@ -25,7 +25,7 @@ function onMobWeaponSkill(target, mob, skill)
     	end
 	local info = MobPhysicalMove(mob,target,skill,numhits,accmod,dmgmod,TP_NO_EFFECT);
 	local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,MOBSKILL_PHYSICAL,MOBPARAM_NONE,info.hitslanded);
-	target:delHP(dmg);	
+	target:delHP(dmg);
 
 	local typeEffect = EFFECT_PETRIFICATION;
 
--- a/scripts/globals/mobskills/Medusa_Javelin.lua
+++ b/scripts/globals/mobskills/Medusa_Javelin.lua
@@ -16,16 +16,23 @@ end;
 
 function onMobWeaponSkill(target, mob, skill)
 
+	local mobID = mob:getID();
 	local numhits = 1;
 	local accmod = 1;
 	local dmgmod = 1.5;
+	if (mobID == 16912876) then 
+   	dmgmod = 10.5;
+    	end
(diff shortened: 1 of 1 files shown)


scripts/globals/mobskills/Medusa_Javelin.lua
```diff
@@ -16,16 +16,23 @@ end;
 
 function onMobWeaponSkill(target, mob, skill)
 
+	local mobID = mob:getID();
 	local numhits = 1;
 	local accmod = 1;
 	local dmgmod = 1.5;
+	if (mobID == 16912876) then
```
TeoTwawki: Like I was just telling someone else, static mobID checks aren't such a great idea.
https://github.com/DarkstarProject/darkstar/pull/2471/files#diff-b0e2bde3c212fb2418d5e7a6df1f62e9R14
We can actually check the skill list via a mobMod though. `mob:getMobMod(MOBMOD_SKILL_LIST) == LIST_ID_HERE`

scripts/globals/mobskills/Medusa_Javelin.lua
```diff
@@ -16,16 +16,23 @@ end;
...
 	local numhits = 1;
 	local accmod = 1;
 	local dmgmod = 1.5;
+	if (mobID == 16912876) then 
+   	dmgmod = 10.5;
+    	end
 	local info = MobPhysicalMove(mob,target,skill,numhits,accmod,dmgmod,TP_NO_EFFECT);
 	local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,MOBSKILL_PHYSICAL,MOBPARAM_NONE,info.hitslanded);
-	target:delHP(dmg);
+	target:delHP(dmg);
```
TeoTwawki: Trailing tab, and extra whitespace near the return below, kill with fire pretty please. :smile:

scripts/globals/mobskills/Medusa_Javelin.lua
teschnei: Or the pool id.. Something that we have control over

TeoTwawki: @teschnei I think we have some pre existing skills in trunk done by mobID that already got merged, I'm going to go find and correct them right now so future editors don't see them and try to do same thing.

Hozu:
Wouldn't it be a better idea to get AV's stats done properly before adjusting mobskill damage? Or is it already matching retail?

scripts/globals/mobskills/Medusa_Javelin.lua
TeoTwawki: @Kthulupwns  `mob:getPool()` will work after #2482 gets merged

xdemolish:
conflicts, reopen when rebased

PR #2482, opened 2015-12-12
mob:getPool() is now a thing

TeoTwawki:
Also converted over the static mobID checks in smobskill scripts to check either the skill list ID or the pool ID...And I ran them through my auto format thing >.>

Source changes in first (and third) commit , script changes in second commit. Everything after that is because you sat on it to long :stuck_out_tongue:


TeoTwawki:
Sorry, have warnings off since there are so many of them already. Fixing.

teschnei:
visual studio will usually only warn once until that line is changed

teschnei:
however I'd like to go get rid of them all some time :)

TeoTwawki:
I had my return in the wrong line. I didn't go far with testing.

TeoTwawki:
-*nudge*-

TeoTwawki:
Fixed conflicts

PR #2493, opened 2015-12-16
The Sandworm Lua's

Musashi-:
The Sandworm and Lambton LUA's and the zones and the bcnm zones/mobs + updated bubble shower which spawns knight arthros knights. Doomvoid was the major add to make it all work. Everything's been tested and troubleshooted to get to this point. The math.random for sandworm spawn is supposed to be a 2, not a 1. Making it a 1 confuses it and glitches it out and spawns a worm in the same location every time which for some reason takes 0 damage when you atk it with melee. Making that 1 a 2 in the math.random fixed it's confusion and with the other code combined it fixes all possibility of there being more than 1 sandworm up at a time. Spawntype 128 is required. Also notice the doomvoid zone is now predetermined on spawn via Var. When the math.random was decided on use of doomvoid it would split the party and reroll the math random for every person in the party. This has been fixed via the random var on spawn.

### DIFF
--- a/scripts/globals/mobskills/Bubble_Shower.lua
+++ b/scripts/globals/mobskills/Bubble_Shower.lua
@@ -1,6 +1,6 @@
 ---------------------------------------------------
 -- Bubble Shower
--- Deals Water damage in an area of effect. Additional effect: STR Down
+-- Deals Water damage in an area of effect. Additional effect: Spawns Adds on King Arthro Fight
 ---------------------------------------------------
 
 require("scripts/globals/settings");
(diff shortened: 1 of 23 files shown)


xdemolish:
paragraph/bullet points :D

also, need to squash and rebase into one commit

scripts/globals/mobskills/Bubble_Shower.lua
```diff
@@ -9,7 +9,25 @@ require("scripts/globals/monstertpmoves");
 
 ---------------------------------------------------
 
-function onMobSkillCheck(target,mob,skill)
+function onMobSkillCheck(target,mob,skill,zone)
+	mob = mob:getID();
+	if (mob == 17129519) then
+	local mobids = {17129520,   -- King Arthro's Knights
```
TeoTwawki: same formatting probs I pointed out in your other PR, just so you know. And you need to rebase this off newer DSP, you are missing the onMobDeath script changes. `killer:` is replaced by `ally:` in most cases

Musashi-: Hey thanks ill try to remember that for future projects for sure :)

xdemolish: oh and rip out your tab key (use 4 spaces instead of tabs)

PR #2509, opened 2015-12-21
added item state

xdemolish:
idk why item finish animation wont start
also, tryinterrupt sux hence the error message pushed as CMessageBasicPacket instead of in the action packet itself

### DIFF
--- a/src/map/ai/ai_container.cpp
+++ b/src/map/ai/ai_container.cpp
@@ -27,6 +27,7 @@ This file is part of DarkStar-server source code.
 #include "states/attack_state.h"
 #include "states/death_state.h"
 #include "states/despawn_state.h"
+#include "states/item_state.h"
 #include "states/magic_state.h"
 #include "states/mobskill_state.h"
 #include "states/raise_state.h"
@@ -169,6 +170,19 @@ void CAIContainer::Trigger(uint16 targID)
     }
 }
 
+void CAIContainer::UseItem(uint16 targid, uint8 loc, uint8 slotid)
+{
+    auto PlayerController = dynamic_cast<CPlayerController*>(PEntity->PAI->GetController());
+    if (PlayerController)
+    {
+        PlayerController->UseItem(targid, loc, slotid);
+    }
+    else
+    {
+        Internal_UseItem(targid, loc, slotid);
+    }
+}
+
 bool CAIContainer::Internal_Engage(uint16 targetid)
 {
     //#TODO: pet engage/disengage
@@ -277,6 +291,13 @@ void CAIContainer::Internal_Raise()
     auto entity {dynamic_cast<CBattleEntity*>(PEntity)};
     if (entity)
         ForceChangeState<CRaiseState>(entity);
+}
+
+void CAIContainer::Internal_UseItem(uint16 targetid, uint8 loc, uint8 slotid)
+{
+    auto entity{ dynamic_cast<CBattleEntity*>(PEntity) };
+    if (entity)
+        ChangeState<CItemState>(entity, targetid, loc, slotid);
 }
 
 void CAIContainer::Reset()
--- a/src/map/ai/ai_container.cpp
+++ b/src/map/ai/ai_container.cpp
@@ -177,10 +177,6 @@ void CAIContainer::UseItem(uint16 targid, uint8 loc, uint8 slotid)
     {
         PlayerController->UseItem(targid, loc, slotid);
     }
-    else
-    {
-        Internal_UseItem(targid, loc, slotid);
-    }
 }
 
 bool CAIContainer::Internal_Engage(uint16 targetid)
@@ -297,7 +293,7 @@ void CAIContainer::Internal_UseItem(uint16 targetid, uint8 loc, uint8 slotid)
 {
     auto entity{ dynamic_cast<CBattleEntity*>(PEntity) };
     if (entity)
-        ChangeState<CItemState>(entity, targetid, loc, slotid);
+        ChangeState<CItemState>(static_cast<CCharEntity*>(entity), targetid, loc, slotid);
 }
 
 void CAIContainer::Reset()
--- a/src/map/ai/states/item_state.cpp
+++ b/src/map/ai/states/item_state.cpp
@@ -1,0 +1,303 @@
+﻿/*
+===========================================================================
+
+Copyright (c) 2010-2015 Darkstar Dev Teams
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see http://www.gnu.org/licenses/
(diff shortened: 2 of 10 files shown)


src/map/ai/states/item_state.h
```diff
@@ -0,0 +1,58 @@
...
+
+    CItemUsable* GetItem();
+
+    void InterruptItem(action_t& action);
+    void FinishItem(action_t& action);
+
+protected:
+    bool HasMoved();
+
+    CBattleEntity* const m_PEntity;
```
teschnei: switch this to a CCharEntity* and you won't have to static cast it to a CharEntity everywhere

src/map/ai/states/item_state.h
```diff
@@ -0,0 +1,58 @@
...
+ This file is part of DarkStar-server source code.
+
+ ===========================================================================
+ */
+
+#ifndef _CITEM_STATE_H
+#define _CITEM_STATE_H
+
+#include "state.h"
+#include "../../items/item_usable.h"
```
teschnei: replace this with a forward declaration if you can

src/map/ai/states/item_state.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+        auto PChar = static_cast<CCharEntity*>(m_PEntity);
+        PChar->pushPacket(m_errorMsg.release());
+
+        PChar->pushPacket(new CInventoryAssignPacket(m_PItem, INV_NORMAL));
+        PChar->pushPacket(new CInventoryItemPacket(m_PItem, m_PItem->getLocationID(), m_PItem->getSlotID()));
+    }
+}
+
+void CItemState::FinishItem(action_t& action)
```
teschnei: this function should be in charentity with the rest of the state callbacks (in case we ever need it to be polymorphic)

src/map/ai/states/item_state.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    if (PItem && PItem->isType(ITEM_USABLE))
+    {
+        if (PItem->isType(ITEM_ARMOR))
+        {
+            // todo: if item is locked, check if its equipped
+        }
+        else if (PItem->isSubType(ITEM_LOCKED))
+        {
+            return;
```
teschnei: throw exception instead of returning if there's an error (or else the state will construct and run like normal)

src/map/ai/states/item_state.cpp
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        m_PItem = PItem;
+    }
```
teschnei: probably want to throw if PItem is null or not usable

src/map/ai/states/item_state.cpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+CItemUsable* CItemState::GetItem()
+{
+    return m_PItem;
+}
+
+void CItemState::InterruptItem(action_t& action)
+{
+    TryInterrupt(dynamic_cast<CBattleEntity*>(GetTarget()));
```
teschnei: static_cast, should never be anything less than a battle entity

src/map/ai/states/item_state.cpp
```diff
@@ -0,0 +1,303 @@
...
+        }
+    }
+    else // разблокируем все предметы, кроме экипирвоки
+    {
+        m_PItem->setSubType(ITEM_UNLOCKED);
+
+        charutils::UpdateItem(PChar, m_PItem->getLocationID(), m_PItem->getSlotID(), -1);
+    }
+
+    action.actionid = m_PEntity->id;
```
teschnei: this should be action.id = m_PEntity->id I think, and I think actionid is the item ID

src/map/ai/states/item_state.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    actionList_t& actionList = action.getNewActionList();
+    actionList.ActionTargetID = PTarget->id;
+
+    actionTarget_t& actionTarget = actionList.getNewActionTarget();
+    actionTarget.reaction = REACTION_NONE;
+    actionTarget.speceffect = SPECEFFECT_NONE;
+    actionTarget.animation = m_PItem->getAnimationID();
+    actionTarget.param = 0;
+    actionTarget.messageID = 0;
```
teschnei: doesn't luautils::OnItemUse set these? or else I have no idea how items ever worked

src/map/ai/ai_container.cpp
```diff
@@ -169,6 +170,19 @@ void CAIContainer::Trigger(uint16 targID)
...
 }
 
+void CAIContainer::UseItem(uint16 targid, uint8 loc, uint8 slotid)
+{
+    auto PlayerController = dynamic_cast<CPlayerController*>(PEntity->PAI->GetController());
+    if (PlayerController)
+    {
+        PlayerController->UseItem(targid, loc, slotid);
+    }
+    else
```
teschnei: remove this else, I changed it recently where I don't want it to run without a controller (jailed char = remove controller)

src/map/ai/states/item_state.cpp
xdemolish: newp, returns 0 or -1

prolly messageSpecial's in the script

teschnei:
the rest looks pretty gucci, maybe the interrupt stuff is a bit wacky (probably from how items were interrupted before, I have no idea how it works)
savin me much time

src/map/ai/states/item_state.cpp
xdemolish: ya you should

src/map/ai/states/item_state.cpp
teschnei: ok... that's probably not how it should work.. but we can live with it for now

src/map/ai/states/item_state.cpp
xdemolish: copy/paste ftw

xdemolish:
tested and should be gud now

dont rename action's members til after merging <3

teschnei:
👌👀👌👀👌👀👌👀👌👀 good shit go౦ԁ sHit👌 thats ✔ some good👌👌shit right👌👌th 👌 ere👌👌👌 right✔there ✔✔if i do ƽaү so my selｆ 💯 i say so 💯 thats what im talking about right there right there (chorus: ʳᶦᵍʰᵗ ᵗʰᵉʳᵉ) mMMMMᎷМ💯 👌👌 👌НO0ОଠＯOOＯOОଠଠOoooᵒᵒᵒᵒᵒᵒᵒᵒᵒ👌 👌👌 👌 💯 👌 👀 👀 👀 👌👌Good shit

xdemolish:
oh wait no it's gud cause cant cast CCharEntity to CBattleEntity in constructor anyways

src/map/ai/ai_container.cpp
```diff
@@ -279,6 +289,13 @@ void CAIContainer::Internal_Raise()
         ForceChangeState<CRaiseState>(entity);
 }
 
+void CAIContainer::Internal_UseItem(uint16 targetid, uint8 loc, uint8 slotid)
+{
+    auto entity{ dynamic_cast<CBattleEntity*>(PEntity) };
+    if (entity)
+        ChangeState<CItemState>(static_cast<CCharEntity*>(entity), targetid, loc, slotid);
```
xdemolish: i changed my mind and copied exactly what you did so when you change your mind you can change this too

get merging

teschnei: oh, what I did was retarded and you're right, I'll fix it later

will merge when I get home

Issue #2502, opened 2015-12-20
Tinnin ZNM all fuqd up: Mixed up TP moves and also too weak


Fiocitrine:
Tinnin is supposed to be much harder than he is now. Curse from Nerve Gas should be much stronger. I remember it dropping a PLD to around 50-60% of his max HP. Also, when he uses Nerve Gas he then immediately follows with either Pyric or Polar Bulwark. 

Also Pyric and Polar Bulwark are mixed up. Pyric is supposed to be a Physical shield and Polar is supposed to be a magical shield, These are reversed.

Hozu:
Er, the two shields aren't reversed when I look at the script here...

Hozu:
Also BGwiki as it so it uses Nerve Gas after the bulwark, rather than the bulwark after Nerve Gas...

Fiocitrine:
Yeah, well I knew it was either one or the other after the other. It's been awhile, I just new he used in tandem.

PR #2510, opened 2015-12-21
Scripts/SQL: Tinnin scripting fixes

Hozu:
-Fixed the swapping of Pyric and Polar Bulwark - they were using the wrong mobskill ID.
-Increased the power of curse so it's -50% HP.
-Added the temporary disabling of no turning behaviour upon head regrowth; corrected the skills used upon head regrowth.
-Added the forcing of Nerve Gas upon using a Bulwark move; allowed Nerve Gas to be used with any amount of heads instead of all three.
-Corrected nameflag ("Tinnin" instead of "The Tinnin").
-Reduced the power of Gravity effect from Barofield because having both it and curse would prevent movement. Something like this would be stated on the wiki if it were a thing on retail...
-Reduced the power of Regen - supposed to be "a fairly weak Auto Regen" (50/tic until all three heads have grown for the first time, 10/tic at that point).
-Removed Regain because it's not supposed to have it (or if it does it's really weak).
-Added attack boost when it goes under 25% HP
-Added single target draw-in; gains 300 TP upon drawing in as it's to use a mobskill after drawing someone in.
-Removed onMagicHit function because it had nothing in it but would throw an error in the console.
-Added breath damage immunity.

Parts of this fixes all of https://github.com/DarkstarProject/darkstar/issues/2502


sql/mob_skills.sql
```diff
@@ -1618,9 +1618,9 @@ INSERT INTO `mob_skills` VALUES (1568,1286,'Amorphic_spikes',0,7.0,2000,1500,4,0
 -- INSERT INTO `mob_skills` VALUES (1570,1570,'Synergism',0,7.0,2000,1500,4,0,0,0);
 -- INSERT INTO `mob_skills` VALUES (1571,1571,'Metastasis',0,7.0,2000,1500,4,0,0,0);
 INSERT INTO `mob_skills` VALUES (1572,1308,'Pyric_Blast',4,15.0,2000,1500,4,0,0,0);
-INSERT INTO `mob_skills` VALUES (1573,1319,'Polar_Bulwark',0,7.0,2000,1500,1,0,0,0);
+INSERT INTO `mob_skills` VALUES (1573,1319,'Pyric_Bulwark',0,7.0,2000,1500,1,0,0,0);
 INSERT INTO `mob_skills` VALUES (1574,1309,'Polar_Blast',4,15.0,2000,1500,4,0,0,0);
-INSERT INTO `mob_skills` VALUES (1575,1318,'Pyric_Bulwark',0,7.0,2000,1500,1,0,0,0);
+INSERT INTO `mob_skills` VALUES (1575,1318,'Polar_Bulwark',0,7.0,2000,1500,1,0,0,0);
```
Hozu: Note that these changes match what's in https://github.com/DarkstarProject/darkstar/blob/master/documentation/monster_tp_moves.txt#L1578 - that is to say, whoever added those two moves here mixed them up.

Fiocitrine:
Thanks Hozu, Tinnin was just too easy. Hopefully I can test Sarameya next with an alliance.

xdemolish:
pull and rebase

if it gives you shit about submodule, reset to origin/master HEAD (mixed, don't hard reset) and it should be gud

Hozu:
Oh I thought this got merged before the tabs -> spaces commit was. Oh well, I'll handle it when I get home.

Hozu:
Also @Fiocitrine - the server I play on found Sarameya fairly difficult - back to back Gates of Hades during Chainspell was quite nasty. However, there is no howl sequence... We ended up having to zerg it with a bunch of melee w/ 2hr BRD support, which was a valid retail strategy I'm told.

Hozu:
Looks like I got the conflicts taken care of.

PR #2514, opened 2015-12-22
Changed Silk Caterpillar spawn mechanics

berghaushouse:
Removed spawning of Silk Caterpillar from onInitialize and removed the despawn logic from onGameHour as I'm not entirely sure what it was trying to achieve.

Added spawning logic to onGameHour so we check every Vanadiel Hour if the Silk Caterpillar should spawn, will only actually spawn at 03:00, 09:00, 15:00 and 21:00 as these are the approximate times the Jeuno-Bastok airship will be flying into Jeuno.

Despawns 210 seconds later if left unclaimed (approximately when the Jeuno-Bastok airship flies out of Jeuno).

All Airship timings were taken from ffxiclopedia.

Tested spawning and despawning and that it definitely drops that damn silkworm egg!

Only thing is I'm not sure about explicitly stating the Caterpillar's ID, is this likely to change?

Also this is my first time using github so anything I'm doing wrong please beat out of me now as I'd like to contribute more!

### DIFF
--- a/scripts/zones/Rolanberry_Fields/Zone.lua
+++ b/scripts/zones/Rolanberry_Fields/Zone.lua
@@ -67,9 +67,6 @@ function onInitialize(zone)
     -- Simurgh
     SetRespawnTime(17228242, 900, 10800);
 
-    -- Spawns Silk Caterpillar (temporary until someone implements a way to make it spawn properly)
-    SpawnMob(17227782,300,660);
-
 end;
 
 -----------------------------------
@@ -117,13 +114,15 @@ end;
 
 function onGameHour()
 
-    local VanadielHour = VanadielHour();
-    local SilkCaterpillar = 17227782;
-        
-    if (VanadielHour % 1 == 0 and GetMobAction( SilkCaterpillar ) == 16) then 
-        DespawnMob( SilkCaterpillar );
+    local vanadielHour = VanadielHour();
+    local silkCaterpillarId = 17227782;
+    --Silk Caterpillar should spawn every 6 hours from 03:00
+    --this is approximately when the Jeuno-Bastok airship is flying overhead towards Jeuno.
+    if (vanadielHour % 6 == 3 and GetMobAction(silkCaterpillarId) == ACTION_NONE) then
+        -- Despawn set to 210 seconds (3.5 minutes, approx when the Jeuno-Bastok airship is flying back over to Bastok).
+        SpawnMob(silkCaterpillarId, 210);
     end
-    
+ 
 end;
 
 -----------------------------------


TeoTwawki:
-*stamps the Teo Seal of Approval*-

In this situation the static ID is unavoidable, we use other methods when we can but even when we can't but we have tools that catch like 90% of these when they change now.

TeoTwawki:
p.s. seems likely whoever put it in onInitialize mistakenly thought it would repop properly. It doesn't.

scripts/zones/Rolanberry_Fields/Zone.lua
```diff
@@ -117,13 +114,15 @@ end;
...
-    end
-    
+    local vanadielHour = VanadielHour();
+    local silkCaterpillarId = 17227782;
+    --Silk Caterpillar should spawn every 6 hours from 03:00
+    --this is approximately when the Jeuno-Bastok airship is flying overhead towards Jeuno.
+    if (vanadielHour % 6 == 3 and GetMobAction(silkCaterpillarId) == ACTION_NONE) then
+    -- Despawn set to 210 seconds (3.5 minutes, approx when the Jeuno-Bastok airship is flying back over to Bastok).
+    SpawnMob(silkCaterpillarId, 210);
+  end
```
TeoTwawki: Missed an indent here after the if.
```lua
    if (vanadielHour % 6 == 3 and GetMobAction(silkCaterpillarId) == ACTION_NONE) then
        -- Despawn set to 210 seconds (3.5 minutes, approx when the Jeuno-Bastok airship is flying back over to Bastok).
        SpawnMob(silkCaterpillarId, 210);
    end
```

berghaushouse:
Indent sorted :)

scripts/zones/Rolanberry_Fields/Zone.lua
```diff
@@ -117,13 +114,15 @@ end;
...
-    end
-    
+    local vanadielHour = VanadielHour();
+    local silkCaterpillarId = 17227782;
+    --Silk Caterpillar should spawn every 6 hours from 03:00
+    --this is approximately when the Jeuno-Bastok airship is flying overhead towards Jeuno.
+    if (vanadielHour % 6 == 3 and GetMobAction(silkCaterpillarId) == ACTION_NONE) then
+        -- Despawn set to 210 seconds (3.5 minutes, approx when the Jeuno-Bastok airship is flying back over to Bastok).
+        SpawnMob(silkCaterpillarId, 210);
+  end
```
TeoTwawki: Look again.
You'll want to use an amend, so your commsit don't stack up trying to adjust one line.
In tortoisegit this is a checkbox on your commit screen. If using command line its `git commit --amend`

If using only the website or that vile app they offer here, switch! :smile:

berghaushouse: been using the website, will have a look into tortoise git and amend this using that as practice then :) thanks.

berghaushouse:
I'm not gonna lie, I have no idea what I just did!

xdemolish:
https://git-scm.com/book/en/v2/Getting-Started-Git-Basics

berghaushouse:
Will give it a good read through thank you.

berghaushouse:
Well, even after following what I thought was correct guides I'm now here with 12 commits, which I'm assuming is not what we want. Whilst the end result is the changes I am intending would it be best I close this and start fresh so we can have a neater looking thing?

TeoTwawki:
short method with less explaining: drop your files into a new clean copy re-commit and then In tortoise's push dialog use "arbitrary url" and specify the branch, and tick the "unknown" box to force push it onto this.

The result will be everything here gets replaced with that local copy's history including the new commit.

TeoTwawki:
The long method involves teaching you to rebase correctly.

teschnei:
You could cherry pick too instead of rebase. Kind of an in between in terms
of easiness

berghaushouse:
Holy balls have I actually manage to follow instructions correctly?  There's only 1 commit!

xdemolish:
looks gewd, thanks for contributing :)

PR #2515, opened 2015-12-22
SQL/Scripts: Correct Equalizer skillflag + power for Chemical Bomb

Hozu:
-Removed the WS skillflag from Equalizer as it's not a Fomor WS - it works as expected now.
-Increased the slow and elegy effect power on Chemical Bomb from 1 each to 512 each, as both are a very severe effect. Increased the duration as well.


Hozu:
Fixes https://github.com/DarkstarProject/darkstar/issues/2450

PR #2517, opened 2015-12-23
SQL: Correct the job/spell list/size of Voidstreaker Butchnotch

Hozu:
It's a NIN NM according to http://www.dynamisbums.com/strategy/san.html and http://ffxiclopedia.wikia.com/wiki/Voidstreaker_Butchnotch

flag and flags field updated as per request by @gedads (made the NM larger than regular mobs).


Hozu:
Whoops, forgot about the mob model.

Hozu:
Updated with size change (flag/flags) as requested by @gedads

PR #2523, opened 2015-12-25
Scripts/SQL: Implement proper behaviour for Carmine Dobsonfly

Hozu:
-Added a spell list of Slowga/Haste/Aero II; disabled casting while out of battle (all videos I've came across had them cast no spells while idle).
-Added superlinking behaviour.
-Gave it proper respawning mechanics - they will never respawn until all 10 have been defeated, at which point they'll all be set to the exact same respawn time.


bendangelo:
Thanks

PR #2526, opened 2015-12-26
Scripts/SQL: Refine mob version of relic WSs.

Hozu:
-Removed their "aftermath" effect - this video shows that they don't get it, or if they do it's silent... https://www.youtube.com/watch?v=P-cpElKuqiM (takes a while for it to land a hit)
-Standardized the inflicting of a status effect on the ones that do so. Made sure to use MobPhysicalStatusEffectMove so it only applies the effect if it lands a hit.
-Unswapped the animations of Randgrith and Mercy Stroke.
-Slightly unrelated, but corrected the nameflags for Zagan's Wyvern and Nebiros' Avatar (removed "The" prefix).


bendangelo:
Thanks for contributing

PR #2527, opened 2015-12-26
Scripts: Remove hardcoded itemid check on Relic WSs

Hozu:
Since the only way to get access to them is to have a completed relic/stage 4 weapon in Dynamis, there's no point in having a check for Aftermath. Whoever decides to add those level 85 latent weapons will have to add a specific check that that point in time.

Other fixes include:
-Refinement of Aftermath application (removed the if/elseif trees)
-Corrected the effect inflicted to the enemy from Randgrith (from weight to evasion down)
-Corrected the Subtle Blow values for Final Heaven and Blade: Metsu Aftermath.

### DIFF
--- a/scripts/globals/weaponskills/blade_metsu.lua
+++ b/scripts/globals/weaponskills/blade_metsu.lua
@@ -35,20 +35,13 @@ function onUseWeaponSkill(player, target, wsID)
 
 
     local damage, criticalHit, tpHits, extraHits = doPhysicalWeaponskill(player, target, params);
-    if damage > 0 and (target:hasStatusEffect(EFFECT_PARALYSIS) == false) then
+    -- TODO: Whoever codes those level 85 weapons with the latent that grants this WS needs to code a check to not give the aftermath effect.
+    if (damage > 0) then
+        local amDuration = 20 * math.floor(player:getTP()/100);
+        player:addStatusEffect(EFFECT_AFTERMATH, 10, 0, amDuration, 0, 1);
         target:addStatusEffect(EFFECT_PARALYSIS, 10, 0, 60);
     end
-        if ((player:getEquipID(SLOT_MAIN) or (SLOT_SUB) == 18312) and (player:getMainJob() == JOB_NIN)) then
-        if (damage > 0) then
-            if (player:getTP() >= 100 and player:getTP() <200) then
-                player:addStatusEffect(EFFECT_AFTERMATH, 5, 0, 20, 0, 1);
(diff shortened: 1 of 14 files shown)


Hozu:
As a side note, the relic WS is only added if the relevant weapon is in the main hand - the check explicitly checks SLOT_MAIN.

scripts/globals/weaponskills/blade_metsu.lua
```diff
@@ -35,20 +35,16 @@ function onUseWeaponSkill(player, target, wsID)
...
-                player:addStatusEffect(EFFECT_AFTERMATH, 5, 0, 40, 0, 1);
-            elseif (player:getTP() == 300) then
-                player:addStatusEffect(EFFECT_AFTERMATH, 5, 0, 60, 0, 1);
-            end
+    -- TODO: Whoever codes those level 85 weapons with the latent that grants this WS needs to code a check to not give the aftermath effect.
+    if (damage > 0) then
+        local amDuration = 20 * math.floor(player:getTP()/100);
+        player:addStatusEffect(EFFECT_AFTERMATH, 10, 0, amDuration, 0, 1);
+
+        if (target:hasStatusEffect(EFFECT_PARALYSIS) == false) then
```
bendangelo: This isn't needed, these checks are done before adding the status

scripts/globals/weaponskills/randgrith.lua
```diff
@@ -30,23 +30,19 @@ function onUseWeaponSkill(player, target, wsID)
...
-        if ((player:getEquipID(SLOT_MAIN) == 18324) and (player:getMainJob() == JOB_WHM)) then
-        if (damage > 0) then
-            if (player:getTP() >= 100 and player:getTP() < 200) then
-                player:addStatusEffect(EFFECT_AFTERMATH, 20, 0, 20, 0, 9);
-            elseif (player:getTP() >= 200 and player:getTP() < 300) then
-                player:addStatusEffect(EFFECT_AFTERMATH, 20, 0, 40, 0, 9);
-            elseif (player:getTP() == 300) then
-                player:addStatusEffect(EFFECT_AFTERMATH, 20, 0, 60, 0, 9);
-            end
+        if (target:hasStatusEffect(EFFECT_EVASION_DOWN) == false) then
```
bendangelo: Not needed

bendangelo:
Fix those things, then I'll merge it. :+1:

Hozu:
Removed the status effect checks on the WSs mentioned and others. Also added the missing attack down effect on Gate of Tartarus.

PR #2529, opened 2015-12-27
Scripts/SQL: More Wamoura moves, resistances, Achamoth spells

Hozu:
-Scripted Erosion Dust, Erratic Flutter, Fire Break, and Proboscis.
-Updated Exuviation to heal more for each debuff removed, but slightly less base heal; fixed a bracket I dropped last time I modified the file.
-Added a new mobskill list for Achamoth since it has Fire Break while others do not.
-Added a spell list for Achamoth containing Protectra V, Shellra V, Hastega, and Curaga IV.
-Updated Protectra V and Shellra V to have the proper base power if the caster has no merits (for mob use).
-Updated MobStatusEffectMove (and physical/gaze) functions to pass flags and subpower so I could add the defense down portion of Dia effect on Erosion Dust; updated Bio effect of Soulflayer's Tribulation mobskill as well.
-Updated family resistances for Wamoura and Achamoth families - both are "immune" to fire damage, while Achamoth is half resistant to wind.


Hozu:
Oh and most -na spells are also in there.

As an aside, I did not handle add spawning because for whatever reason, I can't target the Achamothcampa. Their names are visible, so it's not the stealth flag or whatever various other mobs use.

teschnei:
Their "flags" probably have untargetable set in the db column. Check the
source for what value that is and unset it in the db if it's set

Hozu:
In mob_pools.sql? When I searched for the untargetable thing, it appeared to be handled under roamflag, which is set to 0. If it's also done in the flags column for whatever reason, I have no idea what the flags are there as I've been unable to find where they're defined.

teschnei:
It's under flags. Look for the SetUntargetable function I think it's
called. I'll be home Monday to help if you can't find the flag value by then

Hozu:
Found the flag, 0x800 in 'flags'. Time to script...

Hozu:
Well, now when I try to kill the Achamothcampa, or despawn it, it doesn't despawn. When I despawn it it gets stuck on action 23 (fade out). It does not have no despawn behaviour set. Unless I can fix that, this pull is ready.

teschnei:
how the old AI works isn't really something to worry about at this point

Hozu:
Ok I'll wait until the ai rewrite to do the adds then. Though I updated the pull so the Achamothcampa can be targeted so I don't forget about it later.

Pull is ready to be looked at for merging.

teschnei:
You can probably test it on the ai rewrite branch already

teschnei:
can you request to merge this with airewrite instead of master?

Hozu:
Uh, sure I guess, though I didn't do any scripting of mobs, just added mobskills. I'll have to figure out how to make it to the airewrite branch.

teschnei:
Probably going to want anything AI related to be done there from now on

PR #2532, opened 2015-12-27
SQL: Add sleep resistance to various mob families.

Hozu:
Most Undead, Hecteyes, Greater Birds, and Kindred should have extra dark-based sleep resistance. Removed immunity to sleep from all of the Hecteyes mobs because they're vulnerable to light-based sleep.


Fiocitrine:
Bats?

Hozu:
Oh, forgot about those.

Hozu:
Added Bat/Bats family.

PR #2545, opened 2016-01-01
Scripts: Pso'Xja Gargoyle door able to be picked open

Hozu:
-Made the doors able to be picked open by a THF main with Thief's Tools/Living Key/Skeleton Key. Using wiki estimated success rate of 50%.
-Fix the message displayed for each message (now includes the player name, thanks @teschnei)
-Corrected the positions of the 16 Pso'Xja doors in the header.
-Added sanity checks to the locked gates in Phomiuna Aqueducts (only consumes key if gate is closed and player is on the locked side). Now able to be picked open with Living Key and Skeleton Key in addition to Thief's Tools.


bendangelo:
Looks good thanks.

PR #2546, opened 2016-01-02
Create flask_of_primeval_brew.lua

TwIzTiD2469:


xdemolish:
some of the values don't match wiki's
http://ffxiclopedia.wikia.com/wiki/Primeval_Brew

TwIzTiD2469:
https://www.bg-wiki.com/bg/Primeval_Brew is where I got my info.

xdemolish:
tabs -> 4 spaces

then rebase/squash

TwIzTiD2469:
ok should be fixed properly now. new to git lol, and somehow spacing kept screwing up :/

scripts/globals/items/flask_of_primeval_brew.lua
```diff
@@ -0,0 +1,98 @@
...
+-- Ranged Attack +9000
+
+require("scripts/globals/status");
+
+-----------------------------------------
+-- OnItemCheck
+-----------------------------------------
+
+function onItemCheck(target)
+local result = 0;
```
TeoTwawki: Your indentation and formatting are wrong. Also `== true` isn't needed to check if its true.
```lua
function onItemCheck(target)
    local result = 0;
    if (target:hasStatusEffect(EFFECT_TRANSCENDENCY)) then
        result = 523;
    end
    return result;
end;
```
Its also possible to just else the result if you want (doesn't really matter though, do as you see fit) instead of variablizing it.
```lua
function onItemCheck(target)
    if (target:hasStatusEffect(EFFECT_TRANSCENDENCY)) then
        return 523;
    else
        return 0;
    end
end;
```

The trunk currently has horrible inconsistency all over because many cooks one pot + older code wasn't held to very good (or any?) standards so picking an existing script to model ones work on is a crapshoot.

TeoTwawki:
> ok should be fixed properly now

Still needs collapsed. If you can't do rebase+squash you can use reset in mixed mode to keep files unchanged while setting history to an earlier point, recommit the current files as a single commit, then force push onto branch the branch and the pull request will automatically be updated.

xdemolish:
amend last commit to only commit primeval brew and force push, looks like your rebase went wrong somewhere

TwIzTiD2469:
ok, amended and force pushed.

teschnei:
can you move the onEffectGain/Loss to scripts/globals/effects/transcendency.lua instead of in the brews script? we normally do it this way for food because EFFECT_FOOD can do a thousand different things, but transcendency is always the same

TwIzTiD2469:
yes. would I just amend last commit again when done? still a git nub lol

teschnei:
yep, what you did last time to amend/force push was correct

TwIzTiD2469:
made the changes and tested it, works as it should

xdemolish:
looks good, thanks!

PR #2547, opened 2016-01-02
Scripts: Correct Filamented Hold's power

Hozu:
ffxiclopedia and BG wiki have conflicting power values so I'm going to go with the latter (25%, or 256/1024).


Hozu:
...why am I getting diff not available? All I did was change 25 to 256...

Hozu:
Now it's showing up, and I corrected the file that got submitted (didn't mean settings, whoops!).

PR #2554, opened 2016-01-04
Gathweeda guildshop entry

Scavenge:
Added minium and a few other items on gathweega and gave him a lua, also added the same new items to wahraga as well, since i took his stock in the first place


mspyerzuka:
Warp is using magic so it doesn't count!

teschnei:
Importing guild_shops.sql into the database...ERROR 1062 (23000) at line 1748: Duplicate entry '60425-4443' for key 'PRIMARY'

Scavenge:
yea i just changed that before submitting, i must not have reloaded the page or something

Scavenge:
oh it's the one on wahraga woops

Scavenge:
should be good now..

xdemolish:
rebase/squash

PR #2565, opened 2016-01-06
Add editor config file to help propagate some coding examples

ddumont:
Editor config is a plugin for many text editors including VIM, Atom, Sublime, etc... Try it out!

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+# top-most EditorConfig file
+root = true
+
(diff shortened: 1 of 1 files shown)


bendangelo:
Now we just have to get everyone to install the plugin

teschnei:
line endings too would be nice (lf)

ddumont:
@teschnei done.

PR #2566, opened 2016-01-06
Safe, cacheable requires in lua.

ddumont:
So, I found out that some of the incorrect text I was seeing was the result of some pretty bad global variable name re-use.  One module would load with a global named ```FOO```, and then another would also set ```FOO``` to another value and clobber the previous one.

This works pretty well, and allow us to let lua keep the modules cached.

What do you think?


xdemolish:
death to textid globals

@starlightknight's textid update script will need updating if this is merged, though
https://github.com/DarkstarProject/darkstar-utils/blob/master/id-update-batch-jobs/source/darkstar-batch/src/main/java/org/darkstar/batch/TextIdUpdate.java#L180

starlightknight:
@xdemolish If we are going to do this i would prefer consistency across all zones, as that makes it easier for automation (and guarantees the problem is permanently solved). It would not be hard to write a script to batch convert all zones to this style automatically. I am currently working on some text ids myself that require updates to the batch - in the last update, some textids actually shifted backwards - somethat that has never happened in the last 2 years that we've been batch updating. Once I fix those, I could look at doing a batch conversion to locals but it might be over the week or two

xdemolish:
sorry i meant if we choose to go with this (for all zones, not just the ones in the pull request)

TeoTwawki:
To add to what starlight said, not only do we need all zones consistency but you need to verify this works the way you think it does when done to all zones rather than just one zone.

There are several alternative methods to removing the need to reload the IDs on a zone by zone basis. Tabling like this is similar to a suggestion I made awhile back, except I was also tossing around the idea of a single file with each table being named for the zone name. Problem with a single file would be that it might be to large. And I know this may sound crazy but: I can see a way to handle these (and also easily keep them updated them on client updates) with an sql table loaded once at server start as well. 

Every other way I've thought about doing it without lua tables kept the messageSpecial() calls identical to the trunks current form, but if what you have works I am not opposed to it.

teschnei:
we actually did do something similar with selbina many years ago, but it was changed to the current way because it was different.  so while this is pretty much the preferred format (i'd probably call it text instead of tids though..), we'd want to have it on every zone (or at least most) asap

TeoTwawki:
I recall things not working as expected way back when, and that being the reason for being changed.

ddumont:
I did a bit of testing and verified that the local var is not leaked in
global scope.  However I'm not surprised you had issues, there are many
refs to the globals without zone textid files, relying accidentally on the
previous set value.

sent using my moto x

ddumont:
Tbh, an SQL table also sounds like a great solution for this...  I'm open
to either solution as long as we get rid of these globals.

sent using my moto x

TeoTwawki:
> I did a bit of testing and verified that the local var is not leaked in global scope.

No offense but everyone tells me they texted XYZ. I don't know what *way* you tested this and would rather just see multiple zones done at once and then see for myself.

ddumont:
That's fine, I can convert a few more.  My main goal was to get this done
as a simple test case and start this discussion.  I can most certainly add
to it :)

sent using my moto x

TeoTwawki:
Thanks.

teschnei:
@TeoTwawki if you look at who changed the selbina IDs, you'll understand why it "wasn't working as expected" ;)

TeoTwawki:
@teschnei I almost couldn't find the commit anymore, so old now. My system just sorta hangs for awhile trying to load all that history.. `#WindowsUserProblems`

TeoTwawki:
@teschnei looks like dia was trying to do what I suggested yesterday, but still left it per zone.

ddumont:
@TeoTwawki Ok in the latest commit I've also converted west saru.  I've removed a few text ids that weren't used, and added some for the ```To catch a falling star``` quest that had changed and were before using only numerical ids.

I tested out the quest and it appeared fine to me.  Also, I've done some testing with some code I've not committed where I've tried to access vars defined in the other modules and things seem to be scoped properly.

TeoTwawki:
@ddumont stop. Get with me on irc you are killing my attempt to help you with a big sweeping replace.

TeoTwawki:
Btw the reason I had some much trouble testing yesterday was the zone lua was still the old version, so it was axing your table values for west saruta.

TeoTwawki:
@ddumont so you know I think there are gonna be a few arrayed TextIDs my mass replace didn't catch, and I also forgot to handle anything outside the zones subdirectory. some GM command scripts like `@addkeyitem` have a concatenated strings to get the current zones TextIDs that will need modified.

xdemolish:
reopen when conflicts are resolved

PR #2568, opened 2016-01-06
SCRIPTS: Text ID Updates - Fix some misaligned text ids

starlightknight:
In the December patch, some zones had Text IDs shift backwards. With
the current version of the batch job, this would mark them for manual
review instead of shifting them. When I saw this in December I lazily
assumed without really checking that someone had just added a handful
of new TextIDs without proper comments, since no TextIDs had shifted
forward (that is what usually causes the review marking). 

This is the first time in the 2 years of batch updates that TextIDs have shifted backwards

Fix to the batch job will be pushed to utils at some point soon

Note: I also updated my POLUtils to the new version which fixed some issues with SoA table dumps, so some of the SoA zones that were being missed due to this got their couple of IDs updated as well.


gedads:
Do your magic with npc_list too please!

nesstea:
<3 @starlightknight

PR #2592, opened 2016-01-10
Add trim trailing to editorconfig

TeoTwawki:
..Think I did that right. >.>

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,22 @@
 # top-most EditorConfig file
 root = true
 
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -1,7 +1,22 @@
...
 indent_style = space
 indent_size = 4
 end_of_line = lf
+
+[*.lua]
+trim_trailing_whitespace = true
+
+[*.{cpp,h}]
+trim_trailing_whitespace = true
+indent_brace_style = Allman # This is unfortunately ignored in the Visual Studio add on
```
bendangelo: What's allman?

TeoTwawki: Its what the style of indent and bracing DSP uses is actually called. Named for the guy that started it.

PR #2597, opened 2016-01-11
Add wsID to doWhateverWeaponskill()

TeoTwawki:
Used mine not yours, parameters in diff order.


TeoTwawki:
Screw you travis and appveyor, always logging my amends for all to see :pout:

PR #2608, opened 2016-01-12
Fixed bug where is was possible to spawn a Jug pet whose level exceed…

dickeyf:
…s master's level.

Possible only in these circumstances :
- Master have to be lvl 75 with BEAST_AFFINITY merits.
- Master have to spawn a pet whose lvl limit + BEAST_AFFINITY exceeds master lvl.

The master's level was picked as a base.  Then merits got added on top of that, and that allowed for a pet whose level exceed master's.

Fix : Compute pet max level w/ merits.  Then ceil it to master's lvl, then randomize -[0,2]

### DIFF
--- a/src/map/utils/petutils.cpp
+++ b/src/map/utils/petutils.cpp
@@ -1298,21 +1298,21 @@ namespace petutils
         }
         else if (PPet->getPetType() == PETTYPE_JUG_PET){
             PPet->m_Weapons[SLOT_MAIN]->setDelay(floor(1000.0f*(240.0f / 60.0f)));
-            //TODO: Base off the caps + merits depending on jug type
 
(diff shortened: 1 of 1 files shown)


Hozu:
I wanted to do this earlier, but then I found out that it actually IS possible on retail for jug pets to exceed master's level, much to my dismay.

teschnei:
It is, but not in this way (ie. beast affinity going over the cap)

they only go over the beastmaster's level if the beastmaster is using a weapon with item level

Hozu:
So that means... I should have fixed this over two months ago or whenever I implemented Monster Gloves. Sigh...

teschnei:
yes :)

Issue #1353, opened 2015-04-24
crash getting leader of party, after having already checked to be sure wasn't null...


TeoTwawki:
adfadsafdsa m(_ _)m 
Alright @teschnei, I give up. Here's a dump. https://yadi.sk/d/Xu4fFDOsgDC24

*Getting laughed at on IRC later* : +1

teschnei:
pdb wasn't right, couldn't get debug info :p

TeoTwawki:
Newer dump, this one should work https://yadi.sk/d/ngvQnVGKgdsja that's from revision cef9810560af103a3699c1f4f802159cb41ed5e4
@maxtherabbit since he's not going to be back for awhile yet, you mind taking a look at this and helping me prevent it? The trigger is trying to get the name of a party leader who's not on the same map server, but is otherwise valid. I incorrectly thought I could just make sure they weren't null first.

teschnei:
GetLeader should give you null if they are on another server. If they are a
debug memory location instead of null, its a different problem

PR #2619, opened 2016-01-14
Fix issue #1353 (cross map server nexus cape crash)

TeoTwawki:
@teschnei soon as I put `lua_pushlightuserdata(L, (void*)PLeader);` it finally behaved. No void* no worky.

I did try the [alternate push method we discussed](https://github.com/TeoTwawki/darkstar/commit/31daad3b1f4e60d46e6e28f4e9a65a25b5489770?w=1), which stopped the cross server problems but then broke it in same zone (I don't even know why that would be and didn't investigate farther).

Server using multiple map servers can now allow nexus capes again - it'll just fail to do any teleporting across map servers.


teschnei:
I can't comment on specific lines (per line commenting is disabled with ?w=1 for some reason), so just remove that comment in lua_baseentity about the casting

TeoTwawki:
sure

TeoTwawki:
amended

PR #2646, opened 2016-01-20
Core/SQL: Update Jailer weapons to use level check for mods.

Hozu:
-They will now not add an additional swing while the player's level is temporarily reduced to below the level they can be equipped.
-Fixed the issue where the extra swing would be applied to a non-Jailer weapon while dual-wielding (whoops, that one is my fault).


Hozu:
My tests have shown that this will cause the following behaviours:
Faith Baghnakhs can punch up to two additional times (perfect)
If you have one stone left while equipped with Faith Baghnakhs, only one additional punch is possible (tested by setting the swing rate to 100%)
When equipped with two weapons, one of them being a Jailer weapon, only the Jailer weapon can get the extra swing, regardless of the hand that the Jailer weapon is equipped in.
When under level cap status, below level 73, no extra swings are possible.

Note that this does not support dual wielding two Jailer weapons at once, as that is not possible in retail.

Deadwing888:
The jailer weapon swings on retail stack with triple attack double attack brutal etc.

I'm sure it's on the wiki but I thought I'd mention.

Hozu:
Ah yes, that behaviour is also accounted for. Stacks with those, but does not work on WS nor jump.

PR #2651, opened 2016-01-21
SQL: Updated Drop List

BRFord88:
Update to drop list to include items for obtaining Opo-opo Crown.  Also altered some drop rates that have been consistently reported higher from multiple sources.


bendangelo:
thanks

PR #2652, opened 2016-01-21
implement eagle eye shot

xdemolish:
tested and werks but fkifiknow if its retail


xdemolish:
also the ability damages before bow is even drawn in animation

teschnei:
i already did the EES script though!

xdemolish:
gdi kayjay as if i ever check commits

teschnei:
my script didn't do anything with OnAbilityCheck though, so I think that might be needed

PR #2659, opened 2016-01-22
DNC af quest

ProfessorCattington:
First pass at the rest of the DNC AF quests. Mostly .lua changes and a small change to the lua_baseentity class.

### DIFF
--- a/scripts/zones/Bastok_Markets/npcs/Matthias.lua
+++ b/scripts/zones/Bastok_Markets/npcs/Matthias.lua
@@ -203,7 +203,6 @@ function onEventFinish(player,csid,option)
 
         if(player:getFreeSlotsCount() == 0) then --check to see if the player has enough inventory space before rewarding them.
             player:messageSpecial(ITEM_CANNOT_BE_OBTAINED, dancerAFID);
-            return;
         else
             local playerGender = player:getGender();
             player:addItem(dancerAFID - playerGender);
--- a/scripts/zones/Bastok_Markets/npcs/Matthias.lua
+++ b/scripts/zones/Bastok_Markets/npcs/Matthias.lua
@@ -31,7 +31,7 @@ function onTrade(player,npc,trade)
     if (player:getVar("dancerTailorCS") == 4) then
         local playersAFChoice = player:getVar("dancerAFChoice");
 
-            if(playersAFChoice == 1 and
+            if (playersAFChoice == 1 and
                 trade:hasItemQty(imperialSilk, 1) == true and
                 trade:hasItemQty(wolfFelt, 1) == true and
                 trade:hasItemQty(silverBrocade, 1) == true and
@@ -40,7 +40,7 @@ function onTrade(player,npc,trade)
 
                 rewardThePlayer(player);
 
-             elseif(playersAFChoice == 2 and 
+             elseif (playersAFChoice == 2 and 
                 trade:hasItemQty(karakulCloth, 1) == true and
                 trade:hasItemQty(rainbowCloth, 1) == true and
                 trade:hasItemQty(rainbowVelvet, 1) == true and
@@ -49,7 +49,7 @@ function onTrade(player,npc,trade)
 
                 rewardThePlayer(player);
 
-             elseif(playersAFChoice == 3 and
+             elseif (playersAFChoice == 3 and
                 trade:hasItemQty(wamouraCloth, 1) == true and
                 trade:hasItemQty(moblinWeave, 1) == true and
                 trade:hasItemQty(goldBrocade, 1) == true and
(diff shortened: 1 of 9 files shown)


scripts/zones/Bastok_Markets/npcs/Matthias.lua
```diff
@@ -15,17 +15,125 @@ require("scripts/zones/Bastok_Markets/TextIDs");
...
+        if(player:getFreeSlotsCount() == 0) then
+		
+            player:PrintToPlayer( "Not enough free inventory space." );
+			
+        elseif(currentVanaDay > tailorStartedWorkDay ) then
+			
+            local dancerAFID = 1; -- variable used to convert player's choice into an Item ID. gender is actually important here because it displays the item on screen for you
+            local playerGender = player:getGender();
+			
+            switch (playersAFChoice) : caseof{
```
teschnei: just use a if/elseif chain, I'm going to deprecate the switch eventually (it doesn't perform better like it does in C/C++)

oscsoftware:
:+1:

src/map/lua/lua_baseentity.cpp
```diff
@@ -8734,6 +8734,17 @@ inline int32 CLuaBaseEntity::isSpellAoE(lua_State* L)
     return 1;
 }
 
+inline int32 CLuaBaseEntity::getGender(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype == TYPE_NPC);
```
teschnei: since you're casting it to a CharEntity, you should be checking if it's not a TYPE_PC

scripts/zones/Bastok_Markets/npcs/Matthias.lua
xdemolish: stick a space between ')' and 'then', and remove any extra ' ' before the closing ')'
also, stick a space between if and parentheses so it looks like this
```
if (player:getFreeSlotsCount() == 0) then
```

scripts/zones/Bastok_Markets/npcs/Matthias.lua
```diff
@@ -15,17 +15,125 @@ require("scripts/zones/Bastok_Markets/TextIDs");
...
+			
+        player:startEvent(0x01EE, playersAFChoice -1); -- event params indexed from 0
+	
+    elseif(player:getVar("dancerTailorCS") == 5 )then
+	
+        local currentVanaDay = VanadielDayOfTheYear();
+		
+        if(player:getFreeSlotsCount() == 0) then
+		
+            player:PrintToPlayer( "Not enough free inventory space." );
```
teschnei: Quests should never be using PrintToPlayer, since retail never ever does this (it's always message packets)

teschnei:
nah, you just push to this branch again and then post here that you did (since there's no notification for a new commit, just for comments)

ProfessorCattington:
Ah OK. I thought I already pushed.

teschnei:
you did :)

ProfessorCattington:
I did it! :3

teschnei:
your editor replaced a bunch of trailing whitespace with tabs for some reason (check harmodios's diff to see what I mean), and there are a few tabs hanging around instead of spaces (matthias, for example).  if you could a) remove trailing whitespace and b) convert tabs to 4 spaces, that'd be good

scripts/zones/Bastok_Markets/npcs/Matthias.lua
```diff
@@ -46,7 +161,64 @@ end;
...
+        if(playersAFChoice == 1) then
+            dancerAFID = dancersTiara;
+        elseif(playersAFChoice == 2) then
+            dancerAFID = dancersBangles;
+        elseif(playersAFChoice == 3) then
+            dancerAFID = dancersToeshoes;
+        end;
+		
+        if(player:getFreeSlotsCount() == 0) then --check to see if the player has enough inventory space before rewarding them.
+            player:messageSpecial(ITEM_CANNOT_BE_OBTAINED, dancerAFID);
```
teschnei: if the inventory is full, the function should immediately return - or else you set the var for having completed the quest, but they won't obtain the item

ProfessorCattington: Did you mean just add a return statement after the messageSpecial call or add the return statement and move the block up in the function/event code?

teschnei: just do whatever to make sure that you can't complete the quest without getting the item

ProfessorCattington:
I went through the dancer AF scripts and removed a bunch of white space and tabs. Also changed Matthias's script per request.

What editor would you recommend for lua? I've been playing around with some I found online but I don't really like any of them and notepad++ is a bit limited.

helixhamin:
Sadly, I would still use Notepad++, I cannot think of much else that I would recommend.
What limitations are you seeing?
btw, to auto remove trailing whitespace, save with ctrl+s (only if you are using notepad++)
It shows under the "Macro" menu at the top.

teschnei:
I use vim, but notepad++ and sublimetext for example would still be fine.
Just change the settings to insert spaces instead of tabs and it'll be fine

helixhamin:
For the settings, to replace tab, select Settings -> Preferences -> Tab Settings -> make sure box showing "Replace by space" is checked.

xdemolish:
space between if/elseif and parentheses :D

aizenmyou:
Eclipse has a Lua plugin with autocompletion (can be a mixed blessing), though I have not tried it as I'm in the Vim crowd for things like scripts.

ProfessorCattington:
It would be nice to see and jump to references. I'm going to take a look at vim and the eclipse plugin. Thanks everyone.

ProfessorCattington:
placed spaces between if/elseif and parenthesis.

xdemolish:
thanks!

PR #2662, opened 2016-01-22
NPC IDs (current as of Jan 22nd)

TeoTwawki:
Looks like both of the Escha zones and Reisenjima need redone.

Also need this merged: https://github.com/DarkstarProject/darkstar-utils/pull/11


TeoTwawki:
p.s.
I've noticed mistakes tend to occur when the IDs are out of sequence in the NPC list where multiple similar NPCs appear. So i would like to inform pains in my rear that my desire to keep the IDs sequential is NOT just me being nit-picky about aesthetics.

TeoTwawki:
amending for a few borked IDs

TeoTwawki:
ok no, this is fucking bullshit, the tool tripped over every fucking cs nps/ blank/bullshit we don't even use.

PR #2663, opened 2016-01-22
NPC IDs (current as of Jan 22nd)

TeoTwawki:
redid #2662


starlightknight:
@TeoTwawki thanks for taking a shot at this - sorry I've been slow this time. Very lacking on free time this month :-(

As you noted there are deficiencies in the NPC job, that's why i say it is only "semi-auto" instead of auto - it regularly needs some manual review and tweaking.

The following are weaknesses in the job:
1) Ids not being ordered sequentially, as you noted
2) Areas of the files with large gaps in ids
3) Blanks, as there is no POLUtils data to compare against
4) POLUtils name column being wrong
5) The header comments must match the batch config or bad stuff will happen

1) can be fixed with some manual TLC doing some reordering - and a little effort to keep it ordered
2) could be fixed if we added stub npc entries for the ones we aren't using yet to match polutils
3) is handled by detecting the shift that the rest of the zone is going in - that's why it can be burned by gaps in the file, or out of order ids. The job marks each id as used as it goes along so it won't use the same on again. if the ids are out of order, and a npc with the same name is closer, it may shift to the wrong one and be unable to reuse the id later, throwing things off as it scans further for the next match.
4) is generally indicative of a id being out of sync, in which case the polutils name should reflect what that id number really belongs to
5) isn't an issue unless there is a typo when adding new zones

What i usually do is:
1) Run the job, then immediately run POLUtilsNameScraper
2) Look at the results in a diff client - check the highlighted changes
3) Any column that changed and has a mismatched POLUtils name vs name in the diff was not updated correctly. I fix those manually. It's generally obvious where the disconnect is by comparing the two in the highlighted portion
4) If a zone only has shifted blanks, i usually revert them - this is generally the combination of a gap and no data to compare against for those, and is generally a false positive. If the whole zone shifted, the blanks are generally updated correctly

This updates most things correctly in about 15-20 mins of effort. It is generally weak against battlefields because of a large amount of blanks coupled with a large amount of gaps. Buffing out the info in npc_list would be the easiest way to make those more reliable. The lua job can be run afterward the npc job and that will auto-fix conquest, helm, tomes, etc. Those npcs have reliable names and are always shifted right

TeoTwawki:
It went by pretty fast once I resequenced the IDS in the original file then redid the batch job (I started over once I saw how F'd up my 1st one went). I was a little po'd about how many IDs were out of sequence.

TeoTwawki:
hmm I forgot about the lua job entirely.

starlightknight:
@TeoTwawki yes that is part of what was taking me so long - i hadn't had time to deal with it. the file had gotten messy. 

The lua job uses a properties file that is generated when you run the npc job. it is a 1 time use. it will find/replace all the detected shifts in all the lua files, updating any long ids in any lua file that was detected as changed. 

This is what has been keeping conquest / helm / tomes etc working correctly each update for months :-)

TeoTwawki:
btw, I experimentally tried direct conversion of xml dumps into a new sql table. if it weren't for positional data being stored in the same place and mob ID's being seperated, I could fully eliminate the need for updating and just generate it.

TeoTwawki:
after I get dinner I'll add the lua changes to this. Everybody needs to stop beings cared to try this, waiting around for you like its your job or something.

starlightknight:
@TeoTwawki i thought about doing that before, but didnt for the same reason. 

But what we could potentially do, is have it place a stub row in the file that is marked invisible/disabled and then make a comment noting it as a stub that needs positional data gathered. We could even leave it commented out with a minor tweak to the npc job, and just have the job evaluate commented out lines too.

We could then make a script that fills in the gaps with stubs, then the job would see the numbers on the stubs and have that data toward shifting correctly.

TeoTwawki:
seems nothing changed that had its ID in a lua script as I got no changes in my working copy after running the batch. So this *should* be good to merge as is.

PR #2679, opened 2016-01-27
Scripts: Correct mods on Expiacion and Tachi: Rana

Hozu:
https://www.bg-wiki.com/index.php?title=Expiacion&oldid=269807
https://www.bg-wiki.com/index.php?title=Tachi:_Rana&oldid=270583

Also fixed two broken Katana WSs.

### DIFF
--- a/scripts/globals/weaponskills/blade_ei.lua
+++ b/scripts/globals/weaponskills/blade_ei.lua
@@ -30,6 +30,6 @@ function onUseWeaponSkill(player, target, wsID, tp, primary)
     end
 
     local damage, tpHits, extraHits = doMagicWeaponskill(player, target, wsID, params, tp, primary);
-    return tpHits, extraHits, damage;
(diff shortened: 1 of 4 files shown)


helixhamin:
Have you confirmed which is correct, bg or ffxiclopedia?
http://ffxiclopedia.wikia.com/wiki/Tachi%3A_Rana

They both give different STR mod values, so ffxiclopedia values were probably used.

teschnei:
bgi-wiki is the correct choice for pretty much everything nowadays - especially since the big WS changes last year

helixhamin:
Okay, got it. And looking through, I see some of the material referenced in the ffxiclopedia that goes over how they estimated at 35%.
Odd to think that it is outdated now. I used to use it for everything...

Hozu:
Check out the older version of the Tachi: Rana page - http://ffxiclopedia.wikia.com/wiki/Tachi:_Rana?oldid=1388363 35% there too.

PR #2683, opened 2016-01-28
CORE: Add new latent check

BRFord88:
Added LATENT_WEAPON_DRAWN_MP_OVER for use with the runic weapons.


xdemolish:
tabs > 4 spaces :p

BRFord88:
Sorry, my first time ever working with C++ or H.  Is this worth going back and fixing? Also is this on the line where it says LATENT_WEAPON_DRAWN_MP_OVER      = 56, part?

xdemolish:
yep, that and where you added the cases for it
in terms of functionality it won't make a difference but tabs have different sizes across editors and so generally come out ugly so we ask to replace with 4 spaces instead

BRFord88:
Ok, I see where I messed up.  I'll repull here soon.

PR #2686, opened 2016-01-28
SQL: Added item latents that required ID 56

BRFord88:
Added rune equips and 1 other that required the latent ID 56
implementation.


bendangelo:
CI failing, otherwise looks good

BRFord88:
What could be the possible cause of this?

Hozu:
Wonder if it's because the latent doesn't exist in this particular branch, since it's in the other pull.

teschnei:
Importing item_latents.sql into the database...ERROR 1136 (21S01) at line 2749: Column count doesn't match value count at row 1

teschnei:
the takeaway here is that you do actually need to compile/run/test your changes - I certainly don't have the time to do them for you

BRFord88:
I understand. I'll wait until the latent Id is merged and then try to upload this again.

PR #2707, opened 2016-01-31
Update slug_shot.lua

Swiss-y:
changed mod to correct

### DIFF
--- a/scripts/globals/weaponskills/slug_shot.lua
+++ b/scripts/globals/weaponskills/slug_shot.lua
@@ -7,7 +7,7 @@
 -- Aligned with the Aqua Gorget, Light Gorget & Breeze Gorget.
 -- Aligned with the Aqua Belt, Light Belt & Breeze Belt.
 -- Element: None
--- Modifiers: AGI:70%
+-- Modifiers: AGI:30%
 -- 100%TP    200%TP    300%TP
 -- 5.00      5.00      5.00
 -----------------------------------


whasf:
You updated the comment, not the actual code. Was that intentional?

TeoTwawki:
and the comment was already correct according to this line https://github.com/DarkstarProject/darkstar/pull/2707/files#diff-7b7af5b8b082818473988f5a1c7500a9R31

helixhamin:
Hmm, I look at this, and see:
params.agi_wsc = 0.3

The 70% appears to be here:
     if (USE_ADOULIN_WEAPON_SKILL_CHANGES == true) then
         params.agi_wsc = 0.7;

BG wiki still shows 70%, so should this comment even be changed (Although their old data, pre-Adoulin, shows 30%)? If so, how would it be handled since it has a different value depending on if it is pre-Adoulin or post Adoulin?
Maybe just add that to the comment, then it is correct?
--- Modifiers: AGI:30% (Pre-Adoulin)
--- Modifiers: AGI:70% (Post-Adoulin)

TeoTwawki:
post adoulin = current retail.
It doesn't even really matter.

helixhamin:
Noted. So is this even a valid pull request? Seems like it should remain commented as 70%, not changed to 30%.

TeoTwawki:
I wouldn't complain if someone wanted to add both, but yeah this change is pointless.

Swiss-y:
sorry, was thinking just from the server i play on specifics.

PR #2708, opened 2016-01-31
to obtain limit breaker only if Lv75+

b03605079:


xdemolish:
@teschnei

teschnei:
Need to go verify the cs..

PR #2717, opened 2016-02-02
Attempting Mission 5-2 BCNM (Shadow Lord) freezes char in CS.

dickeyf:
Fixed BCNM winning prematurely at very start.  This makes the server start another CS while client enters BCNM.  Clients says it can't start a cutscene while one is in progress, and client stays stuck until logout.

Caused by BCNM having victory condition set on mobs that spawns later on in the BCNM Fight.  The winning condition would return true because isDead on these mobs are all true (hp ==0, and status disapear).

Adding a flag to mob entity to prevent winning until those mobs did spawn.

Fixes Mission 5-2.

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -36,6 +36,7 @@
 #include "status_effect_container.h"
 #include "ai/ai_container.h"
 #include "enmity_container.h"
+#include "ai/states/death_state.h"
 
 CBattlefield::CBattlefield(CBattlefieldHandler* hand, uint16 id, BATTLEFIELDTYPE type){
 	m_Type = type;
@@ -342,7 +343,7 @@ void CBattlefield::addNpc(CBaseEntity* PNpc){
 bool CBattlefield::allEnemiesDefeated(){
 	bool allDefeated = true;
 	for(auto&& Condition : m_EnemyVictoryList){
-        if (Condition.MobEntity->isDead()){
+        if (Condition.MobEntity->PAI->IsCurrentState<CDeathState>()){
             Condition.killed = true;
 		}
 		if(Condition.killed == false){


teschnei:
move the flag to MobVictoryCondition_t, don't want to be spending kilobytes on 95% of the mobs who won't be using it

dickeyf:
Sure thing, will do the changes tonight (EST).

I don't see any obvious way to update MobVictoryCondition_t on spawn.  Any
suggestions?

I'm considering adding a onMobSpawn method to battlefield and from there,
fetch MobVictoryCondition_t for that mob, and update the flag.

teschnei:
That's probably fine.  Entities have a PBCNM pointer to get the BCNM
they're in, should make it easy.

dickeyf:
Ah, didn't see that. That'll be easy then.  Thanks!

teschnei:
actually.. if you change Condition.MobEntity->isDead() to Condition.MobEntity->PAI->IsCurrentState<CDeathState>() and revert everything else it should work like it used to

dickeyf:
Now I'm seeing issues when exiting the BCNM, either by running away, or by winning.

I don't see any CS at all, just a black screen.

I'm still able to complete other BCNMs though with no such issues.

Fiocitrine:
I have the same issue with CoP Dawn/Promathia fight. Four of my characters have clear already and one is currently on mission. The one character just blacks out while the other four go into final CS for the fight.

teschnei:
Dawn uses the same system to spawn the second phase as 5-2, so it will be
fixed when 5-2 is fixed.

dickeyf:
With this fix 5-2 now works fine until the end :
In the onBcnmLeave() LUA handler, this line runs :
player:startEvent(0x7d01,1,1,1,instance:getTimeInside(),1,0,0);

The client screen goes black, and remains so, until I do @zone ThroneRoom to unstuck myself.

Possible that the event id 0x7d01 changed to something else in a previous update?

I'm trying to find the correct event ID for the end of this battle.

teschnei:
Lets do it in another issue, so I can get this in now

Issue #2439, opened 2015-11-25
Magic Pot mobs are taking 0 magic damage all of the time.


Hozu:
Nuked a pot that was idle since boot so no magic shield effect, took 0 damage, not no effect. I wonder if a change somewhere made other mobs that are resistant to magic have stronger resists too.

teschnei:
check the console for errors (0 damage is the fallback on errors)

Hozu:
Nothing gets printed to the console despite high level spells doing 0 damage. Strange...

Scavenge:
fprint their stats?

Cloudef:
https://github.com/Cloudef/darkstar/blob/personal/scripts/commands/scan.lua You can also use this (You need to be engaged to enemy).

Hozu:
The problem is that when xdt was refactored to be out of 100 instead of 256, various things were forgotten, like mob mdt% mods. So how should xdt%s be done? I was under the impression that being out of 256 was the way to go because it's more accurate, but it was changed to be out of 100 recently.

Hozu:
This was resolved long ago.

PR #2727, opened 2016-02-06
Scripts/SQL: Update mdt-% values from shell/mob family mods

Hozu:
Since they were forgotten when xdt-% was refactored to all be out of /100 (yay for accuracy loss! If anything, it should all have been out of /256 imo). Also removed the flat damage reduction from Fealty though it's still wrong - should reduce the accuracy of any enfeebling effect, but not just the ones DSP has resistances for, and also certain non-elemental damage. Currently it's just magic eva +200. Added a check for Protectra/Shellra V for having no merits, for mobs who have the spell.

### DIFF
--- a/scripts/globals/effects/fealty.lua
+++ b/scripts/globals/effects/fealty.lua
@@ -11,7 +11,6 @@ require("scripts/globals/status");
 -----------------------------------
 
 function onEffectGain(target,effect)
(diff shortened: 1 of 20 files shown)


Hozu:
Oh right, this fixes https://github.com/DarkstarProject/darkstar/issues/2439

bendangelo:
Looks good thanks.

PR #2728, opened 2016-02-06
SQL: Add missing crafting torques.

Hozu:
I'm surprised there wasn't a console error for an item mod being added to a non-existing item...


xdemolish:
> I'm surprised there wasn't a console error for an item mod being added to a non-existing item...

@whasf you know what to do :D

PR #2732, opened 2016-02-07
Scripts/SQL/Core: Implement Bar- spell magic defense bonus

Hozu:
Bar- spells all set the subpower to 1 for whatever reason, and a search through all of the files that contain getSubPower() had nothing in regards to bar- spells. So I'm using it to hold the bonus magic defense bonus for the element, and I removed it for the status resistance spells.

Source: https://www.bg-wiki.com/bg/Category:Barspell#Increasing_Elemental_Barspell_Potency


Hozu:
Also changed the description of the force day/weather bonus mod to be what the code actually checks for. I should have wrote it like that to begin with.

Hozu:
Added a really small change - noticed that there was a spell:getElement() despite there being a previously defined variable containing that value, so I replaced it.

scripts/globals/monstertpmoves.lua
```diff
@@ -396,7 +400,12 @@ function mobAddBonuses(caster, spell, target, dmg, ele)
 
     dmg = math.floor(dmg * burst);
 
-    mab = (100 + caster:getMod(MOD_MATT)) / (100 + target:getMod(MOD_MDEF)) ;
+    local mdefBarBonus = 0;
+    if (ele > 0 and ele <= 6 and target:hasStatusEffect(barSpells[ele])) then -- bar- spell magic defense bonus, held in subid
+        mdefBarBonus = target:getStatusEffect(barSpells[ele]):getSubType();
+		print("test")
```
xdemolish: sneaky print

bendangelo: I thought bar spells don't give magic defense bonus but only a higher resist rate.

Hozu: Whoops I forgot to remove that. As for bar spells: https://www.bg-wiki.com/bg/Category:Barspell#Increasing_Elemental_Barspell_Potency

Hozu:
Removed the print that I forgot to remove. Also included the source in the commit message.

teschnei:
can you switch the barspells to use subpower instead of subid? (or even tier)
subid is kind of reserved and the core does unexpected things with it if you don't use it right

Hozu:
I asked about subpower before, since it's always being set to a specific value (1), you said to use something else. I'll switch it to tier when I get the chance. Or, why do they always have subpower 1?

teschnei:
I have no idea. check the barspell effect script, but there might be other
places

Hozu:
I searched for getSubPower and looked through every file that it was in, and none of them was looking at the barspell effect. Strange... I suppose it was used for something in the past, but that got removed. I'll change it over to subpower then.

Hozu:
Yeah I'm going to use subpower for this mod. Nothing that getSubPower is involved in is getting it from a bar- spell. It must have been from something that got removed at some point.

PR #2741, opened 2016-02-09
Fix for mission 5-2 : Client get stuck in CS after Shadow Lord is def…

dickeyf:
…eated.  Letting the mob die after despawn after 15s fixes the issue.

I'm not sure what the exact cause was, but it probably had to do with timing or order in which packets gets sent to client?


teschnei:
I don't think the BCNM ending 15 seconds after the mob dies is retail, though.  If we're gonna fix it, it should be the same as retail

dickeyf:
In retail, the mob do a death animation.  It takes about 7 seconds.  (24:27 to :24:35)  Then the cs 6 starts.
https://www.youtube.com/watch?v=qO1hf4FBtpo

I'm still trying to make the death animation work, it seems that entity_update with ANIMATION_DEATH doesn't do it ):  I checked, and we sent the entity_update with ANIMATION_DEATH twice (For every mob it gets sent twice).  The client does not animate the Shadow Lord that way it seems...

dickeyf:
If I change the CBattlefield::allEnemiesDefeated() method to use ->isDead as previous, with  && Condition.spawned, I think the previous behavior of ending BCNM right away will not hang the client anymore.  I'm giving that a try.  This is not retail as death_animation will not occur, but, we won't wait 15 seconds looking at the boss just standing still.

teschnei:
the death animation is part of dying, you would not do that manually

dickeyf:
I changed the fix to use isDead as was previously done instead of CDeathState.  I am surprised that the Client doesn't freeze with isDead() but does with CDeathState.

isDead() returns true as soon as hp == 0.  I assume that it takes a tick before the state of the mob changes to CDeathState.  Thus sending the CS one tick earlier also prevents the client lockup.

scripts/zones/Throne_Room/mobs/Shadow_Lord.lua
```diff
@@ -71,8 +71,11 @@ function onMobFight(mob,target)
         end
     else
         -- second phase AI: Implode every 9 seconds
-        if (mob:getBattleTime() % 9 == 0) then
```
dickeyf: This was evaluated twice per second.  Thus implode twice every 9s, instead of once.

dickeyf:
Also, is there a way to have Darkstart print all packets it sends to the client?  I'm looking for a way to analyse whats it being sent.  That'd help me understand more the packet system.

teschnei:
use a client side packet logger for that

PR #2742, opened 2016-02-09
Aht Urhgan Mission 6: Easterly Winds

Gravenet:
No reason to check for free slots if you aren't going to let them make another attempt at getting the item, So the assumption is you want them to be able to get the item in which case you would NOT end the mission on a failed attempt to get the item.  

If this is wrong, then you can remove the get free slots and put the mission ending back in regardless.

### DIFF
--- a/scripts/zones/RuLude_Gardens/Zone.lua
+++ b/scripts/zones/RuLude_Gardens/Zone.lua
@@ -134,8 +134,6 @@ function onEventFinish(player,csid,option)
         if (option ==1) then
             if (player:getFreeSlotsCount() == 0) then 
                 player:messageSpecial(ITEM_CANNOT_BE_OBTAINED,2184);
-                player:completeMission(TOAU,EASTERLY_WINDS);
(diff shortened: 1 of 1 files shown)


xdemolish:
pull request title/messages don't show in commit logs
in future, please include details of what's been changed in the commit message (since Update <file> doesn't really tell us a whole lot about what's been changed at a glance)

Gravenet:
No reason to check for free slots if you aren't going to let them make another attempt at getting the item, So the assumption is you want them to be able to get the item in which case you would NOT end the mission on a failed attempt to get the item.

If this is wrong, then you can remove the get free slots and put the mission ending back in regardless.

xdemolish:
commit message as in when you commit the file and are prompted to enter a commit title and brief description of what was changed (not creating pull request)

TeoTwawki:
Betcha it was done via website. github loves to make "update thing" commit messages. these are editable but most normal people would click through without even noticing.

PR #2748, opened 2016-02-10
bug fixes for RNG AF quest line.

ProfessorCattington:
Small fixes to the RNG AF quest line;

Fixed cases where items weren't showing up in the chat log properly during events.
Fixed cases where events could fire or repeat before requirements were met.
Changed the 'Old Sabertooth' encounter so players on the correct quest step can only advance if they are in the cave with the mob when it dies.
Fixed the Scavenge script to report results properly to chat log. note: action:param() doesn't seem to behave the way the script originally assumed.

### DIFF
--- a/scripts/zones/Sauromugue_Champaign/mobs/Old_Sabertooth.lua
+++ b/scripts/zones/Sauromugue_Champaign/mobs/Old_Sabertooth.lua
@@ -20,7 +20,7 @@ end;
 -- onMobDeath
 -----------------------------------
 
-function onMobDeath(mob, killer)
+function onMobDeath(mob,killer,ally)
 
     if (killer == nil) then
     
--- a/scripts/zones/Sauromugue_Champaign/mobs/Old_Sabertooth.lua
+++ b/scripts/zones/Sauromugue_Champaign/mobs/Old_Sabertooth.lua
@@ -22,25 +22,18 @@ end;
 
 function onMobDeath(mob, killer)
 
-    local players = mob:getZone():getPlayers();
+    if (killer == nil) then
     
-    for i, player in pairs(players) do
+        local players = mob:getZone():getPlayers();
+        for i, player in pairs(players) do
 
-        --check to see if the player has accepted this quest.
-        if (player:getQuestStatus(WINDURST,THE_FANGED_ONE) == QUEST_ACCEPTED) then
+            if (player:getQuestStatus(WINDURST,THE_FANGED_ONE) == QUEST_ACCEPTED) then
 
-        --check to see if the player's x,y,z coordinates are near enough to the cave where the RNG quest mob dies.
-        --point 1: 690, -15, -338
-        --point 2: 654, -9, -383
-        --these are outside the playable space but used to form a box around the cave where this quest takes place
-            local playerX = player:getXPos();
-            --local playerY = player:getYPos(); we shouldn't really care about the y axis for this
-            local playerZ = player:getZPos();
+            local playerDistance = player:checkDistance(mob);
+                if (playerDistance < 32) then
 
-            if (playerX < 690 and playerX > 654 and
-                playerZ < -338 and playerZ > -383) then
-
-                player:setVar("TheFangedOneCS", 2);
+                    player:setVar("TheFangedOneCS", 2);
+                end;
             end;
         end;
     end;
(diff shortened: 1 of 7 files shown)


teschnei:
Can you elaborate on your action:param note?

ProfessorCattington:
Oh sorry,
In the original script for scavenge we call 6 setters on action. All of those calls had syntax errors so none of them worked. I updated the calls so the ability as a whole behaves as you would expect except for action:param(). In this case, I assume the function is for reporting what item the player found with the ability but it always returns a blank to the chat log. example;
scavengereport. Unless I'm misusing action:param I'm sure what the problem is and It doesn't look like we use this function for any other ability.

teschnei:
oh yes, I rewrote the lua_action file and forgot to check if anything was using the old stuff.  Anyways, the action's param is probably overwritten by the return value.  So you can probably just remove the action:param and just return the arrow id, and I think that will work

ProfessorCattington:
Looks like you were right. The ability no properly reports the item you found to the chat log.

scripts/zones/Sauromugue_Champaign/mobs/Old_Sabertooth.lua
```diff
@@ -13,12 +13,35 @@ require("scripts/globals/status");
...
+
+        --check to see if the player's x,y,z coordinates are near enough to the cave where the RNG quest mob dies.
+        --point 1: 690, -15, -338
+        --point 2: 654, -9, -383
+        --these are outside the playable space but used to form a box around the cave where this quest takes place
+            local playerX = player:getXPos();
+            --local playerY = player:getYPos(); we shouldn't really care about the y axis for this
+            local playerZ = player:getZPos();
+
+            if (playerX < 690 and playerX > 654 and
```
teschnei: why not check the distance of the player instead? it's much more readable

ProfessorCattington: Do you mean distance from the a point within the cave? That seems reasonable assuming there's a vector magnitude function. If not I can probably just add it.

bendangelo: It should be `player:distance(x, y, z)`. Although I don't think it's implemented.

teschnei: you can probably just use the distance from the ??? (or even the tiger itself, maybe).

player:checkDistance(entity)

ProfessorCattington: I like these ideas. I will get at this when I have a chance.

ProfessorCattington:
So I tried out and tested checkDistance. it seems to work fine but then I realized there's a small part of this quest missing. I believe on live this quest will not progress unless the mob dies without aggro. The whole theme of the quest is witnessing natural death rather than killing something. I'm not seeing any checkAggro or aggro status functions that will tell me whether or not a mob is currently in an aggro state. Does such a function exist already and if not would the next best thing be use something related like current TP?

teschnei:
Don't do weird workarounds. If a function you need doesn't exist, ask for
it. There is an isengaged in the core, but I don't know if it's exposed and
whether or not it will work in onMobDeath. I'll try to remember to look
when I get home tonight

ProfessorCattington:
Looked at this a moment ago. Not sure isEngaged will work since it's returning a animation state. It seems like putting an aggro flag on battleentity and turning it on or off in the OnEngage/OnDisengage then exposing that flag would be pretty harmless.

teschnei:
No, we will not be adding another field to battleentity. I don't think a
byte for all ten thousand entities for one quest is gonna be that useful.
The most probable solution will be to check claim, I think. That's probably
what retail does.

ProfessorCattington:
Do you mean this 'PMaster' field?

teschnei:
just check if killer is nil or not - if it's not nil, then someone claimed it and killed it

ProfessorCattington:
ah ok. thanks!

scripts/zones/Sauromugue_Champaign/mobs/Old_Sabertooth.lua
```diff
@@ -13,12 +13,28 @@ require("scripts/globals/status");
...
 
 function onMobSpawn(mob)
+
 end;
 
 -----------------------------------
 -- onMobDeath
 -----------------------------------
 
-function onMobDeath(mob, killer, ally)
```
xdemolish: why did you remove ally

ProfessorCattington: That was probably just a mistake. I added back in.

PR #2751, opened 2016-02-11
Scripts: Add skill bonus to avatar pet magical skills

Hozu:
Not sure if it should go there instead of each individual skill.


bendangelo:
I like it, thanks for contributing

Issue #2756, opened 2016-02-12
ZM8 - Kam'lanaut Fight CS issue


tagban:
Seems to get stuck on black screen CS like the previously reported issues. This one doesn't seem fixed but you can eventually click "Exit" just not obvious. Seems like just CS is missing or ID got shifted.

tagban:
release command doesn't work.

dickeyf:
Just gave it a try, I'll try figure out what the exact root cause behind this.  I don't where you can click Exit though.  THis is after 5 minute wait.

tagban:
I try my best to compile bugs from whats posted on my forums. if its not something core to DSP I don't post it here, or if its past level 75 stuff, or past CoP. I had someone post under http://bnet.cc/forums/bugs-issues-reporting.40/ exactly what was going on. I believe this one is under "Zilart"

dickeyf:
I can reproduce it easily.  I'm trying to understand what exactly causes the client to freeze and will fix it once I understand.

dickeyf:
I was able to fix it by letting the BCNM end only 7 seconds after the last mob is defeated.  This is closer to retail (According to many youtube videos that I watched.  The BCNM ends 7 sec after the "is defeated" message shows up).  With the 7 seconds, the client doesn't hang (When an entity_update packet with ANIMATION_DEATH is sent with the event 32001 packet, the client hangs for some reason, the 7 seconds delay make this situation impossible to happen).

Next thing I found out :
When the first param of cs 32001 is a '1', then, the client will remain black at the end of the cs (Unless skip event is chosen) albeit I can hear the character's footsteps, and the compass moves around, but the screen is black.  I believe that the '1' means that the screen must remain black because the server will send another CS after 32001 ends.  But the ending CS wasn't scripped for ZM-8.

I will submit a PR soon with these changes after I test this against 5-2, ZM8 and a BCNM :
- battlefields with no chest will end 7 seconds after all mobs are defeated to prevent the client hang.
- Will add the ending CS to ZM8, so that ZM8 becomes like retail.

Btw, I found this issue:
https://github.com/DarkstarProject/darkstar/issues/1972

Which is very similar, we should close them both once this is fixed.

hookstar:
This is also happening on Through the Quicksand Caves...

PR #2765, opened 2016-02-16
Fix for ZM8 - Kam'lanaut Fight CS issue #2756 :

dickeyf:
End BCNM 7 seconds after winning.
Play CS 6 (The final cutscene) after ZM8 BCNM is won.

Tested against Mission 5-2, ZM8 and a BCNM (Under observation).

### DIFF
--- a/scripts/zones/Throne_Room/Zone.lua
+++ b/scripts/zones/Throne_Room/Zone.lua
@@ -63,7 +63,7 @@ end;
 function onEventFinish(player,csid,option)    
     --printf("CSID: %u",csid);
     --printf("RESULT: %u",option);
-    if (csid == 0x06) then
+    if (csid == 0x07) then
         -- You will be transported back to the entrance of Castle Zvahl Baileys
         player:setPos(378.222,-12,-20.299,125,0xA1);
     end
--- a/src/map/battlefield_handler.cpp
+++ b/src/map/battlefield_handler.cpp
@@ -155,9 +155,9 @@ void CBattlefieldHandler::handleBattlefields(time_point tick){
 					}
 				}
 				//handle time remaining prompts (since its useful!) Prompts every minute
-                auto Tremaining = std::chrono::duration_cast<std::chrono::seconds>(tick - PBattlefield->getStartTime());
+                auto Tremaining = (tick - PBattlefield->getStartTime());
 				if(Tremaining % 60 == 0s){
-					PBattlefield->pushMessageToAllInBcnm(202,(PBattlefield->getTimeLimit()-Tremaining).count()/60);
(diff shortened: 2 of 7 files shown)


src/map/battlefield_handler.cpp
```diff
@@ -174,8 +174,18 @@ void CBattlefieldHandler::handleBattlefields(time_point tick){
...
 					else{
-						ShowDebug("BCNM %i battlefield %i : Winning conditions met. Exiting battlefield.\n",PBattlefield->getID(),PBattlefield->getBattlefieldNumber());
-						PBattlefield->winBcnm();
+						// BCNMs with no treasure chest ends around 7 seconds after winning.
+						if (!PBattlefield->won()){
+							ShowDebug("BCNM %i battlefield %i : Winning conditions met. Waiting 7 seconds before exiting battlefield.\n",
+									PBattlefield->getID(), PBattlefield->getBattlefieldNumber());
+							PBattlefield->win();
+						} else {
+							if (std::chrono::duration_cast<std::chrono::seconds>(tick - PBattlefield->getWinTime()).count() > 6){
```
teschnei: if (tick - PBattlefield->getWinTime() > 6s)

teschnei:
what about bcnms where victory isnt a kill (ouryu, tenzen)?

dickeyf:
Hmmm... I didnt think about those.  I can take a look tomorow (WHM level
cap too fits here I think)

dickeyf:
Added a clause that ends the BCNM right away when victory condition is met but mobs didn't die.  Tested on Ouryu BCNM

scripts/zones/Throne_Room/Zone.lua
```diff
@@ -63,4 +63,8 @@ end;
 function onEventFinish(player,csid,option)    
     --printf("CSID: %u",csid);
     --printf("RESULT: %u",option);
+    if (csid == 0x06) then
+        -- You will be transported back to the entrance of Castle Zvahl Baileys
+        player:setPos(378.222,-12,-20.299,125,0xA1);
```
teschnei: the cutscene should move you without the position needing to be set manually

dickeyf: I just tried and the CS moves to the entrance of Throne room.  It won't zone you at the entrance of Castle Zvahl Baileys.  I don't CS can zone you, only move you.  According to original script, we want to zone.

dickeyf:
I just tried and the CS moves to the entrance of Throne room. It won't zone you at the entrance of Castle Zvahl Baileys. I don't CS can zone you, only move you. According to original script, we want to zone. I also just fixed the CS id, should have been 7,not 6.

Issue #1491, opened 2015-05-29
Non retail behavior of signet staves


TeoTwawki:
All 3 signet staff have same issues.

http://ffxiclopedia.wikia.com/wiki/Republic_Signet_Staff
- The Signet effect gained from this enchantment will overwrite Sanction and Sigil.
- The enchantment can only grant Signet to Bastokan characters.
- The enchantment can only be used in areas subject to Conquest.
- The enchantment can only be used on characters in the caster's party.
- ~~The duration of Signet gained from this item is 5 hours.~~ _fixed duration in #1536_

I currently cannot use republic staff on a party member when we are both Bastokan and standing in Labyrinth of Onzozo which is a Conquest zone, ~~the duration is incorrect,~~ and it doesn't remove sanction or sigil. I don't think any checking is even done for players nation or the zone/region type for conquest right now.

TeoTwawki:
Oh..And the animation is wrong, and I just discovered I can target it on myself if and only if I select myself before going into the item menu.

xdemolish:
i take it this was fixed in #1536?

TeoTwawki:
@xdemolish No, not completely. I had edited the original post I made to reflect the part I fixed (its crossed out and referenced the pull request).

TeoTwawki:
These should be fairly easy fixes for anyone with the time to play with it.

helixhamin:
For this and the removal of sigil and sanction, can it be fine similar to the guards?

`target:delStatusEffect(EFFECT_SIGIL);
        target:delStatusEffect(EFFECT_SANCTION);
        target:delStatusEffect(EFFECT_SIGNET);
        target:addStatusEffect(EFFECT_SIGNET,0,0,18000);`

TeoTwawki:
@helixhamin yup

helixhamin:
@TeoTwawki Noted. I will see what I can do. I have been away too long and want to get back to trying to learn and contribute. And sadly, next week I will be bored out of my mind, so good time to try to help give back.

TeoTwawki:
@helixhamin  I had forgotten all about this till github pinged me on it.

helixhamin:
Eh, stuck waiting for my work notebook to have C formatted and the OS reinstalled, so I was just browsing the old issues.

helixhamin:
One quick question @TeoTwawki to make sure I fix these properly, since the same problem is present in all staves.
I see this code:
[...]

PR #2771, opened 2016-02-18
Fix the three signet staffs to match retail-well, more than before...

helixhamin:
Restrict to player using and target being of the proper nation, and removal of Sanction and Sigil.

### DIFF
--- a/scripts/globals/items/federation_signet_staff.lua
+++ b/scripts/globals/items/federation_signet_staff.lua
@@ -19,6 +19,12 @@ end;
 -----------------------------------------
 
 function onItemUse(target)
-    target:delStatusEffect(EFFECT_SIGNET);
-    target:addStatusEffect(EFFECT_SIGNET,0,0,18000);
+    if (player:getNation() == WINDURST)
+        if (target:getNation() == WINDURST)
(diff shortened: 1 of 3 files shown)


helixhamin:
https://github.com/DarkstarProject/darkstar/issues/1491

scripts/globals/items/federation_signet_staff.lua
```diff
@@ -19,6 +19,12 @@ end;
 -----------------------------------------
 
 function onItemUse(target)
-    target:delStatusEffect(EFFECT_SIGNET);
-    target:addStatusEffect(EFFECT_SIGNET,0,0,18000);
+    if (player:getNation() == WINDURST)
```
xdemolish: did you test?

helixhamin: Need to wait until I get home. At work right now. Will report back once I have.

helixhamin:
@xdemolish Yeah, you are right, it fails. Completely. I can get it to remove sigil/sanction/signet, and apply a new signet, but only by removing the if statements. And it will not buff others, just the person using, so not sure it is any better than before.
The logic is lost on me on how to apply the restrictions. I should just close.

PR #2774, opened 2016-02-19
Added removal of sigil and sanction to signet staff activation

helixhamin:
Tested this and it works. Until I get better at the logic, I cannot completely retail the staffs, but at least they are one step closer.


bendangelo:
Thanks for contributing

PR #2792, opened 2016-02-22
Scripts/SQL: Jump/High Jump/DRG AF helm mods, High Jump enmity loss fix

Hozu:
-Force Jump/High Jump crit and "enhances wyvern breath" mod (latter removes direct ID checks)
-High Jump is supposed to remove total enmity, not just enmity gained on the ability itself. Also it doesn't get the Jump attack bonus effect.
-Corrected High Jump enmity modifier value on certain equipment (it shouldn't be negative).
-Corrected Jump/High Jump/Super Jump distances as per BG wiki.


bendangelo:
Thanks for contributing

PR #2809, opened 2016-02-23
SQL: Equalize currency drops across non-NM Dynamis Beastmen

Hozu:
Made them all have 6 singles at 15% each. Some zones like Glacier and Qufim had them dropping only a single single, at either 9% or 1%. Some Goblins were dropping up to 9 singles, and I kind of messed up on the Kindred last time.


bendangelo:
Thanks

PR #2825, opened 2016-02-26
Core: Fix limbus time chests

Hozu:
Not sure when addTimeToSpecialBattlefield was changed to seconds instead of minutes, but I suppose this will do.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -7220,7 +7220,7 @@ inline int32 CLuaBaseEntity::addTimeToSpecialBattlefield(lua_State *L)
 
     DSP_DEBUG_BREAK_IF(PZone->m_BattlefieldHandler == nullptr);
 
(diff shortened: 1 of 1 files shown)


teschnei:
Probably an accident. I converted them all to chrono library which is
explicit, so sometimes I had to guess whether the original was seconds or
minutes

Hozu:
Should I change it to minutes in the core then? I assumed the change there was intentional.

teschnei:
Minutes I suppose makes more sense here. Should be an easy fix (change the
cast from chrono milliseconds to chrono seconds)

Hozu:
https://github.com/DarkstarProject/darkstar/blob/6c95f022c6b16128d0cdfdb849dc9df01649d203/src/map/lua/lua_baseentity.cpp#L7223

It's already in seconds - I'm guessing it used to be working in minutes. That's why I only needed to multiply by 60 in the scripts there.

Hozu:
There that works. Dynamis timers already working properly.

teschnei:
Sorry, I meant the other way around

Hozu:
Wait, do you mean change the scripts to use seconds for the time extension? I don't think there's any kind of extension that doesn't use minutes.

teschnei:
does that answer your question

PR #2829, opened 2016-02-28
Core: Remove Detectable Effects 

### LINKED ISSUE #2806: Sneak is not removed when attacked by a mob
Client Version (type /ver in game) :
30160203_0

Server Version (type @revision in game) :
3a852af

Source Branch (master/stable) :
master

Additional Information (Steps to reproduce/Expected behavior) :
Should be removed by any action a mob takes on the player, or any time a player takes an action on a mob.

MalRD:
Resolve #2806 by removing detectable effects in OnCastFinished and OnAttack.

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -1263,8 +1263,8 @@ void CBattleEntity::OnCastFinished(CMagicState& state, action_t& action)
         }
     }
 
-    if (this != PActionTarget && PActionTarget->objtype == TYPE_PC 
-        && !PSpell->isBuff() && !PSpell->isHeal() && !PSpell->isNa())
+    // TODO: Pixies will probably break here, once they're added.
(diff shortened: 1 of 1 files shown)


src/map/entities/battleentity.cpp
```diff
@@ -1263,6 +1263,14 @@ void CBattleEntity::OnCastFinished(CMagicState& state, action_t& action)
         }
     }
 
+    if (this != PActionTarget && PActionTarget->objtype == TYPE_PC
```
teschnei: just check if their allegiances are not the same, and you can skip the isbuff/isheal/isna, I think

Hozu:
Should any status with EFFECTFLAG_ATTACK be removed as well in those same checks?

teschnei:
depends if attack is on receiving an attack or an outgoing attack

PR #2830, opened 2016-02-28
SQL: Add drops to Dynamis - Qufim NMs

Hozu:
The three non-beastmen ones. Drop data from ffxidb.


sql/mob_droplist.sql
```diff
@@ -32767,6 +32767,97 @@ INSERT INTO `mob_droplist` VALUES (4556,0,6019,34); -- end of foret mobs
 INSERT INTO `mob_droplist` VALUES (4557,0,18767,97);
 INSERT INTO `mob_droplist` VALUES (4558,0,15942,128);
 INSERT INTO `mob_droplist` VALUES (4559,0,16185,220);
+INSERT INTO `mob_droplist` VALUES (4560,0,5903,926);
```
bendangelo: Why is the rate so high for this and others?

Hozu: http://ffxidb.com/items/5903

That's what FFXIDB has. They're items that weaken the mega boss, though they're not scripted... I added them because similar items are on the NMs in Dynamis - Buburimu.

bendangelo:
Cool, thanks.

PR #2849, opened 2016-03-02
BST stay command now works per retail.  The amount healed may need to be

lenonk:
BST stay command now works per retail. The amount healed may need to be …

tweaked, but as far as I can tell from research, it is at least close to
retail.

### DIFF
--- a/scripts/globals/abilities/stay.lua
+++ b/scripts/globals/abilities/stay.lua
@@ -14,6 +14,10 @@ require("scripts/globals/status");
 -----------------------------------
 
 function onAbilityCheck(player,target,ability)
+    if (player:getPet() == nil) then
+        return MSGBASIC_REQUIRES_A_PET,0;
+    end
+    
     return 0,0;
 end;
 
@@ -21,6 +25,30 @@ end;
 -- onUseAbility
 -----------------------------------
 
-function onUseAbility(player,target,ability)
-    player:petStay();
+function onUseAbility(player,target,ability,action)
+    pet = player:getPet();
+    
+    if (pet == nil or pet == '') then
+        return
+    end
+    
+    if (not pet:hasPreventActionEffect()) then
+      -- reduce tick speed based on level. but never less than 5 and never
+      -- more than 10.  This seems to mimic retail.  There is no formula
+      -- that I can find, but this seems close.
+        level = 0
+        if (player:getMainJob() == JOB_BST) then
+            level = player:getMainLvl()
+        elseif (player:getSubJob() == JOB_BST) then
+            level = player:getSubLvl()
+        end
+        
+        tick = 10 - math.ceil(math.max(0, level / 20))
+        --printf('tick: %d', tick)
+        pet:addStatusEffectEx(EFFECT_HEALING, 0, 0, tick, 0)
+        -- I'm not sure how to set the pet animation to NONE or HEALING
+        -- and I can't find an example.  This results in the perpetuation 
+        -- of the walking/moving animation if you use stay while the pet is moving.
+        --action:animation(pet:getId(), ANIMATION_HEALING)     
+    end
 end;
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -8343,24 +8343,6 @@ inline int32 CLuaBaseEntity::uncharm(lua_State* L)
     return 0;
 }
 
-/************************************************************************
-*                                                                       *
-*   Makes your pet stay put                                             *
-*                                                                       *
-************************************************************************/
-
-inline int32 CLuaBaseEntity::petStay(lua_State *L)
-{
-    if (m_PBaseEntity != nullptr)
-    {
-        if (m_PBaseEntity->objtype != TYPE_MOB)
-        {
-            petutils::MakePetStay((CBattleEntity*)m_PBaseEntity);
-        }
-    }
-    return 0;
-}
-
 inline int32 CLuaBaseEntity::getObjType(lua_State *L)
 {
     DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
@@ -8682,7 +8664,7 @@ inline int32 CLuaBaseEntity::setSpellList(lua_State* L)
 
     DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
 
-    mobutils::SetSpellList((CMobEntity*)m_PBaseEntity, lua_tonumber(L, 1));
+    mobutils::SetSpellList(static_cast<CMobEntity *>(m_PBaseEntity), lua_tonumber(L, 1));
 
     return 0;
 }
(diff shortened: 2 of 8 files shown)


Altalus:
Mind a suggestion ? Perhaps the amount healed should be set by the lua an not hard coded.

teschnei:
Hi, 
as Altalus said, all abilities need to be done via script unless there is a very, very good reason that's not possible (I've just spent the last 4 months refactoring the AI to get rid of abilities done in the core)

lenonk:
That's cool.  I looked at the existing lua, saw how it was implemented, and assumed there was a reason for it.  Would you like me to reimplement this entirely in lua?

teschnei:
Yes, that'd be great

lenonk:
Ok, it's all in lua now except the one part in pet_controller.  I'm a lua noob though.  This is actually the first lua I've done.  Please feel free to point out any issues.

src/map/lua/lua_baseentity.cpp
```diff
@@ -8697,6 +8679,16 @@ inline int32 CLuaBaseEntity::hasSpellList(lua_State* L)
     return 1;
 }
 
+inline int32 CLuaBaseEntity::hasPreventActionEffect(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB && m_PBaseEntity->objtype != TYPE_PET && 
+        m_PBaseEntity->objtype != TYPE_PC);
```
xdemolish: TYPE_NPC

src/map/lua/lua_baseentity.cpp
```diff
@@ -8697,6 +8679,16 @@ inline int32 CLuaBaseEntity::hasSpellList(lua_State* L)
...
+
+    lua_pushboolean(L, ((CMobEntity*)m_PBaseEntity)->StatusEffectContainer->HasPreventActionEffect());
```
xdemolish: static_cast<CBattleEntity*>(m_PBaseEntity)->StatusEffectContainer->HasPreventActionEffect());

src/map/lua/lua_baseentity.cpp
lenonk: Thanks, good catch

src/map/lua/lua_baseentity.cpp
lenonk: This should probably be changed in other places as well then, since I copied this code from....hasSpellList() I think.

src/map/lua/lua_baseentity.cpp
```diff
@@ -8692,11 +8674,21 @@ inline int32 CLuaBaseEntity::hasSpellList(lua_State* L)
...
+    lua_pushboolean(L, (static_cast<CMobEntity *>(m_PBaseEntity))->SpellContainer->HasSpells());
 
     return 1;
 }
 
+inline int32 CLuaBaseEntity::hasPreventActionEffect(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB && m_PBaseEntity->objtype != TYPE_PET && 
+        m_PBaseEntity->objtype != TYPE_NPC);
```
xdemolish: sorry i meant just have it hit if it's TYPE_NPC instead of checking each type

lenonk: Ah, I gotcha.  That makes sense.

src/map/lua/lua_baseentity.cpp
xdemolish: yeah a lot of the old code still uses C-style casts, will prolly get changed eventually

src/map/lua/lua_baseentity.cpp
lenonk: Fixed

src/map/lua/lua_baseentity.cpp
```diff
@@ -8692,7 +8674,17 @@ inline int32 CLuaBaseEntity::hasSpellList(lua_State* L)
...
-    lua_pushboolean(L, ((CMobEntity*)m_PBaseEntity)->SpellContainer->HasSpells());
+    lua_pushboolean(L, (static_cast<CMobEntity *>(m_PBaseEntity))->SpellContainer->HasSpells());
+
+    return 1;
+}
+
+inline int32 CLuaBaseEntity::hasPreventActionEffect(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_NPC);
```
xdemolish: == TYPE_NPC
since npcs (ie those in cutscenes) shouldn't have a status effect

!= would have it hit for everything apart from npc

lenonk: Wait.  This code should only take effect on NPCs.  Pets, Mobs, etc.  Therefore we should DEBUG_BREAK_IF type != NPC.  To use == TYPE_NPC would make this code only run on PCs.  Where am I confused.

teschnei: An NPC in the server is an entity that has no battle functions and can only
be interacted with. Mobs and pets aren't of type_npc

lenonk: Ok, then it was right the way I wrote it the first time.  It needs to change back to:
 DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB && m_PBaseEntity->objtype != TYPE_PET);

teschnei: Just check if it's not an NPC. That's all you need.

teschnei: If it is*, sorry. NPCs don't have status effect containers, everything else
does

xdemolish: applies to players too though? (battle entity)

lenonk: Sure, I guess it probably should.  I was just thinking of it in the context of the code I was writing, but I suppose this function could be useful for players as well.  I’ll change it to DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype == TYPE_NPC);

lenonk: Ok, I think that mess is finally sorted out.  Sorry for the miscommunication.  It's fixed now.

src/map/lua/lua_baseentity.cpp
```diff
@@ -8692,7 +8674,17 @@ inline int32 CLuaBaseEntity::hasSpellList(lua_State* L)
...
+
+    return 1;
+}
+
+inline int32 CLuaBaseEntity::hasPreventActionEffect(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype == TYPE_NPC);
+
+    lua_pushboolean(L, (static_cast<CMobEntity *>(m_PBaseEntity))->StatusEffectContainer->HasPreventActionEffect());
```
xdemolish: CBattleEntity

lenonk: Ah, yes.  If we're going to change the check to work on all battle entities, the cast to be to any battle entity as well.

Fixed

xdemolish:
looks good! thanks for contributing :)

teschnei:
sorry - it wasn't good yet... there's a bunch of global variable declarations in the script

scripts/globals/abilities/stay.lua
```diff
@@ -21,6 +25,30 @@ end;
 -- onUseAbility
 -----------------------------------
 
-function onUseAbility(player,target,ability)
-    player:petStay();
+function onUseAbility(player,target,ability,action)
+    pet = player:getPet();
```
xdemolish: whoops missed this, should be local

lenonk: Can you explain to me exactly what that means?  Like I said, I’m a lua noob, and I thought that *was* local.

 

Does it need to be declared like this?

local pet = player:getPet();

teschnei: If you declare a variable without the local keyword, it's a global.

bendangelo: Yes it has to be defined as `local pet` or else it's defined as a global variable. Which I think it stupid on luas part.

lenonk: Ok, I’ll make it local, commit, and do a new PR.

 

Thanks!

PR #2854, opened 2016-03-03
Noticed some mob names longer than max size allows in mob_skill_lists…

TeoTwawki:
Example: _[Murk-Veined_Baneberry](http://ffxiclopedia.wikia.com/wiki/Murk-Veined_Baneberry)_ is longer than the current limit of 20 when we follow our usual naming convention, though not by much. I've bumped the field to 40 because I don't feel like searching to find what the longest name in existence in game is, and changed it to  varchar like is use din other tables (according to mysql documentation char is intended for fixed length text).


bendangelo:
Thanks

PR #2858, opened 2016-03-04
The "pet" variable in stay.lua is now local

lenonk:
As requested, the "pet" variable in stay.lua is now local.


teschnei:
what about level? and tick?

lenonk:
You’re right, my bad.

lenonk:
Local variables fixed.

bendangelo:
Thanks

PR #2861, opened 2016-03-06
Added test suite

bendangelo:
Hey so I spent some setting up a test suite.
## Reasoning

The challenge with managing a code base this big is things break, and often simple things. By writing unit tests (isolated simple tests) we can ensure all the nuts and bolts work before the server boots up.

Test failures will show up on CI as well.

I so far have one test for a simple method. This test suite will really shine when we can test the enmity controller in complex scenarios or test alliances with fake players.
## TODO
- [ ] Add flag to skip tests
- [ ] Add tests for enmity container
- [ ] Setup lua test suite
- [ ] Add lua tests for Herald (aka the devil)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,7 +7,7 @@ include sources.am
 ## Set up all of our dependency flags
 SRC_ALL                 = $(SRC_DARKSTAR)
 CFLAGS_ALL              = $(DARKSTAR_CFLAGS)
-CPPFLAGS_ALL            = $(DARKSTAR_CPPFLAGS)
+CPPFLAGS_ALL            = $(DARKSTAR_CPPFLAGS) @DARKSTAR_TESTS@
 CXXFLAGS_ALL            = $(DARKSTAR_CXXFLAGS)
 LIBS_ALL                = $(DARKSTAR_LIBS)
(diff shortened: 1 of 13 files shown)


bendangelo:
I think adding virtual is a small price to pay to ensure things work. The only other option would be to refactor to a point where mocking isn't needed, but can only take you so far.

teschnei:
you're not seriously going to suggest making every function we want to test virtual - it's not happening unless you only do it with a testing compiler flag

bendangelo:
It's only for stubbing out methods. Unfortunately stubbing has to happen for some tests. I think I can make a macro that'll only make that method (or any method) virtual when building with tests. That would resolve the issue.

bendangelo:
Ok so next discussion, should the test suite be enabled by default? Meaning tests get compiled and ran on every bootup.

I want to say yes but it should be off by default, the tests would only be useful for developers. In that case  someone will have to enable it for CI and write documentation on how to enable / disable it.

teschnei:
absolutely not, the test suite should be completely invisible to a non-developer
we are most likely only going to run it in CI

whasf:
Thanks, this will be revisited in the future.

PR #2862, opened 2016-03-06
Scripts/SQL: Ix'Aern DRG pet level/drop rate/spawn point corrections

Hozu:
-Increased the level of the spawned pets (there's no data on them, but 65-68 is definitely too low)
-Changed the rate of Vice/Deed from 50/50 to 25/75 respectively
-Corrected the mobid check for setting the spawn in one of the rooms.


bendangelo:
Good stuff thanks

PR #2870, opened 2016-03-06
Core/SQL: Implement Magic Damage Taken -% II

Hozu:
Applied after the standard -50% cap, it caps out at a total of -87.5%.
https://www.bg-wiki.com/bg/Damage_Taken#Magic_Damage_Taken

Updated all existing Aegis entries in the database, and added any missing mods that they should have had (defense mostly). Also removed the comment about MDT-% being /256 because that hasn't been the case for a while now. Also removed the comment about MDT-% being /256 because that hasn't been the case for a while now.


Hozu:
Getting pretty annoying that AppVeyor decides to cancel building whenever a new commit is made.

PR #2874, opened 2016-03-07
updated fame values based on San d'Oria's NPC.  …

waterlgndx:
Seems that SE reduced fame requirements by 1/4th, this is based on the responses given by Namonutice in Southern San d'Oria.

### DIFF
--- a/conf/map_darkstar.conf
+++ b/conf/map_darkstar.conf
@@ -65,6 +65,9 @@ fov_party_gap_penalties: 1
 fov_allow_alliance: 1
 vanadiel_time_offset: 0
 
+#For old fame calculation use .25
+fame_multiplier: 1.00
+
 #Percentage of experience normally lost to keep upon death. 0 means full loss, where 1 means no loss.
 exp_retain: 0
 
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -25,7 +25,6 @@
 #include "../../common/timer.h"
 #include "../../common/utils.h"
 #include "../../common/kernel.h"
-
 #include <math.h>
 
 #include "lua_baseentity.h"
@@ -3903,7 +3902,7 @@ inline int32 CLuaBaseEntity::getFame(lua_State *L)
 
     uint8  fameArea = (uint8)lua_tointeger(L, 1);
     uint16 fame = 0;
-
+    float fameMultiplier = map_config.fame_multiplier;
     CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
 
     switch (fameArea)
@@ -3911,16 +3910,16 @@ inline int32 CLuaBaseEntity::getFame(lua_State *L)
         case 0: // San d'Oria
         case 1: // Bastok
         case 2: // Windurst
-            fame = PChar->profile.fame[fameArea];
+            fame = PChar->profile.fame[fameArea]*fameMultiplier;
             break;
         case 3: // Jeuno
-            fame = PChar->profile.fame[4] + ((PChar->profile.fame[0] + PChar->profile.fame[1] + PChar->profile.fame[2]) / 3);
+            fame = PChar->profile.fame[4] + ((PChar->profile.fame[0] + PChar->profile.fame[1] + PChar->profile.fame[2]) * fameMultiplier / 3);
             break;
         case 4: // Selbina / Rabao
-            fame = (PChar->profile.fame[0] + PChar->profile.fame[1]) / 2;
+            fame = (PChar->profile.fame[0] + PChar->profile.fame[1])*fameMultiplier / 2;
             break;
         case 5: // Norg
-            fame = PChar->profile.fame[3];
+            fame = PChar->profile.fame[3]*fameMultiplier;
             break;
     }
     lua_pushinteger(L, fame);
@@ -3943,21 +3942,21 @@ inline int32 CLuaBaseEntity::getFameLevel(lua_State *L)
     uint16 fame = (uint16)lua_tointeger(L, -1);
     uint8  fameLevel = 1;
 
-    if (fame >= 2450)
+    if (fame >= 613)
         fameLevel = 9;
-    else if (fame >= 2200)
+    else if (fame >= 550)
         fameLevel = 8;
-    else if (fame >= 1950)
+    else if (fame >= 488)
         fameLevel = 7;
-    else if (fame >= 1700)
+    else if (fame >= 425)
         fameLevel = 6;
-    else if (fame >= 1300)
+    else if (fame >= 325)
         fameLevel = 5;
-    else if (fame >= 900)
+    else if (fame >= 225)
         fameLevel = 4;
-    else if (fame >= 500)
+    else if (fame >= 125)
         fameLevel = 3;
-    else if (fame >= 200)
+    else if (fame >= 50)
         fameLevel = 2;
 
     lua_pushinteger(L, fameLevel);
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -1047,6 +1047,10 @@ int32 map_config_read(const int8* cfgName)
         {
             map_config.vanadiel_time_offset = atoi(w2);
         }
+        else if (strcmp(w1, "fame_multiplier") == 0)
+        {
+            map_config.fame_multiplier = atof(w2);
+        }
         else if (strcmp(w1, "lightluggage_block") == 0)
         {
             map_config.lightluggage_block = atoi(w2);
--- a/src/map/map.h
+++ b/src/map/map.h
@@ -114,6 +114,7 @@ struct map_config_t
     int8   CoP_Battle_cap;            // Default is 0. Disable/enable old lv caps on Chains of Promathia mission battles.
     uint8  max_merit_points;          // global variable, amount of merit points players are allowed
     uint16 yell_cooldown;             // Minimum time between uses of yell command (in seconds).
+    float  fame_multiplier;           // Fame multiplier
     bool   audit_chat;
     bool   audit_say;
     bool   audit_shout;


Hozu:
Oh so this is why people were having issues with starting Gobbiebag quests...

dazusu:
inb4 you can cap fame in 2-3 quests.

bendangelo:
Yeah let's put a toggle on this. I wouldn't mind seeing a generic `enable_75_era` to handle all these cases. Then people can just switch it on or off.

waterlgndx:
The San d'Oria npc, Namonutice, returns a value and the client reads it so if you're running the current client version you should divide the value returned by getFame() so that the client displays the proper fame message.  The other fame npcs seem to have player events for each.  (Either that or that event packet needs another look lol).

waterlgndx:
Is there anything I should toggle for `enable_75_era`?  is this what you had in mind?  I know it's not in the lua   Are there admins locked into a certain version of ffxi and should I include logic to check client version to see how to talk to the client (as far as which fame number should be returned by `getFame()`)?

helixhamin:
But, where does the toggle chose one set of values and the other? Also, it has nothing to with client version. It should check that toggle, then select either the newer values or the older values.

waterlgndx:
It has a lot to do with the client version... sandy fame just returns the value instead of the level of fame.   So I could set it to the values but no one would be able to accurately check their fame in Sandy which I'm ok with because I can fix it for myself.  Either that or we have the packets wrong somewhere.  So instead I think we should divide the fame value by 4 instead of changing which values to choose.

waterlgndx:
Anyway to answer your question directly, instead of changing the values that the fame (points?) are compared to (that is, dividing the fame values compared to by 4 if the toggle is enabled).  I instead divide the fame points the person has before checking them against the values.  

So that's what this block does...
`uint16 fameDivisor;`
 `map_config.enable_75_era ? fameDivisor = 4 : fameDivisor = 1;   //75 era needs corrected`

and then the math was updated to reduce by that factor if `enable_75_era` is `true`:
`fame = PChar->profile.fame[fameArea]/fameDivisor;`
`fame = (PChar->profile.fame[4] + ((PChar->profile.fame[0] + PChar->profile.fame[1] + PChar->profile.fame[2]) / 3))/fameDivisor;`
`fame = (PChar->profile.fame[0] + PChar->profile.fame[1]) / (fameDivisor*2);`
`fame = PChar->profile.fame[3]/fameDivisor;`

Pros: 
- Admins don't have to configure in both the settings.lua and in the map config.
- Fame points are divided by 4 if `enable_75_era` is enabled without impacting the amount of already accrued fame.  So this way it accounts for the new fame values while not breaking the fame NPC in Sandy.  (Granted to fix that with the other toggle method there'd be a config in both settings.lua and in map config so the script knows when to divide by 4 unless we throw in a new function for that one call.... I dunno)

Cons:
- It does obscure the code a little so it may become more difficult to read.
- It impacts the fame rate acquisition when `enable_75_era` is `false` because the fame multiplier is now in reference to the current retail setting with this setting.  This may be undesirable.

Anyway I'm open to suggestions, does anyone have any thoughts on this matter?  Also this is my first merge pull request (ever lol)... so if there's anyway I can be more helpful let me know!  Maybe I could just clarify this code better with comments?

xdemolish:
@teschnei

teschnei:
Sorry, it'll take a while to get on this.  I don't know how fame works at all

TeoTwawki:
Thought:

Rather than an on/off switch, it could be adjusted with a multiplier default to 1.00 (100% of current retail), then nuke the fame the bloat settings out of [setting.lua](https://github.com/DarkstarProject/darkstar/blob/aa839e59371d03cb2416e607f7ae1d3f5edc3e0c/scripts/globals/settings.lua#L53) (who even makes just one nation get more fame/why do these even exist as separate settings.lua entries?) since server who want faster or slower (old school) fame can just dial it/up down in the map conf then to adjust the what amounts equal which tier rather than adjusting fame gained per quest.

That's 5 lines less crud in settings.lua replaced by 1 new line that was going in the map conf anyway, some code that was already often forgotten and left out of quest scripts, and one less thing for scriptwriters to try and remember when they make new quests.

TeoTwawki:
That divisor of 4 would just become a setting of 0.25 instead of a bool.
Hows that sound?

TeoTwawki:
edited my 2 comments into one, for you email readers out there.

teschnei:
Math looks right.  I don't like the idea of the "enable 75 era" switch - there is waaay too many things that would fit in that category - and nobody would ever know exactly what it does.  Teo's idea is a good one.
You wouldn't need to touch the add fame functions with this multiplier - the idea is likely that they eased fame requirements on retail (or else why would they randomly change the tier requirements?)

TeoTwawki:
I'm willing to handle the settings.lua removal of the old settings separately myself, btw. Or you can. Or it can even stay. Up to you guys.

teschnei:
actually.. not sure if we can just do a multiplier.  The fame levels must match the tiers given by the client, or it quests will not be given out when the appropriate fame level is reported by the NPC

TeoTwawki:
@teschnei this is how the NPC currently works https://github.com/DarkstarProject/darkstar/blob/169f2a31a9bf6772555150acd0d0719615b8d74d/scripts/zones/Port_Bastok/npcs/Flaco.lua#L23 its just spitting back the level the server says to decide what value to pass into the client (convert servers fame points into fame tier) so fame points times fame multiplier to decide that tier would work, since what this PR does not is change the fame amounts that trigger each tier (and I am assuming this PR is retail accurate in those amounts as it is right now)

TeoTwawki:
The client decides which text to display based on the number fed into that cs.

waterlgndx:
It's important to note though that the sandy npc gets the raw fame value stored in the server.  This could be corrected by simply dividing the returned value by 4.  I'm not sure if there are other instances where the client receives this in raw form. https://github.com/DarkstarProject/darkstar/blob/169f2a31a9bf6772555150acd0d0719615b8d74d/scripts/zones/Southern_San_dOria/npcs/Namonutice.lua

TeoTwawki:
That npc does the same thing as the one I just linked. And that's not the raw fame value, its not calculating anything at all. `getFameLevel` returns the tier, so that NCP is literately spitting out whatever tier you tell it too. :smile:

waterlgndx:
look again 
`player:startEvent(0x001f, player:getFame(SANDORIA));`

TeoTwawki:
Then NPC scritps are wrong and the PR doesn't address that: see Flaco linked above.

waterlgndx:
I don't think the NPC scripts are wrong... I tried to change it to use getFameLevel instead and didn't have any luck.

TeoTwawki:
I'm checking myself in game.

TeoTwawki:
Ok, I'm done peeking, and while there are indeed 2 methods for the cutscenes (jeuno and sandy do same thing, while bastok and windy do the other..even for SE, this shocks me at how stupid it is), if you stick the multiplier where you have the divisor, its accomplishes the exact same thing but adjustable. In both cases it'd adjust what is passed to both types of events. So if sticking that divisor in there didn't pose any problems, neither will this:
`fame = PChar->profile.fame[fameArea]*fameMultiplier;`
0.25 = 1/4
The results of getFame() would change, and getFameLevel is calling that. The question is: what else calls getFame, and would your divisor or my multiplier break it? https://github.com/DarkstarProject/darkstar/search?utf8=%E2%9C%93&q=getFame
NOPE! looks like we're golden.

helixhamin:
So, maybe instead of a level_75_era flag, turn it into a divisor flag for those who want more customizability?

waterlgndx:
Ok it should now be ready lol

TeoTwawki:
Looks legit. Can you squash commits?

waterlgndx:
I'm not sure what squash commits means.  So maybe??

TeoTwawki:
Collapse it into a single commit, is what it means. If you've ever done a rebase, you basically know how already.

teschnei:
We still can't just guess on the new fame amounts per tier.  It has to be consistent with the client.  I don't really care if the old values are off, that is not my problem.  But the default values must be correct and consistent.

TeoTwawki:
@teschnei my bad I thought he had the new retail values already

waterlgndx:
How would you suggest getting the correct values?  I mean I went through and checked at each point lowered it by one (noticed that it resulted in lowering the fame tier by one).  I suppose this is guessing and checking but maybe there's a better way to confirm these values?  Or do you mean the work that needs to be done in order to acquire a new tier?

teschnei:
If you tested each tiers exact limits  with the San dorian cutscene, then
that's right

waterlgndx:
Ok I'll go through and double check this.  I want to be sure :)

TeoTwawki:
@waterlgndx I don't suppose you have a link to where SE announced their change? I can't seem to find it.

waterlgndx:
I think their phrasing was relaxed fame... it was really vague.  I'll check and see.

waterlgndx:
well there's this... I'm not sure if this is the one or not... http://forum.square-enix.com/ffxi/threads/40059-Feb-18-2014-%28JST%29-Version-Update?p=493162&highlight=fame#post493162

TeoTwawki:
Thanks.

dazusu:
So the fame requirement for 'certain quests' was reduced; i.e., you might have needed rank 5 fame previously, but now you only need rank 3. It also states only 'certain' quests - so it's not a global thing.

Would this not warrant further investigation before a global multiplier/divisor is added for all fame?

Fame per tier doesn't appear to have changed (given SE's wording). Just that a lower tier is now required.

TeoTwawki:
@dazusu they then go on to possibly imply its every quest that needed fame - not all quests do. Its classic SE: be as vague as possible. If he's seeing teh NPC in sandy change its cs result, then that implies they did adjust the tiers though I'm surprised that there is no info on either wiki regarding it almost 2yrs later..

regardless, once he compares against that NPC we'll have current retail.

waterlgndx:
I've verified these tiers of fame with sandy npc.

dazusu:
So the only way to maintain the older more grindy fame, is to divide the raw fame value before it's compared and assigned a rank. gotcha.

waterlgndx:
@dazusu yes that is correct.  You will probably want to use .25 as the fame multiplier as opposed to the 1.00 default.  If anyone gets this incorrect they'll just have to take down the server, update the value, and restart it and then they should be golden.  A side effect might be that someone who managed to start a quest they wouldn't have normally been able to start may not be able to turn that quest in until they meet the proper fame requirement.  I think in most cases the npc would just let them turn it in.

waterlgndx:
@TeoTwawki ok I believe I have squashed it into 1 commit like tomatoes... while making spaghetti sauce.  @teschnei I have verified the fame values against the sandy npc.  Is this thing actually ready now?  lol

TeoTwawki:
_aaawriiiight_ :smile:

PR #2881, opened 2016-03-07
Core/SQL/Scripts: Refactor elemental affinity + implement elemental affinity accuracy modifer for Charm

Hozu:
-Removed the check for the affinity that reduces the power/acc/perp of the spell/avatar, and added negative mods on the appropriate items. Example, Fire Staff has +2 fire damage/acc/perp mods, and -2 ice damage/acc/perp mods. Iridial/Chatoyant Staff simply has +2/+3 to the mods of every element instead of the (now removed) all affinity damage/acc/perp mods.
-+10%/15% Charm for NQ/HQ Light/Iridal and Apollo's/Chatoyant staff respectively. Making the reasonable assumption that Dark/Pluto's staff reduces accuracy.
https://www.bg-wiki.com/bg/Charm


teschnei:
i wonder why we never just checked positive multipliers and gave the elemental staves negative affinity modifiers for their ascendant element.

then we wouldn't even need a case for chatoyant/iridal...

Hozu:
Probably for the eventual implementation of the magian staves - that's why I separated the three types of elemental affinity at least (damage, acc, and perp reduction).

teschnei:
that doesn't really make a difference - they'd also just have a negative mod for the ascendant element

Hozu:
Oh I must have interpreted what you were saying... you mean for Chatoyant/Iridal staves having all 8 elemental affinities instead of a single all affinity mod?

teschnei:
yeah. And fire staff having +fire affinity and -water affinity (or ice or whatever it is) as mods - and the scripts would only check the matching element instead of both matching and ascendant (ie. water spell would see -water affinity from wearing a fire staff instead of trying to subtract fire affinity itself)

xdemolish:
@teschnei

teschnei:
I want to do what was discussed before I merge this

Hozu:
Done and tested, works as expected.

PR #2882, opened 2016-03-08
SQL: Add Store TP mod for Hachiman Hakama (+1)

Hozu:
I guess somebody missed that mod.


Hozu:
http://www.ffxiah.com/item/15392/hachiman-hakama
http://www.ffxiah.com/item/15394/hmn-hakama-1

bendangelo:
Thanks

Issue #2867, opened 2016-03-06
AbsorbTP and Reverse Flourish report the wrong amount of TP drained


lenonk:
**_I have:_**
- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160203_0

**_Server Version_** (type `@revision` in game) **:**
e99fb5c

**_Source Branch_** (master/stable) **:**
All

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
These two skills, and probably anything else that drains TP, report the actual amount drained divided by 10 in the resulting message.  It's because the addTP() method in lua_baseentity.cpp multiplies all tp gains by 10.  

I'll be happy to fix it, but before I do, I'd like to know if there is a reason why addTP() multiplies it's parameter by 10.  If there's not a good reason for it, then probably the best thing to do is to remove that feature and modify all lua scripts that call addTP() to drain 10x more.

lenonk:
nvm, I understand why it's the way it is.  On retail TP goes from 0-300, and is considered to be in terms of %,  Here, damage is 0-3000.  So technically the message is correct I guess.

helixhamin:
Because retail does that now.
http://ffxiclopedia.wikia.com/wiki/Tactical_Points
	
Abbreviation: TP

Note: TP displayed on a scale of 0 to 300% prior to June 2014. It now displays on a scale of 0 to 3000 (no % sign) to allow more precise display of TP values. 1000 TP is required to initiate a weapon skill.

lenonk:
Ah, nice, so the functions that multiply * 10 should probably be changed then.

Hozu:
You could probably just have the returned value be multiplied by 10, at the very end. Assuming it's supposed to always give you an amount that's a multiple of 10. Can it give like, 1234 TP return on retail?

lenonk:
Yeah, you could.  I tested that and it works, but it seems unnecessarily hackish.  The current behavior is highly non-intuitive for those writing scripts.

Hozu:
It can't take more TP than the mob has, as it specifically checks for that.

https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/spells/absorb-tp.lua#L43

lenonk:
[...]

PR #2883, opened 2016-03-08
Fix issue #2867

lenonk:
Removed the *10 and /10 from addTP(), getTP(), delTP() and setTP(), and tracked down every occurrence in all the lua files and changed them accordingly.

I tested everything that was more than a trivial and straightforward change, and compared the behavior before my change, to the behavior after.

I probably still broke something...


scripts/globals/abilities/heel.lua
```diff
@@ -26,6 +26,12 @@ end;
 -----------------------------------
 
 function onUseAbility(player,target,ability)
+    local pet = player:getPet()
+
+    if (pet == nil or pet == '') then
```
TeoTwawki: The onAbilityCheck() already fails if there isn't a pet.

scripts/globals/abilities/high_jump.lua
```diff
@@ -52,7 +52,7 @@ function onUseAbility(player,target,ability,action)
     if (tpHits + extraHits > 0) then
         -- Under Spirit Surge, High Jump reduces TP of target
         if (player:hasStatusEffect(EFFECT_SPIRIT_SURGE) == true) then
-            target:delTP(damage * 2)
+            target:delTP((damage * 2) / 10)
```
TeoTwawki: target:delTP(damage * 0.2) just seems more natural to me.

scripts/globals/abilities/stay.lua
```diff
@@ -46,9 +46,6 @@ function onUseAbility(player,target,ability,action)
         local tick = 10 - math.ceil(math.max(0, level / 20))
         --printf('tick: %d', tick)
         pet:addStatusEffectEx(EFFECT_HEALING, 0, 0, tick, 0)
-        -- I'm not sure how to set the pet animation to NONE or HEALING
-        -- and I can't find an example.  This results in the perpetuation 
-        -- of the walking/moving animation if you use stay while the pet is moving.
-        --action:animation(pet:getId(), ANIMATION_HEALING)     
+	pet:setAnimation(0)
```
TeoTwawki: indent is off

TeoTwawki:
Thanks for also fixing that bad compare operator and the entirety of whatever the hell was going on in absorbTP (`math.random()` called 3 times back to back, what the..?)

scripts/globals/abilities/stay.lua
TeoTwawki: Oh that's weird..Github previews tabs differently on the commit page than it does here. It looks like a 2 space indent there, looks like 4 here. Didn't realize it was a tab. They both used to display tabs as 8.

scripts/globals/abilities/heel.lua
lenonk: True, but don't you create a race condition if you don't check in both places?

scripts/globals/abilities/high_jump.lua
lenonk: Ugh, you're right.  I did that in other places.  I guess I was tired when I did that one.

scripts/globals/abilities/heel.lua
teschnei: No - onAbilityCheck and onUseAbility are done in the same tick

lenonk: Oh, awesome.  I was afraid of the pet going away between the two.  Good to know!

lenonk:
RE: absorbTP, I had the same reaction.  What the hell is going on here?  Anyway, I addressed your comments.  Thanks for looking at the diff so closely.

Hozu:
Nice, but you missed many weaponskills. All of the ones from this old pull need to be modified: https://github.com/DarkstarProject/darkstar/pull/2779/files

I think there might be more as well.

teschnei:
he didn't change the tp going into weaponskills afaict (luautils)

lenonk:
Correct.  Those should work exactly as they did before my change.  I only changed the data provided to lua scripts when one of addTP(), getTP(), delTP(), or setTP() is called.  Those scripts don't use any of those, so they didn't change.

teschnei:
however, you should probably change them, because now it's really inconsistent

lenonk:
I'll be happy to, but:
1) Should it be a different issue/pull request?
2) How exactly should I change them?  TP is passed into those skills as a parameter.  I assume it's passed in on the 0-300% scale and you're want to make them work on the 0-3000 scale.

teschnei:
1) no
2) https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/luautils.cpp#L3021
I don't want half the scripts to use 0-300 and the other half to use 0-3000, it's just confusing for no reason

lenonk:
I totally agree with the confusing part.  The fact that it was confusing already was a big reason why I started down this path.  Ok, I'll get to work on converting those scripts.  It will probably take a couple of days.

teschnei:
in reality there aren't that many places that require changes - the fTP functions in weaponskill.lua, and any WS that uses TP as a modifier for the duration of an effect (additional effect WS, aftermath, etc).
you can probably just grep for tp in scripts/globals/weaponskills and look for lines not calling a weaponskill.lua function

lenonk:
I already changed the fTP functions in weaponskill.lua.  So these are probably broken now if they depend on those.  Told you I probably broke something!

lenonk:
Ok, I went through every file in scripts/globals/weaponskills/ and modified them as necessary.  Also, I added an applyAftermathEffect() function to weaponskills.lua to replace all the repetitive aftermath related code.

scripts/globals/weaponskills.lua
```diff
@@ -959,3 +959,61 @@ function takeWeaponskillDamage(defender, attacker, params, finaldmg, slot, tpHit
...
+	player:addStatusEffect(EFFECT_AFTERMATH_LV2, 52, 0, 180, 0, spower_2);
+    elseif ((tp >= 2801) and (tp <= 2900)) then
+	player:addStatusEffect(EFFECT_AFTERMATH_LV2, 56, 0, 180, 0, spower_2);
+    elseif ((tp >= 2901) and (tp <= 2999)) then
+	player:addStatusEffect(EFFECT_AFTERMATH_LV2, 59, 0, 180, 0, spower_2);
+
+    -- AFTERMATH LEVEL 3
+
+    elseif ((tp == 3000)) then
+	player:addStatusEffect(EFFECT_AFTERMATH_LV3, 45, 0, 120, 0, spower_3);
```
TeoTwawki: A tab snuck back in. We'll learn him!

lenonk: Already fixed.  I don't know where they came from, it was copy and paste, so maybe it was already there.

xdemolish:
wow aftermath was a shitload of duplicate code, thanks for cleaning that up :D

scripts/globals/weaponskills.lua
lenonk: btw, I don't use tabs.  We *hates* them precious.  So either something in one of my editors needs tweaking, or they're getting copied and pasted.

scripts/globals/weaponskills.lua
```diff
@@ -959,3 +959,61 @@ function takeWeaponskillDamage(defender, attacker, params, finaldmg, slot, tpHit
...
 end
+
+function applyAftermathEffect(player, tp, spower_1, spower_2, spower_3)
+    spower_1 = spower_1 or 1
+    spower_2 = spower_2 or 1
+    spower_3 = spower_3 or 1
+
+    -- AFTERMATH LEVEL 1
+
+    if ((tp >= 1000) and (tp <= 1100)) then
```
teschnei: can you fix this abomination while you're here? it doesn't need to check every 100 tp for a linearly scaling function...

lenonk: So, I started to fix that, but then I noticed that the power param of addStatusEffect is not linear.  For level 1, it goes from 10-19, incrementing by 1, then jumps to 24 for the first 100 tp of level 2, and increases by 4 for evey level until the last one, where it only increases by 3.  Then, the level 3 aftermath effect then goes backward and applies a power level of 45, instead of something greater than 59.  Instead of trying to guess what the intention was, I decided to leave it alone.  If someone can explain what's going on there, I'll do my best to fix it.

lenonk: Also, for some reason, the duration on the the level 3 effect is 120, where all the others are 180.  Is that intentional?

TeoTwawki: I know `spower_2 = spower_2 or 1` is legit lua syntax but actually seeing it sure throws me for a loop. :dizzy:

lenonk: Do you have a preferred way to default those parameters to applyAftermathEffect()?  Again, I'm a lua noob, so I'm learning as I go.

Hozu: https://www.bg-wiki.com/bg/Mythic_Aftermath Maybe that will help?

lenonk: Thanks Hozu, I'll take a look at it.

TeoTwawki: `utils.clamp()` might be applicable there. But if what you have works its not really a problem (pretty sure that works as you have it, I'm just not used to seeing it done like that),

teschnei: they are linear in their 1000 ranges (1000-2000, 2000-3000).  They jump between ranges because the different tiers of aftermaths have different effects.  However, I would like to point out that they are not always the same values (melee weapons AM2 is 20-60 attack, but tupsimati is 20-30 MAB)

teschnei: both the ranges and the durations need to be variable anyways because of the higher level mythic weapons

teschnei: and "x = x or 1" is the preferred way in lua

TeoTwawki: Noted.

lenonk: Ok, so if they're linear within their ranges, then is it a bug that level 2 increments by 4 at every level except the last, where it increments by 3?

Hozu: I'm more inclined to believe it was improper implementation.

lenonk: Ok, have a look at applyAftermathEffect() now and see what you think.

lenonk: Wait, I think I screwed soemthing up on the commit

lenonk: Yeah, I missed a file.  It's right now

scripts/globals/weaponskills.lua
```diff
@@ -959,3 +959,22 @@ function takeWeaponskillDamage(defender, attacker, params, finaldmg, slot, tpHit
 
     return finaldmg;
 end
+
+function applyAftermathEffect(player, tp, power, subpower, duration)
+    power = power or {10, 24, 45}
```
teschnei: 10, 20, 45

scripts/globals/weaponskills/death_blossom.lua
```diff
@@ -45,64 +46,11 @@ function onUseWeaponSkill(player, target, wsID, tp, primary)
...
-
---        AFTERMATH LV3
-
-        elseif ((tp == 300)) then
-            player:addStatusEffect(EFFECT_AFTERMATH_LV3, 45, 0, 120, 0, 1);
-
-            end
+    	printf('damage: %d', damage)
+        if (damage > 0) then 
+            applyAftermathEffect(player, tp, {10, 24, 45}, {2, 3, 1});
```
teschnei: also 10,20,45

scripts/globals/weaponskills.lua
```diff
@@ -959,3 +959,22 @@ function takeWeaponskillDamage(defender, attacker, params, finaldmg, slot, tpHit
...
+    subpower = subpower or {1, 1, 1}
+    duration = duration or {180, 180, 120}
+
+    -- AFTERMATH LEVEL 1
+
+    local apply_power = 0
+    if (tp == 3000) then
+        player:addStatusEffect(EFFECT_AFTERMATH_LV3, power[3], 0, duration[3], 0, subpower[3])
+    elseif (tp >= 2000) then
+        apply_power = power[2] + (4 * ((tp - 2000) / 100))
```
teschnei: this is not always 4 (it's 1 for omniscience, death blossom and vidohunir)

scripts/globals/weaponskills.lua
```diff
@@ -959,3 +959,22 @@ function takeWeaponskillDamage(defender, attacker, params, finaldmg, slot, tpHit
...
+        player:addStatusEffect(EFFECT_AFTERMATH_LV2, apply_power, 0, duration[2], 0, subpower[2]);    
+    elseif (tp >= 1000) then
+        apply_power = power[1] + ((tp - 1000) / 100)
+        player:addStatusEffect(EFFECT_AFTERMATH_LV2, apply_power, 0, duration[1], 0, subpower[1]);
```
TeoTwawki: Got _LV2 twice.

lenonk: Issues fixed, and I'm much happier with this anyway.

scripts/globals/weaponskills.lua
```diff
@@ -959,3 +959,62 @@ function takeWeaponskillDamage(defender, attacker, params, finaldmg, slot, tpHit
...
+    if (tp == 3000) then
+        player:addStatusEffect(EFFECT_AFTERMATH_LV3, params.power.lv3, 0, 
+	    params.duration.lv3, 0, params.subpower.lv3)
+    elseif (tp >= 2000) then
+        apply_power = params.power.lv2 + (params.power.lv2.inc * ((tp - 2000) / 100))
+        player:addStatusEffect(EFFECT_AFTERMATH_LV2, apply_power, 0, 
+	    params.duration.lv2, 0, params.subpower.lv2);    
+    elseif (tp >= 1000) then
+        apply_power = params.power.lv1 + (params.power.lv1.inc * ((tp - 1000) / 100))
+        player:addStatusEffect(EFFECT_AFTERMATH_LV1, apply_power, 0,
```
lenonk: Damnit!  Where the eff are these tabs coming from???

TeoTwawki: What editor do you use? There is usually a setting you can use to enforce spaces and indent width. You can also use editorconfig if your editor supports it (notpad++ and visual studio both have plugins for it)

lenonk: I'm currently using ZeroBrane Studio and vim, both under Windows.  I don't *usually* do any coding work at all in Windows, so nothing is really configured.  It's probably vim.  I just haven't taken the time to figure out where to put a .vimrc in Windows. :)

TeoTwawki: Vim has editorconfig support, so it can auto handle some formatting issues like those tabs for you.

lenonk: I've got a .vimrc along with a ton of other dotfiles that I keep on github.  I just need to figure out where to put it.  Easy enough.  I should do that now...

teschnei: It just goes in your user folder. C:\users\username. Gvim will look for
both .vimrc and _vimrc

lenonk: Yeah, I tried that and it didn't work.  It's probably because I'm not using gvim.  I'll figure it out. :)

scripts/globals/weaponskills.lua
```diff
@@ -959,3 +959,62 @@ function takeWeaponskillDamage(defender, attacker, params, finaldmg, slot, tpHit
...
+        params = initAftermathParams()
+    end
+
+    local apply_power = 0
+    if (tp == 3000) then
+        player:addStatusEffect(EFFECT_AFTERMATH_LV3, params.power.lv3, 0, 
+        params.duration.lv3, 0, params.subpower.lv3)
+    elseif (tp >= 2000) then
+        apply_power = params.power.lv2 + (params.power.lv2_inc * ((tp - 2000) / 100))
+        player:addStatusEffect(EFFECT_AFTERMATH_LV2, apply_power, 0,
```
lenonk: This needs to be:
math.ceil(((tp - 2000) / 100))

The reason is, if the power is supposed to be 20-60, and you start at 20, then:
20 + 4 * ( 2999 - 2000) / 100) == 20 + 36 == 56

This is the original code started at 24, I think, which in turn, is why I had this code starting at 24.  If you use math.ciel(), anything over 2950 should result in:

20 + 4 * math.ceil((2999 - 2000) / 100)) == 20 + 40 == 60

Thoughts?

scripts/globals/weaponskills.lua
```diff
@@ -959,3 +959,62 @@ function takeWeaponskillDamage(defender, attacker, params, finaldmg, slot, tpHit
...
+    if (tp == 3000) then
+        player:addStatusEffect(EFFECT_AFTERMATH_LV3, params.power.lv3, 0, 
+        params.duration.lv3, 0, params.subpower.lv3)
+    elseif (tp >= 2000) then
+        apply_power = params.power.lv2 + (params.power.lv2_inc * ((tp - 2000) / 100))
+        player:addStatusEffect(EFFECT_AFTERMATH_LV2, apply_power, 0, 
+        params.duration.lv2, 0, params.subpower.lv2);    
+    elseif (tp >= 1000) then
+        apply_power = params.power.lv1 + (params.power.lv1_inc * ((tp - 1000) / 100))
+        player:addStatusEffect(EFFECT_AFTERMATH_LV1, apply_power, 0,
```
lenonk: Same as above.

scripts/globals/weaponskills.lua
teschnei: no, all that does is basically revert starting from 24 (the only time you could ever hit 20 is if you were at exactly 2000 tp).  Chances are it's really 20-59, but "20-60" is just for simplicity.  
I think the step increments should be smooth rather than jump by inc, though. (2975 tp = 59 attack).
applied_power = power + ((tp - 2000) / (100 / inc))

lenonk: math.ceil() will round up .1 to 1?  I thought it would only round up if it was .5 or greater.  I guess what I really wanted was math.round(), which doesn't exist.  I like your solution better anyway.  I'll implement and commit it shortly.

lenonk:
Smooth scaling on aftermath implemented.

xdemolish:
need to rebase, @teschnei can merge once it's mergeable :D

lenonk:
Well, I just screwed that up royally....

I'll fix it...

lenonk:
There, that's better.  No idea what happened in the first rebase

PR #2885, opened 2016-03-09
Added skilllist for Dea ZNM

Fiocitrine:
Dea didn't have a skill list so I created one.


Hozu:
I don't think any of the Wivre moves are scripted? I know one of them can't be done at the moment, the one that gives damage reduction while the attacker is in front.

Fiocitrine:
I'll test them first...

Fiocitrine:
I guess they don't work. It's not even trying to use them.

PR #2905, opened 2016-03-13
Scripts/SQL/Core: Proper implementation of Quick Draw bonus damage

Hozu:
-Repurposed the existing mod to be used for AF(+1) hat.
-Added a new mod for a percentage boost for Death Penalty, converted the old mod to new one.
-Removed the existing mod from things it shouldn't have been on.

I noticed there's MOD_QUICK_DRAW_TRIPLE_DAMAGE but nothing actually checks for it. I have no idea how it works so I left it out.


teschnei:
isn't this just "added quick draw percent damage bonus"? it doesn't look like you changed how MOD_QUICK_DRAW_DMG work

Hozu:
I didn't - the issue was that AF(+1) hat didn't have that mod when it was supposed to, while the mythic gun had it erroneously.

PR #2911, opened 2016-03-15
Scripts: Set initial spawn time for Duke Decapod

dazusu:
Forgot to do this in last pull


dazusu:
@xdemolish fixes my previous checkin, which caused Duke Decapod to never pop.

PR #2937, opened 2016-03-25
Core/SQL: Fix Quick Draw recast merits

Hozu:
Was some real weirdness going on with it missing, was possible to use different element Quick Draws infinitely, without regards to charges. Works properly now with and without those merits.


Hozu:
This was working properly before, I guess it was broken when Sic was added.

teschnei:
I'll need to review this later. This shouldn't be right

Hozu:
Actually I'm not so sure why the Quick Draw ability itself had a recast of 199 seconds (or if that even did anything), but it seems that right now it's checking if the ability is a charge type, and not Sic, and then it doesn't apply the merit reduction. Right now it doesn't work, and with this change it does. Still, if you want it to be done differently, that's fine.

teschnei:
No.. It should already have the merited duration when it gets the charges.
Otherwise the recast container doesn't have the right value.

Hozu:
Oh, I see what you mean. Hm...

Hozu:
https://github.com/DarkstarProject/darkstar/blob/4b5fe131ca077a809382a2c3f9f5869a1d8024bf/src/map/ability.cpp#L495

Should it be in there somewhere?

teschnei:
I'm out of town, I'll check on Sunday or Monday what it's supposed to do.
It was pretty recent where I redid this

Hozu:
Any word on this? I just realized the only reason my change worked is because the merit mod ID is set both in abilities and abilities_charges.

teschnei:
Remind me in irc

PR #2938, opened 2016-03-26
Updated Item_Weapon

Ninjistix:
fixed some delays and added missing h2h weapons


bendangelo:
Do you know how to squash this?

Ninjistix:
nope, how?

TeoTwawki:
Also you tested?

Ninjistix:
yeah tested them

Ninjistix:
figured out how to rebase commits, will prob have to do it again in a few days...

bendangelo:

Ninjistix:
thanks a lot, will work on adding/fixing more during the week

tagban:
Thanks Ninji!

R3P0FFXI:
@Ninjistix @bendangelo can we be sure this is correct, I dont remember which as its been a while since anyone talked about H2H delays but was the base delay taken into account I thought H2H was always a lower delay due to the delay associated with not having a weapon equipped and just using bare knuckles. If this is the case then they may have been previously correct at the 300 delays they were at before.

Ninjistix:
I was told by kj that all h2h start wth 480+ whatever delay is on the
weapon , so nothing should be under 480 for h2h

tagban:
I'm guessing 480 is the default delay IE 0 on a weapon itself..?

TeoTwawki:
Probably 480 is what your delay is with nothing at all equipped, before any reduction from martial arts etc.

PR #2950, opened 2016-03-28
Fixed hyakinthos GroupID so it'll spawn

Fiocitrine:
just changed the GroupId to the one matching in mob_group.sql


bendangelo:
thanks :heart:

PR #2953, opened 2016-03-29
Fei'yin shadow NM's pop mechanics

ymkc:
Fixed the Northern/Eastern/Southern/Western Shadow pop type according to wikia:
random pop from PH after 10h/16h
added NM's spawn points
added missing Northern Shadow
shortened PH repop time from 16min to 5min


Fiocitrine:
Like what you've done! Bon travaille!

bendangelo:
Thanks

PR #2958, opened 2016-03-30
updated blu cast time

Swiss-y:
to match retail as of update 1235 oct 2010, Triumphant Roar was already modified to the update prior to me doing the rest


bendangelo:

Issue #2075, opened 2015-09-14
TOAU mission Variable name


Turttle:
TOAU Mission crated one variable for each mission for each player. As mentioned on the IRC chat the variable "TOAU#" needs to be replaced for "TreasureStatus" and be standardized. Ofc players with active mission or mid mission needs to be fixed with @setplayervar command as suggested by @TeoTwawki.

Issue #1894, opened 2015-08-07
Someone Broke TOAU Mission 1


R3P0FFXI:
Date & Time: august 4th 2015
FFXI Client Version: 30150619_2
Area & Coordinates: Aht Urghan Whitegate
Party or Solo: Party
NPC Name: Naja Salaheem
Monster Name: not applicable

Steps To Reproduce The Issue:
Talked with Naja Salaheem to start mission 1. Got cutscene approaching her door, telling me i have received a "Supplies Package" Checked my KI and nothing, and the NPC will not tell me anything. I tried talking to the NPC Mayeevaat I think the name is , at Ilrusi Atoll Staging Point , didnt say anything. Used the runic portal to return to whitegate.

I had tried last night to use the Ilrusi Atoll Staging Point to return home forgetting that i didnt have access yet, and it told me It can't let me pass.

waytim:
Not sure if related or coincidence but I had someone stuck on this earlier today, they had not completed the Road to Aht Urgan quest. They had never went back to Faursel after being warped into Wajaom Woodlands. After completing the quest they were able to get the KI from the cs.

TeoTwawki:
@waytim even if that were all that is wrong, that quest isn't even part of the mission and shouldn't be included - the mission is pre-flagged on retail on new characters because the "mission" literally begins in whitegate regardless of how you got there (if you manage to cheat your way there on retail, that cs would still fire off). 

All the appropriate places in script to prevent you from boating over there ahead of time etc. should already work correctly leaving zero reason for anyone to check that in mission. So if that's the case, someone still messed up.

TeoTwawki:
Just a brief glance at the involves scripts and it becomes clear that someone took a previously working mission and made changes that they did not test. There is code in the wrong functions, unnecessary additional player variables, variables attempting to be cleared in onRegionLeave, and breaks with normal DSP convention by using new vars for every. single. mission. **even when that mission is literately only a single unavoidable cs that you can't escape without getting it completed.**

TeoTwawki:
I caved in to temptation and ranted. Sue me.

waytim:
This mission is a super pain in the butt at the moment because if you aren't standing in the right spot on CS trigger, it will keep looping the CS over and over. These changes must be the reason why it's suddenly been acting up the past month or so.

TeoTwawki:
@waytim Yep.

The loop is from this:
[...]

PR #2960, opened 2016-03-31
Fix some of the issues with multiple ToAU scripts

TeoTwawki:
Which have lain broken for over 8 months. Ever since 6667e96

I and 3 other ppl have run brand new characters all the way to 15th mission with these changes.

The diff may be hard to read but I'd be happy to explain in detail (sarcasm, but I actually will do it).

### DIFF
--- a/scripts/globals/missions.lua
+++ b/scripts/globals/missions.lua
@@ -1,21 +1,21 @@
 -----------------------------------
---  Areas ID
+-- Areas  ID     mission step var
 -----------------------------------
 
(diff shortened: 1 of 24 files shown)


TeoTwawki:
P.S.
This fixes both #2075 and  #1894

TeoTwawki:
Amended because I spotted yet another variable that didn't need to go to the database - used for only 1 cs to tell which selection you already saw (cs ends when you have seen both).

PR #2968, opened 2016-04-01
Scripts: Make Weaponskills check for DA/TA from merits.

Hozu:
Since the mods for DA/TA don't include merits.


Hozu:
Updated to include Quad attack and corrected multi-hit logic.

scripts/globals/weaponskills.lua
```diff
@@ -845,41 +849,44 @@ end;
...
+    -- QA/TA/DA can only proc on the first hit of each weapon or each fist
+    if (attacker:getOffhandDmg() > 0 or attacker:getWeaponSkillType(SLOT_MAIN) == SKILL_H2H) then
+        multiChances = 2;
     end
 
-    for i = 1, numHits, 1 do
-        chance = math.random();
-        if (chance < tripleRate and i <= tripleChances) then
+    for i = 1, multiChances, 1 do
+        if (math.random() < quadRate) then
```
teschnei: you should only need one math.random - if its between 0 and QA, its a quad, if its between QA and TA + QA, its a TA, if its between TA + QA and QA + TA + DA its a DA, otherwise its a single

Hozu: Oh, so you're saying, if you had let's say... 5% QA, 10% TA, and 15% DA, it'd be 0-4 is QA, 5-14 is TA, and 15-29 is DA?

Hozu: Done with that change.

PR #2969, opened 2016-04-01
SQL: Correct a couple of Dynamis mob currencies

Hozu:
Thanks to @deadwing for bring this to my attention. PLD was dropping a jadeshell instead of bronzepiece, BRD dropped jadeshell/100 byne instead of bronzepiece/1 byne respectively, also had improper rates.


bendangelo:

PR #2970, opened 2016-04-01
Item_Weapon Updates

Ninjistix:
added some more weapons, will start adding the rest next week


sql/item_weapon.sql
```diff
@@ -3975,12 +3975,54 @@ INSERT INTO `item_weapon` VALUES (21056,'tomonari',10,0,0,0,0,2,1,420,118,0);
 INSERT INTO `item_weapon` VALUES (21057,'tomonari_+1',10,0,0,0,0,2,1,407,119,0);
 INSERT INTO `item_weapon` VALUES (21058,'shichishito',10,0,54,54,42,2,1,450,126,0);
 INSERT INTO `item_weapon` VALUES (21059,'ichijinto',10,0,0,0,0,2,1,420,121,0);
+INSERT INTO `item_weapon` VALUES (21060,'mjollnir',11,0,242,242,228,3,1,308,160,0);
 INSERT INTO `item_weapon` VALUES (21061,'mjollnir',11,0,242,242,228,3,1,308,160,0);
-INSERT INTO `item_weapon` VALUES (21062,'yagrush',11,0,242,242,228,3,2,267,133,0);
-INSERT INTO `item_weapon` VALUES (21063,'yagrush',11,0,242,242,228,3,2,267,133,0);
+INSERT INTO `item_weapon` VALUES (21062,'yagrush',11,0,242,242,228,3,3,267,133,0);
```
TeoTwawki: if I am not mistaken, that field is the max hit count for multi hit weapons. Yagrush occ. attacks twice but only as part of its aftermath effect, so that should be one.

TeoTwawki: p.s. 
You seem to be merging dsp into your fork to update. Set DSP as a remote and just pull from DSP master instead, and then you won't get all these merges in your history showing on your pull requests. 

It also pays to use branches other than master to make pull requests with, and leave master 100% unchanged from stock DSP, since by default new branches are 1:1 copies of your master branch. This is how I so often have 3-4 pull requests open at once with different things in them.

Ninjistix: as for the multi-hit i was following how the other weapons were listed, and the ones that were changed from 2-3 is because the weapon stated "twice or thrice"

TeoTwawki: All of them are wrong, both before and after you touched them.

That isn't part of the weapon. Look at the item help-text:

Yagrush

That is in the aftermath effects, not the items basic stats.

Ninjistix: Okay then I'll have to go over all of them when I get a chance unless
someone corrects them first

PR #2971, opened 2016-04-01
End ceaseless complaints about Misison 2-3 when bastokan player visits windurst first

TeoTwawki:
condition check was looking for the same var it sets in the event finish..


TeoTwawki:
...gdi, how do half these ppl have var 7 and half 8, yet are on same effing step of that same mission. Don't merge yet, this is getting an amend

Hozu:
wat. I was thinking that the check would probably have to be missionstatus == 7 or missionstatus == 8 because of something like that.

TeoTwawki:
turns out this PR is correct, I just have idiots for testers -_-

Issue #2977, opened 2016-04-02
BRD songs are able to be interrupted by attacks


Hozu:
**_I have:_**

- [X] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [X] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160203_0

**_Server Version_** (type `@revision` in game) **:**
https://github.com/DarkstarProject/darkstar/commit/8f2765e5f9a29ac5f9bd75cb2da6de36a610a8bc

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
https://www.bg-wiki.com/bg/Spell_Interruption_Rate says physical damage interrupting spells does not apply to songs.

Hozu:
Resolved with that merged pull.

PR #2979, opened 2016-04-02
Core/Scripts/SQL: Spell interrupt adjustments

Hozu:
-Changed Aquaveil from increasing interrupt rate to a counter that prevents interrupts while it remains. Also added the modifier for the gear that currently exists in the DB. The interrupt rate setting was modified to be the base counter so people can change the amount of hits it will absorb if they'd like.
-Made songs unable to be interrupted by physical attacks.

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1562,6 +1562,13 @@ namespace battleutils
             return false;
         }
 
+        // Songs cannot be interrupted by physical attacks.
+        if ((SKILLTYPE)PSpell->getSkillType() == SKILL_SNG)
+        {
+            // ShowDebug("Is song, interrupt prevented.\n");
+            return false;
+        }
+
         //Reasonable assumption for the time being.
         int base = 40;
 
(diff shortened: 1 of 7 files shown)


Hozu:
Oh right, this will resolve https://github.com/DarkstarProject/darkstar/issues/2977

src/map/utils/battleutils.cpp
```diff
@@ -1562,6 +1562,29 @@ namespace battleutils
...
 
+        // Songs cannot be interrupted by physical attacks.
+        if ((SKILLTYPE)PSpell->getSkillType() == SKILL_SNG)
+        {
+            // ShowDebug("Is song, interrupt prevented.\n");
+            return false;
+        }
+
+        // Also cannot interrupt when Aquaveil is active, however the counter is reduced per hit taken.
+        if (PDefender->StatusEffectContainer->HasStatusEffect(EFFECT_AQUAVEIL))
```
teschnei: it doesn't consume aquaveil unless the hit would have interrupted (ie. this needs to be after the random change)

Hozu:
Moved the Aquaveil check to only kick in if the spell were to be interrupted.

src/map/utils/battleutils.cpp
```diff
@@ -1615,7 +1622,22 @@ namespace battleutils
 
         if (chance < check)
         {
-            //Interrupt the spell cast.
+            // Prevent interrupt if Aquaveil is active, if it were to interrupt.
+            if (PDefender->StatusEffectContainer->HasStatusEffect(EFFECT_AQUAVEIL))
+            {
+                uint8 aquaCount = PDefender->StatusEffectContainer->GetStatusEffect(EFFECT_AQUAVEIL)->GetPower();
```
teschnei: there's no reason to downcast this to a uint8.. just use auto as the type

Hozu: Oh, ok.

PR #2984, opened 2016-04-03
Core: Correct the implementation of minimum recast time + other corrections

Hozu:
-Put the existing 80% recast time cap after everything (including the song flat recast time reduction) except the Light/Dark Arts recast time modification.
-Corrected a bug I accidentally introduced when combining very high recast reductions via haste and the flat recast time reduction (underflowed the uint).
-Corrected the Hasso/Seigan casting time penalty (2x, I misread it as 1.5x when I added it originally a long time ago).
-Added an entity type check before getting the merit value.

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -4937,7 +4937,7 @@ namespace battleutils
 
         if (PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_HASSO) || PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_SEIGAN))
         {
-            cast = cast * 1.5f;
+            cast = cast * 2.0f;
         }
 
         if (PSpell->getSpellGroup() == SPELLGROUP_BLACK)
@@ -5106,7 +5106,7 @@ namespace battleutils
 
         bool applyArts = true;
         uint32 base = PSpell->getRecastTime();
-        uint32 recast = base;
+        int32 recast = base;
 
         //apply Fast Cast
         recast *= ((100.0f - dsp_cap((float)PEntity->getMod(MOD_FASTCAST) / 2.0f, 0.0f, 25.0f)) / 100.0f);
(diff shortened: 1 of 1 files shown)


Hozu:
Oh right the source http://ffxiclopedia.wikia.com/wiki/Recast_Time

Hozu:
Turns out the source is outdated, cap is 80%, with Light/Dark Arts (and Celerity/Alacrity) bypassing that. Updated to reflect that.

Hozu:
Damn it, I didn't realize that doing a merit value grab in the core would cause a crash if it wasn't a player. Updated this pull with a check for that.

src/map/utils/battleutils.cpp
```diff
@@ -5205,30 +5229,7 @@ namespace battleutils
...
-            }
-            if (PSpell->getID() == 376 || PSpell->getID() == 377 || PSpell->getID() == 463 || PSpell->getID() == 471) // apply Lullaby recast merits
-            {
-                recast -= ((CCharEntity*)PEntity)->PMeritPoints->GetMeritValue(MERIT_LULLABY_RECAST, (CCharEntity*)PEntity) * 1000;
-            }
-            recast -= PEntity->getMod(MOD_SONG_RECAST_DELAY);
-            // ShowDebug("Recast after merit reduction: %u\n", recast);
-        }
 
-        if (recast < 0)
```
bendangelo: You sure we don't need this? Recast should never be below 0 right?

Hozu: I put that in in my previous commit when I added the subtraction mods for BRD songs. Since they're done before the recast gets set to a minimum of 20% of base, if it were to be negative it'd still get set to 20% of base.

Hozu: Anything after line 5151 is multiplicative.

bendangelo: Right it looks good.

Hozu: Meh I'll put it back in anyway, probably prevent someone from doing the same stupid thing I did here.

bendangelo:

PR #2986, opened 2016-04-03
SQL: Correct haste latent on Wyvern Earring

Hozu:
Changed to gear haste instead of magic haste, and 51/1024 from 5/1024.


bendangelo:

PR #2988, opened 2016-04-04
Added proper spell list, according to wiki, for Diabolos Waking Dream

Fiocitrine:
This is for the quest Waking Dream, the uncapped Diabolos fight.

I added four spells, Aspir, Drain, Sleepga II, and Dispelga. According to ffxiwiki, these are the only spells he uses.


bendangelo:

PR #2990, opened 2016-04-04
Item Weapon changes v1

Ninjistix:
fixed multi-hit, added missing weapons


Ninjistix:
messed-up something will repost

R3P0FFXI:
@Ninjistix You don't need to open and close the request, if you made a mistake just amend a previous commit and it will go through the process of recertifying again automatically.

Ninjistix:
@R3P0FFXI was trying to squash them as @bendangelo suggested, so there wouldn't be so many commits.. but messed-up somewhere

xdemolish:
doesn't matter, force push rewrites the history
@teotwawki does it all the time :)

PR #2992, opened 2016-04-04
Update wizards_roll.lua

Swiss-y:
values were set to pre 07 values. updated to match BG/retail


bendangelo:

Issue #2943, opened 2016-03-26
BST can charm uncharmable mobs


ymkc:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160203_0 

**_Server Version_** (type `@revision` in game) **:**
8f2765e5

**_Source Branch_** (master/stable) **:**
Master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
As a BST it's possible to charm morbols, wyverns, goobbues, manticores, hippogryphs

Afaik this was not possible on retail

helixhamin:
I want to charm Fafnirs xD

ymkc:
NM's aren't charmable, but Storms of Fate Bahamut is checked as TW and is charmable :D 

Unfortunately not mobs around to have fun with a bahamut pet ...

ymkc:
btw, you can "zone" bahamut, but you can't make him fight

PR #2996, opened 2016-04-05
Fixes for Charm

ymkc:
- make certain mob families uncharmable to match retail (#2943)
- duration of charm-bind randomized

Used wikia as the main source for the families, left abyssea mostly untouched


teschnei:
We're not going to stick a giant array of every uncharmable family into the core, sorry.  We can add a flag to one of the sql tables and check that, though.

src/map/utils/battleutils.cpp
```diff
@@ -3900,19 +3901,58 @@ namespace battleutils
...
                 PVictim->m_EcoSystem == SYSTEM_ELEMENTAL || PVictim->m_EcoSystem == SYSTEM_EMPTY ||
                 PVictim->m_EcoSystem == SYSTEM_LUMORIAN || PVictim->m_EcoSystem == SYSTEM_LUMINION ||
-                PVictim->m_EcoSystem == SYSTEM_UNDEAD || PVictim->PMaster != nullptr) {
+                PVictim->m_EcoSystem == SYSTEM_UNDEAD || PVictim->m_EcoSystem == SYSTEM_DRAGON ||
+                PVictim->m_EcoSystem == SYSTEM_HUMANOID ||
+                std::find(std::begin(families), std::end(families), ((CMobEntity*)PVictim)->m_Family) != std::end(families) || 
+                PVictim->PMaster != nullptr) {
                 PVictim->StatusEffectContainer->AddStatusEffect(
-                    new CStatusEffect(EFFECT_BIND, EFFECT_BIND, 1, 0, 5));
+                    new CStatusEffect(EFFECT_BIND, EFFECT_BIND, 1, 0, dsprand::GetRandomNumber(0, 5)));
```
bendangelo: It binds for at least one second.

bendangelo:
Thanks for your work. Adding this as a flag on the mob families table would work better.

ymkc:
ok cool, np. just want to have it moving forward :)

@teschnei gonna take a look or want me to add it?

xdemolish:
mob_family_mods, add a new MOBMOD and check that

bendangelo:
A mob mod works as well

teschnei:
I don't like using a mobmod for this. I thought about it earlier, but it
makes it kind of hard to remember when adding new families. Mob mods are
like.. Adding specific functionality to something, but uncharmable is a
pretty generic thing that should be in your face when adding a new family
(you should be actively presented with "is this family charmable?" Rather
than having the possibility of forgetting to check the family mobmods, if
that makes sense

xdemolish:
ya
was under the impression its not really a significant number of families which are uncharmable hence the mobmod to avoid adding a 'charmable' member to every mobentity object (ya i realise you didnt say that but w/e ive already typed this far cantstopwontstop)

anyways, mobmod but set from the new field instead of mob_family_mods and check that in battleutils

teschnei:
Yeah, I like that compromise. It keeps it overwritable by pool/script if
needed.

Hozu:
Should the beastmen ecosystem check be part of the uncharmable flag set in the DB as well?

Hozu:
Actually I just remembered that Nightmare mobs in dreamlands Dynamis are able to be charmed on DSP... might be a good idea to make the creation of the mobs for Dynamis/BCNMs/instances set the mobmod automatically, like how true aggro is currently enforced for Dynamis. That way it can be overridden if necessary.

bendangelo:
Beast ecosystem should be moved to a flag in the family mob table. Good idea Hozu, dynamis mobs shouldn't be charmed and the flag should be set probably in mobutils.

xdemolish:
nah it'll get set when adding the mob to the battlefield (when i eventually get round to finishing the rewrite)
though i guess master can have it set on gaining a confrontation effect

ymkc:
> mobmod but set from the new field instead of mob_family_mods
add a new field to which table?

xdemolish:
mob_family_system

TeoTwawki:
This may have changed in retail since it was somewhat exploitable: back when I played placeholders for an NM were not able to be charmed, even though mobs of that family are normally able to be charmed. So the mobmod makes sense to me for script-ability for the reasons above and this, should we need to make PH's uncharmable in the future (not sure we do, but the ability will be there).

xdemolish:
leaving for kj cause not sure about checking by ecosystem instead of just setting the flag and leaving it at that
also remember to leave a comment after updating a pull request so we know to have another look :)

ymkc:
I added the ecosystem checks back to reduce the sql queries. I would also exclude battlefield mobs and flag them from the scripts (is there even any chamable battlefield mob apart from the furies?)

teschnei:
there are giant queries that load mobs in zoneutils and instance_loader: just add the charmable column to the select list there and set it where every other parameter is set

ymkc:
thanks for the input, works like a charm :)

code updated

teschnei:
now remove m_charmable and set the mobmod in zoneutils/instanceutils

ymkc:
could have seen that one ... done

PR #2997, opened 2016-04-05
Major Fixes to Item_Weapon

Ninjistix:
fixed some weapon values, and added missing weapons


Ninjistix:
why is the check taking so long?

xdemolish:
it imports so whatever, not gonna check each item

Ninjistix:
okay, thanks for the quick reply @xdemolish

PR #2998, opened 2016-04-05
SQL/Scripts: Better handling of Pamamas-related latents

Hozu:
Put the bonuses as latents for the gear instead of the food effect. Prevents equipping the gear then using Pamamas, then unequipping the gear to keep the bonus. Also fixes the issue where Pamamas yield no effect (might need to have one of the offending items equipped).

### DIFF
--- a/scripts/globals/equipment.lua
+++ b/scripts/globals/equipment.lua
@@ -37,28 +37,3 @@ function ChocoboShirt(player)
     end
     return power;
 end;
-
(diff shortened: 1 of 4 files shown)


Hozu:
Something weird is going on with Github today. My late modifications aren't shown despite me force pushing to fix them.

Hozu:
There we go, though the build failed for whatever reason.

PR #3005, opened 2016-04-07
Scripts: Remove some redundancy with BRD enfeebles/refine scripts

Hozu:
-Made Elegy effect check against MOD_SLOWRES for applyResistanceEffect
-Made Elegy use applyResistanceEffect instead of applyResistance
-Remove second resistance check on each of the spells because why is it being checked twice?
-Cleaned up scripts so they're consistent.


bendangelo:
Thanks for contributing :heart: Surprised this wasn't fixed earlier.

PR #3022, opened 2016-04-10
More Charm adjustments

ymkc:
- made only normal mobs charmable, excluding battlefield/dynamis/nm/fished up/etc
- added mobmod_charmable to wings of fury furies.lua

### DIFF
--- a/scripts/zones/Ghelsba_Outpost/mobs/Furies.lua
+++ b/scripts/zones/Ghelsba_Outpost/mobs/Furies.lua
@@ -4,11 +4,13 @@
 -- BCNM20
 -----------------------------------
 
+
 -----------------------------------
 -- onMobSpawn Action
 -----------------------------------
 
 function onMobSpawn(mob)
+    mob:setMobMod(MOBMOD_CHARMABLE, 1);
 end;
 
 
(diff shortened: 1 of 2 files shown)


xdemolish:
looks good but i'll let @teschnei merge in case he wants anything done differently

ymkc:
rebased on master

anything wrong/could be better with this?

src/map/utils/zoneutils.cpp
```diff
@@ -500,7 +500,9 @@ void LoadMOBList()
                 PMob->m_TrueDetection = Sql_GetUIntData(SqlHandle, 65);
                 PMob->m_Detects = Sql_GetUIntData(SqlHandle, 66);
 
-                PMob->setMobMod(MOBMOD_CHARMABLE, Sql_GetUIntData(SqlHandle, 67));
+                if (GetZone(ZoneID)->GetType() != ZONETYPE_DYNAMIS && GetZone(ZoneID)->GetType() != ZONETYPE_BATTLEFIELD &&
```
bendangelo: I think these checks should be done inside `mobutils::calculatestats`. There you can change the charmable value just before the mob spawns. It's just easier to debug when all these checks are done in one place.

ymkc:
ok so leave setMobMod in zoneutils/instance_loader as it is (set it for all mobs) and remove/change it later in mobutils::calculatestats on special mobs( dyna, battlefield, etc) in mobutils::calculatestats?

bendangelo:
You should move all those checks to `mobutils`. You can already find some examples in it. Once you do that I think it's good. 👍

ymkc:
moved it all to modutils, looks much cleaner

Issue #2826, opened 2016-02-26
Antlions that ambush remain untargetable/spam Pit Ambush


Hozu:
Client Version (type /ver in game) :
30160203_0

Server Version (type @revision in game) :
3a852af

Source Branch (master/stable) :
master

Additional Information (Steps to reproduce/Expected behavior) :
Same issue as chigoes, different method of being hidden, it seems.

tagban:
This issue also happens on retail sometimes...

tagban:
Just disabled aggro for now on these, did the same with Chigoe.

teschnei:
chigoe have been fixed for a while

tagban:
I updated a week ago and had people still getting killed but just updated
today again so will test..

Hozu:
So after implementing a binding for HideModel, calling upon it in the Pit Ambush script, removing the special skill mobmod from the DB AND the core for mobs with the ambush flag, and even trying making them not be set to hide/name hidden on spawn, I still can't get them to unhide. They seem to be set to hide every tick or something and I can't figure out where.

Hozu:
Oh this gets even better, I put a breakpoint on CMobEntity::HideModel and commented it out from being called in the roam check. It triggered when it used Pit Ambush via the lua binding I added, but it never unhid. And the breakpoint was never triggered again. ???????????

Hozu:
Never mind it was the hidename thing... damn it. That should have been obvious.

PR #3025, opened 2016-04-10
Core/Scripts: Fix ambushing Antlions

Hozu:
-Added a lua binding for hideModel(), used it in Pit Ambush mobskill. Also added HideModel for npcentity.
-Created a mixin for ambushing Antlions (Pit Ambush mobskill handles the unhiding entirely to look proper in-game)
-Removed the remaining core/SQL stuff that was doing the Antlion ambush attack (most notably the special mobskill thing that was making them use Pit Ambush repeatedly).
-Some minor fixes to the handling of Alastor Antlion spawning.


Hozu:
Fixes https://github.com/DarkstarProject/darkstar/issues/2826

PR #3026, opened 2016-04-10
Scripts: Fix scripted 2hrs + mob Charm move

Hozu:
Since forced mobskills do the onWeaponskillCheck with the ai rewrite. Standard mob 2hrs still not done though.

Removed Charm skill from the mobs that had it because they're forced called by script.


bendangelo:
Wait so why did you switch the returns from 1 to 0?

Hozu:
Forced mobskills used to ignore the onMobskillCheck, but they don't with the AI rewrite. It used to return 1 for whenever a mob had that skill in their mobskill list. But now the approach has changed to keep forced moves out of their list, and have both the mob script and the onMobskillCheck handle things.

That's why I removed Charm from Vrtra/Ark Angel MR.

Having the scripted mobmod checked first would have it ignore the generic handling of mob 2hrs whenever that gets implemented again.

bendangelo:
Oh ok I see. Thanks

PR #3027, opened 2016-04-10
Fix a typo, add SQL for Distract II, restore some lost commented out data..

TeoTwawki:


TeoTwawki:
I left the trust "spells" out. Everything you see in commented out in the table file now previously was in there but was lost during a redump. Aside from the Geomancer spells most of it just needs scripts made.

PR #3028, opened 2016-04-11
Scripts: Fix DRG AF1 NM

Hozu:
This file escaped the onMobDeath conversion from (mob, killer) to (mob, killer, ally) because it was using (mob, player).


bendangelo:

PR #3031, opened 2016-04-12
Core/Scripts: Properly implement Atonement (mostly)

Hozu:
-Added lua bindings to get current CE and VE a player has on a mob, for the calculation of Atonement's damage.
-Added in the damage formula/type for Atonement vs. mobs. I did not find damage vs. non-mobs (ie. charmed players) so I left in the old damage calculation for that.
-Put WS Gorget/Belt handling into its own function to make things cleaner and so it can be accessed from elsewhere. I couldn't figure out better way to handle the itemid checks though, since you can say... use Flame Gorget + Light Belt for Knights of Round, and if you just add mods to the gorgets/belts it would double count them (once for Light property, and again for Fusion)...
-Updated Spirits Within to check if dual wielding for TP return.

Old Atonement damage source: http://www.ffxionline.com/forums/showthread.php?t=73784
New Atonement damage source: https://www.bg-wiki.com/bg/Talk:Atonement

### DIFF
--- a/scripts/globals/weaponskills/spirits_within.lua
+++ b/scripts/globals/weaponskills/spirits_within.lua
@@ -21,6 +21,7 @@ function onUseWeaponSkill(player, target, wsID, TP, primary)
 
     local HP = player:getHP();
     local WSC = 0;
+    local tpHits = 0;
     -- Damage calculations based on https://www.bg-wiki.com/index.php?title=Spirits_Within&oldid=269806
     if (TP == 3000) then
         WSC = math.floor(HP * 120/256);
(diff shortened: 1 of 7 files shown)


Hozu:
Whoops, the damage cap needs to be applied before it's actually done, otherwise it's meaningless. Fixed that.

scripts/globals/weaponskills/spirits_within.lua
```diff
@@ -42,13 +43,20 @@ function onUseWeaponSkill(player, target, wsID, TP, primary)
     end
 
     local damage = target:breathDmgTaken(WSC);
-    damage = damage * WEAPON_SKILL_POWER
+    if (damage > 0) then
+        if (attacker:getOffhandDmg() ~= 0 and attacker:getOffhandDmg() > 0) then
```
bendangelo: Is the `~=` needed? `> 0` gives the same effect.

bendangelo:
Everything looks good. Just address that one comment, then I'll merge. 👍

scripts/globals/weaponskills/spirits_within.lua
Hozu: I copied that from weaponskills.lua (minus the h2h skill check). I can change that though.

Hozu:
Requested changes made.

bendangelo:

PR #3037, opened 2016-04-13
fixed Shami.lua not giving orbs and taking wrong seals.

Ninjistix:
added orbs missing from item_basic, added missing trade info for shami and removed obsolete lines as per @xdemolish request


Ninjistix:
fixed dup error in item_weapon.sql

Issue #2951, opened 2016-03-28
Exploit: Players can remove BCNM level restriction via level sync


Hozu:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160203_0

**_Server Version_** (type `@revision` in game) **:**
https://github.com/DarkstarProject/darkstar/commit/8f2765e5f9a29ac5f9bd75cb2da6de36a610a8bc

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
Sync party, trigger BCNM entry, get everyone in, disband party. Everyone gets uncapped after sync wears, then reform and destroy the fight that's supposed to be capped.

Hozu:
Whoops, BCNM status itself is never removed, otherwise it wouldn't be much of an exploit, heh. Only the level capping is removed.

PR #3039, opened 2016-04-14
Fix issue #2951

TeoTwawki:
No more exploiting level sync to uncap BCNMs..

I also broke up an extremely long line (4522 of charutils) without changing what it does, because that was crazy and I hate left/right scrolling to read code.

@Hozu thanks for checking in game for me


Fiocitrine:
Thanks, Teo! I like how we can't exploit the game....

bendangelo:

PR #3044, opened 2016-04-14
Scripts/Core: Add command to check/delete item from someone's inventory

Hozu:
Make it easier for GMs to handle things without having to mess with the database. Existing uses of the delItem lua binding work as intended without modifications.

### DIFF
--- a/scripts/commands/delitem.lua
+++ b/scripts/commands/delitem.lua
@@ -1,0 +1,41 @@
+---------------------------------------------------------------------------------------------------
+-- func: @delitem
+-- desc: Deletes a single item held by a player, if they have it.
+---------------------------------------------------------------------------------------------------
+
+cmdprops =
+{
+    permission = 1,
+    parameters = "is"
+};
+
+function onTrigger(player, itemId, target)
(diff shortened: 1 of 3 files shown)


TeoTwawki:
There are situations where we need to know the item is on their person not sitting in the mog safe or storage. Which is the original for the existence of `hasItem()` if I am not mistaken. Perhaps arrange it so if a container is not specified, it'll check them all? I dunno if that'd be the best way to handle it, just one possibility.

One odd case according to [ffxiclopedia](http://ffxiclopedia.wikia.com/wiki/I_Can_Hear_a_Rainbow):
"You may also keep the Carbuncle's Ruby in your Mog Sack or Mog Satchel and still receive the cutscene and credit for the weather, but not in your Mog Case."

TeoTwawki:
er..I think I did not read all your code correctly..

TeoTwawki:
@Hozu nudge me when you return to irc

Ninjistix:
omg i would have loved this a few weeks ago... lol

bendangelo:
This looks good to me.

PR #3050, opened 2016-04-16
heavy_metal_pouch.lua and linen_coin_purse.lua

Kosmos82:
Hoping I did this correctly. I tried to follow TeoTwawki guide as best I could. If I did something wrong please let me know what so I can do it properly next time. Thanks.


TeoTwawki:
Looks good.

In the future you may want to make additional branches instead of using master. Its especially helpful if you want to have multiple pull requests open at the same time for different things.

Kosmos82:
Ok I can do that next time. I was kind of unsure the best way to handle that. Thanks

xdemolish:
thanks for contributing :)

PR #3053, opened 2016-04-16
Scripts: Make WSs get critical hit rate merits when applicable.

Hozu:
The current implementation neglected that.

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -113,7 +113,7 @@ function doPhysicalWeaponskill(attacker, target, wsID, params, tp, primary)
             critrate = critrate + (10 + flourisheffect:getSubPower()/2)/100;
         end
         nativecrit = (attacker:getStat(MOD_DEX) - target:getStat(MOD_AGI))*0.005; -- assumes +0.5% crit rate per 1 dDEX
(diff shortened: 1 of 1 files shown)


scripts/globals/weaponskills.lua
```diff
@@ -113,7 +113,7 @@ function doPhysicalWeaponskill(attacker, target, wsID, params, tp, primary)
             critrate = critrate + (10 + flourisheffect:getSubPower()/2)/100;
         end
         nativecrit = (attacker:getStat(MOD_DEX) - target:getStat(MOD_AGI))*0.005; -- assumes +0.5% crit rate per 1 dDEX
-        nativecrit = nativecrit + (attacker:getMod(MOD_CRITHITRATE)/100);
+        nativecrit = nativecrit + (attacker:getMod(MOD_CRITHITRATE)/100) + attacker:getMerit(MERIT_CRIT_HIT_RATE)/100;
```
bendangelo: I think you have to check it's a player because this will crash for mobs.

`attacker:getObjType() == TYPE_PC`

Hozu: You do in the core, but not lua. See the outdated comments here: https://github.com/DarkstarProject/darkstar/pull/2968

bendangelo: ok cool

bendangelo:

PR #3057, opened 2016-04-17
Scripts: Update TP gain from magic damage.

Hozu:
Wasn't updated when TP was converted to /1000.


Deadwing888:
While we're on the subject, I've noticed poison II giving tp on Nasomi's (admittedly old) build. Spells for 0 and spells that do no damage on cast should not be giving 10.0 tp.

Hozu:
Looking at magic.lua it only adds tp in  function finalMagicAdjustments, which is called for damaging spells only... it's not in applyResistanceEffect so it shouldn't be happening for non-damaging spells.

PR #3060, opened 2016-04-18
Scripts: Correct pos for spawn point for THF AF2 NM

Hozu:
I didn't actually verify the pos last time, just fixed the method. These three towers didn't have the spawn point set to near the tower...


TeoTwawki:
I'm confused due to the fact I actually completed this before..Did someone move them to make them be in wrong pos? Because they -_used to be_- in the right places.

Hozu:
Even if they didn't, before the recent deaggro change the mob would eventually reach the person who spawned it. Also some towers were fine.

bendangelo:

PR #3063, opened 2016-04-18
Fix Assault tag distribution

ymkc:
- made SetPoints accessable from lua (setCurrency)
- minor fixes to NPC 

I know assault is not working, but since 99% is there anyway, why not fix it


TeoTwawki:
@ymkc thanks.

PR #3064, opened 2016-04-18
added some rings to item_armor.sql and item_basic.sql

Ninjistix:
added some missing items


bendangelo:

PR #3065, opened 2016-04-18
Scripts: Rudimentary Amphiptere support

Hozu:
Enough to make it come down upon aggro so people can target/fight it. Goes back up on deaggro.


bendangelo:

PR #3067, opened 2016-04-19
SQL: Set levels for mobs in WotG northlands areas

Hozu:
Glacier [S] and Xarcabard [S] were pretty much entirely devoid. Baileys [S] had some set, Keep [S] had more, but some were incorrect.

Also corrected the spawn type for the elementals in Baileys [S], and the level of the tiger pet in Vunkerl Inlet [S].


bendangelo:

PR #3068, opened 2016-04-19
SQL: Change mobskill list for Corpselights to be 0, rename the one th…

Hozu:
…ey had to be for Pandemonium Lamps

Also set the spell list to BLM for the two that were missing it.


bendangelo:

PR #3070, opened 2016-04-20
Scripts: Setting for various types of magic damage bonus + minor fixes

Hozu:
I noticed the existing ones weren't working, aside for the one for Drain. Which wasn't even being called for MP stealing spells.

Minor fixes include:
-Holy/Holy II now use doDivineNuke, which calls upon a function to do the exact thing it had in the individual spell scripts.
-Added settings for different kinds of magic damage (Dark, Elemental, Divine, and Ninjutsu).
-Capped Drain-type spells at mob's current HP.
-Fixed the double counting of blue magic accuracy merits for BLU nukes.
-Removed the forcing of a specific message on successful use of the BLU drain spells, since the message is already set in the database.


bendangelo:

PR #3071, opened 2016-04-20
Add Zoraal Ja & Zoraal Ja's Pkuucha

ymkc:
Tried to follow wiki as close as possible:

- add pkuucha as a lottery pop from lesser colibris
- force pop bst at 20-50% HP on pkuucha and fully restore pkuucha's HP
- made pkuucha unkillable as long as bst is up
- "autokill" pkuucha if HP is 1% and but is killed first
- removed the "the" from the pkuucha's in-game name


scripts/zones/Wajaom_Woodlands/mobs/Zoraal_Ja_s_Pkuucha.lua
```diff
@@ -11,8 +13,33 @@ function onMobSpawn(mob)
 end;
 
 -----------------------------------
+-- OnMobFight Action
+-----------------------------------
+
+function onMobFight(mob, target)
+    if(mob:getHPP() <= math.random(20,50) and GetMobAction(Percipient_Zoraal_Ja) == ACTION_NONE
```
Hozu: Since onMobFight runs once every second, you should probably set the % to spawn the master at on the spawn, as a local var, then check against that local var instead.

TeoTwawki:
I've actually killed Zoraal Ja's Pkuucha while Zoraal Ja was up. In fact thats almost always how I did it, because of the chance of someone stealing it when it goes unclaimed if I were to target Zoraal Ja. Doesn't matter which you kill first on retail.

TeoTwawki:
ignore my previous post...I'm having a severely off day

scripts/zones/Wajaom_Woodlands/mobs/Zoraal_Ja_s_Pkuucha.lua
ymkc: @Hozu Thanks for the hint!
You mean like this to reduce the math.random calls?

Hozu: Yeah. If you leave it in there, every second, it will roll for random 20-50, which would make it spawn the master closer to 50% way more often than at lower percent.

ymkc:
Anything to improve on this one?

HP of both seems to be off (based on player feedback around the web) but I can't find hard numbers

TeoTwawki:
looks alright to me

ymkc:
Fixed formatting and a typo

PR #3074, opened 2016-04-20
SQL:item_basic adjusted flags to match polutils value

Ninjistix:
a continuation of last commit


bendangelo:

PR #3079, opened 2016-04-22
Correct scripts/globals/items/elshimo_pachira_fruit.lua

TeoTwawki:
In retail all this does is poison effect, not food+poison.

http://www.ffxiah.com/item/5604/el-pachira-fruit
1st comment on the page, if anyone doubts my memory and hasn't used any on retail lately..


bendangelo:

PR #3080, opened 2016-04-22
Core/SQL: Fix Overwhelm check

Hozu:
-Moved the check to before damage was actually dealt to the target.
-Added a check to make Overwhelm only apply to melee Weaponskills.
-Replaced the redundant job/level check (since it's already done when fetching merit value) with an objtype check, since there are some mobskills that use battleutils::TakeWeaponskillDamage (and thus battleutils::getOverWhelmDamageBonus).
-Fixed the merit power value since the value fetched for merits is the total power value, not the amount of upgrades in the category.


bendangelo:

PR #3081, opened 2016-04-22
Core/Scripts: Further implementation of Afflatus Misery damage mod

Hozu:
-Created a lua binding to call battleutils::HandleAfflatusMiseryDamage, and added it to the final checks for magic, blue magic, and mobskills.
-Changed the cap amount for Banish boost to current max HP/4. The page on BG wiki put a range of 250-300 damage, and the source forum post suggested that it's 1/4th max HP.

Cura was untouched because of a previous commit I did long ago made it work as intended, just the Misery damage storage wasn't being called by everything.

### DIFF
--- a/scripts/globals/bluemagic.lua
+++ b/scripts/globals/bluemagic.lua
@@ -244,6 +244,7 @@ function BlueFinalAdjustments(caster, target, spell, dmg, params)
 
     target:delHP(dmg);
     target:updateEnmityFromDamage(caster,dmg);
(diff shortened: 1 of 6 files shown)


Hozu:
Crap I was working on an older branch because I didn't want to update my client. I will fix the conflict.

Hozu:
There, fixed. Turns out it was me fixing the same (lack of) spacing a second time.

PR #3084, opened 2016-04-23
Some more charm fixes

ymkc:
- corrected mobtypes of regular mobs (torama, korrigan, etc)
- split snipper and mousse into normal and fished up mobpools
- added mobmod_charmable to ironshells/cutters


bendangelo:

PR #3089, opened 2016-04-24
sql/item_basic.sql update

some1here:
to 1200 as per Windower, SE, and polutils [windower items.txt](https://github.com/DarkstarProject/darkstar/files/233675/windower.items.txt)


Altalus:
Do you actually test the inserts at all ?

`Importing item_basic.sql into the database...ERROR 1064 (42000) at line 3433: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'INSERT INTO `item_basic` VALUES (3557,0,'athena_orb','athena_orb',1,28752,0,1,0)' at line 2`

some1here:
item (3557,0,'athena_orb','athena_orb',1,28752,0,1,0)' is as it was from DS

Altalus:
But the default DS imports fine, not yours :)

 

______________________________ 
Guillaume Gingras

PR #3091, opened 2016-04-25
item_basic.sql to 1200

some1here:
This version


xdemolish:
you didn't test again
set up a server and test to make sure it imports and works otherwise you're just wasting everyone's time (including your own)

PR #3092, opened 2016-04-25
Blue magic script corrections and additions

TeoTwawki:
Add sudden lunge, add missing stun effect to frypan and tail slap, get rid of potential client crash in head butt

Durations might be off, I erred on the side of being short if data wasn't available (copied head butt duration unless wiki gave me actual duration info).

The stuns don't overwrite, so we don't need to check if the target already has stun effect here.

The crashy message thing has apparently existed since c2d47b3e03308d9d50f0c0bd93b6a8957cda9518 back in the SVN patchfile days. The problem was that it was returning damage values into a message that expected an effect ID, and per discussion with @MalRD in IRC that should never have even been there.

### DIFF
--- a/scripts/globals/spells/bluemagic/frypan.lua
+++ b/scripts/globals/spells/bluemagic/frypan.lua
@@ -12,10 +12,9 @@
 -- Skillchain Element(s): Lightning (can open Liquefaction or Detonation; can close Impaction or Fusion)
 -- Combos: Max HP Boost
 -----------------------------------------
-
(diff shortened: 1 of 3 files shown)


bendangelo:
`addStatusEffect` will first confirm it's valid be applying, so stun effects will not overwrite. 

TeoTwawki:
Yeah thats what I meant to say, the effect were already flagged that way so I didn't need to check in script.

PR #3095, opened 2016-04-26
added/adjusted below

Ninjistix:
item.cpp - adjusted flag ref item.h
item.h - adjusted flags to match polutils
packet_system.cpp - adjusted flag ref to item.h
treasure_pool.cpp - adjusted flag ref to item.h
-- and some tab2space


Ninjistix:
need to fix pr

PR #3098, opened 2016-04-26
Scripts: Add include for settings

Hozu:
I didn't think to do this earlier because I assumed it was there. Whoops.


Hozu:
...and I was beaten to it.

nasomi:
:cat:

Issue #2680, opened 2016-01-27
Worms aggroing underground stay untargetable


ymkc:
If a player gets aggroed by an underground worm (worm while moving, eg land worm's in qufim), the name doesn't appear and they stay untargetable.

also, if this happens and the player puts on sneak+deo (and runs off), it's not possible to lose aggro 
and finally, after returning to check for aggro state, player got hit by rasp but didn't lose sneak+deo

happend on today's master 2d460d7 

sorry for 3 in 1 bug report, but i guess it might be somehow related

armouredking:
Worm issues have been for a while https://github.com/DarkstarProject/darkstar/issues/2138

Didn't personally test it with spell use though.

PR #3099, opened 2016-04-26
Core: Prevent underground worms from aggroing/linking

Hozu:
Fixes https://github.com/DarkstarProject/darkstar/issues/2680


bendangelo:

PR #3103, opened 2016-04-26
Fix missing NPC's in Abyssea-Altepa

TeoTwawki:


bendangelo:

PR #3096, opened 2016-04-26
added/adjusted below pt1

Ninjistix:
item.cpp - adjusted flag ref item.h
item.h - adjusted flags to match polutils
packet_system.cpp - adjusted flag ref to item.h
treasure_pool.cpp - adjusted flag ref to item.h
-- and some tab2space

PR #3105, opened 2016-04-26
added/adjusted below pt1

Ninjistix:
item.cpp - adjusted flag ref item.h
item.h - adjusted flags to match polutils
packet_system.cpp - adjusted flag ref to item.h
treasure_pool.cpp - adjusted flag ref to item.h
-- and some tab2space
 #3096 closed by mistake


Ninjistix:
missed a file in commit

PR #3106, opened 2016-04-26
SQL: Item basic adjusted sort

Ninjistix:
fixed sort to ascending, nothing else changed


Ninjistix:
@xdemolish  fixed the item sort to ascending

PR #3107, opened 2016-04-27
HAAAALLLLVVVEEEERRRRR!

TeoTwawki:
@teschnei I can't locate the issue but there was one where I questioned this type of logic being used, and you told me it was both legit and preferred. I'm now wondering if that only works for variable declarations or if our specific implementation we use just doesn't like it. My google fu only turns up examples where its used in a local variable declaration and nothing else, so I'm thinking that's why it doesn't work here.

@xdemolish thanks for the assist


TeoTwawki:
Oh..or because its not a var, its an arg? Anyway that didn't work for old halver..

teschnei:
First assumption was right. And/or are just expression operators. Check the
PIL book chapter 3.3 for where it talks about "a = x or y"

Issue #3109, opened 2016-04-28
Shadow Mantle null damage overpowered by factor of 3-5


Deadwing888:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **30160329_0**

**_Server Version_** (type `@revision` in game) **5402178**

**_Source Branch_** (master/stable) **Master**

**_Additional Information_** (Steps to reproduce/Expected behavior) **.**

The Item Shadow mantle (itemid 13658) has modID 416 (nullify physical damage) set at 15% in item_mods.sql. As a long time owner of a retail shadow mantle I will say that the proc rate is between 3 and 5% on retail.

Any other shadow mantle owners out there care to add their two cents?

towbes:
Let's focus on citing references when they are easily available rather than relying on memory:

https://www.bg-wiki.com/bg/Shadow_Mantle

https://www.bluegartr.com/threads/88580-Shadow-Mantle-Proc-Rate

Deadwing888:
Well there you have it

Hozu:
Semi-related, Dynamis Lord itself has the effects of the mantle/ring right?

Hozu:
Also does the Shadow Ring not null magic damage? I noticed it only has the mod for deathres in the DB.

Deadwing888:
@Hozu I only ever saw Dynamis Lord zerged so it's hard for me to say what his behavior is really like. I have never heard of him nullifying magic or physical damage though.

Altalus:
It depends on the form no? It alternates between these on his 2nd form. Took ages to kill on Vanilla.

 

______________________________

Guillaume Gingras

PR #3110, opened 2016-04-28
SQL: Correct Shadow Mantle/Ring proc rate

Hozu:
6% for Mantle. I noticed that Ring did not have an entry for mod 476 (MOD_MAGIC_NULL) so I added it at 13%.

Also added the Darksday latent for the two, since I noticed they were missing.

Fixes https://github.com/DarkstarProject/darkstar/issues/3109
Proc rate: https://www.bluegartr.com/threads/88580-Shadow-Mantle-Proc-Rate
MDB value for Ring: http://ffxiclopedia.wikia.com/wiki/Shadow_Ring


Deadwing888:
Yeah that sounds about right.

PR #3111, opened 2016-04-28
SQL: Nocturnus Helm/Mail Phys/Magic absorb chance

Hozu:
5% each. Removed the wrong 15% null phys chance from the helm.

https://www.bg-wiki.com/bg/Nocturnus_Helm
https://www.bg-wiki.com/bg/Nocturnus_Mail


bendangelo:

PR #3113, opened 2016-04-29
Scripts/SQL: Some Dynamis Lord fixes

Hozu:
-Made Dynamis Lord spawn after the 15 NMs are defeated (yes, that's when he's to spawn).
-Fix Ying and Yang respawning other dragon mechanic.
-Fix Dynamis Lord spawning the dragons instantly. Also, set them to spawn at his spawn point when he spawns them (I observed a late 75 era video where this is the case). Because of this, Ying and Yang have their spawn points set prior to initial spawn as well.
-Made Dynamis Lord send the dragons after his target if they have no target (prevents kiting away).
-Set Dynamis Lord's sub to BLM, enabling him to cast the spells he already has set.
-Toned down the TP scaling on Tera Slash and Oblivion Smash, but increased the number of hits of Oblivion Smash from 1 to 3 (aka the alliance wiper).
-Removed Dynamis Lord's immunity to Stun.
-Correct npcid of ??? that spawns when Dynamis Lord is defeated.
-Implemented Dynamis Lord's damage immunity and removal of immunity onto Ying and Yang for initially fighting them. Note: I had to guess at which dragon had which immunity attached to it.


Hozu:
Note: I wasn't sure of a good way to handle the damage immunity (and removal of it) on DL, so it was left out.

scripts/zones/Dynamis-Xarcabard/mobs/Dynamis_Lord.lua
```diff
@@ -52,14 +52,21 @@ function onMobFight(mob,target)
...
 
-    if (mob:getBattleTime() % 90 == 0) then
-        if (GetMobAction(YingID) == ACTION_NONE) then
-            SpawnMob(YingID):updateEnmity(target); -- Respawn Ying after 90sec
-        end
-        if (GetMobAction(YangID) == ACTION_NONE) then
-            SpawnMob(YangID):updateEnmity(target); -- Respawn Yang after 90sec
+    if (mob:getBattleTime() % 90 == 0 and mob:getBattleTime() >= 90) then
+        if (GetMobAction(YingID) == ACTION_NONE and GetMobAction(YangID) == ACTION_NONE) then
+            GetMobByID(YingID):setSpawn(-414.282,-44,20.427); -- These come from DL's spawn point when he spawns them.
```
teschnei: why don't you set their spawn point in the sql?

Hozu: That's not where they initially spawn. At first they spawn in front of the castle/bridge to the castle. When DL spawns them during the fight, they spawn in a different location. Hence, the new spawn point.

Hozu: That is to say, their current spawn point in the DB is correct for their initial spawning.

Hozu: I changed it so their spawn point is also set for their initial spawning so they spawn in the correct location for multiple runs without a server reboot.

Deadwing888:
I believe Dynamis Lord should not spawn until Ying and Yang are killed within a few seconds of each other. Ying and Yang spawn when the prison of souls sets free its captive spirits.

Hozu:
Ying and Yang are currently spawned when the 15 NMs are defeated. Dynamis Lord is also spawned at that point - Ying and Yang are required to be defeated not because DL is not there, but because he's immune to damage until they're killed. Though I was not able to implement that part.

teschnei:
why not?

Hozu:
I don't know which dragon does which immunity - supposedly one removes phys immunity, the other removes magic immunity.

teschnei:
just pick one if you can't figure it out, i'd rather have it 99% working
than 0% working

Hozu:
Fair enough.

Hozu:
Updated with the damage immunity thing on DL. Also fixed the ??? spawning when DL is defeated.

Deadwing888:
Ah yeah, that's right.

scripts/zones/Dynamis-Xarcabard/mobs/Dynamis_Lord.lua
teschnei: you asked me if it was okay to not set their spawn point and just move them - what happens if you wipe while they're up?

Hozu: Ah yes, I should reset their spawn points on spawn based on the variable Dynamis Lord has.

teschnei: I was getting at: if they wander back to their initial spawn point (and not Dynamis Lords), then it makes sense to just move them and not touch their spawn point

Hozu: That's what I don't know. In any case, they should probably move back to their spawn point instead of where the one dragon spawned the other. I'm going to make the assumption that they wander back to where DL spawns them in the event of a wipe.

Hozu:
Updated pull to reset spawn point after one dragon spawns the other - this would make the dragons wander back to where they were spawned (either by the defeat of the 15 Kindred NMs or by DL), rather than back to where one dragon spawned the other.

Hozu:
I underestimated the damage that Oblivion Smash can do, so I upped it. Also I set the power of the inflicted statuses to something other than 1.

Hozu:
Huh, not sure how my change to mob_pools.sql managed to not be included when it was before. Oh well, this is complete, unless someone has an issue with something.

Hozu:
Slight alterations to the scripts - made the local vars hold the mob instead of the mobid so it doesn't have to do a bunch of GetMobByID calls.

PR #3122, opened 2016-05-01
Make dsconnect bindable to a specific IP

### LINKED ISSUE #3117: Limit dsconnect to IP
**_I have:_**

not related

**_Server Version_** (type `@revision` in game) **:**
ffc3a95

master

It's not possible to limit dsconnect to only listen to a specific ip address.

this would be of great help to test a modified version versus a vanilla version or to have different branches running or even to have a dev and a live game server on the same box at the same time

ymkc:
First attempt to fix #3117

Not sure it works under windows (cannot test it) and about the parameter names (as I just passed them through)

### DIFF
--- a/conf/login_darkstar.conf
+++ b/conf/login_darkstar.conf
@@ -32,6 +32,17 @@ stdout_with_ansisequence: no
 console_silent: 0
 
 #--------------------------------
+#TCP/IP parameters
+#--------------------------------
+
+login_data_ip: 127.0.0.1
+login_data_port: 54230
+login_view_ip: 127.0.0.1
+login_view_port: 54001
+login_auth_ip: 127.0.0.1
+login_auth_port: 54231
+
+#--------------------------------
 #SQL parameters
 #--------------------------------
 
--- a/src/common/socket.cpp
+++ b/src/common/socket.cpp
@@ -797,7 +797,7 @@ int connect_client(int listen_fd, sockaddr_in& client_address)
 	return fd;
 }
 
-int32 makeListenBind_tcp(uint32 ip, uint16 port,RecvFunc connect_client)
+int32 makeListenBind_tcp(const char* ip, uint16 port,RecvFunc connect_client)
 {
 	struct sockaddr_in server_address;
 	int fd;
@@ -827,7 +827,7 @@ int32 makeListenBind_tcp(uint32 ip, uint16 port,RecvFunc connect_client)
 	//set_nonblocking(fd, 1);
 
 	server_address.sin_family      = AF_INET;
-	server_address.sin_addr.s_addr = htonl(ip);
+	server_address.sin_addr.s_addr = inet_addr(ip);
 	server_address.sin_port        = htons(port);
 
 	result = sBind(fd, (struct sockaddr*)&server_address, sizeof(server_address));
--- a/src/common/socket.h
+++ b/src/common/socket.h
@@ -274,7 +274,7 @@ extern int32 naddr_;   // # of ip addresses
 
 	int32 makeConnection_tcp(uint32 ip, uint16 port);
 
-	int32 makeListenBind_tcp(uint32 ip, uint16 port,RecvFunc connect_client);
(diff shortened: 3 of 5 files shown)


teschnei:
can't merge until it compiles on Windows

if you don't know how to deal with the IP addresses, look at map.cpp (it reads the IP address as command line arg)

also, it's 4 spaces for a tab, not 8

conf/login_darkstar.conf
```diff
@@ -32,6 +32,17 @@ stdout_with_ansisequence: no
 console_silent: 0
 
 #--------------------------------
+#TCP/IP parameters
+#--------------------------------
+
+uiLobbyDataIp: 0.0.0.0
```
TeoTwawki: please default to 127.0.0.1 for IP

ui us? typical DSP user isn't going to know what your abbreviation is unless you explain it in comment above.

ymkc:
Will try to fix the windows side, just have no way to test/compile it :|

As for the ui/us stuff, that's what was already there.
I can just drop them or change the them

TeoTwawki:
point was on the settings end of it, ppl won't know what that means because they aren't familiar with the internal code you got them from. Can name the settings anything that looks reasonable that doesn't make you explain to much. :)

teschnei:
It meant unsigned int/short, but the IP were changed to a string :)

teschnei:
Also the default of 0.0.0.0 matches the old setting (INADDR_ANY)

TeoTwawki:
make more sense to default to math the ip of other settings defaults though

xdemolish:
remember to update xiloader to specify the port too

ymkc:
Ok thanks for the input!

How about reducing/unifying the 3 IP params to just one? Does it make sense/is it reasonable/does it even work to have dsconnect running on 3 different IPs or even hosts?

TeoTwawki:
> remember to update xiloader to specify the port too

feature already exists I thought, and just didn't' work because the lobby was still a static port

xdemolish:
@teotwawki doesnt look like it
https://github.com/DarkstarProject/xiloader/blob/master/xiloader/network.cpp#L215
https://github.com/DarkstarProject/xiloader/blob/master/xiloader/network.cpp#L501

TeoTwawki:
lol, so atom0s had a command line switch for it when he first released the thing, but it did nothing this entire time.

ymkc:
- fixed the win32 build (missing header)
- set 127.0.0.1 as default IP
- used <creativemode> for var names

but can't for the life of baby jesus get that one tab from src/common/socket.cpp to be gone
it's not in my local copy and I even tried to remove/readd the whole line and commit it

src/common/socket.cpp
```diff
@@ -827,7 +828,7 @@ int32 makeListenBind_tcp(uint32 ip, uint16 port,RecvFunc connect_client)
 	//set_nonblocking(fd, 1);
 
 	server_address.sin_family      = AF_INET;
-	server_address.sin_addr.s_addr = htonl(ip);
+        inet_pton(AF_INET, ip, &(server_address.sin_addr));
```
teschnei: server_address.s_addr = inet_addr(ip);

this is what the map server does, and should be what you do unless it's not possible

ymkc: @teschnei thanks, now i got you :)

src/common/socket.cpp
```diff
@@ -827,7 +827,7 @@ int32 makeListenBind_tcp(uint32 ip, uint16 port,RecvFunc connect_client)
 	//set_nonblocking(fd, 1);
 
 	server_address.sin_family      = AF_INET;
-	server_address.sin_addr.s_addr = htonl(ip);
+        server_address.sin_addr.s_addr = inet_addr(ip);
```
teschnei: there are 8 spaces on this indent, and there should be 4

ymkc: resetted the file

teschnei:
gonna do a quick test on windows when I get home then I'll merge it if it works as expected

PR #3124, opened 2016-05-03
Double bust implementation

thanacus:
Implements bust system as retail pre-Abyssea:

1) COR can re-roll same roll after a bust of that roll; the negative and positive effects should stack.
2) If a COR main busts twice, they cannot roll more until a bust wears off.
3) If a COR sub busts once, they cannot roll more until a bust wears off.
4) Bust causes immediately loss of Double-Up.


teschnei:
the code looks fine, but it should be post-abyssea

thanacus:
I'll look up patch changes for COR after that point, but hadn't seen any that affected the dice system; will verify.

Hozu:
There was some really strange things when I tried to implement this before, I'd get two bust effects then I'd still be able to roll, deleting one of the bust effects.

How does this work with Fold? I'm not sure which Bust it's supposed to remove first, but I know it mistakenly removes the oldest roll effect instead of the newest one if you don't have a Bust.

thanacus:
The first point is resolved by not letting the ability go off if you have too many busts.  As for Fold, I did not check it as I saw (probably mistakenly) in the wiki that merit abilities for COR had not yet been implemented, but I'll test it out.

thanacus:
Also, finished patch check and I think my implementation holds for post-Abyssea as well.  The only new roll mechanic that took me be surprise was the "no-bust under effect of 11", which I thought was a bug (and was going to go after next...lol).

Hozu:
Ah, that check makes sense. I must have made a mistake somewhere, but it was so long ago that I forget what it was. But yeah, Fold is usable, though not quite correct it seems. I suppose it will need to be corrected.

Issue #3126, opened 2016-05-03
spirits within and odd visual glitch on TP bar


Scavenge:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160329_1

**_Server Version_** (type `@revision` in game) **:**
fbbd5722188581e3bc099b2be3f835bb5a4a0203

**_Source Branch_** (master/stable) **:**
stable

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

spirits within missed due to a reference to 'attacker' that it did not like, but also when it missed my TP remained at 3000% in the bar in the top left, but trying to use another WS was met with 'not enough tp' despite the bar being visibly full.  some kind of visual glitch.  went away once my tp changed again with @ tp.

Scavenge:
removing the elseif set that checked for attacker and leaving only the second option fixed it from missing

Hozu:
This one was my fault.

PR #3127, opened 2016-05-03
Scripts: Fix wrong target reference

Hozu:
Fixes https://github.com/DarkstarProject/darkstar/issues/3126


Ninjistix:
just curious but wouldn't https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/weaponskills.lua need updating too since it has the same attacker ref ?

teschnei:
the problem is that the parameter to spirits within is not called "attacker", not the actual name of it

PR #3131, opened 2016-05-04
sql/item_mods.sql

bn7:
Added some missing Mods


TeoTwawki:
2 things: 
* This is more than some, its enough changed in a single commit that github can't display the changes. So you'd have to  look at the diff locally like I just did. Huge pita.
* Since there is no commenting, anyone reviewing the changes would have to look up when item the item ID each one is and then what modifiers those mod IDs are being applied.

Could you possibly make this a little easier on us?

bn7:
Sorry the system just took the whole list. Is their a command/way to make it as several? instead of one?

TeoTwawki:
You'd have to make commits as you make changes, instead of after all changes are finished. 

It is possible to roll back the history, but backup the file first if you are going to. I'd hate to see you lose any of your work!

xdemolish:
pull to a new branch, change 100 lines of code at most, commit and push to your fork
then pull request and wait til we review before making a new one

Issue #3024, opened 2016-04-10
Hidden mobs can cast spells.


Hozu:
**_I have:_**

- [X] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [X] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160203_0

**_Server Version_** (type `@revision` in game) **:**
https://github.com/DarkstarProject/darkstar/commit/ec30b55f7457dbfbbe190ed181c4b83e7e07871d

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
While testing out antlion stuff, I noticed that when I deaggroed Alastor Antlion it was attempting to cast Stoneskin repeatedly while hidden, though it could never complete the spell.

Hozu:
Worms too, but when they cast on players they don't unhide their name so you can't even fight back.

xdemolish:
are worms supposed to be able to cast while hidden (and then unhide)?

Hozu:
Worms don't cast spells when hidden, it's only like alastor antlion after it deaggros.

PR #3133, opened 2016-05-04
disallow untargetable/hidden mobs to cast spells

xdemolish:
dunno if it's either or both so pull requested
probably fixes https://github.com/DarkstarProject/darkstar/issues/3024


Hozu:
About this, there are certain situations where an untargetable mob can cast spells on players/self/NPCs - those three taru in the Tenzen fight.

teschnei:
really this whole behaviour should just be a part of worm specific AI - i would really like it to be moved entirely to a mixin and get it all out of the core

Hozu:
I'm pretty sure the bug is currently on only alastor antlion for whatever reason, despite worms having spells too.

PR #3134, opened 2016-05-04
Added files via upload

eraffxi:
SCRIPTS: added Shadows of the Departed quest

### DIFF
--- a/Zone.lua
+++ b/Zone.lua
@@ -1,0 +1,173 @@
+-----------------------------------
+--
+-- Zone: RuLude_Gardens (243)
+--
+-----------------------------------
+package.loaded["scripts/zones/RuLude_Gardens/TextIDs"] = nil;
+-----------------------------------
+require("scripts/zones/RuLude_Gardens/TextIDs");
+require("scripts/globals/missions");
+require("scripts/globals/quests");
+
+-----------------------------------
+-- onInitialize
+-----------------------------------
+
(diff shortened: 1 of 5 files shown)


Zone.lua
```diff
@@ -0,0 +1,173 @@
+-----------------------------------
```
TeoTwawki: You have this file in the root of the repository instead of in `scripts/zones/RuLude_Gardens/`

npcs/_0gc.lua
```diff
@@ -0,0 +1,41 @@
+-----------------------------------
```
TeoTwawki: https://github.com/myera/darkstar/blob/master/npcs/_0gc.lua
Same for all your files..the npc subdir was left in root instead of in the zone when you made the commit.

npcs/qm6.lua
```diff
@@ -0,0 +1,40 @@
...
+-----------------------------------    
+    
+function onTrade(player,npc,trade)    
+ if (trade:hasItemQty(1720,1) and trade:getItemCount() == 1 and player:hasKeyItem(436)==false) then
+ player:addKeyItem(436); --add map
+ player:tradeComplete();
+ player:messageSpecial(KEYITEM_OBTAINED,436);
+ else
+ player:messageSpecial(NOTHING_HAPPENS);
+ end
```
xdemolish: ew indentation

PR #3144, opened 2016-05-06
Core: Implement Mog Wardrobe 2

Hozu:
-Added the ability to move items into/from Mog Wardrobe 2 via client.
-Added the ability to equip items in Mog Wardrove 2 manually and with equipset via client.
-Updated @delitem to account for the extra inventory container
-Removed unused var declaration as per request by @demolishx

### DIFF
--- a/scripts/commands/delitem.lua
+++ b/scripts/commands/delitem.lua
@@ -24,13 +24,13 @@ function onTrigger(player, itemId, target)
 
     local targ = GetPlayerByName(target);
     if (targ ~= nil) then
-        for i = 0, 9 do -- inventory locations, see item_location.h
+        for i = 0, 10 do -- inventory locations, see item_location.h
             if (targ:hasItem(itemId, i)) then
                 targ:delItem(itemId, 1, i);
                 player:PrintToPlayer(string.format("Item with ID %u deleted from player '%s'.", itemId, target));
                 break;
             end
-            if (i == 9) then -- 9 being last inventory location, if it reaches this point then the player does not have the item anywhere.
+            if (i == 10) then -- 9 being last inventory location, if it reaches this point then the player does not have the item anywhere.
                 player:PrintToPlayer(string.format("Player '%s' does not have item with ID %u", target, itemId));
             end
         end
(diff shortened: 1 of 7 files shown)


teschnei:
Reorder the inventory loading like retail does as well

Hozu:
What's the order?

xdemolish:
who even is @demolishx

teschnei:
Move wardrobe 1 and 2 before the rest

MalRD:
Perhaps you should also remove the inventory related magic numbers and replace them with defined values while you're touching this. Maybe then when SE add wardrobe 3 it'll be simple enough that @xdemolish will be able to do it.

Hozu:
So it should be... inventory, wardrobe, wardrobe 2, satchel, sack, case, safe, safe 2, storage, locker? And where does temp items go, before wardrobe?

Will that mean I need to change the order in every single file I modified? Including renumber the locations? ie. LOC_INVENTORY = 0, LOC_WARDROBE = 1?

Hozu:
Also I'm not sure what you mean @MalRD

teschnei:
No, the location IDs are from the client. You can't change them. I can
handle it later if you don't know what to do

Hozu:
I honestly don't know what to do, sorry. :/

teschnei:
Don't worry about it, I'll get it tonight if I remember

Hozu:
@MalRD I already replaced the couple I came across in packet_system.cpp (that was the reason I was having so much trouble finding how to actually be able to equip items from wardrobe 2...). Were there more to change?

Here: https://github.com/DarkstarProject/darkstar/pull/3144/files#diff-8c85289593f4b314f43880f99da4baf5R2339 and here: https://github.com/DarkstarProject/darkstar/pull/3144/files#diff-8c85289593f4b314f43880f99da4baf5R2365 is what I found.

MalRD:
I only saw the delitem.lua values still at 10, but it is possible that they're not used anywhere else in the Lua code. The packet_system ones were the only others that I could think of, but you got those.

Hozu:
@MalRD it goes from 0-10, which is for all 11 of the slots. If there's an easy way to get a core enum for the scripts, I don't know it...

PR #3150, opened 2016-05-08
Core: Reduce cone angle for Overwhelm

Hozu:
Because 90 degrees from centre in both directions is not a cone - it's a rectangle. Seriously, you shouldn't be getting the bonus while beside the mob.

23 degrees is the same angle that is used for Ambush, seems reasonable.

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -3646,7 +3646,7 @@ namespace battleutils
         if (m_PChar->objtype == TYPE_PC) // Some mobskills use TakeWeaponskillDamage function, which calls upon this one.
         {
             // must be facing mob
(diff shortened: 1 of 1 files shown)


src/map/utils/battleutils.cpp
```diff
@@ -3646,7 +3646,7 @@ namespace battleutils
         if (m_PChar->objtype == TYPE_PC) // Some mobskills use TakeWeaponskillDamage function, which calls upon this one.
         {
             // must be facing mob
-            if (isFaceing(PDefender->loc.p, m_PChar->loc.p, 90))
+            if (isFaceing(PDefender->loc.p, m_PChar->loc.p, 23))
```
xdemolish: https://www.youtube.com/watch?v=-iy47Q--EbU&t=35s

PR #3152, opened 2016-05-08
Core/Scripts: Implement the rest of the Innin/Yonin directional checks

Hozu:
-Crit hit bonus/acc bonus from behind with Innin, Enemy crit hit reduction/evasion bonus while in front with Yonin. Note the Enemy crit hit rate bonus was previously missing. Using 23 degrees to match Sneak Attack (seems reasonable). Affects standard melee attacks, physical weaponskills (and ranged for the evasion bonus), and physical mobskills. Note that mobskills can't crit currently...
-Moved the Innin damage bonus to Ninjutsu magic from the individual spell script into doNinjutsuNuke, and changed the angle from 15 to 23.
-Added optional parameter for angle to the isFacing lua binding.
-Changed offsetAccuracy in getHitRateEx from uint8 to int8 to prevent underflow, fixes a bug with Closed Position as well.
-Added enemy crit hit rate merit reduction into weaponskills.lua since there are a handful of mobskills that use it. Also added the entire crit hit rate mod + crit hit rate merits - enemy crit hit rate merits into ranged weaponskill crit hit rate calculation.

### DIFF
--- a/scripts/globals/magic.lua
+++ b/scripts/globals/magic.lua
@@ -1130,6 +1130,12 @@ function doDivineNuke(V,M,caster,spell,target,hasMultipleTargetReduction,resistB
 end
 
 function doNinjutsuNuke(V,M,caster,spell,target,hasMultipleTargetReduction,resistBonus,mabBonus)
+    mabBonus = mabBonus or 0;
+
+    mabBonus = mabBonus + caster:getMod(MOD_NIN_NUKE_BONUS); -- "enhances ninjutsu damage" bonus
+    if (caster:hasStatusEffect(EFFECT_INNIN) and caster:isBehind(target, 23)) then -- Innin mag atk bonus from behind, guesstimating angle at 23 degrees
+        mabBonus = mabBonus + caster:getStatusEffect(EFFECT_INNIN):getPower();
+    end
     return doNuke(V,M,caster,spell,target,hasMultipleTargetReduction,resistBonus,NINJUTSU_SKILL,MOD_INT,mabBonus);
 end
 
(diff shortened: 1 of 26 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -8234,10 +8234,15 @@ inline int32 CLuaBaseEntity::isFacing(lua_State *L)
     DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isuserdata(L, 1));
 
     CLuaBaseEntity* PLuaBaseEntity = Lunar<CLuaBaseEntity>::check(L, 1);
+    uint8 angle = 45;
```
xdemolish: angle = condition ? value_if_true : value_if_false;

scripts/globals/magic.lua
```diff
@@ -1130,6 +1130,14 @@ function doDivineNuke(V,M,caster,spell,target,hasMultipleTargetReduction,resistB
 end
 
 function doNinjutsuNuke(V,M,caster,spell,target,hasMultipleTargetReduction,resistBonus,mabBonus)
+    if (mabBonus == nil) then
```
xdemolish: can also ``mabBonus = mabBonus or 0;`` here instead (you dont need to change this if you dont want to, just for future reference)

scripts/globals/weaponskills.lua
```diff
@@ -347,9 +358,15 @@ function getRangedHitRate(attacker,target,capHitRate,bonus)
...
     local eva = target:getEVA();
 
-    if (bonus) then
-        acc = acc + bonus;
+    if (bonus == nil) then
+        bonus = 0;
     end
+    if (target:hasStatusEffect(EFFECT_YONIN) and target:isFacing(attacker, 23)) then -- Yonin evasion boost if defender is facing attacker
+        bonus = bonus - target:getStatusEffect(EFFECT_YONIN):getPower();
+        print(bonus);
```
xdemolish: admiring the dedication to testing

PR #3154, opened 2016-05-09
SQL: Item_Basic Update AH categories

Ninjistix:
Update AH category to remove some Ex,R/E,iLv Items from AH, some items were in wrong category, removed GM gear from ah category also


R3P0FFXI:
@Ninjistix where did you get the information that some of this should be removed from AH categories, im pretty positive some of the stuff you removed is sellable on the AH specifically the ilvl items.

Ninjistix:
Ffxiah and most of them were rare exclusive or exclusive I'm sure the ah
never sold them, and I'm sure judge gear wasn't meant to be on ah

Ninjistix:
When I get home I can triple check but the only items that might need to be confirmed is the ammo

R3P0FFXI:
Theres ammo and Voay stuff thats wrong now @Ninjistix which was correct before

Ninjistix:
fixed, and also fixing more categories based on BG FFXI WIKI

R3P0FFXI:
Thanks @Ninjistix looks good @xdemolish

PR #3157, opened 2016-05-11
Job variables to table

TeoTwawki:
because having to type JOB_WAR instead of just war for the GM commands is just dumb.

Todo: TABLE ALL OF THE THINGS AND STUFFS (o_O)?

May possibly contain the odd regex find+replace mistake, but I tried to make sure I got them all.

The scripts work as I have them here. `JOBS.WAR` instead of `JOB_WAR` the GM commands look different from that because input strings require extra handling and I made case not matter: war and WAR are the same in the GM commands now.


TeoTwawki:
amended to include some requires the scripts should have already had, and a package nil I had in for testing reasons

PR #3158, opened 2016-05-11
SQL: item_basic adjust ah values

Ninjistix:
adjusted ah values of normal and rare weapons based on https://www.bg-wiki.com/bg/Category:Weapon


Ninjistix:
added updates from may 2016 update -- https://github.com/Windower/Resources/blob/master/xml/items.xml?raw=true

Issue #3163, opened 2016-05-12
NORMAL_MOB_MAX_LEVEL_RANGE Lua Settings not functioning properly


Darkstorm87:
**_I have:_**

- [X] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30160329_1

**_Server Version_** (type `@revision` in game) **:** 914ce5203730fdca17614150936640d083a20431

**_Source Branch_** (master/stable) **:** stable

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

The following Lua settings do not seem to be applied accurately:
NORMAL_MOB_MAX_LEVEL_RANGE_MIN
NORMAL_MOB_MAX_LEVEL_RANGE_MAX

Within zoneutlis.cpp the following function doesn't properly check an NM flag

```
void LoadMOBList()
{
    uint8 normalLevelRangeMin = luautils::GetSettingsVariable("NORMAL_MOB_MAX_LEVEL_RANGE_MIN");
    uint8 normalLevelRangeMax = luautils::GetSettingsVariable("NORMAL_MOB_MAX_LEVEL_RANGE_MAX");

.
.
.

// Cap Level if Necessary (Don't Cap NMs)
if (normalLevelRangeMin > 0 && PMob->m_Type != MOBTYPE_NOTORIOUS && PMob->m_minLevel > normalLevelRangeMin)
{
     PMob->m_minLevel = normalLevelRangeMin;
}

if (normalLevelRangeMax > 0 && PMob->m_Type != MOBTYPE_NOTORIOUS && PMob->m_maxLevel > normalLevelRangeMax)
{
    PMob->m_maxLevel = normalLevelRangeMax;
}
```

The PMob->m_Type field should be doing a bitwise AND check instead of an equality check.  I believe the line should appear as

```
if (normalLevelRangeMin > 0 && !(PMob->m_Type & MOBTYPE_NOTORIOUS) && PMob->m_minLevel > normalLevelRangeMin)
```

This matters for Arena fights and may be applicable elsewhere to other NM types.  As an example of an Arena fight.

```
[...]

PR #3164, opened 2016-05-12
Phalanx II and NORMAL_MOB_MAX_LEVEL_RANGE corrections

Darkstorm87:
1) Phalanx II should have a hard cap.  The following link illustrates the skill necessary to pass 35 damage absorbed.
http://ffxiclopedia.wikia.com/wiki/Phalanx_II

2) NORMAL_MOB_MAX_LEVEL_RANGE_MIN and NORMAL_MOB_MAX_LEVEL_RANGE_MAX are not applied correctly

### DIFF
--- a/scripts/globals/spells/phalanx_ii.lua
+++ b/scripts/globals/spells/phalanx_ii.lua
@@ -28,10 +28,6 @@ function onSpellCast(caster,target,spell)
 
     final = (enhskill / 25) + merits + 1;
 
-    if (final>35) then
-        final = 35;
-    end
-
     if (target:addStatusEffect(EFFECT_PHALANX,final,0,duration)) then
         spell:setMsg(230);
     else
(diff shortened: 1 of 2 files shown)


Hozu:
Curious, what was the issue with the setting to cap the mob levels? The server I play on has it working fine (I specifically checked the levels of normal mobs and NMs).

Darkstorm87:
I outlined the issue in here:

https://github.com/DarkstarProject/darkstar/issues/3163

But basically, the server had those values set too low and it was artificially capping Battlefield NMs that had a mobtype of 18, which is a combination of MOBTYPE_NOTORIOUS and MOBTYPE_BATTLEFIELD.  Checking just != to MOBTYPE_NOTORIOUS was passing and allowing the NMs level to be capped.

teschnei:
It looks good, just waiting for the build to complete.

You should do your commits on a topic branch instead of master though, in case you ever want to pull request multiple things at once (and I do prefer separate things as separate requests, but I'll let it slide for this one since it's so small)

Issue #2311, opened 2015-10-23
BLM category 1 merits: unable to fetch


nesstea:
After doing some tests on damage calculations, I discovered that the BLM cat 1 merits aren't functioning with the player:getMerit() command. 

With an @test command, I found that this is returning 0.
```
player:PrintToPlayer("BLM Merit - THU potency:   " .. player:getMerit(MERIT_LIGHTNING_MAGIC_POTENCY));
```
I verified that the status.lua and the merits.h are correctly named the same. They are in the database correctly as well. magic.lua is defining the merits in the beginning by the array blmMerit and is attempting to add the merit rank or MAB to the formula.
```
        local mab = caster:getMod(MOD_MATT) + bonusmab;
        if (spell:getElement() > 0 and spell:getElement() <= 6) then
            mab = mab + caster:getMerit(blmMerit[spell:getElement()]);
        end
```

Hozu:
That... explains everything. I can verify this as well.

Hozu:
Well this is strange, while the value can't be fetched on the server I play on, it works on my test server. Unlikely to be custom code causing the problem if another server is having the same issue...

nesstea:
I just rebuilt.

Hozu:
Rebuilt the server you play on? Yeah, same thing with the one I play on, still doesn't work.

m241dan:
Any movement on this?

rettahs:
Tested these merits today, do not function still.

Darkstorm87:
I'm like 99% confident that this is an issue with the following code

```
int32 CMeritPoints::GetMeritValue(MERIT_TYPE merit, CCharEntity* PChar)
{
    Merit_t* PMerit = GetMeritPointer(merit);
	uint8 meritValue = 0;

    if (PMerit)
    {
        if (PMerit->catid < 5 || (PMerit->jobs & (1 << (PChar->GetMJob() - 1)) && PChar->GetMLevel() >= 75))
            meritValue = dsp_min(PMerit->count, cap[PChar->GetMLevel()]);

        if (PMerit->catid == 8 && PChar->GetMLevel() < 96)
            meritValue = 0;

        meritValue *= PMerit->value;
    }

	return meritValue;
}
```

Specifically this if Check

```
if (PMerit->catid == 8 && PChar->GetMLevel() < 96)
            meritValue = 0;
```

[...]

PR #3165, opened 2016-05-12
Core: Fix BLM T1 merit fetching

Hozu:
The check to disable merit WS merits below level 96 was using the wrong category ID. Fixes https://github.com/DarkstarProject/darkstar/issues/2311


Hozu:
And the test: http://i.imgur.com/AaZk6RE.png

PR #3167, opened 2016-05-12
Misc corrections for things we don't even use yet.

TeoTwawki:
Animation IDs and case/name corrections. The animations have been verified.

Also took the liberty of changing the name field to varchar like other tables that have variable fields.

I'd have scripts to go with these, but the server using them (Legion) heavily customized all the damage and effects so those aren't retail right now.

### DIFF
--- a/sql/mob_skills.sql
+++ b/sql/mob_skills.sql
@@ -29,7 +29,7 @@ DROP TABLE IF EXISTS `mob_skills`;
 CREATE TABLE IF NOT EXISTS `mob_skills` (
   `mob_skill_id` smallint(4) unsigned NOT NULL,
   `mob_anim_id` smallint(4) unsigned NOT NULL,
-  `mob_skill_name` char(25) CHARACTER SET latin1 NOT NULL,
+  `mob_skill_name` varchar(40) CHARACTER SET latin1 NOT NULL,
   `mob_skill_aoe` tinyint(1) unsigned NOT NULL DEFAULT '0',
   `mob_skill_distance` float(3,1) NOT NULL DEFAULT '6.0',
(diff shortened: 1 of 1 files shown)


sql/mob_skills.sql
```diff
@@ -438,7 +438,7 @@ INSERT INTO `mob_skills` VALUES (630,374,'Kick_Out',4,10.0,2000,1500,4,0,0,0);
 INSERT INTO `mob_skills` VALUES (631,375,'Shock_Wave',4,7.0,2000,1500,4,0,0,2);
 INSERT INTO `mob_skills` VALUES (632,376,'Flame_Armor',0,7.0,2000,1500,1,0,0,0);
 INSERT INTO `mob_skills` VALUES (633,354,'Howl',1,20.0,2000,1500,1,0,0,0);
--- INSERT INTO `mob_skills` VALUES (634,378,'Meteor',0,7.0,2000,1500,4,0,0,0);
+-- INSERT INTO `mob_skills` VALUES (634,2256,'Meteor',0,7.0,2000,1500,4,0,0,0);
```
teschnei: 378 was the right animation here (it is chlevnik's death meteor)

TeoTwawki: explains why when I checked that it looked like humping the ground, must be specific to that one NM I presume. Amend incoming.

TeoTwawki: Amended now. The other animation ID I had is for animating a regular non _hump the ground and die_, use of meteor.

teschnei: not sure who uses a mobskill version of meteor that just looks like a
regular meteor..

PR #3168, opened 2016-05-12
change killer checking back to a bool

TeoTwawki:
Broken into 3 commits for easier reviewing.

1.  core change

2.  global script change (onMobDeathEx)

3.  mob scripts (onMobDeath)

Dear past me: near-present me apologizes for second guessing you when we tried to fix it before

Scripters: `mobs.lua` has a fine example of how to make things only hit happen to the actual killer. The function is identical except for the weaponskill check.


TeoTwawki:
p.s. :
Tested party alliance and solo, tested all arguments, made sure didn't F up stack, made sure every behaved as expected in both mob's function and global and made a test case for things that don't even exist yet just to be 100% sure.

TeoTwawki:
if I still managed to derp something, shoot me.

Hozu:
I hate you for making me have to change all of my custom scripts, yet again.

teschnei:
time to learn bufdo ;)

src/map/lua/luautils.cpp
```diff
@@ -2714,18 +2714,18 @@ namespace luautils
                     }
 
                     CLuaBaseEntity LuaMobEntity(PMob);
-                    CLuaBaseEntity LuaKillerEntity(PChar);
                     CLuaBaseEntity LuaAllyEntity(PMember);
-
+                    bool isKiller = (PMember == PChar) ? true : false;
```
teschnei: bool isKiller = PMember == PChar;

src/map/lua/luautils.cpp
```diff
@@ -2714,18 +2714,18 @@ namespace luautils
...
                     Lunar<CLuaBaseEntity>::push(LuaHandle, &LuaMobEntity);
-                    Lunar<CLuaBaseEntity>::push(LuaHandle, &LuaKillerEntity);
                     Lunar<CLuaBaseEntity>::push(LuaHandle, &LuaAllyEntity);
+                    lua_pushboolean(LuaHandle, isKiller);
+
                     lua_pushboolean(LuaHandle, isWeaponSkillKill);
                     // lua_pushboolean(LuaHandle, isMagicKill);
                     // lua_pushboolean(LuaHandle, isPetKill);
-                    // Upcoming AI rewrite will likely have a better way to handle it..
+                    // Todo: look at better way do do these than additional bools...
```
teschnei: we'll probably just add an event to be triggered under those conditions rather than jamming it into here when we need it

TeoTwawki:
@Hozu I have find+replace instructions ready to post to dsp forums when this gets merged.

teschnei:
I can't tell if you updated on my comment or not (too many changes to view the full diff rather than just the single commit)

src/map/lua/luautils.cpp
TeoTwawki: @teschnei hadn't actually see it yet just woke aup. And yeah I dunno why I did that the way I did...

TeoTwawki: @teschnei fixed now.

TeoTwawki:
@Hozu I went and lost the regex I used last night >.< but heres how to do it in multiple non regex find+replace:

Find this string: ` ally:`
Replace with: ` player:`
Find this string: `(ally`
Replace with: `(player`
Find this string: `, ally`
Replace with: `, player`
Find this string: `,ally`
Replace with: `,player`
Find this string: `onMobDeath(mob,killer,ally)`
Replace with: `onMobDeath(mob, player, isKiller)`
Find this string: `onMobDeath(mob, killer, ally)`
Replace with: `onMobDeath(mob, player, isKiller)`
If `killer:` actually got used anywhere, that was busted to begin with.

TeoTwawki:
@teschnei  in case github didn't show you my earlier reply, I rewrote all 3 commits with the change.

PR #3170, opened 2016-05-12
SQL: item_weapon delay corrections, removed grip/strap

Ninjistix:
corrected h2h and ranged delays... removed straps,grips and 4 other items


Hozu:
Hm? What was wrong with h2h weapon delay? Seems like a fairly significant increase there.

teschnei:
It looks to me like he screwed them up in a previous PR (or someone else
did). O checked a 2014 version of the file and they were all 500ish

Hozu:
Ooh, I see.

Ninjistix:
Those were the original values before my last commit they h2h delay is 480
plus whatever is on the weapon, i think I imported SE delays for h2h. users
on our sever reported h2h weapons with no delay and having hundred fists
all the time

TeoTwawki:
Why were grips and straps there, I already removed had warned everyone making a PR since that they don't belong in the weapon table. :panda_face: The panda is sad. People need to stop upsetting the Panda.

sql/item_weapon.sql
```diff
@@ -1116,54 +1113,54 @@ INSERT INTO `item_weapon` VALUES (17468,'raise_rod',11,0,0,0,0,3,1,288,10,0);
...
-INSERT INTO `item_weapon` VALUES (17513,'hyd.baghnakhs_+1',1,0,0,0,0,4,1,295,8,0);
-INSERT INTO `item_weapon` VALUES (17514,'hydro_claws',1,0,0,0,0,4,1,306,10,0);
-INSERT INTO `item_weapon` VALUES (17515,'hydro_claws_+1',1,0,0,0,0,4,1,301,11,0);
-INSERT INTO `item_weapon` VALUES (17516,'hydro_patas',1,0,0,0,0,4,1,336,19,0);
-INSERT INTO `item_weapon` VALUES (17517,'hydro_patas_+1',1,0,0,0,0,4,1,326,20,0);
-INSERT INTO `item_weapon` VALUES (17518,'manoples_+1',1,0,0,0,0,4,1,346,24,0);
-INSERT INTO `item_weapon` VALUES (17519,'platoon_cesti',1,0,0,0,0,4,1,288,7,0);
+INSERT INTO `item_weapon` VALUES (17472,'cross-counters',1,0,0,0,0,4,1,576,22,0);
+INSERT INTO `item_weapon` VALUES (17473,'torama_cesti',1,0,0,0,0,4,1,520,9,0);
+INSERT INTO `item_weapon` VALUES (17474,'grapnel',0,0,0,0,0,0,1,0,3,0);
```
TeoTwawki: grapnel doesn't need to be here. its not an actual weapon, its equipment but its not used as a weapon.

TeoTwawki: I meant to say "looks like but isn't" equipment.

Ninjistix: removed it... anything else?

TeoTwawki: not right now since i assume th current icky naming matches item basic, but I did warn you about punctuation. the name in item basic sets the script name. while 90% of these will never need a script, when one is needed having `.` `,` `;` or `:` screws those. So cut that out.

So if you change the naming in item basic _again_ remember that..

TeoTwawki: Pardon, only half awake. These are matching the short-name in item basic yeah? They are fine and I don't think we even use this name column in item_weapon (not sure). The long name in item_basic is the one that matters, some folk were annoyed at colon's in there the other day so I remembered it. I'll stfu now.

Ninjistix: No big deal and yes this matches short name from item basic

PR #3173, opened 2016-05-14
Added Argus & Leech King pop mechanism

ymkc:
- only one of them can be up (randomized)
- spawn window open after "maintenance"/server restart
- added nm spawn points
- removed Argus' name_prefix

reference: http://ffxi.gamerescape.com/wiki/Argus


Hozu:
Weren't they separated at some point with new retail?

TeoTwawki:
@hozu my memory if it is they used to be and then were combined afterward just to make us all groan even more about Argus's pop..


Also: nuuuuuuuuuuuuu now how will I exploit argus' pop after every server crash! :cry:

scripts/zones/Maze_of_Shakhrami/Zone.lua
```diff
@@ -21,7 +22,13 @@ function onInitialize(zone)
     SetGroundsTome(tomes);
 
     UpdateTreasureSpawnPoint(17588769);
-    
+
+    local whichNM = math.random(0,19);
+    if (whichNM < 10) then
+        SetRespawnTime(Argus, 0, 43200); -- 0-12 hours 
+    else
```
Hozu: You need to use DeterMob here too, false for the one that will spawn, true for the other one.

scripts/zones/Maze_of_Shakhrami/mobs/Argus.lua
```diff
@@ -0,0 +1,23 @@
...
+
+function onMobDeath(mob,killer,ally)
+    local whichNM = math.random(0,19);
+
+    if (whichNM < 10) then
+        SetRespawnTime(Argus, 64800, 108000); -- 18-30 hours
+        DeterMob(Leech_King, true);
+    else
+        SetRespawnTime(Leech_King, 64800, 108000); -- 18-30 hours
+        DeterMob(Argus, true);
```
Hozu: You need to do DeterMob(Leech_King, false); here as it will not spawn otherwise.

scripts/zones/Maze_of_Shakhrami/mobs/Leech_King.lua
```diff
@@ -0,0 +1,23 @@
...
+-----------------------------------
+-- onMobDeath
+-----------------------------------
+
+function onMobDeath(mob,killer,ally)
+    local whichNM = math.random(0,19);
+
+    if (whichNM < 10) then
+        SetRespawnTime(Argus, 64800, 108000); -- 18-30 hours
+        DeterMob(Leech_King, true);
```
Hozu: Same here, except with Argus. DeterMob(Argus, false);

scripts/zones/Maze_of_Shakhrami/mobs/Argus.lua
```diff
@@ -0,0 +1,23 @@
...
+-----------------------------------
+-- onMobDeath
+-----------------------------------
+
+function onMobDeath(mob,killer,ally)
+    local whichNM = math.random(0,19);
+
+    if (whichNM < 10) then
+        SetRespawnTime(Argus, 64800, 108000); -- 18-30 hours
+        DeterMob(Leech_King, true);
```
Hozu: Come to think of it, if you're deterring the mob in the zone.lua on boot, you don't need to deter the other one again on death. Since Leech King is already deterred if you defeat Argus.

scripts/zones/Maze_of_Shakhrami/mobs/Leech_King.lua
```diff
@@ -0,0 +1,23 @@
...
+    else
+        SetRespawnTime(Leech_King, 64800, 108000); -- 18-30 hours
+        DeterMob(Argus, true);
```
Hozu: Same deal, no need to deter Argus here if it's already set to be deterred in the zone.lua.

scripts/zones/Maze_of_Shakhrami/Zone.lua
ymkc: I hope I got this right, because according to luautils.cpp:

/************************************************************************
    * Set SpawnType of mob to scripted (128) or normal (0) usind mob id     *
    int32 DeterMob(lua_State* L)

I set both of them to 128 in the db so I don't have to call DeterMob here

scripts/zones/Maze_of_Shakhrami/mobs/Argus.lua
ymkc: Tested and not neccessary, as CLuaBaseEntity::setRespawnTime per default sets

PMob->m_AllowRespawn = true;

I know it looks a bit short/incomplete but I tested alot and checked the underlying functions, so it should work as expected

scripts/zones/Maze_of_Shakhrami/mobs/Leech_King.lua
ymkc: But onZoneInitialize is only called once, so if you have more than one kill you need to deter the other mob. Or am I off on this?

ymkc:
I can't find any source saying that they can now be up at the same time, wikia only states that they reduced the repop time to around one hour (like on many other NMs)

scripts/zones/Maze_of_Shakhrami/Zone.lua
xdemolish: three backticks before and after a block of code on github will help make your comment more readable :)

scripts/zones/Maze_of_Shakhrami/mobs/Leech_King.lua
Hozu: You're right about onZoneInitialize. If you defeat the mob that was chosen on boot, and it gets selected again, there's no need to change what mob is deterred - only need to set the respawn time. If it's different, you deter the mob you just killed, and have the other mob not be deterred.

scripts/zones/Maze_of_Shakhrami/mobs/Argus.lua
Hozu: Oh yeah... there's an optional argument that would not change that. Carry on.

scripts/zones/Maze_of_Shakhrami/Zone.lua
ymkc: @xdemolish thanks

scripts/zones/Maze_of_Shakhrami/mobs/Leech_King.lua
ymkc: @Hozu my idea was to always deter one mob, so there is no way they will be up at the same time

I can change the logic if it's the better/prefered way, just thought those calls 2 calls every 12+ hours wouldn't matter much :)

ymkc: @Hozu took a look at it, but this would make the logic more complex
the problem is, that the respawn time changes (from 0-12h to 18-30h) after the first kill of each NM, so you get more checking + 2x deter/1xsetRespawn instead of no checking and 1x setRespawn + 1x deter

bendangelo:
Hey, please fix the conflicts. If there are no more issues I'll like to get this merge in.

ymkc:
@bendangelo conflicts resolved

bendangelo:
If anyone has a problem with this, I'll fix it. Thanks for contributing :heart:

PR #3175, opened 2016-05-15
SCRIPTS/SQL: May Update - Batch Id Updates

starlightknight:
Usual batch updates

-Text ID Updates - handful of shifts in zones
-NPC ID Updates - updates in the following zones:

Nyzul Isle
West Ronfaure
South Gustaberg
West Sarutabaruta
Southern San d'Oria
Bastok Markets
Windurst Waters

-Lua ID Updates - minor updates


starlightknight:
@gedads @TeoTwawki Need some help as a follow-up in the following zones:

West Ronfaure
South Gustaberg
West Sarutabaruta
Southern San d'Oria
Bastok Markets
Windurst Waters

SE has shuffled a fair number of npcs around here, and these zones have gotten out of sync to the point that the batch cannot accurately identify where many of the NPCs should be shifted to. This is largely because there are many NPCs in these zones that do not have a POLUtils name, meaning the batch has nothing to compare against. 

In this case, what the batch normally does is compare the last surrounding NPCs that it could concretely identify in POLUtils, and figure out how it should be shifted based on a combination of how much the last identified npcs went, combined with ids not being allowed to be reused. 

Normally this has worked very well for years now, but in order for this to work accurately, the NPCs in the SQL file need to be kept in sequential order. These zones - and in particular, the Windurst ones - are a mess in this regard. I have spent a fair amount of time analyzing the problems and have fixed as many as could be concretely identified, but I need someone to help and do the following:

1) Recap these zones and make sure the NPCs with no POLUtils name are correct so that cutscenes are not borked. Since this has problem has been growing for awhile now, its near impossible to tell whats in or not in sync in these areas
2) Place the npcs in these zones in sequential order when doing so
3) When new npcs are added, event or otherwise, please add them in sequential order instead of in new chunks at the bottom of the zone

This will ensure smooth and painless batch updates continue to happen in the future. If no one is willing to help with this, then these zones will have to be dropped from the batch update as they cannot be auto updated correctly in the current state (which will lead to the out of sync id complaints of old).

In the meantime - merging this will should fix more than it breaks, so I recommend doing so for now. All files load successfully.

TeoTwawki:
I also need to go through Abyssea-la'Theine..Somethings off there atm, I probably won't be able to get to any of this till the 20th because of moving.


Aaaand just quoting for emphasis:
> **in order for this to work accurately, the NPCs in the SQL file need to be kept in sequential order.**

COUGH COUGH

gedads:
probably some festival going on in starter cities
about the mess in starter cities if i look one year back https://raw.githubusercontent.com/starlightknight/darkstar/ff0d6dd87ec6b3d365cf60ec90417ac355f4b74c/sql/npc_list.sql
all are in order, i don't think npcs were added since them in those area. i'll check npc_list soon compared to previous version to see were npcs are misplaced over time.
nyzul they added the master content (which i can enter and capture too but not alone).

Also that reminds me some event where SE are repoving temporarily npcs to add other then add them back in some update later which was really boring when i was updating npc_list manually, i can't tell if it's something like this i haven't really checked npc_list for a long time now since it's automated.

starlightknight:
@gedads @TeoTwawki 

Yeah its usually the event areas that get messed up, thus these few zones - everything else is generally good elsewhere. 

The event stuff seems to get grouped up toward the end of the zone and while they were at one time also in sequential order, SE likes to move them around - and then at some point, some non event npcs got interwoven as well. It is hard to correlate some of the current blocks versus the current POLUtils file.

The csnpcs, blanks, and common npc names like moogles, doors, and internal names are the ones that are potentially likely to be messed up. If the blocks are out of sequence, the wrong id number may be chosen since the batch only has visibility into the name of the npc. Many npcs match those names, and its scanning forward from the start of the zone - there is no positional data or anything to compare with. This can have cascading effects on the remainder of the zone since ids aren't allowed to be reused.

If we are able to line them up so that they're in sync and sequential, then the approach the batch uses should keep them that way. I started trying to manual fix as much as i could identify versus current polutils, but eventually i hit a wall and gave up after several hours at it - guessing at POLUtils on blanks won't lead to good results :-(

As for them being misplaced - some of it is probably my fault for not noticing this issue early on in the life of the batch process over the past two years. It wasn't always apparent that this was a problem. Some of it may have just happened over time as well as new events and npcs were added, or corrected. Its hard to keep an eye on as any edit to npc is almost always too big to see the diff in github. SE also likes to do lots of reordering on these areas too - where the whole block, as labeled in npc, may switch places with another block of event npcs. After we sort this problem out and are sure the ids are in sync, I will add a step to the batch to make sure they're ordered automatically before and after the compare so that we can hopefully keep this from happening again.

As far as ones that the batch isn't sure about - i can make it add like a -- FIXME comment to the end of the line or something if that works

Also: Any NPC that has a real polutils name, and whose POLUtils_name column matches the NPC name in this commit is guaranteed to be set to the right id right now (assuming the positional data didn't get borked over time for something like a moogle). I have the batch rescrape these values as a cross check after shifting, so that anything that was not done correctly is shown in the diff I review. This covers the vast majority of non-event npcs being correct. It's just the blanks, csids, etc and event npcs that are messy since they can't be verified this way.

Thanks for any help :-)

gedads:
hi, 
FIXME would be perfect. honestly, i don't think it's that hard to correct once they're explicitely displayed as wrong with fixme. Overall your tool is magic, saved a lot of time and removed the really annoying task  to manually update so whatever you need to make it even better i'm ok to help (mostly by capturing again area i suppose).

PR #3176, opened 2016-05-15
Stop Jailers of Prudence, Hope, Justice, and Love from instantly despawning the second they go unclaimed.

TeoTwawki:
Note the comment in Jailer of Prudence's script : I don't have time to research this before I go offline for the weekend, but its no more broken than before I touched it and can easily be adjusted.

If anyone has any input before tomorrow afternoon I can do a quick amend for this if needed.

### DIFF
--- a/scripts/zones/AlTaieu/mobs/Jailer_of_Prudence.lua
+++ b/scripts/zones/AlTaieu/mobs/Jailer_of_Prudence.lua
@@ -1,13 +1,12 @@
 -----------------------------------
 -- Area: Al'Taieu
--- NM:  Jailer of Prudence
+--  NM:  Jailer of Prudence
 -- IDs: 16912846, 16912847
 -- AnimationSubs: 0 - Normal, 3 - Mouth Open
 -- Wiki: http://ffxiclopedia.wikia.com/wiki/Jailer_of_Prudence
 -----------------------------------
-
-require("scripts/globals/status");
 require("scripts/zones/AlTaieu/mobIDs");
+require("scripts/globals/status");
 
 -----------------------------------
 -- onMobInitialize Action
@@ -77,6 +76,13 @@ end;
 -----------------------------------
 
 function onMobDeath(mob, player, isKiller)
+end;
+
+-----------------------------------
+-- onMobDespawn
+-----------------------------------
+
+function onMobDespawn(mob)
     local firstPrudence     = GetMobByID(PrudenceOne);
     local secondPrudence    = GetMobByID(PrudenceTwo);
     if (mob:getID() == PrudenceOne) then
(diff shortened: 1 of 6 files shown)


scripts/zones/AlTaieu/mobs/Jailer_of_Prudence.lua
```diff
@@ -90,4 +89,12 @@ function onMobDeath(mob, player, isKiller)
...
     end;
-end;
\ No newline at end of file
+end;
+
+-----------------------------------
+-- onMobDedpawn
+-----------------------------------
+
+function onMobDedpawn(mob)
```
xdemolish: dedpawn

TeoTwawki: I imagine you delivering this news in your best deadpan snarker voice.

Deadwing888:
Right, so here's how Jailer of Prudence works:

Prudences will spawn with 60 seconds of flee, but if the popper pops from outside the alliance and manages to flee all the way to the southern updraft to zone, the Prudence that was following them will begin pathing slowly back to the ??? and will depop after 3 minutes.

Once it depops, the other prudence will open its mouth and gain 100% triple attack. If it manages to wipe the claimed alliance in this state it will head to where the other prudence depopped, and will itself depop as soon as it gets there.

In every other circumstance they are sharing hate right up until the wipe at which point they will naturally depop together at the ???

TeoTwawki:
@Deadwing888 thanks for clearing that up. I've moved the mouth open logic from death to despawn, since despawn will be called after death this covers both scenarios.

scripts/zones/AlTaieu/npcs/qm2.lua
```diff
@@ -1,25 +1,26 @@
...
     -- Trade the Second Virtue, Deed of Moderation, and HQ Xzomit Organ
-    --[[if (GetMobAction(16912839) == 0 and trade:hasItemQty(1853,1) and trade:hasItemQty(1854,1) and trade:hasItemQty(1785,1) and
+    if (GetMobAction(16912839) == 0 and trade:hasItemQty(1853,1) and trade:hasItemQty(1854,1) and trade:hasItemQty(1785,1) and
     trade:getItemCount() == 3) then
         player:tradeComplete();
-        SpawnMob(16912839,900):updateClaim(player); -- Spawn Jailer of Justice
-    end]]
+        SpawnMob(16912839):updateClaim(player); -- Spawn Jailer of Justice
+    end
+    ]]
```
xdemolish: why's this commented out? better to just remove if it's wrong/unneeded

TeoTwawki: Yeah I just moved the placement of the comment block to be what I thought was more legible/less likely to have its indents f\*cked later

TeoTwawki:
*nudge*

Issue #3120, opened 2016-05-01
Ranged monsters periodically attack players after respawning, even though they have not been aggroed/engaged.


dazusu:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160329_1

**_Server Version_** (type `@revision` in game) **:**
bdfb082

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
This seems to still be an issue, but only if the mob in question has been killed - then respawns; then the random shooting behavior continues.

PR #3179, opened 2016-05-16
added distanceSquared function 

eraffxi:
to allow for lighter distance checks against squared values rather than finding the square root in distance()


teschnei:
where would you use this?

src/common/utils.cpp
```diff
@@ -79,6 +79,27 @@ float distance(const position_t& A, const position_t& B)
 *																		*
 ************************************************************************/
 
+float distanceSquared(const position_t& A, const position_t& B)
+{
+	float one = 0, two = 0, three = 0, four = 0;
```
xdemolish: distance_x, distance_y, distance_z

xdemolish:
@teschnei 
```
<warp> so my friend took a look at the mobroam stuff and said the distance checking is not as optimized as it could be
<warp> he says instead of doing a square root in the distance check, it is way less work for a cpu to square a number than find the square root
<warp> he suggested doing a distanceSquared() function that just returns an absolute value and checking it against a squared version of the original check
<warp> so distanceSquared() <= 100 instead of distance() <= 10
```

eraffxi:
i honestly am not 100% sure the best way to use it and am testing it with the distance checks for mob skills by using
          
  if (currentDistance <= pow(PMobSkill->getDistance(), 2))
  where currentDistance = distanceSquared(PMob->loc.p, PActionTarget->loc.p);

but i'm not sure if that's actually faster or slower than just doing PMobSkill->getDistance() * PMobSkill->getDistance()

max said he's in favor of it so i figured i'd offer it to be used by someone who might know faster than me how best to use it

teschnei:
he's probably right, but why does it matter?

eraffxi:
i'm using it because the analysis i took showed 50-60% of all cpu time was spent on mobroam and hozu suggested it might be due to checking distance for ranged attacks, so i'm testing that.  since max said 'no reason not to use it really' i pushed this for him

teschnei:
if that's actually what the problem is, then we can consider using this (or something similar to this that's easier to use)
but you need to make sure that's actually the problem, because I can't realistically see that being the case (why would a mob be checking for ranged attacks while roaming?)

Deadwing888:
Mobs also don't roam if no one is in the zone. That's not retail but I assume it was a choice made to conserve processing power.

For example if you are the only person in the zone and you pull an NM 500' away from its spawn and log out, it'll still be right next to you when you log back in, no matter how long you wait.

Hozu:
The problem I mentioned is that every tick it's checking attempting to actually use the ability (provided it's not on cooldown) - if it doesn't have a target, there's no distance to check.

bendangelo:
Sure the function is faster and thanks for the work. If we want to improve the performence of mob roaming we have to do profiling. I've never done that before in C++

xdemolish:
@hozu thought i fixed that in https://github.com/DarkstarProject/darkstar/issues/3120 ?

Hozu:
Well they lose their target with the fix, but it doesn't stop them from keep trying to shoot. They just don't shoot with no target.

teschnei:
TrySpecialSkill fails nearly immediately if there's no target

 I'm just going to close this for now unless it can actually be proven (ie using visual studio's profiler) that this SPECIFICALLY is a bottleneck

PR #3180, opened 2016-05-17
dont calculate distance where there is none to be calculated

xdemolish:


teschnei:
why?

xdemolish:
cause checking distance on self triggered me

teschnei:
is it actually a problem?  have you profiled it?

xdemolish:
lolno that's what we have @Scavenge for

teschnei:
closing until it's proven to fix an issue

PR #3182, opened 2016-05-18
Mapconfig option additions

nekolepsy:
Hi Darkstar Team!
I've made 10 patches for dsp to add server customization features if you'd like them. Nine of the patches add options in conf/map_darkstar.conf and one adds a configuration option in scripts/globals/settings.lua . I consider the patches to be tested in the "Works for me!" state. Thanks for making such an awesome project!


src/map/entities/battleentity.cpp
```diff
@@ -733,7 +769,7 @@ void CBattleEntity::SetMLevel(uint8 mlvl)
 
 void CBattleEntity::SetSLevel(uint8 slvl)
 {
-    m_slvl = (slvl > (m_mlvl >> 1) ? (m_mlvl == 1 ? 1 : (m_mlvl >> 1)) : slvl);
+    m_slvl = (slvl > (m_mlvl * map_config.sj_level_ratio) ? (m_mlvl == 1 ? 1 : (m_mlvl * map_config.sj_level_ratio)) : slvl);
```
TeoTwawki: this won't work out in the math the way you think it will. I previously wrote one that will, but its not in the project because of additional work that needed done for the search list to show the correct subjob level at the time.

src/map/map.cpp
```diff
@@ -946,6 +946,9 @@ int32 map_config_default()
     map_config.craft_day_matters = 1;
     map_config.craft_moonphase_matters = 0;
     map_config.craft_direction_matters = 0;
+    map_config.craft_success_chance_multiplier = 1.0f;
```
TeoTwawki: is craft_sucess_chance_multiplier different from craft_chance_multiplier ?

teschnei:
separate pull requests for logically different things, please

src/map/map.cpp
nekolepsy: Yes. The already existing craft_chance_multiplier affects the chance of getting a skill up on a craft when a skill up is possible. My added craft_success_chance_multiplier affects the chance of successfully crafting. I can see how this is confusing, but I did try to name my modifier as close to what it does as possible.

src/map/entities/battleentity.cpp
nekolepsy: Thanks. I'll have to do some more testing on this one, unless you're going to do your patch instead.

nekolepsy:
Sorry about that. I figured since most of the patches affect the same files it would be easier to just make a big pull request rather than merge several branches. I'm still very new to using git and github. Do you need me to change the pull request for these patches?

teschnei:
yes - close this one, make a branch for each logically grouped setting (ie. all the crafting ones, etc) and cherry pick the commits from this branch to those ones, and then open separate pull requests for each one.  that way, we can pick which settings are useful and don't impact the server negatively and refuse the ones that aren't/do

src/map/map.cpp
TeoTwawki: its cool, just looked familiar

src/map/entities/battleentity.cpp
TeoTwawki: I tagged you on a gist that shows you, I don't mind if you want to pr it (github should have sent you a notification, get me on irc if it didn't). There is also my old forum post. https://forums.dspt.info/viewtopic.php?f=20&t=1651&hilit=settings+setting#p8107

I did the same thing as you in my first version of it.

PR #3184, opened 2016-05-18
Add item vendor sell multiplier configurable option

nekolepsy:
This adds the option to modify the price that vendors will pay for items.


whasf:
Please re-push and clean up.

PR #3187, opened 2016-05-18
Casting configuration patches

nekolepsy:
Add spell cast/recast time configurable options
    Create options for nm, mob, and player cast and recast time multipliers.

Add cast interrupt multiplier configurable options
    Create options for nm, mob, and player cast interruption rate

### DIFF
--- a/conf/map_darkstar.conf
+++ b/conf/map_darkstar.conf
@@ -121,6 +121,16 @@ nm_stat_multiplier:     1.0
 mob_stat_multiplier:    1.0
 player_stat_multiplier: 1.0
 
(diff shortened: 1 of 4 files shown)


teschnei:
i don't really like the interrupt rate modifier - it's a lot of work for something very few will use.  if it was just implemented in addition to the existing modifier, it'd be a lot more manageable.

the recast options make no sense - if they are going to still be subject to the fast cast caps, why don't they just add fast cast?

whasf:
Closing, please re-submit after suggestions have been implemented and cleaned up.

thanks,

PR #3188, opened 2016-05-18
Attacking configuration patches

nekolepsy:
Add attack delay multiplier configurable options
    Create options for nm, mob, and player attack delay multiplier

Add tp cost multiplier configurable option
    Create option to adjust the tp spend cost of weapon skills.

### DIFF
--- a/conf/map_darkstar.conf
+++ b/conf/map_darkstar.conf
@@ -121,6 +121,11 @@ nm_stat_multiplier:     1.0
 mob_stat_multiplier:    1.0
 player_stat_multiplier: 1.0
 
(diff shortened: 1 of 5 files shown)


teschnei:
these ones are unlikely to be merged as I don't really think they're that useful, and it wouldn't outweigh the overhead associated with it
i'll see what anyone else has to say about the usefulness, though

TeoTwawki:
I foresee the attack delay masking or exaggerating probs because they misused it (like what happened with NM base stats recently that severely exaggerated a problem in an issue).

The tp cost one could be interesting.

PR #3189, opened 2016-05-18
Add crafting success hq lost item options

nekolepsy:
Create multipliers for crafting success, hq, and item loss odds.


whasf:
Please re-push and clean up.

Issue #3058, opened 2016-04-17
Dolls are always aggressive to sight, barring level difference


Hozu:
**_I have:_**

- [X] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [X] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160203_0

**_Server Version_** (type `@revision` in game) **:**
https://github.com/DarkstarProject/darkstar/commit/ec30b55f7457dbfbbe190ed181c4b83e7e07871d

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
I believe the issue stems from the fact that the mob is aggressive, and the detection type is both magic and sight. If it was set to not aggro, then it wouldn't aggro to magic either.

PR #3191, opened 2016-05-18
Detection correction

eraffxi:
updated ro maeve doll family detection to magic only, changed sweeper family so that it is not affected


eraffxi:
fixes https://github.com/DarkstarProject/darkstar/issues/3058

Hozu:
You know, I don't know why I didn't think of this. I thought the solution should have been make it so magic/JA/WS detection ignores the aggro value in the mob pool and happens regardless (barring level of course). But another family for non-sight aggro dolls works out better.

Scavenge:
there are 3 doll families total, i just switched that one to 84 on a guess, i dont know the difference between 83 and 84. i assume they were made seperate for a reason

PR #3194, opened 2016-05-20
Russian>English comment translations

Ninjistix:
Tab2Space, white space


bendangelo:
Wow thanks, I never thought they would ever be translated. This is a serious piece of history you're cleaning away.

Ninjistix:
Lol I was reading it and was like the fuck this says engrish onry. So yeah
....

teschnei:
i don't mind the comments being translated, but a machine translation is just a waste of time

how does "For the characters stored in the size growth" make any more sense than the russian text it replaces?

Ninjistix:
Dunno I blame Google

Cloudef:
Yeah I don't think machine translation is good and it just kills opportunity for real russian to translate it to english.

wasamasa:
This is pretty pathetic.  Even I can do better than that:

```
флаги перед именем персонажа
flags in front of the person name

для персонажей в size хранится рост,
для npc и монстров что-то вроде типа используемой модели
size contains the height for persons,
otherwise something like the type of the used model for NPCs and monsters

ранг используется только в ремеслах. размер 64 необходим для совместимости ID
rank used only for crafts, size of 64 is mandatory for the ID [array] capacity

таблица ключевых предметов
таблица ключевых предметов, отмеченных как "увиденные"
key item table
key item table of items marked as "seen"

угол поворота сущности относительно своей позиции (используется 255 система, место 360°)
высота расположения сущности относительно "уровня моря"
что-то вроде расстояния перемещения, необходимое для правильной отрисовки в клиенте количества шагов сущности
rotation angle of the object relative to its position (with 255 being equivalent to 360°)
height of the object relative to the "sea level"
something like a travel distance, required for correct rendering in the client, amount of object steps

информация для окна поиска
information for the search box

предпочтительный язык общения
тип комментария
preferred communication language
commentary type

комментарий поиска
search commentary

небольшой комментарий к пакетам ниже, определенным в виде констант
1-й байт - размер пакета
через 4-ре байта начинается заголовк 0x49, 0x58, 0x46, 0x46 - IXFF
после заголовка идет предположительно тип сообщения:
0x03 - положительный результат
0x04 - ошибка (в случае ошибки начиная с 33 байта идет код ошибки uint16)
small commentary about the packets below, of constant (definitive) nature
first byte - packet size
after four bytes the header starts: 0x49, 0x58, 0x46, 0x46 - IXFF
the header is most likely followed by the message type:
0x03 - successful result
0x04 - error (in the error case the error code can be found as uint16 after the 33rd byte)
```

PR #3195, opened 2016-05-23
Detection correction reflection

eraffxi:
was told the change of the value to 32 was causing errors, so instead of trusting my shoddy knowledge of how it works, i set it to be 288 which is what elementals use (which are aggressive to magic, just like these dolls are supposed to).

then being told maybe 32 is not wrong after all. so i will just leave this here, and if it's not needed by all means close it.


maxtherabbit:
288 is scent (which the elementals shouldn't have either) and magic

thanks for doing this, but I'm just going to adjust the spurious error message

PR #3198, opened 2016-05-27
Fix typos

ymkc:
Just a few startup errors and missing semi-colons


teschnei:
There are no missing semi-colons in lua

ymkc:
I know it's not causing errors, but every other line in the file ends with a semi-colon

PR #3200, opened 2016-05-27
CORE: correct max_missionid

eraffxi:
increasing max_missionid to 94 to allow for last verse


bendangelo:

PR #3201, opened 2016-05-27
Adjust a few NPC's required_expansion

ymkc:
Amtacite Refiners -> ABYSSEA
Unity NPC's -> SOA


ymkc:
Couldn't really decide between WOTG and ABYSSEA, but went with ABY as the currency they exchange is considered ABY

http://ffxiclopedia.wikia.com/wiki/Cruor

But can correct it if it's off

PR #3206, opened 2016-06-01
Cardian aggro, Windurst 6-1 despawns

waytim:
Changed Cardian aggro from sight to sound+magic, added despawn timers on mission popped Cardian Jacks per wiki (3 minutes)


bendangelo:

PR #3213, opened 2016-06-03
Fixed issues with the quest 'Mom, the Adventurer?'

HeavensSword:
Fixed some issues with the quest 'Mom, the Adventurer?'. After reaching
Bastok rep 2 or higher the player was able to start the next quest
without completing this one. In addition, even if this quest had been
accepted, the player would be unable to complete it after reaching rep
2+. A check was also added to fix a player who has already had this
quest bugged for them so they should now be able to complete it.

### DIFF
--- a/scripts/zones/Bastok_Markets/npcs/Nbu_Latteh.lua
+++ b/scripts/zones/Bastok_Markets/npcs/Nbu_Latteh.lua
@@ -6,42 +6,53 @@
 -----------------------------------
 package.loaded["scripts/zones/Bastok_Markets/TextIDs"] = nil;
 -----------------------------------
-require("scripts/globals/settings");
-require("scripts/globals/keyitems");
-require("scripts/globals/titles");
-require("scripts/globals/quests");
-require("scripts/zones/Bastok_Markets/TextIDs");
+require( "scripts/globals/settings" );
+require( "scripts/globals/keyitems" );
+require( "scripts/globals/titles" );
+require( "scripts/globals/quests" );
+require( "scripts/zones/Bastok_Markets/TextIDs" );
 
 -----------------------------------
 -- onTrade Action
 -----------------------------------
 
-function onTrade(player,npc,trade)
+function onTrade( player, npc, trade )
 end;
 
 -----------------------------------
 -- onTrigger Action
 -----------------------------------
 
-function onTrigger(player,npc)
-    local pFame = player:getFameLevel(BASTOK);
-    local MomTheAdventurer = player:getQuestStatus(BASTOK,MOM_THE_ADVENTURER)
-    local questStatus = player:getVar("MomTheAdventurer_Event");
+function onTrigger( player, npc )
 
-    if (player:needToZone()) then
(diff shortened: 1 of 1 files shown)


scripts/zones/Bastok_Markets/npcs/Nbu_Latteh.lua
```diff
@@ -6,42 +6,53 @@
 -----------------------------------
 package.loaded["scripts/zones/Bastok_Markets/TextIDs"] = nil;
 -----------------------------------
-require("scripts/globals/settings");
-require("scripts/globals/keyitems");
-require("scripts/globals/titles");
-require("scripts/globals/quests");
-require("scripts/zones/Bastok_Markets/TextIDs");
+require( "scripts/globals/settings" );
```
TeoTwawki: Don't do that

TeoTwawki:
No, the way it works now was already retail. I rewrote it last to fix its bugs and at the time I created new characters on retail to work out exactly what happens. You actually can easily lose out on doing the first quest forever, on retail.

TeoTwawki:
https://github.com/DarkstarProject/darkstar/pull/125

scripts/zones/Bastok_Markets/npcs/Nbu_Latteh.lua
```diff
@@ -6,42 +6,53 @@
...
+            player:startEvent( 0x00e6 );  
+    elseif( MomTheAdventurer >= QUEST_ACCEPTED and questStatus == 2 ) then
+        if( player:seenKeyItem( LETTER_FROM_ROH_LATTEH ) ) then
+            player:startEvent( 0x00ea );
+        elseif( player:hasKeyItem( LETTER_FROM_ROH_LATTEH ) ) then
+            player:startEvent( 0x00e9 );
         end
-    elseif (pFame >= 2 and player:getQuestStatus(BASTOK,THE_SIGNPOST_MARKS_THE_SPOT) == QUEST_AVAILABLE) then
-        player:startEvent(0x00eb);
+    elseif( MomTheAdventurer >= QUEST_COMPLETED and pFame >= 2 and TheSignpostMarksTheSpot == QUEST_AVAILABLE ) then
```
TeoTwawki: reiterating my above "don't do that": `elseif( ` should be `elseif (`

HeavensSword:
Odd. The forum posts I referenced and the wiki entry seemed to indicate that you at least had to accept and finish the first quest before the second could be accepted regardless of rep.
I'll take your word on it though since it's been at least 11-12 years since I completed these quests on retail.
Also, I'll keep the formatting in mind next time.

bendangelo:
Thanks for contributing. Please don't format the whole file, it makes it difficult to see which parts you changed. If @TeoTwawki is good with it, I'll merge this in.

TeoTwawki:
@bendangelo no, not ok with it. Its incorrect. An argument could be made for an acceptable departure form retail in preventing the 1st quest from being missed out on, but this wouldn't be how I'd want to handle that.

PR #3217, opened 2016-06-05
DRK missing MOBMOD_MAGIC_COOL

ymkc:
Not sure why DRK and PLD were left out, but I think at least DRK should have it too.

Probably PLD aswell?


TeoTwawki:
should probably be a default for when it isn't explicitly set (I thought there already was)

ymkc:
updated commit to add it to PLD aswell

bendangelo:

PR #3218, opened 2016-06-07
Added another set of timed spawn NMs

ymkc:
- Metal Shears
- Sappy Sycamore
- Skirling Liger
- Wake Warder Wanda

source: wikia, ffxidb, yt

not quite sure about removing mod_regain on mobDisengage and mobSpawn as some mobs have it removed on mobRoam


scripts/zones/Batallia_Downs/mobs/Skirling_Liger.lua
```diff
@@ -0,0 +1,49 @@
...
+        mob:setMod(MOD_REGAIN,0);
+    end
+end;
+
+-----------------------------------
+-- onMobFight
+-----------------------------------
+
+function onMobFight(mob, target)
+    mob:setMod(MOD_REGAIN, 50);
```
TeoTwawki: that'll needlessly run every tick. if he's only supposed to have regain while in battle, set in in onMobEngaged() and take away in onMobDisengage() uing the precise amount with addMod()/delMod() instead of the ~0 stuff

only IF that is so though (I didn't look them open before posting, and do not recall how it works)

TeoTwawki: I did something like what you have with Ghillie_Dhu.lua but that was because his regain was variable, not a fixed amount. If these also are variable, my previous suggestion will be wrong.

ymkc: @TeoTwawki thanks, was bad c&p without thinking it to the end

what about onMobDeath or will the mob lose the mod on respawn anyway?

TeoTwawki: on repop mobs get reset to their initial values (whatever is in database+onMobinit+onMobSpawn) or should, if they don't then an old bug recurred and would need reported again.

scripts/zones/Valkurm_Dunes/mobs/Metal_Shears.lua
```diff
@@ -0,0 +1,42 @@
...
+-- onAdditionalEffect Action
+-----------------------------------
+function onAdditionalEffect(mob,target,damage)
+    -- Guesstimating 1 in 3 chance to poison on melee.
+    if ((math.random(1,100) >= 33) or (target:hasStatusEffect(EFFECT_POISON) == true)) then
+        return 0,0,0;
+    else
+        local duration = math.random(10,25);
+        target:addStatusEffect(EFFECT_POISON,15,3,duration);
+        return SUBEFFECT_POISON,0,EFFECT_POISON;
```
TeoTwawki: that zero is the message, does this mob's add effect do its poisoning silently? if so this is correct, if not you can find the message ID in status.lua :)

ymkc: I actually don't know, since the yt fights are pretty short or in japanese :|

guess default is to display something, will change it

ymkc: done, also corrected it on Numbing Norman

scripts/zones/Buburimu_Peninsula/mobs/Wake_Warder_Wanda.lua
```diff
@@ -0,0 +1,48 @@
...
+    mob:setMod(MOD_REGAIN,0);
+end;
+
+-----------------------------------
+-- onMobDeath
+-----------------------------------
+
+function onMobDeath(mob, player, isKiller)
+    UpdateNMSpawnPoint(mob:getID());
+    mob:setRespawnTime(math.random(3600,4200)); -- repop 60-70min
```
xdemolish: probably dont want this running for every player in the party, better to have it run once (isKiller)

scripts/zones/Batallia_Downs/mobs/Skirling_Liger.lua
```diff
@@ -0,0 +1,37 @@
...
+function onMobDisengage(mob)
+    mob:setMod(MOD_REGAIN,0);
+end;
+
+-----------------------------------
+-- onMobDeath
+-----------------------------------
+
+function onMobDeath(mob, player, isKiller)
+    UpdateNMSpawnPoint(mob:getID());
```
xdemolish: same here

scripts/zones/Buburimu_Peninsula/mobs/Wake_Warder_Wanda.lua
```diff
@@ -0,0 +1,48 @@
...
+    mob:setMobMod(MOBMOD_MAGIC_COOL, 50); -- just one spell to spam
+end;
+
+-----------------------------------
+-- onMobSpawn Action
+-----------------------------------
+
+function onMobSpawn(mob)
+    mob:addMod(MOD_SILENCERES,20);
+    mob:addMod(MOD_BINDRES,20);
```
xdemolish: mob would be super resistant after going down a few times, should put this in onMobInitialise if that's not supposed to happen (maybe mob_spawn_mods)

scripts/zones/Batallia_Downs/mobs/Skirling_Liger.lua
teschnei: or onmobdespawn?

xdemolish: ya that too
forgot all about that lol

TeoTwawki: Thats what I did to land kings before for same reason

scripts/zones/Buburimu_Peninsula/mobs/Wake_Warder_Wanda.lua
TeoTwawki: Aren't they supposed to be getting reset before onMobSpawn gets called again?

ymkc:
thanks for all the input, updated the stuff

xdemolish:
looks good :)
i'll let @teschnei decide what to do about regain (im not sure why it needs to be unset in the first place)

scripts/zones/Buburimu_Peninsula/mobs/Wake_Warder_Wanda.lua
```diff
@@ -0,0 +1,48 @@
...
+
+function onMobSpawn(mob)
+end;
+
+-----------------------------------
+-- onMobEngaged
+-----------------------------------
+
+function onMobEngaged(mob,target)
+    mob:setMod(MOD_REGAIN, 25);
```
takhlaq: move to onMobInitialize, prolly no need to set/unset on engage/disengage

scripts/zones/Batallia_Downs/mobs/Skirling_Liger.lua
```diff
@@ -0,0 +1,37 @@
...
+
+function onMobInitialize(mob)
+end;
+
+-----------------------------------
+-- onMobEngaged
+-----------------------------------
+
+function onMobEngaged(mob,target)
+    mob:setMod(MOD_REGAIN, 50);
```
takhlaq: this too

scripts/zones/Buburimu_Peninsula/mobs/Wake_Warder_Wanda.lua
TeoTwawki: Depends on if mob should have regin while idle (usually no). If it only has it during battle, it needs set on engage and removed on disengage

TeoTwawki: which it is https://github.com/DarkstarProject/darkstar/pull/3218/files#diff-7dc3599402cd88948b941c882c5a7f48R38

PR #3221, opened 2016-06-08
Correct onMobDeath to onMobDespawn for previous PR

ymkc:
While correcting these, I noticed that alot of mobs (from AV to Leaping Lizzy and PHs) suffer from the same problem.

wouldn't that make it way more easy to pop these NMs with more ppl in a pt?

### DIFF
--- a/scripts/zones/West_Sarutabaruta/mobs/Numbing_Norman.lua
+++ b/scripts/zones/West_Sarutabaruta/mobs/Numbing_Norman.lua
@@ -33,5 +33,12 @@ end;
 
 function onMobDeath(mob, player, isKiller)
     checkRegime(player,mob,61,2);
+end;
+
+-----------------------------------
+-- onMobDespawn
+-----------------------------------
+
+function onMobDespawn(mob)
     UpdateNMSpawnPoint(mob:getID());
 end;
--- a/scripts/zones/East_Ronfaure/mobs/Rambukk.lua
+++ b/scripts/zones/East_Ronfaure/mobs/Rambukk.lua
@@ -11,9 +11,9 @@ function onMobSpawn(mob)
 end;
 
 -----------------------------------
--- onMobDeath
+-- onMobDespawn
 -----------------------------------
 
-function onMobDeath(mob, player, isKiller)
+function onMobDespawn(mob)
     UpdateNMSpawnPoint(mob:getID());
 end;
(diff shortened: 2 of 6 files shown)


scripts/zones/West_Sarutabaruta/mobs/Numbing_Norman.lua
```diff
@@ -28,10 +28,10 @@ function onAdditionalEffect(mob,target,damage)
 end;
 
 -----------------------------------
--- onMobDeath
+-- onMobDespawn
 -----------------------------------
 
-function onMobDeath(mob, player, isKiller)
+function onMobDespawn(mob)
     checkRegime(player,mob,61,2);
```
takhlaq: regime needs to be checked onMobDeath cause all party members need it

teschnei: Well it just randomizes the respawn time multiple times instead of once,
right? Which should be fixed of course, but it's not exploitable

ymkc: arg :X

ymkc: @teschnei 

scripts/zones/AlTaieu/mobs/Jailer_of_Love.lua:function onMobDeath(mob, player, isKiller)
scripts/zones/AlTaieu/mobs/Jailer_of_Love.lua-    local AV_CHANCE = 25;
scripts/zones/AlTaieu/mobs/Jailer_of_Love.lua-    if (AV_CHANCE > math.random(0,99)) then
scripts/zones/AlTaieu/mobs/Jailer_of_Love.lua-        SpawnMob(16912876, 180);
scripts/zones/AlTaieu/mobs/Jailer_of_Love.lua-    end
scripts/zones/AlTaieu/mobs/Jailer_of_Love.lua-end;

or in Rock_Lizard.lua under onMobDeath:
...  
          if (math.random(1,20) == 5) then
...

would be rerolled for each party member would'nt it?

teschnei:
in those cases yeah, but for the most part they would have been the same either way (ie doing math.random and then setRespawnTime)

ymkc:
updated PR

ymkc:
Dynamis mobs are doing alot of SpawnMob/Despawn in onMobDeath, could that impact performance or are those cheap calls?

takhlaq:
everything's cheap unless proven otherwise! (prolly negligible performance impact)
can always profile it though

also since you mentioned it can you move those to onMobDespawn too :D

ymkc:
I will take a look at it, maybe it's not that bad... Or atleast sed-able

PR #3222, opened 2016-06-09
June item_basic.sql Update

Ninjistix:
just saw June updates, fixed some flags, and corrected some names.


TeoTwawki:
> 940 additions, 921 deletions not shown because the diff is too large.

PR #3223, opened 2016-06-10
Batch update onMobDeath->onMobDespawn PH's & NM's

ymkc:
This contains all the cheap & easy Leaping Lizzy like NM's and PH's

### DIFF
--- a/scripts/zones/Valkurm_Dunes/mobs/Valkurm_Emperor.lua
+++ b/scripts/zones/Valkurm_Dunes/mobs/Valkurm_Emperor.lua
@@ -4,29 +4,12 @@
 -----------------------------------
 
 -----------------------------------
--- onMobDeath
------------------------------------
-
-function onMobDeath(mob, player, isKiller)
-
-    -- Set VE's Window Open Time
-    SetServerVariable("[POP]Valkurm_Emperor", os.time(t) + 3600); -- 1 hour
-    DeterMob(mob:getID(), true);
-
-    -- Set PH back to normal, then set to respawn spawn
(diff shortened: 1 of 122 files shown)


TeoTwawki:
1 small caveat: the `setDropRate()` commands won't work there. Everything else is good that I can see.

Actually, I am near certain they should never have been in onMobDeath anyway and cannot work in either function, so I guess this doesn't break anything! I'm not sure they could be moved into `onMobSpawn()` but I've been meaning to move them to sql anyway (a change to the table to handle drops in "slots" is required, but I screwed up my branch I was doing that in :frowning: ).

TeoTwawki:
tl:dr I say merge as is anyway, coz those lines days are numbered

scripts/zones/Valkurm_Dunes/mobs/Valkurm_Emperor.lua
```diff
@@ -4,10 +4,10 @@
 -----------------------------------
 
 -----------------------------------
--- onMobDeath
+-- onMobDespawn
 -----------------------------------
 
-function onMobDeath(mob, player, isKiller)
+function onMobDespawn(mob)
```
takhlaq: there's two onMobDespawns in this file doing the exact same thing

takhlaq: other than that its gud

ymkc: didn't catch that, fixed

PR #3226, opened 2016-06-13
Fixing SCH unlock step

Kushdr:
Fixing this file as this particular instance of JOB_SCH was missed in commit: https://github.com/DarkstarProject/darkstar/commit/2bd029e1e70a2616be83dc50784b79a7905ac5dc
Allows the SCH job to actually be unlocked now


TeoTwawki:
Thanks

PR #3230, opened 2016-06-16
Fixed avatars using wrong animations

Vadavim:
After testing out Summoner, I noticed none of the avatars were not using the correct animations. I dug around in the code and I found that lines 640-641 could not possibly evaluate to true. This led to avatars and automatons using ACTION_MOBABILITY_FINISH, which results in the wrong animation being used.

 The problem doesn't exist for jug bets because they should be using ACTION_MOBABILITY_FINISH (which is why I don't check for == PETTYPE_JUG_PET in my edit).


Fiocitrine:
wooo thanks!

PR #3238, opened 2016-06-21
Fix some mob lua's

ymkc:
- moved left over mobs  from killer to isKiller
- updateEnmity -> updateClaim (verified on yt)
- Cerberus initial pop was set too high
- Bastok 9-1 Mission mob popped with updateClaim

### DIFF
--- a/scripts/zones/VeLugannon_Palace/mobs/Detector.lua
+++ b/scripts/zones/VeLugannon_Palace/mobs/Detector.lua
@@ -77,18 +77,7 @@ end;
 -----------------------------------
 
 function onMobDeath(mob, player, isKiller)
-
     checkGoVregime(player,mob,743,1);
-
-    local Detector = mob:getID();
-    local Caretaker = mob:getID() + 1;
-
-    GetMobByID(Detector):resetLocalVars();
-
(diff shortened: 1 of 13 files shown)


scripts/zones/VeLugannon_Palace/mobs/Detector.lua
```diff
@@ -77,18 +77,7 @@ end;
...
-
-    local Detector = mob:getID();
-    local Caretaker = mob:getID() + 1;
-
-    GetMobByID(Detector):resetLocalVars();
-
-    if (GetMobAction(Caretaker) ~= 0) then
-        DespawnMob(Caretaker);
-    end
-
```
takhlaq: why was this removed? (no idea what's supposed to happen in retail but assuming whatever was here before was retail-ish)

ymkc: the same code was in there twice, once under onMobDeath and once under onMobDespawn

takhlaq: derp didn't see that lol

TeoTwawki:
Thank you for taking care of these.

PR #3240, opened 2016-06-22
add to Treasure Pool binding

TeoTwawki:
This was way less effort than finding out why setdroprate is broken (~~it seems to do nothing, no matter where you place it, this was already reported and I have no idea and no time to figure out whats wrong or how to fix it - don't ask me to~~ may have to do with people calling the wrong drop ID so they set rate on things that don't even exist cough cough) or fixing my broken attempt at a new "slot/group" column in the mob_drops table which I may try again on later. This is more reliable and harder for people to screw up in the meantime..

Fair warning, I only did minimal testing. Seems to work fine after an odd delay before the messages hit client.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -130,6 +130,7 @@
 #include "../ai/states/magic_state.h"
 
 #include "../transport.h"
+#include "../treasure_pool.h"
 #include "../mob_modifier.h"
 
 CLuaBaseEntity::CLuaBaseEntity(lua_State* L)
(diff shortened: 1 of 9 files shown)


teschnei:
I was just about to say - not sure if doing the NM loot like this is the best idea, but this would be ideal for things that get added to the pool via an actual scripted event (ie. treasure chests)

For the NMs, there's a really easy way to make "drop slots" work if you want to talk about it later (or here)

src/map/lua/lua_baseentity.cpp
```diff
@@ -715,6 +716,28 @@ inline int32 CLuaBaseEntity::getSpawnPos(lua_State* L)
...
+
+    CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
+
+    if (PChar->PTreasurePool != nullptr)
+    {
+        PChar->PTreasurePool->AddItem((uint16)lua_tointeger(L, 1), PChar);
+    }
+    else
+    {
+        ShowError(CL_RED"Lua::addTreasure: Tried to add item to a treasure pool that was nullptr! \n" CL_RESET);
```
teschnei: shouldn't this just add the item to the character's inventory directly?

TeoTwawki:
what happened with the column method I wanted, was I had it 90% done and then f'd up my branch royally and couldn't recover it in git. I still want to do that (and remove what you see here) but figured I could at least make these "work" for now. I'll catch you on irc later today.

src/map/lua/lua_baseentity.cpp
TeoTwawki: when solo, it counts as a pool but just auto falls to you.

PR #3242, opened 2016-06-23
NPC ID corrections for Abyssea - La Theine 

TeoTwawki:
NOTE: this is not for the most recent client, but for the version DSP is currently still on. I'm PRing this so the next run of the tool has correct offsets to work with.


TeoTwawki:
@gedads @starlightknight I started to sort out bastok markets and then quit. too many out of order IDs combined with the typical plethora of of even NPCs that get shuffled in/out.



@teschnei would it be possible to rig some sort of fake NPC object the way you can fake mob objects, so that it would not matter for us that SE deletes seasonal NPCs?

gedads:
I'll try to get some time to capture cities again but from time to time, they add events npcs that eventually disappear. bastok market is messed where npcs christmas usually are?

TeoTwawki:
its more than that. somebody placed lines out of order, left obsolete garbage behind when npcs got replaced etc etc.
if you go 1,2,5,3,4 instead of 1,2,3,4,5 (sequential id) the tool trips up on that and reordering to compensate makes the diffs impossible to compare

gedads:
like the previous topic take a look at an older file, you'll see all is in order. i think that's not someone, that's the softaware not finding names or something

TeoTwawki:
no. not the name. the godamn id. look. at. it.

gedads:
where?

TeoTwawki:
what part of what I have said was confusing?

PR #3243, opened 2016-06-26
Added UpdateNPCSpawnPoint/FallenEgg/Antican Tag

ymkc:
- Added UpdateNPCSpawnPoint, a similar function to UpdateTreasureSpawnPoint, to move ???-like NPC's around 
- Added FallenEgg using UpdateNPCSpawnPoint
- Added Nihniknoovi popped by FallenEgg (also corrected mob level and attack speed)
- Added the moving ??? in QSC for the Antican Tag
- Added the recursing lua function RotateNPCSpawnPoint, which I am not quiet sure off, as there might be a better way to do it


scripts/zones/La_Theine_Plateau/npcs/FallenEgg.lua
```diff
@@ -0,0 +1,41 @@
...
+    local X = npc:getXPos();
+    local Y = npc:getYPos();
+    local Z = npc:getZPos();
+
+    if (random <= 20 and GetMobAction(Nihniknoovi) == ACTION_NONE) then
+        GetMobByID(Nihniknoovi):setSpawn(X,Y,Z);
+        SpawnMob(Nihniknoovi):updateClaim(player);
+        UpdateNPCSpawnPoint(FallenEgg,1800);
+    elseif (random > 20 and random <= 61) then
+        player:messageSpecial(NOTHING_OUT_OF_ORDINARY); -- missing correct TextID
```
Hozu: <field name="index">7811</field>
    <field name="text">There is a broken egg on the ground here.
Perhaps there is a nest in the boughs of this tree.≺Prompt≻

Is that what you're looking for?

sql/npc_spawn_points.sql
```diff
@@ -0,0 +1,124 @@
...
+/*!40014 SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0 */;
+/*!40014 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 */;
+/*!40101 SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='NO_AUTO_VALUE_ON_ZERO' */;
+/*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;
+
+--
+-- Table structure for table `npc_spawn_points`
+--
+
+DROP TABLE IF EXISTS `npc_spawn_points`;
```
TeoTwawki: I'd rather we not have an additional location where we have to watch out for shifted IDs. We have method to move an NPC in lua that don't use the database. It shouldn't be to hard to table the positions in lua.

scripts/zones/La_Theine_Plateau/npcs/FallenEgg.lua
ymkc: @Hozu Thanks! Is this from the resource mass extractor?

sql/npc_spawn_points.sql
ymkc: I totally see your point, yet I still got 2 arguments/questions:
- the table structure is the same as the treasure_spawn_points, so if that one get's batched, this one should be the same 
- there are many more "moving/jumping npc's/???", and maybe the HELM locations could be moved into this table aswell. that way most of the npcid would be in a central, easily batchable location

nevertheless, if you guys say move it to lua and do it like HELM, I do so and rewrite it

scripts/zones/La_Theine_Plateau/npcs/FallenEgg.lua
Hozu: Nah POLutils, in the data browser > dialog tables.

sql/npc_spawn_points.sql
TeoTwawki: Helm are lua'd, and treasure spawn points I very much want to obliterate but I think I'm alone on that one. I spoze we'd best ask da big chief.

@teschnei opinion, oh lead dev?

TeoTwawki: p.s.:
I'm exactly not opposed to an "everything that moves all in one place" table. As long as it covered all of it like you suggested.

teschnei:
why is there a distinction between treasure spawn points and npc spawn points?  treasures are npcs already.  I don't think it matters whether you want to put them in sql or lua, but they should be the same thing (and in the same place) because they are the same object type.
to be honest, i think it's easier to put the positions in the object's lua script, meaning it never has to refer to a hardcoded id

ymkc:
moved it all to lua :)

scripts/globals/npc_util.lua
```diff
@@ -7,25 +7,34 @@ npcUtil = {};
 -- Picks a new position for an NPC and excluding the current position.
 -- INPUT: npc = npcID, position = 2D table with coords: index, {x, y, z}
 -- RETURN: table index
-function npcUtil.pickNewPosition(npc, positionTable)
-
+function npcUtil.pickNewPosition(npc, positionTable, allowCurrentPosition)
     local npc = GetNPCByID(npc)
     local positionIndex = 1 -- Default to position one in the table if it can't be found.
     local tableSize = 0
+    local allowCurrentPosition = allowCurrentPosition or false;
```
takhlaq: no need to redefine

scripts/globals/npc_util.lua
```diff
@@ -7,25 +7,34 @@ npcUtil = {};
...
+
     for i, v in ipairs(positionTable) do   -- Looking for the current position
 
-        -- Finding by comparing the NPC's coords
-        if (math.floor(v[1]) == math.floor(npc:getXPos()) and math.floor(v[2]) == math.floor(npc:getYPos()) and math.floor(v[3]) == math.floor(npc:getZPos())) then
-            positionIndex = i; -- Found where the NPC is!
+        if (allowCurrentPosition == "false") then
```
takhlaq: not sure why you're comparing as string, should be able to just `if not allowCurrentPosition then`

scripts/globals/npc_util.lua
```diff
@@ -7,25 +7,34 @@ npcUtil = {};
...
+            end
         end
+
         tableSize = tableSize + 1; -- Counting the array size
     end
 
-    -- Pick a new pos that isn't the current
-    local newPosition = positionIndex
-    repeat
+    if (allowCurrentPosition == "false") then
```
takhlaq: same here

takhlaq:
other than that, looks good


one other (optional) change would be to change the tables to store pos like so
```
fallenEggPositions = {
         [1] = {x = 211.000, y = 24.000, z = 258.000},
         -- etc
```
and access like so `npc:setPos(fallenEggPositions[newPosition].x, fallenEggPositions[newPosition].y, fallenEggPositions[newPosition].z);`

scripts/globals/npc_util.lua
ymkc: I read this is the lua way to pass optional parameters, so other scripts using it don't need to be modified

scripts/globals/npc_util.lua
ymkc: I tried in many different ways, comparing as a string was the only way it seemd to work. I will give this a try, maybe I just didn't see the obvious

ymkc: doesn't work :|

TeoTwawki: `if (allowCurrentPosition == "ziggy") then` winds up being the same thing. Unless somewhere you are assigning the string to that variable, it can never be true. It only "works" because the comparison always returns false there. Since that variable is not that string. I don't think this is how you wanted that to work.

ymkc: I don't really know why, but this worked while I tested and I just redid a printf-based test and it still works:
[30/Jun] [10:09:00][Status] do_init: loading zones[30/Jun] [10:09:01][LUA Script] allowCurrentPosition: true - newPosition: 3
[30/Jun] [10:09:00][Status] do_init: loading zones[30/Jun] [10:09:02][LUA Script] allowCurrentPosition: false - newPosition: 51
...
[30/Jun] [10:19:04][LUA Script] allowCurrentPosition: true - newPosition: 1
[30/Jun] [10:20:21][LUA Script] allowCurrentPosition: true - newPosition: 1
[30/Jun] [10:22:11][LUA Script] allowCurrentPosition: true - newPosition: 4
[30/Jun] [10:23:50][LUA Script] allowCurrentPosition: true - newPosition: 4
[30/Jun] [10:25:21][LUA Script] allowCurrentPosition: true - newPosition: 3
[30/Jun] [10:27:01][LUA Script] allowCurrentPosition: true - newPosition: 2
[30/Jun] [10:28:15][LUA Script] allowCurrentPosition: true - newPosition: 2
[30/Jun] [10:29:29][LUA Script] allowCurrentPosition: true - newPosition: 2
...

ymkc: found the bug, works now in a clean way

takhlaq:
looks good, just need it to return the actual position subtable instead of its index in main table and i'll merge

ymkc:
@takhlaq didn't see your edit, changed it

takhlaq:
ah sorry about that, i should've made a new comment (was hoping i edited before you had a chance to look at it)
i'll let @teschnei decide what to do about `UpdateAnticanTagSpawnPoint` as in whether that should go to npcUtil or not since we don't wanna repeat code if other scripts do the same but other than that it looks good, thanks for making the requested changes :)

scripts/zones/Quicksand_Caves/Zone.lua
```diff
@@ -196,3 +205,21 @@ function onEventFinish(player,csid,option)
...
+    local newPosition = npcUtil.pickNewPosition(npc:getID(), anticanTagPositions, true);
+
+    if (GetServerVariable("[POP]Antican_Tag") <= os.time(t)) then
+        npc:hideNPC(1); -- hide so the NPC is not "moving" through the zone
+        npc:setPos(newPosition.x, newPosition.y, newPosition.z);
+    end
+    npc:timer(respawnTime * 1000, function(npc)
+        UpdateAnticanTagSpawnPoint(id, minTime, maxTime);
+    end)
+end;
```
takhlaq: rename to UpdateNPCSpawnPoint
stick in npcUtil, add a parameter for respawnTime for npc:timer function
move the if check to onGameHour in this zone's lua, 2-3min seems reasonable (see <http://ffxi.allakhazam.com/forum.html?forum=10&mid=111271398059447339>)

ymkc: ok will do!

what about the "pausing" by using a server var, add it as a parameter?

takhlaq: ya, with an option for localvar/server var in case it doesn't need to persist across reboot for whatever reason

takhlaq: hmm wouldn't npc:timer just keep getting called since it's being called inside UpdateAnticanTagSpawnPoint

ymkc: yes, it's a recursive function. that's why it's checking the server var

takhlaq: move the timer so it only runs when the check is gud then and forget what i said about onGameHour earlier

takhlaq: derp. leave the timer where it is
just move the function to npcUtil and it's gud

ymkc: @takhlaq done

Issue #2733, opened 2016-02-07
Kirin spawning two pets at a time


Fiocitrine:
Kirin has a high percentage to spawn two pets at a time. We've had Seiryu and Genbu up, as well as Suzaku and Byakko up.

dazusu:
Do you mean both pets appear at the exact same time after his summoning animation has completed?

Kirin can have all four pets up, and they should pop 3 minutes(ish) apart

Fiocitrine:
Sorry, yes he spawns them simultaneously. Not one at a time.

neosatus:
I just tested this fix on a personal server and it works:

**/scripts/zones/The_Shrine_of_RuAvitau/mobs/Kirin.lua**
```
-----------------------------------
-- onMobSpawn Action
-----------------------------------

function onMobSpawn(mob)
    mob:setMod(MOD_WINDRES, -64);
    mob:setLocalVar("numAdds", 1);
end

-----------------------------------
-- onMobFight Action
-----------------------------------

function onMobFight( mob, target )
    if (mob:getHPP() < math.random(50,60) and mob:getLocalVar("astralFlow") == 0) then
        mob:useMobAbility(734);
        -- Spawn Avatar
        mob:spawnPet();
        mob:setLocalVar("astralFlow", 1);
    end
    local numAdds = mob:getLocalVar("numAdds");
    if (mob:getBattleTime() / 180 == numAdds) then
        -- Ensure we have not spawned all pets yet..
        local genbu = mob:getLocalVar("genbu");
        local seiryu = mob:getLocalVar("seiryu");
        local byakko = mob:getLocalVar("byakko");
        local suzaku = mob:getLocalVar("suzaku");

        if (genbu == 1 and seiryu == 1 and byakko == 1 and suzaku == 1) then
            return;
        end

        -- Pick a pet to spawn at random..
        local ChosenPet = nil;
        local newVar = nil;
        repeat

            local rand = math.random( 0, 3 );
            ChosenPet = 17506671 + rand;

            switch (ChosenPet): caseof {
                [17506671] = function (x) if ( genbu == 1) then ChosenPet = 0; else newVar = "genbu";  end end, -- Genbu
                [17506672] = function (x) if (seiryu == 1) then ChosenPet = 0; else newVar = "seiryu"; end end, -- Seiryu
                [17506673] = function (x) if (byakko == 1) then ChosenPet = 0; else newVar = "byakko"; end end, -- Byakko
[...]

PR #3244, opened 2016-06-26
Fixed Kirin spawning multiple adds at a time

neosatus:
Fixed issue #2733 

The intention of the code is clearly to spawn one lesser god every 3 minutes. In game we were seeing 2-3 spawn at a time. I believe it was because the function was being run multiple times per second so the condition mob:getBattleTime() % 180 == 0 was passing multiple times in a row. I reworked the condition and it now behaves correctly in game.


teschnei:
you need to replace the tabs with spaces

neosatus:
Will do.

PR #3253, opened 2016-06-29
Add mission packet/tracking for SOA + ROV, and mission ID globals for them

ibm2431:


src/map/packets/quest_mission_log.cpp
```diff
@@ -31,248 +31,262 @@
...
 void CQuestMissionLogPacket::generateCurrentMissionPacket(CCharEntity * PChar)
 {
-	uint16 add_on_scenarios = 0;
+    uint16 add_on_scenarios = 0;
 
-	add_on_scenarios +=	PChar->m_missionLog[9].current;
-	add_on_scenarios +=	PChar->m_missionLog[10].current << 0x04;
-	add_on_scenarios +=	PChar->m_missionLog[11].current << 0x08;
-	// Not perfect, but they display and missions DO progress. Can fix properly later. There is a delay before when the menu updates. Zoning will force it.
+    add_on_scenarios += PChar->m_missionLog[MISSION_CRISTALLINE_PROPHECY - 12].current;
```
takhlaq: i realise you didnt make the typo, but could you change this to MISSION_CRYSTALLINE_PROPHECY instead

ibm2431:
Yeah, I saw that and debated doing so. I'll do so.

ibm2431:
I noticed that in [missions.lua](https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/missions.lua) the add-ons are referred to simply by acronyms, and that for other expansions it's "COP", "TOAU", and "WOTG" instead of their full names. So while I was changing variables I made add-ons and SOA/ROV match the style.

ibm2431:
Actually, now that I've slept and noticed how Zilart worked, the whole half-ROV thing seems rather silly. It's not like you'd be using the defined ID values (that's the whole point of the named constants), so I might as well drop the generic multiply and just increase the mission log offsets by sets of 2, save for 'halves'. Will recommit once I'd done finding the "new" values.

PR #3254, opened 2016-06-29
Implemented Ikishoten and fixed Shikikoyo upgrades.

### LINKED ISSUE #3251: Shikikoyo additional upgrades are bugged
**_I have:_**

- [X] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [X] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** N/A

**_Server Version_** (type `@revision` in game) **:** N/A

**_Source Branch_** (master/stable) **:** Master

I noticed this in the Shikikoyo lua script:

local pTP = player:getTP() - 1000 + (player:getMerit(MERIT_SHIKIKOYO) - 12);

Additional merits of Shikikoyo should increase the pTP by 12%, not 12. It should be:

local pTP = (player:getTP() - 1000) * (1 + (player:getMerit(MERIT_SHIKIKOYO) - 12) / 100);

neosatus:
Fixed #3251 

I recently became the owner of a Nanatsusayanotachi on a 75 cap server only to learn Ikishoten was not implemented yet, which was a bit of a let down. So I implemented it.

http://ffxiclopedia.wikia.com/wiki/Ikishoten

"Adds 30 TP per upgrade to the base weapon TP (before Store TP is applied) gained from a Zanshin attack."

Note: the page actually says 3 because it was never updated after TP changes. The SQL also had 3, probably because it was copied right off that page, so I changed it.

Hasso's wiki page reads: "Occasionally triggers Zanshin after landing a normal attack" so I believe it is correct that Ikishoten's bonus should proc for both a Hasso Zanshin "double attack" as well as a regular ole' Zanshin.

I tested it on my test server with a Nanatsu (base tp gain of 107), with only the base 25% STP from being a 75 SAM, a regular Zanshin gained me (107 + 150) * 1.25 = 321 TP while a Hasso Zanshin gained me (107 * 1.25) + ((107 + 150) * 1.25) = 455. In game I actually saw 454, so I believe we lose 1 TP from integer math.

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -710,7 +710,7 @@ namespace battleutils
                 // FINISH HIM! dun dun dun
                 // TP and stoneskin are handled inside TakePhysicalDamage
                 Action->spikesMessage = 536;
-                Action->spikesParam = battleutils::TakePhysicalDamage(PDefender, PAttacker, dmg, false, SLOT_MAIN, 1, nullptr, true, true, true);
+                Action->spikesParam = battleutils::TakePhysicalDamage(PDefender, PAttacker, ATTACK_NORMAL, dmg, false, SLOT_MAIN, 1, nullptr, true, true, true);
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/map/utils/battleutils.cpp
```diff
@@ -1808,7 +1808,7 @@ namespace battleutils
     *																		*
     ************************************************************************/
 
-    int32 TakePhysicalDamage(CBattleEntity* PAttacker, CBattleEntity* PDefender, int32 damage, bool isBlocked, uint8 slot, uint16 tpMultiplier, CBattleEntity* taChar, bool giveTPtoVictim, bool giveTPtoAttacker, bool isCounter)
+    int32 TakePhysicalDamage(CBattleEntity* PAttacker, CBattleEntity* PDefender, uint16 attackType, int32 damage, bool isBlocked, uint8 slot, uint16 tpMultiplier, CBattleEntity* taChar, bool giveTPtoVictim, bool giveTPtoAttacker, bool isCounter)
```
teschnei: PHYSICAL_ATTACK_TYPE instead of uint16 (how do you know what type it is? it's implicit in the enum...)

neosatus: I actually had it that way initially and it yelled, a lot. One of the errors was "cannot convert uint16 to PHYSICAL_ATTACK_TYPE so I changed it to a uint16 and it worked fine afterwords. I'll try to get it to work as a PHYSICAL_ATTACK_TYPE again.

PR #3256, opened 2016-07-01
Fix the command - wasn't running at all

Kushdr:
The text 'req' inserted at line 6 as part of the last commit (a8969109828a88c5455a0de3ae1c8c4b0a44ff12) was causing it to not run

Removing that allowed it to work normally

### DIFF
--- a/scripts/commands/checkmission.lua
+++ b/scripts/commands/checkmission.lua
@@ -3,8 +3,6 @@
 -- desc: Prints current MissionID for the given LogID and target Player to the in game chatlog
 ---------------------------------------------------------------------------------------------------
 
-req
-
 cmdprops =
 {
     permission = 1,


takhlaq:
derp
should be `require("scripts/globals/missions");`

Kushdr:
I thought of that but it works fine without as far as I can tell. Better
safe than sorry though

On Friday, July 1, 2016, Tahir Akhlaq <notifications@github.com> wrote:

> derp
> should be require("scripts/globals/missions");
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/DarkstarProject/darkstar/pull/3256#issuecomment-230069581>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AS_Yc0gFyf991aryeGokmuc71-JU3bqgks5qRadpgaJpZM4JDjRA>
> .
>


-- 
"sic transit gloria mundi"
~Thus passes away the glory of the world

takhlaq:
yeah, it'll work if a script has already loaded missions.lua but if the command was run prior to the global vars in mission.lua being loaded (or if the vars werent in memory) you wouldnt be able to `@checkmission SANDORIA MISSION_NAME` since the vars arent defined

Issue #3252, opened 2016-06-29
sleeping fafnir (and perhaps others) causes loss of claim and deaggro


eraffxi:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160427_0

**_Server Version_** (type `@revision` in game) **:**

dd31b5fc1da371803c8f5cb57cfc63374c8091a5

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

attack fafnir
sleep fafnir
wait

eraffxi:
i have just discovered that if you engage with sleep off the bat, nothing goes wrong, but when you attack first and then sleep or just sleep again after it wakes, it will lose aggro.

also this only seems to happen on HNMs perhaps, tried it on a darter next to fafnir and got nothing out of the ordinary, but with fafnir it's reproduceable easily

takhlaq:
server revision?

neosatus:
I actually have video evidence of this bug! I was recording our LS doing a Fafnir when a lone RDM ran up and engaged, with a Sleep, while we were buffing. Fafnir de-aggro'd after being reslept following nearly killing the RDM during a Terror and we claimed it from him. Technically against the rules of the server but sleeping Fafnir solo with no raid force coming to back you up is a bit of a dick move as well.

https://www.twitch.tv/tuptain/v/72204439

eraffxi:
woops forgot to put in the revision.  edited post to include it.

neosatus:
Ya, you can watch the log. The RDM sleeps Fafnir then slows and paralyzes it when Sleep falls off and Fafnir Terrors the RDM. Fafnir then eats two shadows, eats the Stoneskin, gets two paralyze procs in a row (much to our dismay), then gets slept again. The RDM casts Stoneskin, Blink, and starts a Cure 4 when Fafnir de-aggros.

neosatus:
I fixed this bug last night on my test server. One of the GMs of Era tested it on his as well. I'll get a pull request up tonight, it's a one line fix.

PR #3257, opened 2016-07-02
Fixed sleep unclaim bug (or at least one cause of it)

neosatus:
Ref #3252

I added a bunch of debug messaging, spawned Fafnir, punched him in the mouth and slept him. A handful of seconds later he unclaims and my debug messaging showed it was because the CheckDetection method returned true causing TryDeaggro to remove me from the enmity list, leading to an empty list and an unclaimed Fafnir. I added additional conditions to extend the deaggro time, similar to the already existing condition for Bind, for each type of Sleep. In further testing, I was unable to get him to deaggro while asleep.


teschnei:
That is not how it is supposed to work, though

PR #3259, opened 2016-07-02
Added pop-mechanics to La Theine Plateau NM's

ymkc:
- Made Tumbling Truffle a lottery pop
- Added Lumbering Lambert & Bloodtear Baldurf a lottery pop from Battering Ram
- Added nm_spawn_points to Slumbering Samwell


scripts/zones/La_Theine_Plateau/mobs/Battering_Ram.lua
```diff
@@ -0,0 +1,51 @@
...
+-- onMobDespawn
+-----------------------------------
+
+function onMobDespawn(mob)
+    local mobID = mob:getID();
+    local chanceForLambert = 0;
+    local chanceForBaldurf = 0;
+    local nmToPop = 0;
+
+    if (mobID == Battering_Ram) then
```
takhlaq: why do you need to check its id? isn't there only one per zone

TeoTwawki: @takhlaq there are multiples

scripts/zones/La_Theine_Plateau/mobs/Bloodtear_Baldurf.lua
```diff
@@ -19,4 +30,28 @@ end;
...
+-----------------------------------
+-- onMobDespawn
+-----------------------------------
+
+function onMobDespawn(mob)
+    local mobID = mob:getID();
+    local chanceForLambert = 0;
+
+    if (GetServerVariable("[POP]Lumbering_Lambert") <= os.time(t)) then
+        local chanceForLambert = math.random(1,100);
```
takhlaq: no need to local this again (at least i don't think there is)

ymkc: @takhlaq will test it

TeoTwawki: local should only be used when the variable is declared. Assigning new values to it just replace it.

ymkc: i thought local limit's it's scope to the current function/block and it's not guaranteed which one is included first :| but will test it in the afternoon and adjust it

TeoTwawki: Already in the scope of the function and your if block already has that one. There is no "which one" - there's only one, not two of it. :)

ymkc: ah lol, didn't see that one, thought you meant that it's in the other ram file aswell:)

TeoTwawki: lol nope we meant just above on line 41

ymkc: fixed

PR #3265, opened 2016-07-07
Exclude certain NPC's from widescan

ymkc:
Added an extra field `widescan` to the npc_list table, which is checked before returning the npclist.
At the moment I only set it to 0 for the most obvious types: Treasures, QM/???, Door's, HELM Points

### DIFF
--- a/src/map/entities/npcentity.cpp
+++ b/src/map/entities/npcentity.cpp
@@ -100,6 +100,7 @@ CNpcEntity::CNpcEntity()
 {
 	objtype = TYPE_NPC;
 	look.face = 0x32;
+        widescan = 1;
 	allegiance = ALLEGIANCE_MOB;
     PAI = std::make_unique<CAIContainer>(this);
 }
--- a/src/map/entities/npcentity.h
+++ b/src/map/entities/npcentity.h
@@ -39,6 +39,7 @@ public:
 
     uint32      m_flags;
     uint8       name_prefix;
+    uint8       widescan;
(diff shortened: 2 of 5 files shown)


gedads:
i never paid attention to npcs being displayed with widescan but that's a really good thing if they were (like christmas decorations doors like you mentionned etc) basically only npcs displayed are those spawned in cities/outdoors and waypoint homepoint books...)

TeoTwawki:
Retail has a flag we should be using to handle this. I forget which one it is though :/
(way to not help, Teo's brain!)

ymkc:
does it even matter which flag, can't it just be added to the flag/flags field as a new value?

I was thinking about using a flag, but thought it would be easier to customize what ppl want on their servers if it's an extra field

TeoTwawki:
> but thought it would be easier to customize what ppl want on their servers if it's an extra field

Hmm this is true. Lets mot worry bout the flag for now...The advantage of using whatever flag retail does is then whenever we use whatever flag value they have, we'd already have it filled in. 

> does it even matter which flag, can't it just be added to the flag/flags field as a new value?

We can technically do that, but then we'd have the wrong flags on/off whenever we grabbed a value from retail. Most of them are just "unused" and though several might already have been changed to doing whatever someone wanted...In any case I take back my previous comment about what we "should" be using. Its more trouble than I wanna deal with and I doubt anyone else does either.

TeoTwawki:
Btw thanks for placing the new column at the end, every time somebody stuffs one in the middle it makes me homicidal :wink:

PR #3268, opened 2016-07-07
Various bug fixes from previous commit

neosatus:
The Trick Attack fix got put live on Era recently and mass testing caught a few more things.

Jump and High Jump got missed when doPhysicalWeaponskill params were changed.
Spirits Within used a different variable syntax that was broken by a mass edit.

### DIFF
--- a/scripts/globals/abilities/high_jump.lua
+++ b/scripts/globals/abilities/high_jump.lua
@@ -46,8 +46,13 @@ function onUseAbility(player,target,ability,action)
         end
         target:lowerEnmity(player, enmityShed + player:getMod(MOD_HIGH_JUMP_ENMITY_REDUCTION)); -- reduce total accumulated enmity
     end
+    
+    local taChar = nil
+    if (player:hasStatusEffect(EFFECT_TRICK_ATTACK)) then
+        taChar = player:getTrickAttackChar(target)
+    end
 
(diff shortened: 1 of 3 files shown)


teschnei:
jump/high jump should check for TA

neosatus:
Alright, I'll add that in.

teschnei:
not that it matters anymore (since you're replacing it), but nullptr isn't a thing in lua.. you probably meant nil

neosatus:
Yea, I had just noticed that. It's actually running on Era now as nullptr (cause I copied the fix we put in when we found it) and it works. Presumably an uninitialized defaults to nil. Obviously not good practice though.

Quick question, the getTrickAttackChar checks for TrickAttack status as it's first step. The previous code from weaponskills.lua (that I replaced recently) checked for it as well before calling it. Is there a performance savings to do the check twice in order to not have to go back to core or should I remove the check as it's redundant?

takhlaq:
performance hit is prolly negligible but ya, redundant code be redundant so check isnt necessary :)

TeoTwawki:
There is this wonderful feature call "amend" you should look into :smiley: 
It lets you change the very last commit you made, and then you can do whats called a force push to replace it, rather than stacking up additional commits.

teschnei:
also check: scripts/mobskills/freezebite.lua

PR #3271, opened 2016-07-08
Adjusted available NPC's on widescan

ymkc:
Lots of generic NPC's (which I forgot on the first commit) don't belong on widescan:
- Auction Counter's
- Excavation Point's
- Gate's excluding Gate Sentry's
- Elevator's


teschnei:
are there any npcs with no models that show on widescan? is it possible to just disregard all no model npcs?

ymkc:
great tip!
the list of npc's without models is 3200+ long, so pretty hard to check and there is little info on which npc's are on widescan or not

ymkc:
not sure about stuff like:
- Large Apparatus
- Unstable Displacement
- Torches
- Banishing Gates

ymkc:
added them all to the pr, if they do belong on widescan I will change it

PR #3274, opened 2016-07-09
Fame for Abyssea and Adoulin

ibm2431:
Adding fame tracking and levels for Abyssea areas and Adoulin.

Only thing of real note is the addition of FAME_AREA constants in `scripts/globals/quests.lua`.

At the moment, all NPCs that deal with fame use the area's name defined in `quests.lua` or `shop.lua` (BASTOK, WINDURST, etc), which works for them because the quest log IDs and fame area IDs just happen to be the same - you can do `player:addFame(BASTOK)`, `showShop(player, BASTOK, stock)`, and `player:completeQuest(BASTOK, QUEST)` without worrying about conflicts between `quests.lua` and `shop.lua` depending on which you've included (either one, or both, depending on the NPC).

However, each Abyssea area has its own separate fame, with all sharing the same quest log ID. This also posed an issue for Adoulin, because with 9 areas of fame between it (quest log = 9) and Norg (quest log = 5), you couldn't very well have ADOULIN be the same value in reference to both Fame and Quests (ie: Adoulin fame can't also be 9 like its quest log).

So I tacked on FAME_ as a prefix (similar to how a couple C headers use MISSION_ and QUEST_) to be used when distinctly referencing Abyssea/Adoulin Fame (but not Quests):

```
player:completeQuest(ABYSSEA, QUEST_NAME)
player:addFame(FAME_ABYSSEA_LATHEINE)
player:completeQuest(ADOULIN, QUEST_NAME)
player:addFame(FAME_ADOULIN, 60)
```

Existing NPCs are unaffected.

Part of me feels that currently existing NPC references should be made more specific (ie: FAME_BASTOK, QUEST_BASTOK, MISSION_BASTOK instead of just BASTOK in all cases), but if the same value works for all three types, there's really no added benefit to doing so other than maybe consistency with Abyssea/Adoulin.

### DIFF
--- a/scripts/globals/quests.lua
+++ b/scripts/globals/quests.lua
@@ -26,27 +26,6 @@ ADOULIN     = 9;
 COALITION   = 10;
 
 -----------------------------------
---  Fame IDs
------------------------------------
-
-FAME_SANDORIA           = 0;
-FAME_BASTOK             = 1;
-FAME_WINDURST           = 2;
-FAME_JEUNO              = 3;
-FAME_SELBINA_RABAO      = 4;
-FAME_NORG               = 5;
-FAME_ABYSSEA_KONSCHTAT  = 6;
-FAME_ABYSSEA_TAHRONGI   = 7;
-FAME_ABYSSEA_LATHEINE   = 8;
-FAME_ABYSSEA_MISAREAUX  = 9;
-FAME_ABYSSEA_VUNKERL    = 10;
-FAME_ABYSSEA_ATTOHWA    = 11;
-FAME_ABYSSEA_ALTEPA     = 12;
-FAME_ABYSSEA_GRAUBERG   = 13;
-FAME_ABYSSEA_ULEGUERAND = 14;
-FAME_ADOULIN            = 15;
-
------------------------------------
 --  San d'Oria - 0
 -----------------------------------
 
--- a/scripts/globals/quests.lua
+++ b/scripts/globals/quests.lua
@@ -26,6 +26,27 @@ ADOULIN     = 9;
 COALITION   = 10;
 
 -----------------------------------
+--  Fame IDs
+-----------------------------------
+
+FAME_SANDORIA           = 0;
+FAME_BASTOK             = 1;
+FAME_WINDURST           = 2;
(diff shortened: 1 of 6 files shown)


TeoTwawki:
> Part of me feels that currently existing NPC references should be made more specific (ie: FAME_BASTOK, QUEST_BASTOK, MISSION_BASTOK instead of just BASTOK in all cases)

I was planning on making it all tables, where the functions in question would lookup in its specific table (that way you could still say "bastok" and not have to worry about context). But I haven't had time to do that and probably won't for a long time.

ibm2431:
That would probably have more utility. And seemingly not too difficult to implement. I'll see what I can do.

ibm2431:
Okay, so what started out as a simple jaunt to add some lookup tables has resulted in me taking an axe to a good portion of quest_mission_log itself, which is probably outside the scope of this PR.

It might be best to go ahead and merge the PR to add fame tracking for the new areas (fame-related function changes and SQL updates), because those won't change. The fame lookup tables can be added later. No currently existing NPCs use Abyssea/Adoulin fame, so not having Lua constants defined for those areas should be alright until I make a PR adding said tables.

TeoTwawki:
> EDIT2: So it seems that calls to quest/mission/fame functions don't go through a Lua intermediary (which would make this a trivial change), but instead go straight to the C methods exposed via Lunar. So this might take a little longer than expected. 

The easy out for that sort of situation is to make a new lua function inside the quest/mission/fame globals (which is where tables would go anyways) the that calls the existing C++ function inside it. The index to number conversion happens on lua's and, and sends the number to the C++ end of it. Any way it happens is going to result in some mass find+replace magic being required and some servers groaning at their custom bits breaking, which is why nobody has done this yet.

TeoTwawki:
> It might be best to go ahead and merge the PR to add fame tracking for the new areas (fame-related function changes and SQL updates), because those won't change. The fame lookup tables can be added later. No currently existing NPCs use Abyssea/Adoulin fame, so not having Lua constants defined for those areas should be alright until I make a PR adding said tables.

No objections from me. Thanks for doing any of this.

PR #3276, opened 2016-07-10
CORE: Wardrobe 3&4 core Implementation including customization via st…

7cipherx7:
CORE: 
Fix Mob casting as it was interrupted by spell range check and mobs can cast up to 40yalms.
Implemented Wardrobe 3 & 4.
Wardrobes 1, 2, 3, 4 can be now customized via char_storage but sql default is 80.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -477,7 +477,11 @@ namespace charutils
             "locker,"     // 2
             "satchel,"    // 3
             "sack,"       // 4
-            "`case` "     // 5
+            "`case`,"     // 5
+            "wardrobe,"     // 6
+            "wardrobe2,"     // 7
+            "wardrobe3,"     // 8
+            "wardrobe4 "     // 9
             "FROM char_storage "
             "WHERE charid = %u;";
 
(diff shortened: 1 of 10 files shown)


teschnei:
"Fix Mob casting as it was interrupted by spell range check and mobs can cast up to 40yalms."
put that in a separate pull request

src/map/utils/charutils.cpp
```diff
@@ -496,8 +500,10 @@ namespace charutils
             PChar->getStorage(LOC_MOGSACK)->AddBuff((uint8)Sql_GetIntData(SqlHandle, 4));
             PChar->getStorage(LOC_MOGCASE)->AddBuff((uint8)Sql_GetIntData(SqlHandle, 5));
 
-            PChar->getStorage(LOC_WARDROBE)->AddBuff(80); // Always 80..
-            PChar->getStorage(LOC_WARDROBE2)->AddBuff(80); // Always 80..
+            PChar->getStorage(LOC_WARDROBE)->AddBuff((uint8)Sql_GetIntData(SqlHandle, 6));
+            PChar->getStorage(LOC_WARDROBE2)->AddBuff((uint8)Sql_GetIntData(SqlHandle, 7));
+            PChar->getStorage(LOC_WARDROBE3)->AddBuff((uint8)Sql_GetIntData(SqlHandle, 8));
+            PChar->getStorage(LOC_WARDROBE4)->AddBuff((uint8)Sql_GetIntData(SqlHandle, 9));
```
teschnei: in what case would we need 4 separate columns for mog wardrobe size?

7cipherx7:
Mobs should not be interrupted when player run away at 204 but after 400 so this check we have is useless [https://github.com/7cipherx7/darkstar/blob/e93f5277115066386186fd613301492107e810ad/src/map/ai/states/magic_state.cpp#L186-L190](url) when getRange will overwrite is every single time.

As to the Wardrobe columns it should be customizable since not all wardrobes are ON by default on retail but you purchase it so why use one column for all 4 that would increase 1-4 together or using one column with bitwise check column would not be very friendly with non coders, so this makes customization more friendly, and its done with same select from DB not like we do multiple selects.

teschnei:
separate fixes must be in separate pull requests

for wardrobe, I thought about it later and I agree with that

7cipherx7:
well how about you commit this one since its only one fix and not multiple ones together in one pull and new feature implementation so it will be fine as all will get both  anyways and we move on forward with this and do this next time, agreed?

teschnei:
I split them for you.  Make sure you keep each logical change in a separate pull request (or at least a separate commit!) next time

Issue #3269, opened 2016-07-08
charybdis does not multi attack, because CheckMobMultiHits never gets called, ever


eraffxi:
**_I have:_**

- [ x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [ x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160427_0

**_Server Version_** (type `@revision` in game) **:**
dd31b5fc1da371803c8f5cb57cfc63374c8091a5

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

charybdis does not multhit like normal, put a breakpoint on the CheckMobMultiHits exception for 'unique mobs' where charybdis ID is highlighted, never got set off. set breakpoint on the function itself, nothing.

HeavensSword:
I found that CheckMobMultiHits is not actually called anywhere. I went ahead and split it into a new method that check multi-hits for special mobs like charby. Tested and works correctly; creating a pull request now.

HeavensSword:
We should be able to close this issue now after change #3278

PR #3277, opened 2016-07-11
Fix for mobs that have special multi-hit capabilities like Charybdis.…

HeavensSword:
… Split the existing CheckMobMultiHits method (which wasn't being called anywhere)  into a new method, CheckSpecialMobMultiHits, that handles calculating number of attacks for mobs specified.

Fix for issue #3269 [https://github.com/DarkstarProject/darkstar/issues/3269](url)

I created the fix based off of the existing code, but perhaps min-max attacks for mobs should be broken out into a lua script or placed in the DB?

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -2521,7 +2521,6 @@ namespace battleutils
 
     uint8 CheckMobMultiHits(CBattleEntity* PEntity)
     {
-
         if (PEntity->objtype == TYPE_MOB || PEntity->objtype == TYPE_PET)
         {
             uint8 num = 1;
@@ -2530,17 +2529,7 @@ namespace battleutils
             if (PEntity->GetMJob() == JOB_MNK)
             {
                 num = 2;
-            }
-
-            //check for unique mobs
-            switch (PEntity->id)
-            {
-                case 17498522:// Charybdis 2-6
-                    return (1 + getHitCount(5));
-
-                default:
-                    break;
-            }
+            }           
 
             int16 tripleAttack = PEntity->getMod(MOD_TRIPLE_ATTACK);
             int16 doubleAttack = PEntity->getMod(MOD_DOUBLE_ATTACK);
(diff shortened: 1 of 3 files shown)


teschnei:
"I created the fix based off of the existing code, but perhaps min-max attacks for mobs should be broken out into a lua script or placed in the DB?"
yeah... if we're here already, lets just get rid of this abomination before it grows
i know there are some mods related to multi attacks that probably aren't being used by mobs, it might be possible to use one of those for this case

eraffxi:
in the old function of checkmobmultihits there was a line about checking if it's a mnk and then making num = 2.  is that covered in a different way now?

HeavensSword:
> in the old function of checkmobmultihits there was a line about checking if it's a mnk and then making num = 2. is that covered in a different way now?

Yes. AttackRound's constructor checks if the attacker is using h2h and will create a left hand attack. The original function also checked for double/triple attack, etc, but that is all handled by CreateAttack now.

(Sent from my phone)

eraffxi:
ok yea i saw that part. just asking because i'm still trying to understand the bug where maat seems to attack too fast, as well as other mobs like alky gigas, we've resorted to switching their attack type to 7 (scythe) instead for now but they still attack at super fast speeds

src/map/utils/battleutils.cpp
```diff
@@ -2554,8 +2543,38 @@ namespace battleutils
...
+    uint8 CheckSpecialMobMultiHits(CBattleEntity* PEntity)
+    {
+        if (PEntity->objtype == TYPE_MOB || PEntity->objtype == TYPE_PET)
+        {
+            uint8 num = 1;
+            
+            // Check for unique mobs
+            switch (PEntity->id)
+            {
+                case 17498522: // Charybdis 2-6
```
takhlaq: MOBMOD_MULTIHIT or something
don't want this to be affected by id shifts (and i just hate hardcoded id checks in general)

takhlaq: nvm just read kayjays comment

HeavensSword: > MOBMOD_MULTIHIT or something
don't want this to be affected by id shifts (and i just hate hardcoded id checks in general)

I agree; I was using what was there as a quick fix, but wanted to get opinions on how it should be refactored. 

I can look into a proper fix later today.

HeavensSword:
I've instead created a new MOBMOD and have implemented a fix utilizing that.

Issue #3207, opened 2016-06-01
mandragoras not attacking twice


eraffxi:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160427_0

**_Server Version_** (type `@revision` in game) **:**

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
fight a mandragora

could this potentially be related to the other issue where mobs are attacking twice when they should not?  mandragoras are monks, and are countering attacks properly, but only attacking once when normally i believe they attack twice at a 100% rate.  perhaps the values representing single/double attacks are flipped?

HeavensSword:
Did some quick debugging; these particular mob's main hand is set to Scythe instead of H2H.

HeavensSword:
I notice all the mandragora are set to scythe. iirc they were set to that because of delay issues?

I set the mainhand of the Yhoator mandy to H2H in the sql and left the cmbDelay alone; everything seems fine to me.
Used : [](https://www.youtube.com/watch?v=qFqktcqvIuE) as a reference for attack speed.

bendangelo:
This issue fixed? @HeavensSword

HeavensSword:
Not yet; I'll take care of it tonight. I was holding off in case there was any reasons against me making the changes. Sounds like there are none. :)

HeavensSword:
Update:
I think I'm going to make this a part of a larger Mandragora Family fix. Going through the DB, I'm seeing other issues like incorrect jobs and missing spell lists for certain NMs in the mandragora family.

abriasffxi:
Any luck completing that?

HeavensSword:
Sorry I've been very busy irl and had to take a break from darkstar
updates. I may not get to it in a few more weeks.

PR #3278, opened 2016-07-11
Created new mob mod 'MOBMOD_MULTI_HIT'

HeavensSword:
Created new mob mod 'MOBMOD_MULTI_HIT' for specifying if a mob has the capability of attacking more than once per round. e.g. Charybdis.

I also went ahead and deleted the old and unused method **CheckMobMultiHits**.

Now, if a mob has special mutli-hit requirements, **setMobMod** just needs to be called in the mob's **onMobInitialize** of its associated lua script.

Fixes issue #3269

### DIFF
--- a/src/map/attackround.cpp
+++ b/src/map/attackround.cpp
@@ -199,11 +199,22 @@ void CAttackRound::DeleteAttackSwing()
 void CAttackRound::CreateAttacks(CItemWeapon* PWeapon, PHYSICAL_ATTACK_DIRECTION direction)
 {
     uint8 num = 1;
+    
+    bool isPC = m_attacker->objtype == TYPE_PC;
 
     // Checking the players weapon hit count
     if (PWeapon->getReqLvl() <= m_attacker->GetMLevel())
     {
         num = PWeapon->getHitCount();
+    }
+    
+    // If the attacker is a mobentity or derived from mobentity, check to see if it has any special mutli-hit capabilties
+    if (dynamic_cast<CMobEntity*>(m_attacker))
+    {
+        auto multiHitMax = static_cast<CMobEntity*>(m_attacker)->getMobMod(MOBMOD_MULTI_HIT);
+        
+        if (multiHitMax > 0)
+            num = 1 + battleutils::getHitCount(multiHitMax);
     }
 
     AddAttackSwing(PHYSICAL_ATTACK_TYPE::NORMAL, direction, num);
@@ -214,7 +225,7 @@ void CAttackRound::CreateAttacks(CItemWeapon* PWeapon, PHYSICAL_ATTACK_DIRECTION
     int16 quadAttack = m_attacker->getMod(MOD_QUAD_ATTACK);
 
     //check for merit upgrades
-    if (m_attacker->objtype == TYPE_PC)
+    if (isPC)
     {
         CCharEntity* PChar = (CCharEntity*)m_attacker;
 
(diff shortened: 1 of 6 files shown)


ibm2431:
Would this also fix #3207?

teschnei:
Mandragoras are monks, so they wouldn't need this

src/map/attackround.cpp
```diff
@@ -199,12 +199,23 @@ void CAttackRound::DeleteAttackSwing()
...
+    bool isPC = m_attacker->objtype == TYPE_PC;
 
     // Checking the players weapon hit count
     if (PWeapon->getReqLvl() <= m_attacker->GetMLevel())
     {
         num = PWeapon->getHitCount();
     }
+    
+    // If the attacker is a mob, check to see if it has any special mutli-hit capabilties
+    if (!isPC)
```
bendangelo: If mobentity would be a better check. A pet entity shouldn't run this code?

teschnei: a pet could (no reason to not), but it certainly shouldn't be !isPC - you've no guarantee there won't be some other entity type that could end up in here that can't be cast to a mob entity

HeavensSword: Sorry, this was an oversight on my end.

Should it be:
TYPE_MOB && TYPE_PET?

teschnei: you can either dynamic_cast it to a mobentity or do that

I think dynamic_cast is slower but if anything new derives from mobentity it'll already work without having to change that line..

HeavensSword: I've updated the check to use dynamic_cast and tested.

src/map/attackround.cpp
```diff
@@ -199,12 +199,23 @@ void CAttackRound::DeleteAttackSwing()
...
     // Checking the players weapon hit count
     if (PWeapon->getReqLvl() <= m_attacker->GetMLevel())
     {
         num = PWeapon->getHitCount();
     }
+    
+    // If the attacker is a mobentity or derived from mobentity, check to see if it has any special mutli-hit capabilties
+    if (dynamic_cast<CMobEntity*>(m_attacker))
+    {
+        uint8 multiHitMax = (uint8)static_cast<CMobEntity*>(m_attacker)->getMobMod(MOBMOD_MULTI_HIT);
```
teschnei: it's a bit weird to use a c-style cast and a static_cast on the same line... you don't really even need to cast to uint8, just leave it as whatever it is:
auto multiHitMax = static_cast<CMobEntity*>(m_attacker)->getMobMod(MOBMOD_MULTI_HIT);

HeavensSword:
Removed the c-style cast and use auto for the multiHitMax var.

src/map/utils/mobutils.cpp
```diff
@@ -991,7 +991,7 @@ void InitializeMob(CMobEntity* PMob, CZone* PZone)
     PMob->defaultMobMod(MOBMOD_2HOUR_PROC, 60);
     PMob->defaultMobMod(MOBMOD_SIGHT_RANGE, CMobEntity::sight_range);
     PMob->defaultMobMod(MOBMOD_SOUND_RANGE, CMobEntity::sound_range);
-
+    PMob->defaultMobMod(MOBMOD_MULTI_HIT, 0);// Mobs default to 0 multi-hit
```
takhlaq: this isnt needed (returns 0 if cant find the mod for mob)

HeavensSword: Thanks, I didn't know that. I'll go ahead and remove that line.

PR #3279, opened 2016-07-12
Fix build issue on g++ 4.9 (cf. N3922)

kjl2y:
The project will not build using gcc 4.9 because it deduces `auto x { y }` to the type `std::initializer_list<y>` instead of type `y`.  The alternative syntax `auto x = y;` deduces the correct type.  

In any case, both forms were being used throughout the codebase.  It may be clearer to stick with just one.


teschnei:
Hello, we don't support gcc less than 5.0. Sorry.

Issue #2942, opened 2016-03-26
charmed pets pettp not working


ymkc:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160203_0

**_Server Version_** (type `@revision` in game) **:**
8f2765e5

**_Source Branch_** (master/stable) **:**
Master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
Using <pettp> on charmed pets always returns 0, but sic seems to work correctly. 
Also works fine on jug pets.

ymkc:
Using pettp on charmed pets...

Hozu:
I noticed with summons, whenever I used exec to set the tp to an amount, pettp was reporting a different value than what I set. Like, I set it to 800 and it would have 1400.

PR #3282, opened 2016-07-13
Fix for <pettp> command

HeavensSword:
Charmed mobs were not sending the PetSync packet which caused the <pettp> command to always report the TP value from the last PetSync packet received (sent from the ApplyCharm function).

Fixes issue #2942


teschnei:
pet sync should be after entity update, I think

HeavensSword:
I based it on CPetEntity::PostTick, where the PetSyncPacket is being pushed before EntityUpdate as well. 
I can change the order for both though if it makes more sense.

teschnei:
yeah, do that

teschnei:
(unless it breaks things)

HeavensSword:
Ok, i'll test that now.

HeavensSword:
Looks like it works fine, though in testing i found out that choosing a jug pet ability with "Ready" seems to be off. It seems to use the ability index +1. I'll look into that as a part of my other branch dealing with some ability issues.

teschnei:
on second thought... can you try moving the petsync to charentity and checking PPet instead?

HeavensSword:
Sure; I'll be out of town this weekend, but can probably get to it Sunday night.

HeavensSword:
Actually looks like i'm leaving a few hours later today than expected, so i've made the changes and am testing now.

HeavensSword:
Scratch that, there are some issues with the new way. Not updating enough. I'll debug some more sunday.

teschnei:
if you put it inside the players updatemask check, it'd only update if the player had to update

I forgot about that, so it's probably better to keep it in mobentity/petentity since it's only needed when  the pet needs to update

PR #3284, opened 2016-07-17
Implement Auction House fees - no more free ride kids.

TeoTwawki:
**I finally got around to it.**

Tested. Displays correctly, deducts correct amount of gil at correct time.

Comes with settings, defaults are what both wiki's say the retail fees are.

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -2105,8 +2105,8 @@ void SmallPacket0x04E(map_session_data_t* session, CCharEntity* PChar, CBasicPac
             !(PItem->isSubType(ITEM_LOCKED)) &&
             !(PItem->getFlag() & ITEM_FLAG_NOAUCTION))
         {
-            PItem->setCharPrice(price);
-            PChar->pushPacket(new CAuctionHousePacket(action, PItem, quantity));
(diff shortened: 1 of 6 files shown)


src/map/packet_system.cpp
```diff
@@ -2163,18 +2163,33 @@ void SmallPacket0x04E(map_session_data_t* session, CCharEntity* PChar, CBasicPac
             !(PItem->isSubType(ITEM_LOCKED)) &&
             !(PItem->getFlag() & ITEM_FLAG_NOAUCTION))
         {
-            if (quantity == 0 &&
-                (PItem->getStackSize() == 1 ||
-                PItem->getStackSize() != PItem->getQuantity()))
+            uint32 AUCTION_FEE = 0;
```
teschnei: don't use caps for a local variable (caps are usually reserved for macros and enums)

TeoTwawki: I'll change it, it was copy pasta from the previous AUCTION_FEE garbage

TeoTwawki: p.s. this is why I asked before when doing the style guide for the repo, I forget these things

TeoTwawki:
Amended.

PR #3286, opened 2016-07-18
Correct some small mob issues

ymkc:
- Added MOBMOD_ALWAYS_AGGRO to Morbolger
- Variable Hares shouldn't be aggressive and have an added +10% xp bonus
- Groundkeepers on blue platforms (in sky) should aggro on sight
- Added droplist to Leshy
- Set repspawn to 300s on the edited mobs


bendangelo:

ymkc:
@Deadwing888 The 2nd to last column is the detects field, the 4th from the back is the dark element.
I just wanted to split the groundkeepers family because in sky all groundkeeps aggro by magic but only some aggro by sight. but I guess you know that and probably just missread the column :)

PR #3287, opened 2016-07-18
Take into account gear under level restriction

Kushdr:
AF head / relic hands giving tp bonus under level restriction. Wrapping the extra tick check with a player level >= 60 will account for all sync as there is no level restriction post 60. Just to be safe


teschnei:
the real question is: is there any situation where getEquipId should return it if it's synced? if not, i'd just change getEquipId to return 0 when its synced, because I am sure there are way more instances of this than just meditate

Kushdr:
I'm sure, but it's not as simple as "if has level sync, ignore this". At 60 cap, SAM AF helm would work as it's level 60. That was the the though.
We could do an item level and player level check in getEquipID I suppose?

teschnei:
yes, I meant when the item was synced (not the player)

PR #3288, opened 2016-07-19
Drg and Bst Ability Fixes (Initial)

HeavensSword:
- Implemented the emnity reduction and super climb triggering functionality of Super Jump
- Implemented Gauge.
- Added new blank tame script.
- Fixed wyvern types for GEO and RUN.

NOTE:
I originally had Gauge taking the light/apollo staves into account, but have read that they didn't effect Gauge's message on retail. EDIT: The GetCharmChance function takes a boolean (defaulted to true) to specify whether to use Charm affinity/chance mods in the calculation. Gauge.lua passes in false to ignore those mods, which should reflect retail more. If we ever want to diverge a bit and have Gauge report a more accurate analysis, we can just remove the "false".

TODO:
I was planning on adding logic for the invincibility frames of Super Jump and Super Climb, but wanted to get some input on how best to implement it. I initially was thinking of adding a new effect e.g "EFFECT_SUPER_JUMP". This will be added in a later PR.

### DIFF
--- a/scripts/globals/abilities/super_jump.lua
+++ b/scripts/globals/abilities/super_jump.lua
@@ -8,6 +8,7 @@
 
 require("scripts/globals/settings");
 require("scripts/globals/status");
+require("scripts/globals/pets");
 
 -----------------------------------
 -- onAbilityCheck
@@ -22,4 +23,21 @@ end;
 -----------------------------------
 
 function onUseAbility(player,target,ability)
+
+    -- Reduce 99% of total accumulated enmity
+    if (target:isMob()) then
+        target:lowerEnmity(player, 99);
+    end
+    
+    ability:setMsg(0);
+    
+    -- Prevent the player from performing actions while in the air
+    player:stun(5000);
+    
+    -- If the Dragoon's wyvern is out and alive, tell it to use Super Climb
+    local wyvern = player:getPet();
+    if (wyvern ~= nil and player:getPetID() == PET_WYVERN and wyvern:getHP() > 0) then 
+        wyvern:useJobAbility(636, wyvern);
+    end
+
 end;
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -366,6 +366,28 @@ inline int32 CLuaBaseEntity::isJugPet(lua_State* L)
 
 //======================================================//
 
+inline int32 CLuaBaseEntity::getCharmChance(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr || m_PBaseEntity->objtype == TYPE_NPC);
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isuserdata(L, 1));
+
+    CLuaBaseEntity* PLuaBaseEntity = Lunar<CLuaBaseEntity>::check(L, 1);
+    
+    auto PCharmer = static_cast<CBattleEntity*>(m_PBaseEntity);
+    auto PTarget = static_cast<CBattleEntity*>(PLuaBaseEntity->GetBaseEntity());
+    
+    bool includeCharmAffinityAndChanceMods = true;
+    if (!lua_isnil(L, 2) && lua_isboolean(L, 2))
+        includeCharmAffinityAndChanceMods = lua_toboolean(L, 2);;
+    
+    float charmChance = battleutils::GetCharmChance(PCharmer, PTarget, includeCharmAffinityAndChanceMods);
+    lua_pushnumber(L, charmChance);
+
+    return 1;
+}
+
+//======================================================//
+
 inline int32 CLuaBaseEntity::getMP(lua_State *L)
 {
     DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
@@ -11008,6 +11030,7 @@ Lunar<CLuaBaseEntity>::Register_t CLuaBaseEntity::methods[] =
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,familiar),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,getPetID),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,isJugPet),
+    LUNAR_DECLARE_METHOD(CLuaBaseEntity,getCharmChance),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,needToZone),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,getContainerSize),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,changeContainerSize),
--- a/scripts/globals/abilities/deep_breathing.lua
+++ b/scripts/globals/abilities/deep_breathing.lua
@@ -15,7 +15,7 @@ require("scripts/globals/status");
 function onAbilityCheck(player,target,ability)
     if (player:getPet() == nil) then
         return MSGBASIC_REQUIRES_A_PET,0;
-   elseif (player:getPetID() ~= 48) then
+   elseif (player:getPetID() ~= PET_WYVERN) then
       return MSGBASIC_NO_EFFECT_ON_PET,0;
     else
       return 0,0;
--- a/scripts/globals/abilities/gauge.lua
+++ b/scripts/globals/abilities/gauge.lua
@@ -13,7 +13,7 @@ require("scripts/globals/status");
 -- onAbilityCheck
 -----------------------------------
 
-function onAbilityCheck(player,target,ability)
+function onAbilityCheck(player, target, ability)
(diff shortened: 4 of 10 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -366,6 +366,106 @@ inline int32 CLuaBaseEntity::isJugPet(lua_State* L)
...
 //======================================================//
 
+inline int32 CLuaBaseEntity::getCharmChance(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isuserdata(L, 1));
+
+    CLuaBaseEntity* PLuaBaseEntity = Lunar<CLuaBaseEntity>::check(L, 1);
+    
+    auto PCharmer = dynamic_cast<CBattleEntity*>(m_PBaseEntity);
```
teschnei: just static cast these.. they're never going to be anything else

src/map/lua/lua_baseentity.cpp
```diff
@@ -366,6 +366,106 @@ inline int32 CLuaBaseEntity::isJugPet(lua_State* L)
 
 //======================================================//
 
+inline int32 CLuaBaseEntity::getCharmChance(lua_State* L)
```
teschnei: is this copied from somewhere? can you just call the other function?

HeavensSword: Not quite; I used some of the code from both "tryToCharm" and "TryCharm" in battleutils.cpp to get the charm percentage chance.
Those functions actually try to charm a mob, so I needed a function that just gave me the chance the player had to charm a particular target.

src/map/lua/lua_baseentity.cpp
HeavensSword: Fixed.

teschnei: Make a function in battleutils that gets the charm chance and have both
call that

src/map/lua/lua_baseentity.cpp
HeavensSword: > Make a function in battleutils that gets the charm chance and have both call that

Will do

HeavensSword: Refactored so that the charm and lua functions are now using the new GetCharmChance function to reduce the amount of duplicate/similar code.

scripts/globals/abilities/super_jump.lua
```diff
@@ -22,4 +23,18 @@ end;
...
+    end
+    
+    ability:setMsg(0);
+    
+    -- If the Dragoon's wyvern is out and alive, tell it to use Super Climb
+    local wyvern = player:getPet();
+    if (wyvern ~= nil and player:getPetID() == PET_WYVERN and wyvern:getHP() > 0) then 
+        wyvern:useJobAbility(636, wyvern);
+    end
+
```
Hozu: I wanted to get this working a long time ago, but I couldn't figure out how to make it so the player is unable to move/act for a bit. player:stun(2000) would crash, probably still does.

HeavensSword: Moving isn't an issue since the player is locked in the animation, but acting still is. I was able to activate a WS for instance.

Unless we already have one, it would be useful to have a function to disable an entity's movement and actions.

teschnei: movement is client side, we can't restrict that server side (the super jump animation itself will of course stop movement)

other than that, player:stun does exactly what you're asking

HeavensSword: I'll try adding a stun. 

Also, I mentioned it in my initial comment, do you guys have any preferred methods/ideas for getting the invincibility working for Super Jump/Climb? My initial thought was a new Effect that can be added to the player/wyvern for 5 secs.

teschnei: the only reason I didn't implement super jump when I did the rest of the earlier drg refactors was because I don't know how it behaves exactly (can party members cast on you? can they even target you?)
My first thought was to just exclude them from all targetfinding, but I don't remember if I tried it or not (and I guess it would probably be a custom status effect or some kind of flag on a timer, but it doesn't really matter which of those is chosen)

HeavensSword: iirc the Drg can't be targeted by pt members while in the air. I played a drg for many years on retail, but still I'm going off of memory.

Ffxiclopedia has some decent info in the notes section:
http://ffxiclopedia.wikia.com/wiki/Super_Jump

src/map/lua/lua_baseentity.cpp
```diff
@@ -366,6 +366,33 @@ inline int32 CLuaBaseEntity::isJugPet(lua_State* L)
...
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isuserdata(L, 1));
+
+    CLuaBaseEntity* PLuaBaseEntity = Lunar<CLuaBaseEntity>::check(L, 1);
+    
+    auto PCharmer = static_cast<CBattleEntity*>(m_PBaseEntity);
+    auto PTarget = static_cast<CBattleEntity*>(PLuaBaseEntity->GetBaseEntity());
```
takhlaq: you want to dynamic_cast here
dynamic_cast returns null pointer if cast is invalid (which is what you're checking a few lines below anyways)

HeavensSword: Heh, I originally had dynamic_cast if you look at the comments on the outdated diff above. XD

teschnei: might as well get rid of the PCharmer && PTarget check then, since those are never going to be null

HeavensSword: I'll go ahead and make that change now.

takhlaq: oh didnt see lol
think you should dynamic_cast anyways since getMod (which is being used in the battleutils function) wont work for npc/baseentity

takhlaq: or you could check that in GetCharmChance

HeavensSword: I originally had dynamic_cast and the null check in CLuaBaseEntity::getCharmChance because I figured someone could possibly pass in an npc/baseentity from lua (even if accidental).

teschnei: you could, but that's the scripter's problem - it's not a valid thing to check, so why should it silently fail?

HeavensSword: Fair enough. :)

I've removed the null check and left the static casts.

src/map/lua/lua_baseentity.cpp
```diff
@@ -366,6 +366,28 @@ inline int32 CLuaBaseEntity::isJugPet(lua_State* L)
 
 //======================================================//
 
+inline int32 CLuaBaseEntity::getCharmChance(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
```
takhlaq: `DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr || m_PBaseEntity->objtype == TYPE_NPC);`

HeavensSword: Updated, thanks!

scripts/globals/abilities/super_jump.lua
Hozu: From what I'm told, the user of Super Jump is unable to be targeted, and is an ineligible target for any AoE that they would be in range of during the duration of the jump.

HeavensSword: Just an update; i've added the player:stun and it works.

Hozu: Interesting, I guess the issue with it was resolved at some point. It caused a crash similar to the one that used to be caused by using a sleep potion.

TeoTwawki: To add to above: anything that was mid cast on a drg that just popped super jump will cancel, and any targeting that had them focused is lost.

HeavensSword:
Btw, I don't plan to make other changes to this PR unless it's something one of you guys needs me to change; stuff like the invulnerability for SJ I plan to add in a later PR. 
For now, this will get Gauge working and the Emnity Shedding functionality of SJ working which is better than nothing.

bendangelo:
@HeavensSword thanks for your work. Just a note, it's better to open smaller PRs, big changes like this tend to drag on too long.

teschnei:
actually it only dragged on cause I thought he was doing more to super jump

HeavensSword:
Yea, I'll finish up the Super Jump changes when I have some more time.

PR #3290, opened 2016-07-20
Added a subjob check to mobutils::SetupJob

ymkc:
Some mobs (atm only Ahriman's (war/blm) that I know of, but probably some NMs aswell) are missing MOBMOD_MAGIC_COOL/DELAY/etc because their main job doesn't have mp.

### DIFF
--- a/src/map/utils/mobutils.cpp
+++ b/src/map/utils/mobutils.cpp
@@ -588,6 +588,7 @@ void CalculateStats(CMobEntity * PMob)
 void SetupJob(CMobEntity* PMob)
 {
     JOBTYPE mJob = PMob->GetMJob();
+    JOBTYPE sJob = PMob->GetSJob();
 
     switch(mJob)
     {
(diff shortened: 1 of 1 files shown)


src/map/utils/mobutils.cpp
```diff
@@ -705,6 +706,33 @@ void SetupJob(CMobEntity* PMob)
...
+    switch(sJob)
+    {
+        case JOB_BLM:
+        case JOB_WHM:
+        case JOB_RDM:
+        case JOB_DRK:
+        case JOB_BRD:
+        case JOB_BLU:
+        case JOB_SMN:
+        case JOB_BRD:
```
takhlaq: you listed JOB_BRD twice (gj travis https://travis-ci.org/DarkstarProject/darkstar/builds/146070689#L1986)

HeavensSword:
I know you didn't do this, but I noticed that in the MainJob switch, BRD, BLU, RDM, and SMN are missing breaks.

`         case JOB_BRD:
             PMob->defaultMobMod(MOBMOD_MAGIC_COOL, 35);
             PMob->defaultMobMod(MOBMOD_GA_CHANCE, 25);
             PMob->defaultMobMod(MOBMOD_BUFF_CHANCE, 60);
             PMob->defaultMobMod(MOBMOD_MAGIC_DELAY, 10);
         case JOB_BLU:
             PMob->defaultMobMod(MOBMOD_MAGIC_COOL, 35);
         case JOB_RDM:
             PMob->defaultMobMod(MOBMOD_MAGIC_COOL, 35);
             PMob->defaultMobMod(MOBMOD_GA_CHANCE, 15);
             PMob->defaultMobMod(MOBMOD_BUFF_CHANCE, 40);
             PMob->defaultMobMod(MOBMOD_MAGIC_DELAY, 10);
         case JOB_SMN:
             PMob->defaultMobMod(MOBMOD_MAGIC_COOL, 70);
             // smn only has "buffs"
             PMob->defaultMobMod(MOBMOD_BUFF_CHANCE, 100);`

src/map/utils/mobutils.cpp
ymkc: @takhlaq fixed

ymkc:
great eyeballing :)

added the breaks

TeoTwawki:
I think there used to be a default value for those, and then the job switch statement would replace the value if a matching job was hit. Either I'm misremembering something else or it was accidently lost in some refactor. /shrug

ymkc:
I just saw the "mob has magic, but no cool down set" in the logs related to ahrimans and thought this might be helpful on some other mobs aswell

TeoTwawki:
Yeah definitely. I'm just guessing at how it happened.

src/map/utils/mobutils.cpp
```diff
@@ -692,19 +693,48 @@ void SetupJob(CMobEntity* PMob)
...
+        case JOB_WHM:
+        case JOB_RDM:
+        case JOB_DRK:
+        case JOB_BLU:
+        case JOB_SMN:
+        case JOB_BRD:
+        case JOB_NIN:
+            if(PMob->getMobMod(MOBMOD_MP_BASE))
+            {
+                if (!PMob->getMobMod(MOBMOD_MAGIC_COOL))
```
bendangelo: You don't need these if statements. `defaultMobMod` does the check for you.

bendangelo:
I'll merge this in after you make that small change.

src/map/utils/mobutils.cpp
ymkc: @bendangelo thanks for the hint

bendangelo:

src/map/utils/mobutils.cpp
```diff
@@ -692,19 +693,41 @@ void SetupJob(CMobEntity* PMob)
...
+    {
+        case JOB_BLM:
+        case JOB_WHM:
+        case JOB_RDM:
+        case JOB_DRK:
+        case JOB_BLU:
+        case JOB_SMN:
+        case JOB_BRD:
+        case JOB_NIN:
+            if(PMob->getMobMod(MOBMOD_MP_BASE))
```
TeoTwawki: aaaand what happens for blm/blm mobs? unlike players this actually happens with mobs they legit have both jobs..

teschnei: er, i thought i commented about this before, but I guess i thought about it and forgot..

it also depends on which mod we're talking about - a smn/whm should probably have the whm's magic cooldown, for example

TeoTwawki: Someone else did but I had missed it https://github.com/DarkstarProject/darkstar/pull/3290#discussion_r72143359
so what happens is they get the first one to trigger (the main job) currently.

PR #3291, opened 2016-07-21
Fixes for MOBMOD_IDLE_DESPAWN

ymkc:
It's using getMobMod instead of getBigMobMod, and it should be applied to every mob on init/spawn(not just onDisenage) as some mobs are using it that way.

### DIFF
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -992,7 +992,7 @@ void CMobEntity::OnDisengage(CAttackState& state)
 
     if (getMobMod(MOBMOD_IDLE_DESPAWN))
     {
-        SetDespawnTime(std::chrono::milliseconds(getMobMod(MOBMOD_IDLE_DESPAWN)));
+        SetDespawnTime(std::chrono::seconds(getMobMod(MOBMOD_IDLE_DESPAWN)));
     }
     // this will let me decide to walk home or despawn
     m_neutral = true;
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -918,7 +918,7 @@ void CMobController::Disengage()
 
     if (PMob->getMobMod(MOBMOD_IDLE_DESPAWN))
     {
-        PMob->SetDespawnTime(std::chrono::milliseconds(PMob->getMobMod(MOBMOD_IDLE_DESPAWN)));
+        PMob->SetDespawnTime(std::chrono::seconds(PMob->getMobMod(MOBMOD_IDLE_DESPAWN)));
     }
 
     PMob->delRageMode();


src/map/entities/mobentity.cpp
```diff
@@ -603,7 +603,14 @@ void CMobEntity::Spawn()
         }
     }
     
-    m_DespawnTimer = time_point::min();
```
TeoTwawki: Perhaps I'm misunderstanding but..This mobmod is for NMs that stay up for so many seconds after the alliance wipes. And anything flagged as an NM should already ignores the regular depop timer even without this mobmod set at all. So the conditional check here should not be required.

As far as bigMobMod...I was shocked to look here and see `milliseconds` when the normal value is in seconds  (300 for a mob that will stay up at its ??? spot for 5 min before depopping).

teschnei: adding on to the second part... why use bigmobmod instead of just using std::chrono::seconds?

TeoTwawki:
When I saw the PR title I was momentarily overjoyed, thought it was someone finally adding the missing mobmod to all those NMs it belongs on that currently don't have it and instead have an erroneous depop time set on their `spawnMob()` command in script (coz those scripters didn't know what that argument actually did) 😢

HeavensSword:
@TeoTwawki 
I can do that sometime this week if you'd like lol.

TeoTwawki:
@HeavensSword oh yes please. Just seek out all the instances of it using an extra param after the mobID. that value causes stuff to depop before it should (its a timer that's started when the mob spawned, rather than after it declaims). Every server with a group that sucks at holding their claim (or had their tank hit by charm..) will love you for it.

src/map/entities/mobentity.cpp
ymkc: @teschnei I had it that way but saw that all the others are doing it this way

ymkc:
That's what I was thonking of. Some mobs already have it (eg lumbering jack). 

But for that to work it's not enough to just set it on onDisengage

src/map/entities/mobentity.cpp
teschnei: use seconds... that's the point of using the chrono library

TeoTwawki:
I've been setting the mobMod on spawn. https://github.com/DarkstarProject/darkstar/commit/89b8dedd43e334ff7bab116061dfaaacb22e354a#diff-037c4036bce941a85ce6be9233a7805eR246

ymkc:
I did the same, but it's only applied when the mob hits onDisengage. Or did I miss something in the code?

TeoTwawki:
I dunno but back when kj added it, it was working in game. Its not working now if set in the sql? I never tried it in script.

ymkc:
I tested it with silk caterpillar. Removed the 210 depop from the spawnMob and added he mobmod. Never depopped :|

TeoTwawki:
You tested via the sql table, restarting server, or you tried to script it? 

I checked one of the ZNM I set it on, and they depopped a bit after I zoned myself out and back using `@goto <me>`

Looking at the code it actually shouldn't matter where that mobmod was set (whether script or not, nor which function it was set in). I do see why it does matter where in the core its being checked though, there is a reason it was checked at disengage - that's when we want the timer to start.

src/map/entities/mobentity.cpp
ymkc: Moved it to sec

ymkc:
I added it via a scriot in onMobInit. I know it works on"losing" aggro/onDisengage. But a mob can spawn idle and should then despawn aswell, because it fits the description of the mobmod. Mob is idle, so if it's set, it should despawn. 

Maybe I just missinterpreted the name/description, but that's why why I did a pr, because it makes more sense to me :)

TeoTwawki:
yeah, you did. and we can't have it work that way, breaks the point of having it. Retail behavior is the mobs depop timer doesn't start till it "goes idle" out of its battle AI with no more players to attack. The other behavior you expectd should not happen on the NMs we are using this mobmod on, and that is accomplished by a different method (that I know of, its only there to auto depop mobs that spawn under fog/night conditions, at least that's what teschnei  said when I asked why that even did what it does because I thought that param should do what the mobmod now handles)

ymkc:
Oh ok, i see. My bad, I will close the pr

TeoTwawki:
I think we could still use the milliseconds >> seconds fix? I could be wrong but that seemed like it should be seconds.

ymkc:
I will drop the other commit in a min

teschnei:
"I dunno but back when kj added it, it was working in game."

here I am thinking "man I have no idea how this works, I hope whoever implemented it knows what it does"

TeoTwawki:
lmao

PR #3292, opened 2016-07-21
Internal table for quest logs, mission logs, and fame.

ibm2431:
This PR addresses the issue(s) raised in #3274 in regards to Lua constants for city/areas having different values depending on context, and potentially conflicting. It does so by using a lookup table, as @TeoTwawki mentioned.

~~Although instead of a Lua global, this uses an internal table in `quest_mission_log.h`. Because having it on the C-side:~~
- Doesn't harm any existing scripts!
- Doesn't require scripts to use a Lua lookup table/function [ex: `player:completeQuest(getlog(BASTOK, QUEST), QUEST_NAME)` ] going into the future. It's still `player:completeQuest(BASTOK, QUEST_NAME)` like normal
- Allows us to greatly simplify `quest_mission_log.cpp` itself! I cut down the crazy-long switch statement, and was able to clarify confusing additions/subtractions like `[MISSION_SOA - 11].current` into `[LOG_SOA].current`
- Can let us abuse context a bit: `player:completeQuest(KAZHAM, QUEST_NAME)`, which I'll explain in a bit.

~~Basically, I defined a set of city/area/mission IDs, taking all quests/missions/fame areas into account. I referred to this as a "content ID" because I couldn't come up with a better variable name. These are synced up with corresponding Lua globals, included into quest/mission/shop files. This gets mapped to the true ID needed via the `LOG_TYPES` table in the `quest_mission_log.h` depending on the function exposed/called via Lunar.~~

I made sure to include all (differing) cities in these content IDs, so they can look up a correct value if used in an appropriate context. For example, Kazham doesn't have its own unique quest log, but all its quests are Outlands quests. So we can (and I did) link Kazham's "quest log" to Outlands, and fame to Windurst, so if someone who is working on a Kazham quest "erroneously" does:
```
player:completeQuest(KAZHAM, QUEST_NAME);
player:addFame(KAZHAM);
```
The correct quest will be completed in the Outlands log, with fame being added for the right Windurst fame area.

Aliases work too, by defining the alias to the same "content ID" as its equivalent. ADOULIN and SOA can be used interchangeably without worry (`player:addFame(ADOULIN), player:getCurrentMission(SOA)`) because they're both LOG_ADOULIN's ID of 31. A quest added in WOTG can be referenced through "WOTG", or by the proper quest log name of "CRYSTAL_WAR".

I considered adding aliases for Nashmau (TOAU) and Shadowreign Cities (CRYSTAL_WAR), but didn't. Someone can do so later if they feel it worthwhile to be able to do something like `completeQuest(WINDURST_S, THE_TIGRESS_STIRS)`.

Other things I did while I was messing with all this:
- Added some verification for fame functions to make sure they were called with a valid fameArea
- Cut down the function(s) for completed Campaign missions from two down to one by adding a variable.
- Made references to current/complete status as a constant instead of non-descriptive hard-coded 0x01s and 0x02s.
- Some more comments explaining how quest_mission_log packets work, in case anyone else finds themselves working with them later.

### DIFF
--- a/scripts/commands/checkmission.lua
+++ b/scripts/commands/checkmission.lua
@@ -13,7 +13,15 @@ cmdprops =
 
 function onTrigger(player,logId,target)
     
+    local logName;
     logId = tonumber(logId) or _G[logId];
+    if ((type(logId) == "table")) then
+        logName = logId["full_name"];
+        logId = logId["mission_log"];
+    else
+        local missionAreas = {SANDORIA, BASTOK, WINDURST, ZILART, TOAU, WOTG, COP, ASSAULT, CAMPAIGN, ACP, AMK, ASA, SOA, ROV};
+        logName = missionAreas[logId + 1]["full_name"];
+    end
     
     if (logId == nil) then
         player:PrintToPlayer( "You must enter a valid LogID!" );
@@ -28,52 +36,18 @@ function onTrigger(player,logId,target)
     local targ = GetPlayerByName(target);
     local Log = targ:getCurrentMission(logId)
     -- Todo: loop through missionIDs and output mission name, not just a number.
-    -- Hopefully can use same method to replace this annoying pile of elseif's.
     if (targ ~= nil) then
-        if (logId == 12) then    -- Seekers of Adoulin
-            player:PrintToPlayer( string.format( "Current Seekers of Adoulin Mission ID is: '%s' !", Log ) );
-        elseif (logId == 11) then -- A Shantotto Ascension
-            player:PrintToPlayer( string.format( "Current Shantotto Ascension Mission ID is: '%s' !", Log ) );
-        elseif (logId == 10) then -- A moogle Kupo d'Etat
-            player:PrintToPlayer( string.format( "Current moogle Kupo dEtat Mission ID is: '%s' !", Log ) );
-        elseif (logId == 9) then -- A Crystalline Prophecy
-            player:PrintToPlayer( string.format( "Current Crystalline Prophecy Mission ID is: '%s' !", Log ) );
-        elseif (logId == 8) then -- Campaign
-            player:PrintToPlayer( string.format( "Current Campaign OPs ID is: '%s' !", Log ) );
-        elseif (logId == 7) then -- Assault
-            player:PrintToPlayer( string.format( "Current Assault Mission ID is: '%s' !", Log ) );
-        elseif (logId == 6) then -- Chains of Promathia
-            player:PrintToPlayer( string.format( "Current Chains of Promathia Mission ID is: '%s' !", Log ) );
-        elseif (logId == 5) then -- Wings of the Goddess
-            player:PrintToPlayer( string.format( "Current Wings of the Goddess Mission ID is: '%s' !", Log ) );
-        elseif (logId == 4) then -- Treasures of Aht Urgan
-            player:PrintToPlayer( string.format( "Current Treasures of Aht Urgan Mission ID is: '%s' !", Log ) );
-        elseif (logId == 3) then -- Rise of the Zilart
-            if (Log == 255) then
-                player:PrintToPlayer( "No current Rise of the Zilart mission." );
+        if (logId <= 13) then
+            if ((logId <= 3) and (Log == 255)) then
+                player:PrintToPlayer( string.format( "No current %s mission.", logName ) );
             else
-                player:PrintToPlayer( string.format( "Current Rise of the Zilart Mission ID is: '%s' !", Log ) );
-            end
-        elseif (logId == 2) then -- Windurst
-            if (Log == 255) then
-                player:PrintToPlayer( "No current Windurst mission." );
-            else
-                player:PrintToPlayer( string.format( "Current Windurst Mission ID is: '%s' !", Log ) );
-            end
-        elseif (logId == 1) then -- Bastok
-            if (Log == 255) then
-                player:PrintToPlayer( "No current Bastok mission." );
-            else
-                player:PrintToPlayer( string.format( "Current Bastok Mission ID is: '%s' !", Log ) );
-            end
-        elseif (logId == 0) then -- San d'Orea
-            if (Log == 255) then
-                player:PrintToPlayer( "No current San d'Orea mission." );
-            else
-                player:PrintToPlayer( string.format( "Current San dOrea Mission ID is: '%s' !", Log ) );
+                player:PrintToPlayer( string.format( "Current %s Mission ID is: '%s' !", logName, Log ) );
             end
         else     -- Everything not valid or not currently handled
-            player:PrintToPlayer("Invalid or unsupported LogID. The LogID must be 0-12.");
+            player:PrintToPlayer("Invalid or unsupported LogID. The LogID must be 0-13.");
         end
+    else
+        player:PrintToPlayer( string.format( "Player named '%s' not found!", target ) );
+        player:PrintToPlayer( "@checkmission <logID> <player>" );
     end
 end
--- a/scripts/commands/checkmission.lua
+++ b/scripts/commands/checkmission.lua
@@ -16,11 +16,11 @@ function onTrigger(player,logId,target)
     local logName;
     logId = tonumber(logId) or _G[logId];
     if ((type(logId) == "table")) then
-        logName = logId["full_name"];
-        logId = logId["mission_log"];
+        logName = logId.full_name;
+        logId = logId.mission_log;
     else
         local missionAreas = {SANDORIA, BASTOK, WINDURST, ZILART, TOAU, WOTG, COP, ASSAULT, CAMPAIGN, ACP, AMK, ASA, SOA, ROV};
-        logName = missionAreas[logId + 1]["full_name"];
+        logName = missionAreas[logId + 1].full_name;
     end
     
     if (logId == nil) then
--- a/scripts/commands/checkmission.lua
+++ b/scripts/commands/checkmission.lua
@@ -19,8 +19,13 @@ function onTrigger(player,logId,target)
         logName = logId.full_name;
         logId = logId.mission_log;
     else
-        local missionAreas = {SANDORIA, BASTOK, WINDURST, ZILART, TOAU, WOTG, COP, ASSAULT, CAMPAIGN, ACP, AMK, ASA, SOA, ROV};
-        logName = missionAreas[logId + 1].full_name;
+        if (logId <= 2) then
+            local logNames = {"San d'Oria", "Bastok", "Windurst"}
+            logName = logNames[logId + 1];
+        else
+            local missionAreas = {ZILART, TOAU, WOTG, COP, ASSAULT, CAMPAIGN, ACP, AMK, ASA, SOA, ROV};
+            logName = missionAreas[logId - 2].full_name;
+        end
     end
     
     if (logId == nil) then
--- a/src/map/packets/quest_mission_log.cpp
+++ b/src/map/packets/quest_mission_log.cpp
@@ -37,70 +37,71 @@ CQuestMissionLogPacket::CQuestMissionLogPacket(CCharEntity * PChar, uint8 logID,
 
     // FFXI packs different TOAU information in the same packet as certain other content, so we'll have to work around it.
     // First, deal with all quest areas which aren't TOAU:
-    if ((logType <= LOG_QUEST_COMP) && (logID != QUESTS_TOAU))
+    if ((logType <= LOG_QUEST_COMPLETE) && (logID != QUESTS_TOAU))
     {
         // We're updating any non-TOAU quest log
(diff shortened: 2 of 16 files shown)


src/map/packets/quest_mission_log.h
```diff
@@ -53,38 +98,60 @@
...
-#define MISSION_CAMPAIGN2               0x14
-
-#define MISSION_ACP                     0x15
-#define MISSION_AMK		                0x16
-#define MISSION_ASA		                0x17
-
-#define MISSION_SOA					    0x18
-#define MISSION_ROV                     0x19
+// Log Types
+const int LOG_TYPES[34][7] =
```
teschnei: i'm not gonna lie, this is exactly the kind of thing that we have Lua for.. i really don't like this being in here. at least 50% of it is not even valid

ibm2431: By invalid do you mean references called in the wrong context linking to invalid data (Tavnazia, Fame: -1)? All valid references point to the correct values in a character's set of mission logs.

Regardless of which C/Lua side the table is on, we'll still need some sort of indication (-1) that a subject was called in an inappropriate context. Transitioning to a key-based lookup (in Lua, prior to sending it to the Lunar binding somehow) with individual tables (mission only, quest only, etc) and flagging an error if the key wasn't found would avoid explicitly listing invalid references. But that requires mass-replacing calls in all existing scripts (from Lunar-bound player:completeQuest to something like a Lua-intermediary player_util:completeQuest), which is what I was attempting to avoid.

The replacing itself wouldn't be an issue, but all non-vanilla-DSP scripts would break and probably result in several unhappy server operators. Not that I personally mind; just trying to avoid making waves.

I will also admit a personal preference to referencing/verification on C-side as opposed to script side because that's less likely to get screwed up. The lookup occurring on the back-end means it'll always result in the correct value. But if it's on Lua's end, someone could use the old (current) player:completeQuest instead of the new one (maybe they're using an old guide/documentation, or simply didn't get the memo) to result in the server pulling/saving in the wrong log location. But I fully acknowledge that's just personal preference based on a latent attitude of distrust.

Is mass-replacing the existing calls an avenue you'd rather pursue?

TeoTwawki: > ut that requires mass-replacing calls in all existing scripts (from Lunar-bound player:completeQuest to something like a Lua-intermediary player_util:completeQuest), which is what I was attempting to avoid.

I know how to avoid that, and also a mass replace isn't too terrible just hassles ppl with pen prs and custom content on their servers. I've done that to this project several times now (dance puppets, dance! mwahahaha!)

when I am free tomorrow I'll look for you on irc

ibm2431: > I know how to avoid that

You know, I'm feeling a bit sheepish right now... because there probably is a way to not go straight to the declared Lunar method and I'm not familiar enough with Lunar to recognize it (my guess: renaming the current binding, and having new Lua-intermediary take its place, calling the new binding?). Going to see if I can figure it out before you tell me!

teschnei: you aren't gonna fill up a lua table with a bunch of values that mean nothing, you just won't put anything there and you'll get nil (which is totally fine)

TeoTwawki: you are over thinking it.and appear to not really grasp how the connecting glue actually works.

the log ids in the current trunks global, to lua's perspective, are nothing but global variables.

TeoTwawki: @joncloud I wish I had time right now to help

TeoTwawki: ...thanks for auto filling the wrong name github -_-

ibm2431:
Lua Include:
```
ADOULIN = {
    ['quest_log']= quest_log_id,
    ['mission_log']= mission_log_id,
    ['fame_area'] = fame_id
}
```

C:
```
inline int32 CLuaBaseEntity::getCurrentMission(lua_State *L)
{
    lua_getfield(L, 1, "mission_log");
    uint8 MissionLogID = lua_tointeger(L, -1);
    if (MissionLogID < MAX_MISSIONAREA)
    {
        MissionID = (uint8)((CCharEntity*)m_PBaseEntity)->m_missionLog[MissionLogID].current;
    }
}

```

NPC Script:
`local SOA_Mission = player:getCurrentMission(ADOULIN);`

Apologies for my earlier silliness.

teschnei:
I forgot to mention this earlier, but changing all the lua calls is probably the preferable solution here - since this is a FOSS project, I'm going to always err on understandable code over things like breaking custom changes (or breaking anything really)

ibm2431:
Okay, so I've moved the table out of the quest_mission_packet header and put it in Lua where it belongs. And it's a lot better now! Now Lua scripts can use it too, as you can see in the updated commands - [checkmission](https://github.com/ibm2431/darkstar/blob/19c1b346ad30459d25d95b3c65525834965a7c2c/scripts/commands/checkmission.lua) probably appreciates it.

Basically, the binding checks the first argument. If it's a table (from [log_ids.lua](https://github.com/ibm2431/darkstar/blob/19c1b346ad30459d25d95b3c65525834965a7c2c/scripts/globals/log_ids.lua), with the function call `player:completeQuest(ADOULIN, EMPTY_NEST)` ), it'll try to pop the appropriate log ID from the Lua table onto the top of the stack. It then takes either this new item, or if logID is just a number instead of a table, it'll go with the original passed value. So the binding can still take integer log IDs if need be.

src/map/packets/quest_mission_log.cpp
```diff
@@ -28,211 +28,123 @@
...
-        if (status == 0x01)
-        {
-            generateCurrentMissionPacket(PChar);
-            logType = MISS_CURRENT;
-            break;
+        else {
+            // Completed TOAU Quests share a packet with completed Assault Missions
+            generateAssaultMissionPacket(PChar); // Writes in same packet
         }
+        packetType = QUEST_PACKET_BYTES.at(logID * 2 + logType);
```
teschnei: structure your map so you don't need to do this - unordered_map<int, pair>, maybe

src/map/packets/quest_mission_log.h
```diff
@@ -24,67 +24,81 @@
...
+#define MISSION_ASSAULT      7
+#define MISSION_CAMPAIGN     8
+#define MISSION_ACP          9
+#define MISSION_AMK         10
+#define MISSION_ASA         11
+#define MISSION_SOA         12
+#define MISSION_ROV         13
+
+// Quest Log Packet Bytes
+const std::unordered_map<uint8, uint16> QUEST_PACKET_BYTES =
```
teschnei: this should not be capitalized and it should be static

teschnei:
i don't really like the idea of using a pair as the key instead of value, but I haven't had much time to look into how the alternative would work yet

ibm2431:
Another alternative is two maps. One for the current quest byte values, and a second for the complete byte values. Then just choose which one to pick from based off logType, probably with a ternary.

teschnei:
thats the same as having one map with the value as a pair

ibm2431:
Oh right, sorry. Wasn't thinking. So just have the key be the value of the quest log (0 to max_quest area), return a pair of current quest byte value + complete quest byte value, then just take off the one we need?

teschnei:
that's right

scripts/commands/checkmission.lua
```diff
@@ -13,7 +13,15 @@ cmdprops =
...
 function onTrigger(player,logId,target)
     
+    local logName;
     logId = tonumber(logId) or _G[logId];
+    if ((type(logId) == "table")) then
+        logName = logId.full_name;
+        logId = logId.mission_log;
+    else
+        local missionAreas = {SANDORIA, BASTOK, WINDURST, ZILART, TOAU, WOTG, COP, ASSAULT, CAMPAIGN, ACP, AMK, ASA, SOA, ROV};
+        logName = missionAreas[logId + 1].full_name;
```
ibm2431: They should be included by the require for [missions](https://github.com/ibm2431/darkstar/blob/953cedb579fc20dbd74db3b92f55c115dcecdde8/scripts/globals/missions.lual)

takhlaq: o ya
didnt notice that

ibm2431:
So bit of an oversight on my part. The Conquest system and a couple NPCs check against static integers for the three nations and/or the player's current nation. Depending on where/when the variables are defined/used, this could lead to conflicts.

Since the three nations are unique in that their mission, quest, and fame IDs are in perfect alignment and don't conflict with other areas, I've commented their tables out and set them back to static integers. Core works with both integers and tables, so there shouldn't be any ill effects. Non-starting nations are still tables.

src/map/packets/quest_mission_log.h
```diff
@@ -25,66 +25,73 @@
...
+#define MISSION_SOA         12
+#define MISSION_ROV         13
+
+// Log Types
+enum LOG_TYPE
+{
+    LOG_QUEST_CURR = 1,
+    LOG_QUEST_COMP = 2,
+    LOG_MISS_CURR = 3,
+    LOG_MISS_COMP = 4,
```
takhlaq: LOG_MISSION_COMPLETE etc
need to get rid of the shorthand enums at some point (like for skills)

src/map/packets/quest_mission_log.h
```diff
@@ -25,66 +25,73 @@
...
+};
+
+// Mission Log Packet Bytes
+#define MISS_COMPLETE       0xD0
+#define MISS_CURRENT        0xFFFF
+#define EXP_MISS_COMPLETE   0xD8
+#define EXP_MISS_CURRENT    0x80
+#define ASSAULT_COMPLETE    0xC0
+#define CMPGN_MISS_UN       0x30
+#define CMPGN_MISS_DEUX     0x38
```
takhlaq: CAMPAIGN_MISSION_DEUX etc

teschnei: you could replace UN and DEUX with one and two while you're at it..

ibm2431:
I have since addressed the shared conquest/questArea variables issue for the three nations in #3322. 

So I've now made the values for those variables (SANDORIA, BASTOK, and WINDURST) in `log_ids.lua` back into tables, under the assumption that the PR being merged will resolve any problems with said contexts (conquest + player nation _vs._ quest + mission logIDs) being used in the same script.

takhlaq:
i can't see anything i want changing really so i guess i'll merge when you get time to resolve conflicts (or i can wait til @teschnei takes another look)

scripts/commands/checkmission.lua
```diff
@@ -13,7 +13,20 @@ cmdprops =
...
 function onTrigger(player,logId,target)
     
+    local logName;
     logId = tonumber(logId) or _G[logId];
+    if ((type(logId) == "table")) then
+        logName = logId.full_name;
+        logId = logId.mission_log;
+    else
+        if (logId <= 2) then
+            local logNames = {"San d'Oria", "Bastok", "Windurst"}
```
TeoTwawki: @ibm2431 we have a prob here in need of fixing. `San d'Oria` can never be input because the command interpretation would see it as multiple arguments. Better off being `sandoria`, `bastok`, `windurst` and fed through `string.upper()` to kill any case sensitivity. I don't know where else these strings got used exactly, no time to go look.

ibm2431: I think those strings were slapped in because prior to #3322 the nation variables were still integers due to conquest scripts. So a command of `@checkmission player SANDORIA` would always come in as an integer, and didn't have a full name in a would-be table to grab. Hence quick if check with local array of strings just to get the "San d'Oria" for the "Current San d'Oria Mission is 5".

Although since they're actually tables now, `SANDORIA`, `BASTOK`, and `WINDURST` could probably be tossed into the missionAreas array and used like all the others to fetch a full log name, killing the `if (logId <= 2)` check entirely.

Regarding killing case sensitivity, I think each QMF area is its own variable and isn't part of a parent, so key-based reference (ie: `log_information[string.upper(text_from_gm)]`) might need a lookup table added
```
log_information = {
    ["SANDORIA"] = SANDORIA
```
to link a given string to a table.

Another potential issue is that I think at the moment all the GM commands just treat/invoke the argument as if they were a raw variable (ie: literally passing the `ADOULIN` table instead of a string of "ADOULIN").

TeoTwawki: well if it didn't ride in with yer changes, must have been @takhlaq DEMOLISH! 😄 I volunteer you to see wtf is up and do whatever is needed, coz I know it was you that went _G happy with commands.

takhlaq: maaaybe
but commit log says i did that waaay before this was merged https://github.com/DarkstarProject/darkstar/commit/a8969109828a88c5455a0de3ae1c8c4b0a44ff12

changing it so its like how teo said should work fine though

PR #3295, opened 2016-07-23
Batch fixup SpawnMob

ymkc:
Moved the depop time parameter from SpawnMob to MobMods.
A few specially syntax'ed mobs are missing, but the bulk of the mobs should be in here


TeoTwawki:
Somebody with merge powers, please make this happen sometime soon. Solves multiple issues where an NM will depop under conditions where it would not on retail.

PR #3299, opened 2016-07-28
Deprecated `@` symbol use in GM commands. It still works for now alongside its replacement 

TeoTwawki:
Deprecated `@` symbol use in GM commands. It still works for now alongside its replacement `$` and all documentation has been updated.

I cannot guarantee that the automation didn't mangle the human readability of a line or two someplace, but I did make sure the console log was clean of errors after.

Also, during this process a great many script errors were spotted. Most incidents were missing requires in quest scripts and these are now corrected.

Reasoning:
Many modern sites and services use the AT symbol (`@`) to direct messages to users, including the service currently hosting our repository. Someone unfamiliar with this (or simply doesn't consider it or doesn't care) can easily create situations where persons not involved in the project could be spammed by our issue tracker or even our commit messages. **There are already instances of people being auto-messaged by the project and its forks.**

@teschnei you said you didn't care, so here it is.

**`@`anyone else reading this:** if you have an objection (better be a good one) to the key I chose that doesn't amount to "*I don't like change / its not my problem*", you should speak up now. 

For the real diehard "I refuse to change ever!" people, its trivial to make script add on in either windower or ashita to catch and replace the command character you are used to using.


TeoTwawki:
Aside from [this](https://github.com/TeoTwawki/darkstar/blob/GM/src/map/packet_system.cpp#L3692) and the missing requires/bad vars I fixed, its all comment changes.

PR #3302, opened 2016-07-31
Killer Effect merits and idea for Circle Effects

ciradan:
Added the killer effect merits to be taken into account (like anyone would ever merit that anyway ^^')

Also tried to start implementing the Circle Effects, the 15% damage reduction/increase.
I'm not sure whether my solution is good or not, so this is basically just a prototype to gather some feedback. ^^'
For now I only did Ancient Circle (the ability lua should be fine, not sure about the takePhysicalDamage thing).

If it's fine how I tried to implement it I can do the other Circle Effects as well.

### DIFF
--- a/scripts/globals/abilities/ancient_circle.lua
+++ b/scripts/globals/abilities/ancient_circle.lua
@@ -7,7 +7,7 @@
 
 --According to BGWiki:
 --	When cast as Subjob DRG it gives 5% damage reduction/increase
---	as Mainjog it gives 15%
+--	as Mainjob it gives 15%
 --	when wearing certain Brais it gets another +2% power
 -----------------------------------
 
@@ -27,19 +27,19 @@ end;
 -----------------------------------
 
 function onUseAbility(player,target,ability)
-	local power = 5;			--when cast as Subjob
-	local duration = 180;
-	
-	if (player:getMainJob() == 14) then
-		local sLegs = player:getEquipID(SLOT_LEGS);
-		
-		if (sLegs == 14227 or sLegs == 15574 or sLegs == 28103) then
-			power = 17;			--when cast as Mainjob w/ equipment
-			duration = 240;		--Brais increase duration by 50%
-		else
-			power = 15;			--when cast as Mainjob w/o equipment
-		end
-	end
+    local power = 5;			--when cast as Subjob
+    local duration = 180;
+
+    if (player:getMainJob() == JOB_DRG) then
+        local sLegs = player:getEquipID(SLOT_LEGS);
+
+        if (sLegs == 14227 or sLegs == 15574 or sLegs == 28103) then
+            power = 17;			--when cast as Mainjob w/ equipment
+            duration = 240;		--Brais increase duration by 50%
+        else
+            power = 15;			--when cast as Mainjob w/o equipment
+        end
+    end
 	
     target:addStatusEffect(EFFECT_ANCIENT_CIRCLE,power,0,duration);
 end;
--- a/scripts/globals/abilities/ancient_circle.lua
+++ b/scripts/globals/abilities/ancient_circle.lua
@@ -4,6 +4,11 @@
 -- Obtained: Dragoon Level 5
 -- Recast Time: 5:00
 -- Duration: 03:00
+
+--According to BGWiki:
+--	When cast as Subjob DRG it gives 5% damage reduction/increase
+--	as Mainjog it gives 15%
+--	when wearing certain Brais it gets another +2% power
 -----------------------------------
 
 require("scripts/globals/settings");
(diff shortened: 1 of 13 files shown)


scripts/globals/abilities/ancient_circle.lua
```diff
@@ -22,5 +27,19 @@ end;
 -----------------------------------
 
 function onUseAbility(player,target,ability)
-    target:addStatusEffect(EFFECT_ANCIENT_CIRCLE,1,0,60);
+	local power = 5;			--when cast as Subjob
+	local duration = 180;
+	
+	if (player:getMainJob() == 14) then
```
takhlaq: use the job enum

takhlaq:
tabs > 4 spaces

ciradan:
No idea what you mean with enum?

I did use tabs.

takhlaq:
variable even
i mean replace tabs with 4 spaces (we dont use tab character here)

ciradan:
Like this?

takhlaq:
see how https://github.com/DarkstarProject/darkstar/blob/eae431fc033b2be0bd5a8d96d2bf17fbd4c2aca5/scripts/globals/mobskills/Grim_Halo.lua#L15 does it (i meant global lua table, not c++ enum, my bad)

teschnei:
I think the damage reduction should probably be a modifier instead of a status effect check

src/map/utils/battleutils.cpp
```diff
@@ -2653,26 +2689,36 @@ namespace battleutils
...
-        }
+		int16 KillerEffect = 0;
+		int16 MeritBonus = 0;
+
+		if (PDefender->objtype == TYPE_PC)
+		{
+			CCharEntity* PChar = (CCharEntity*)PDefender;
+
+			if (PDefender->GetMJob() == 9 && PDefender->GetMLevel() >= 75)	//Don't want to make a gorillion hasTrait checks
+				MeritBonus = PChar->PMeritPoints->GetMeritValue(MERIT_KILLER_EFFECTS, (CCharEntity*)PDefender);
```
teschnei: doesn't GetMeritValue already do this job/level check?

ciradan:
Something along this?

And I really hope I have this variable thing right now. ^^'

>doesn't GetMeritValue already do this job/level check?
True, it does. I saw in other examples that people still checked whether the trait was there or not, but on second thought that was probably done to check that it's a TYPE_PC and not a TYPE_MOB. 
I removed that job/level check now.

src/map/modifier.h
```diff
@@ -620,6 +620,9 @@ enum MODIFIER
     // For example, +10% damage to Chant du Cygne would be ID 570 + 225 (795)
     MOD_WEAPONSKILL_DAMAGE_BASE   = 570,
 
+	// Damage/reistance bonus from Circle Effects
+	MOD_ANCIENT_CICRLE_BONUS	  = 571, // Increases/decreases damage against/from dragons
```
teschnei: use the first mod_spare (64), and remove the commented out mod_spare - 571 is already in use for combo dmg%

teschnei: also, name the mod what it does, not what it's from (DRAGON_DAMAGE_BONUS or something like that)

teschnei:
my last 2 comments got marked as outdated after the typo fix, so make sure you look at those

ciradan:
I saw them and changed it. Hope the code itself is OK now, though I have no idea about the position. Right now this is handled as the very last thing before the HP gets subtracted, but I'm sure it needs to be handled waaaay sooner (before Phalanx, block, etc.) but I don't know where.

ciradan:
Alright.

I'll add the other Circles later then, after that it's simply a switching around of the code.

src/map/utils/battleutils.cpp
```diff
@@ -2696,8 +2684,7 @@ namespace battleutils
 		{
 			CCharEntity* PChar = (CCharEntity*)PDefender;
 
-			if (PDefender->GetMJob() == 9 && PDefender->GetMLevel() >= 75)	//Don't want to make a gorillion hasTrait checks
-				MeritBonus = PChar->PMeritPoints->GetMeritValue(MERIT_KILLER_EFFECTS, (CCharEntity*)PDefender);
+			MeritBonus = PChar->PMeritPoints->GetMeritValue(MERIT_KILLER_EFFECTS, (CCharEntity*)PDefender);
```
takhlaq: i realise you didnt write this but
```
auto PChar = dynamic_cast<CCharEntity*>(PDefender);

if (PChar)
{
    MeritBonus = PChar->PMeritPoints->GetMeritValue(MERIT_KILLER_EFFECTS, PChar);
}
instead of c style casting PDefender to PChar again anyways

File: src/map/utils/battleutils.cpp [hunk 5]
```diff
@@ -1905,7 +1905,35 @@ namespace battleutils
             damage = HandleStoneskin(PDefender, damage);
             HandleAfflatusMiseryDamage(PDefender, damage);
         }
-        damage = dsp_cap(damage, -99999, 99999);
+
+		if (PAttacker->objtype == TYPE_MOB && PDefender->objtype == TYPE_PC)	//Attacker is the Monster, damage gets decreased
```
teschnei: skip the TYPE_PC checks - there's no reason to restrict this bonus to players (for example, pets)

ciradan:
Tried to think up how to use Killer Instinct.

It also seems that the Beast Strength stuff isn't handled yet (outside of mob tp moves), so I thought it could be done via this MODs as well. Like, give all Vermins on initialization a addMod(MOD_PLANTOID_DAMAGE_BONUS, 5) and addMod(MOD_LIZARD_DAMAGE_BONUS, -5) if that's possible.

whasf:
Please clean and re-push

Issue #1618, opened 2015-06-22
run out of range of mob tp move = mob gets very glitched


TeoTwawki:
things I have seen happen after outrunning a mobs tp attack as it readies it, soem or all of these may or may not happen, seemingly randomly but only if I ran out of range of a tp attack it was preparing:

- mob ceases all action: no melee, no movement, no nothing

- mob loses ability to do anything but melee (if I am standing in range...it seems unable to follow me).

- mob becomes unable to despawn or die properly. I can empty its HP bar and obtain drops, but it stays standing there and never keels over.

- mob turns invisible/has no model, but can still be targeted. if the mob has not ceased action, it will continue to invisibly melee me.

Hozu:
A similar thing happens if a tp move gets stunned. It can also call various player WS animations, stretching out the mob model strangely. It's always the same animation for the same move, though, so I think it's just using the mob animation value but ends up calling a different animation type.

dazusu:
I had these issues when attempting to implement Dark Ixion.

helixhamin:
I have seen these issues when stunning them also. They become modern art, with some strange effects sometimes randomly going off.

TeoTwawki:
@Loki79 tells me despawning a mob while its busy/people fighting it can produce similar effects

nasomi:
despawning a mob under any circustamnces while engaged causes the monster to disappear.

OnMobFight() is executing every 1 second, not 3 as intended.

Loki79:
Ok to be a bit more clear on this despawning a mob during fight is causing the model to disappear, while  the name is still visible and the mob is target-able and continues to take dmg while not performing any actions towards the players.

TeoTwawki:
> OnMobFight() is executing every 1 second, not 3 as intended.

1 is actually intended, it just doesn't work very well compared to having it at 3 (remember the collibri animation bug? see its onMobFight where I worked around that).

bendangelo:
Just confirmed the mob skill interrupt packet is broken and always plays a random mob ability instead of clearing the charging up animation. I'll keep testing to reproduce the other issues.

bendangelo:
@TeoTwawki Can you look into this. Last time I checked the packet for interrupt mobskill was broken.

TeoTwawki:
If its really broken packet.. here's a cap https://yadi.sk/d/Hodv1fw0yse7R

I dunno wtf to do with it

PR #3303, opened 2016-08-02
New Pandemonium Warden Implementation

neosatus:
- Working phases
- Working Astral Flows
- Unskippable Adds
- Several new LuaBaseEntity functions

Notes:
- This is building on Era Devs Warp and Lee's old PW script where they modified the current DSP version to make PW unkillable with no healthbar until all the phases have been completed. This allows us to use realistic phase HP amounts since we can just reset his HP for each phase and trigger a change on HPP <= 1. The current version tries to squish all the phase HPs into PW's leaving him with none at the end when the fight should enter it's true 2nd phase.
- I reworked the phases from a long string of if statements to several arrays and a single logic check, as the previous note suggested. I tested and modified each TP move to make sure it worked and was retail accurate. I also split Chariot into 4 phases each doing one of the TP moves like in retail.
- I reworked the final phase Astral Flows. In retail, these were 8x simultaneous Astral Flow Blood Pacts every 25%. In practice that causes clients to crash so I changed it to 1x Astral Flow every 10% with accompanying pet respawn. It also uses the array/single logic check approach. The Avatars are correctly matched with their abilities. I reuse a spawning pet as an Avatar and just keep them around. Carbuncle is last, PW is weakened by then...
- I made the pets alternate between two sets of 8 pets so that players couldn't kill the adds directly before forcing a phase change to prevent them from respawning or otherwise glitch them. ~~I wrote a function to clone HP and status effects (presently just sleep, more can be added) that I use for these swaps.~~ This is an abundance of caution. If they get around this whatever.
- Further discussion about the pets, since nothing on DSP gains immunity to sleep overtime like on retail, the lamps are fully sleepable. I think this is okay because they have 9k HP each while the first few phases have 10k and then 15k. It still requires a coordinated group of BLMs who are multitasking sleeping and bursting. I think if they're made immune to sleep they should have much less HP so a coordinated group of BLMs could -aga them down. 
- ~~I tried to make PW sleep immune but my attempt doesn't work and I don't know what else to try.~~
- Brainjack charms now. I enjoyed it when one of my test characters got charmed and the other got one shot by Discoid.
- Searing Light mobskill wasn't a thing and is now and "Homing Missle" is now spelled correctly so it matches up with the DB.
- When he disengages he fully resets his phase variables, should he do that?

New LuaBaseEntity Functions:
- ~~despawn (calls PAI function)~~
- isSpawned (calls PAI function)
- isEngaged (calls PAI function)
- isAsleep (more effects? another condition?) (I just discovered this is a clone of a status.lua function)
- ~~clone (needed for PW)~~

This is still not retail accurate although a lot closer than the current version. It's tested, functional and challenging. I'm pushing it up here so Era can easily patch it in for some live testing.

EDIT: Updated


scripts/zones/Aydeewa_Subterrane/mobs/Pandemonium_Warden.lua
```diff
@@ -314,4 +191,29 @@ end;
...
+-----------------------------------
+
+function handlePet(mob, newPet, oldPet, target, modelId)
+    if oldPet:isSpawned() then
+        oldPet:despawn();
+    end
+    newPet:setModelId(modelId);
+    newPet:spawn();
+    newPet:updateEnmity(target);
+    newPet:clone(oldPet);
```
takhlaq: no need for this, since you're wanting to copy status effects from one entity to another, you can just grab the effect object `local effect = entity:getStatusEffect(EFFECT_ID)`, and `target:addStatusEffect(effect)` to the target entity

neosatus: Alright, I can just get rid of clone and bake it into handlePet then. I made it a method in case it was ever useful to anyone else but it probably wouldn't be.

neosatus:
While there are a few smaller questions I put forward in the above Notes section, the big ones are:

1. Is the way I'm handling the pets acceptable?
2. Is my change to the Astral Flow phase acceptable?

src/map/lua/lua_baseentity.cpp
```diff
@@ -10001,6 +10001,25 @@ inline int32 CLuaBaseEntity::spawn(lua_State* L)
...
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB);
+
+    CMobEntity* PMob = (CMobEntity*)m_PBaseEntity;
+
+    if (PMob->PAI->IsSpawned())
+    {
+        PMob->PAI->Despawn();
+        return 1;
```
teschnei: you returned 1 but didn't push anything on the stack, that will probably corrupt the stack

src/map/lua/lua_baseentity.cpp
```diff
@@ -10766,6 +10785,62 @@ int32 CLuaBaseEntity::getAutomatonHead(lua_State* L)
     return 1;
 }
 
+int32 CLuaBaseEntity::isSpawned(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    return ((CMobEntity*)m_PBaseEntity)->PAI->IsSpawned();
```
teschnei: same for all these

src/map/lua/lua_baseentity.cpp
```diff
@@ -10766,6 +10785,62 @@ int32 CLuaBaseEntity::getAutomatonHead(lua_State* L)
...
+    
+    return 0;
+}
+
+int32 CLuaBaseEntity::isAsleep(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+
+    CStatusEffectContainer* statuses = ((CBattleEntity*)m_PBaseEntity)->StatusEffectContainer;
+    EFFECT sleepEffects[3] = { EFFECT_SLEEP, EFFECT_SLEEP_II, EFFECT_LULLABY };
```
teschnei: isn't there already a function that checks sleep? in battleutils or statuseffectcontainer?

neosatus: I actually just found it last night, in fact. It's in status.lua hidden in the middle of a lot of enums. That one is hasSleepEffects(target) while this is target:isAsleep(). They're functionally the same though, would you like me to just use the old one?

src/map/lua/lua_baseentity.cpp
neosatus: Noted, I'll remedy this tonight.

src/map/lua/lua_baseentity.cpp
teschnei: i mean, there might be another one... i'll check when I get home since I don't have it in my IDE at work

neosatus: Since this is tangentially related, how do you go about making something sleep immune? I tried setting MOD_SLEEPRES to 100 but Elemental Seal still makes it land.

Hozu: Immunity field in the db, IIRC mob_pools. Flags are: https://github.com/DarkstarProject/darkstar/blob/a0396e2e4cbbe93bd4949c9277f4e4b847b919d1/src/map/entities/battleentity.h#L372

src/map/lua/lua_baseentity.cpp
takhlaq: `lua_pushTYPE(L, var);` where `TYPE` is `boolean`, `integer` etc (see how other functions return)
and then `return 1;`

src/map/lua/lua_baseentity.cpp
neosatus: Thanks!

scripts/zones/Aydeewa_Subterrane/mobs/Pandemonium_Warden.lua
```diff
@@ -43,268 +104,84 @@ end;
...
+    if (os.time(t) > depopTime) then
+        for i = 1, 2 do
+            for j = 1, 8 do
+                if pets[i][j]:isSpawned() then
+                    pets[i][j]:despawn();
+                end
+            end
         end
-        PWardenAstralFlows = PWardenAstralFlows + 1;
+        mob:despawn();
```
TeoTwawki: `DespawnMob(mob:getID)` woudl do teh excat same thing. Both of these would have the same flaw: mob randomly getting stuck as per [#1618](https://github.com/DarkstarProject/darkstar/issues/1618#issuecomment-115251959). It doesn't happen everytime but happens enough to be annoying (like the auto homepoint thing) The only way to avoid that I have found is to put the despawn command into a dummy mobskill.

src/map/lua/lua_baseentity.cpp
TeoTwawki: we already have functionality to do this exact thing. before you go adding new functions you should at least do a quick search of the trunk for the thing you are trying to do.

src/map/lua/lua_baseentity.cpp
teschnei: there is an IsAsleep in CStatusEffectContainer.. so just call that instead of checking the effects yourself

neosatus:
Thanks for taking the time to look over it all, I'm working on the changes.

As for those methods, I code and think in OO so I thought the object could use those methods but I can see the code duplication concerns so I'll call the existing functions instead.

neosatus:
I removed the OO version of those methods, fixed the return for the other functions, and ditched cloning entirely so now mobs respawn full health and wide awake every phase.

scripts/zones/Aydeewa_Subterrane/mobs/Pandemonium_Warden.lua
```diff
@@ -21,9 +44,41 @@ end;
...
+-- onMobDisengage Action
+-----------------------------------
+
+function onMobDisengage(mob)
+    -- Make sure model is reset back to start
+	mob:setModelId(1839);
+    
+    -- Prevent death and hide HP until final phase
+    mob:setUnkillable(true);
+	mob:hideHP(true);
```
takhlaq: tab to 4 spaces

neosatus: No idea how that happened, changed my settings a long time ago. Whatever, fixing.

scripts/zones/Aydeewa_Subterrane/mobs/Pandemonium_Warden.lua
```diff
@@ -32,9 +87,12 @@ end;
...
 function onMobEngaged(mob,target)
     -- pop pets
-    for i = 17056170, 17056177, 1 do
-        SpawnMob(i,180):updateEnmity(target);
-        GetMobByID(i):setModelId(1841);
+    for i = 1, 8 do
+        local pet = GetMobByID(petIDs[1][i]);
+        pet:setModelId(1841);
+        pet:spawn();
+        pet:updateClaim(target);
```
TeoTwawki: You are spawning them all pre-claimed. https://youtu.be/BPbsHhbi4LE?t=35s
Nawt retail bro.
You should just plain spawn them, then have them link - they will be in range if you set the position correctly. That'll also stop them from casting before they have faded into view.

neosatus: Hmm... so there's also that block of code that was left over from previous version where if it's not time for a phase change or an astral flow it just checks to make sure the pets are engaged and if not it teleports them to PW and engages them on the tank. This is to prevent people from dragging the adds off and leaving them somewhere. If I spawn them unclaimed, they'll get hit by that code and engaged anyways (I think, at work, can't test). Should I remove that block of code too? They spawn on top of PW each phase regardless of where they are so they'll be able to engage normally. If they get drug off it'll just be for one phase which isn't a big deal I guess. The longest phase is only 20k hp.

TeoTwawki: Are you talking about in handlePet() where I see the same auto claim? (axe that one too)
Pro-tip: you can get links directly to a line by clicking the line number on the left

I don't recall what all hate mechanics these guys use, can't look it up atm (its *possible* updateEmntity may need to stay or super-linking happens, I am not the guy to answer that though), It doesn't sound like there would be much risk of someone depopping or otherwise losing them all though.

I just wanted to point out you aren't supposed to have the adds auto claimed. Normal claiming is you only hold 1 claim from one player on one mob, so additional party/alliance members would need to act to claim other mobs. And only PW himself will spawn pre-claimed (to the person that popped).

neosatus: I was talking about this else block (kinda hard to see cause all the old code is mixed up in it): https://github.com/DarkstarProject/darkstar/pull/3303/files#diff-024cb4c707543574f17510da7bec95dcR155

I'll remove that claim stuff, was just trying to prevent any add shenanigans because the current version has buggy adds which caused a lot of drama on our server. I need to test it as a non GM too, I think GM status keeps them from aggroing.

TeoTwawki: yeah like 3, maybe 4 ppl, have tried to do this..I helped clean up a past version of it format wise (so it has me listed as a committer involved on the files) and at that point it kinda worked halfassedly, but then some stuff elsewhere changed and broke it all to hell again.

And I think you are right, that code looks redundant.

TeoTwawki: in case you are an email reader: I just edited my comment.

neosatus: I noticed he looked different in that video you posted. Maybe I'll play around with model IDs a bit.

No one has commented on the non-retail Astral Flow setup too, I guess it's acceptable given it crashes otherwise? We had a Kirin Fenrir AF/BP one shot half our raid last night so I think having to deal with that every 10% on top of pets repopping is a good compromise.

TeoTwawki: I think it is, been waiting to see what anyone else thought on the matter.

neosatus: Was just watching that video again and noticed him meleeing the adds, so I looked at how much hp they lost and how much damage he did. The one I saw was from 46% -> 31% with a hit of 522. So 15% of their health was 522. 522 * 100 / 15 = maxHP. maxHP = 3480 (probably 3500). They're currently set at 9k. Thoughts?

neosatus:
Removed the update Claim/Enmity functions except for the case of the AF mob during the final phase, it needs to have updateEnmity called in order for the ability to actually go off. I also changed the models so that now PW is the red model and the adds are the old one for the final phase. The red model isn't any bigger than the old model so it's a bit confusing, they all jumble together.

I ran through the full fight, felt good, aggro worked great without those calls. I'm ready to have my raid group test it. That final phase is gonna be hell.

Hozu:
Eh? Wouldn't you want updateEnmity for the pets? Make them spawn aggressive to whoever has hate on PW, but not claimed? Then again, there is a weird issue with that where it appears that any spawned add that uses that (or updateClaim) seems to copy the enmity list of the entity that called it, making it impossible for someone else to pull hate.

TeoTwawki:
@hozu he was having an issue where they'd cast before they were fully on screen, and since they will all link anyway I suggested removing the auto-agro.

neosatus:
Is there something else I need to change before this can be accepted?

teschnei:
I haven't really looked at it yet :)

LocoFFEra:
Any chance, this can be reviewed? Been waiting a few weeks now.

Thanks in advance!

Ninjistix:
what ever happened to this...

neosatus:
I stopped playing/developing on the server because the new PoE league came
out. Last I heard a dev on Era had implemented it and was going to do live
tests but I haven't heard since then.

nyczducky:
So I just tested this. Here are my findings so far:

The pets on spawn in all phases open with GA3's/AM/Tier 4 nukes its a bit insane lol. 

Lamps shouldn't be attacking as soon as PW spawns. PW should start with 2 TP moves and then phase change immediately into first chariot.

Phase changes work properly as do Astral Flows.

Mamool Ja TP animations are bugged out.

Client crashed a few times. Can't seem to determine what caused it.

Discoid shared damage works.

Hozu:
Any info on when the client crashes happened? They may be due to improper animations on mobskills.

nyczducky:
I'll have to test it again and limit the adds to 1 to prevent chat spam. I'll test more this week and see if I can narrow it down.

whasf:
Needs to be cleaned and re-pushed

whasf:
fix conflicts and repush please

Issue #3204, opened 2016-05-31
15 Character name makes DSP broken


R3P0FFXI:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**

30160427_0

**_Server Version_** (type `@revision` in game) **:**

9624926bac121c75464925b9c309ebae28f659d5

**_Source Branch_** (master/stable) **:**

Master and Stable

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

15 character name breaks DSP, client allows up to 15 characters DSP however does not. Cannot send tells to the character name was sucidaldeathtar. Everything started working fine no character recreation after renaming the character in the database to sucidaldeath

aizenmyou:
Just quickly looking through, it may have something to do with the data[5] value being passed in SmallPacket0x0B6. Everywhere else the charname is 15.

TeoTwawki:
I don't think data[5] has anything to do with it. I think this is just the [packet size](https://github.com/DarkstarProject/darkstar/blob/a0396e2e4cbbe93bd4949c9277f4e4b847b919d1/src/map/packets/basic.h#L62) and doesn't effect the packets contents.

PR #3306, opened 2016-08-04
Fixed 15-character name issue #3204

kjl2y:
The packet only contains 15 characters for the player's name.  Thus, the old data[5] does not contain a null terminator if the player's name is 15 characters.


TeoTwawki:
Did you confirm this was the issue before fixing it, or write code based off someones comment in the issue?

PR #3307, opened 2016-08-05
Batch text id update for August 2016 version update.

thedax:
Also bumped version.info.

*Note*: I don't have retail access, so I can't guarantee polutils did this all 100% right.


thedax:
Also note this is my first time doing this, and was told that you apparently have to do it in a specific order (i.e. by each version update)? Is this correct? I'll probably just close this is if it is, since i don't know how to downgrade the client properly (I just updated the game and ran the batch files "fresh" with no prior batch updates ran before).

starlightknight:
If you ran my batch from utils - no, you can skip and update and it will still work fine. Just make sure you've given it the latest POLUtils dump. 

Check the log at the end for fixme notes - if anyone added new ids with invalid comments, you may need to fix those manually (and fix the comments). If you can run the batch again without any changes, then the comments are all valid. Comments need to match what is in the polutils file - if there is garbage from choices or other special characters, you can use a fragment that doesn't have the special chars.

TeoTwawki:
@thedax just so you know and in case you haven't already found and fixed any probs: any spot where the IDs were not in order will have been borked and require manual fixing. Usually spot with multiple blank/cs npc in a row and multiple duplicate names in a row will have probs too. And starter city zones had been out of sync for awhile.

starlightknight:
That only applies to NPC IDs though, not text. For text ids it is fine to skip versions. 

For NPC, I would just leave alone any starter zone changes after the event npcs start. What I usually do is, run the batch, then run the scraper. Then look at a diff. 

For any line that's changed but Name + POLUtils name doesn't match, then it was borked by the ordering or multiple blank npc bugs and requires manual fixing. 

Everything that does match is good and can be submitted as is. Most of the time this is the majority of the changes, but with less npc shifts in the updates now, usually its the same messed up areas in the starter zone showing up again and few if any real changes.

TeoTwawki:
sorry I didn't see that it was a text id only pr, thought he did both "id"s. I fail at reading pr title.

thedax:
Yeah, this is only for text ids as you saw. Since I don't know how to downgrade properly (i.e. where playonline would see the version as the July update, 301606xx_xx (or whatever it was)), I probably won't be of much help updating the NPCs.

Anyway, I'm fixing up the "FIXME" entries that the batch file created (I had that set to false before, oops). Shouldn't be long.

scripts/zones/Mog_Garden/TextIDs.lua
```diff
@@ -2,16 +2,16 @@
 
 -- General Texts
 ITEM_CANNOT_BE_OBTAINED  = 6379; -- You cannot obtain the item <item> come back again after sorting your inventory
-ITEM_CANNOT_BE_OBTAINEDX = 6379;
+ITEM_CANNOT_BE_OBTAINEDX = 6378; -- (FIXME: Is this right??) You cannot obtain the <item>.
```
thedax: Any insight on what this should actually be is appreciated, I just had to take a guess. Even if Mog Garden's not implemented yet, it'd still be good to know exactly what text this is supposed to be.

scripts/zones/Mog_Garden/TextIDs.lua
```diff
@@ -2,16 +2,16 @@
...
           ITEM_OBTAINED  = 6384; -- Obtained: <item>
         ITEM_OBTAINEDX   = 6390; -- You obtain <number> <item>!
            GIL_OBTAINED  = 6385; -- Obtained <number> gil
        KEYITEM_OBTAINED  = 6387; -- Obtained key item: <keyitem>
     NOT_HAVE_ENOUGH_GIL  = 6389; -- You do not have enough gil
-MOGLOCKER_MESSAGE_OFFSET = 7458;
+MOGLOCKER_MESSAGE_OFFSET = 7508; -- (FIXME: Is this right??) Your particular paid period of Mog Locker patronage has been extended until the following time, kupo!
```
thedax: Had to take another guess here as well, since it had no comment or clues as to which mog locker line is required.

teschnei: look at the green thumb moogles script to understand better the text IDs

thedax: Thanks, turns out I was right with it. Will update the comment (i.e. remove the fixme).

takhlaq:
squash and i'll merge

thedax:
Squashed and rebased (not that it needed it, but I like to keep things up to date anyway).

PR #3314, opened 2016-08-10
Twinkle Tree (West Sarutabaruta): don't use hardcoded text ids.

thedax:
They're now correct, but this NPC is still not 100% retail quality (the event ID is still unknown, I'll try to figure it out in another pull).

Also fixed a couple more indentation problems.

### DIFF
--- a/scripts/zones/West_Sarutabaruta/npcs/Twinkle_Tree.lua
+++ b/scripts/zones/West_Sarutabaruta/npcs/Twinkle_Tree.lua
@@ -19,17 +19,17 @@ require("scripts/zones/West_Sarutabaruta/TextIDs");
 -----------------------------------
 
 function onTrade(player,npc,trade)
-starstatus = player:getQuestStatus(WINDURST,TO_CATCH_A_FALLIHG_STAR);
+    local starstatus = player:getQuestStatus(WINDURST,TO_CATCH_A_FALLIHG_STAR);
     if (starstatus == 1 and VanadielHour() <= 3) then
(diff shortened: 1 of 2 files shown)


scripts/zones/West_Sarutabaruta/npcs/Twinkle_Tree.lua
```diff
@@ -19,17 +19,17 @@ require("scripts/zones/West_Sarutabaruta/TextIDs");
 -----------------------------------
 
 function onTrade(player,npc,trade)
-starstatus = player:getQuestStatus(WINDURST,TO_CATCH_A_FALLIHG_STAR);
+    starstatus = player:getQuestStatus(WINDURST,TO_CATCH_A_FALLIHG_STAR);
```
takhlaq: local, we dont want this to be a global var

scripts/zones/West_Sarutabaruta/npcs/Twinkle_Tree.lua
```diff
@@ -44,10 +44,10 @@ end;
 
 function onTrigger(player,npc)
     if (VanadielHour() <= 3 and player:getVar("QuestCatchAFallingStar_prog") == 0) then
-        player:messageSpecial(7336);
-        player:messageSpecial(7338);
+        player:messageSpecial(FROST_DEPOSIT_TWINKLES);
+        player:messageSpecial(FROST_DEPOSIT_TWINKLES + 2); -- "It looks like it would melt if you touched it with your bare hands..."
```
takhlaq: should make this a var too in case of textid shifts

PR #3317, opened 2016-08-13
Adoulin: Second batch of Adoulin NPCs/Shops + npcUtil.tradeHas() fix

ibm2431:
Pretty self-explanatory. Adding more Western Adoulin NPCs, and all Western Adoulin shops. Remaining NPCs for Western Adoulin are now mostly "hard" ones like those relating to coalitions, colonization, augmenting, GEO/RUN, and Levil. I'm not sure if I'll tackle those next, start on zone CSes, or if I'll start working on Eastern Adoulin.

Also included is a simple fix for `npcUtil.tradeHas()`. It now works as it was intended to (judging by the comments), and has already been useful for me. No preexisting code will be affected by this, because it had yet to be used anywhere before I started utilizing it.

**Note:** This PR assumes ADOULIN now resolves properly for fame additions/checks due to merging #3292. Merging this without that will cause all fame checks to be adding to / comparing against fame for Abyssea - Misareaux.


scripts/zones/Western_Adoulin/npcs/Ansegusele.lua
```diff
@@ -0,0 +1,53 @@
...
+--  Type: Shop NPC
+--  @zone 256
+--  @pos 74 4 -126 256
+-----------------------------------
+package.loaded["scripts/zones/Western_Adoulin/TextIDs"] = nil;
+-----------------------------------
+require("scripts/globals/shop");
+require("scripts/zones/Western_Adoulin/TextIDs");
+
+local stock =
```
TeoTwawki: This should be inside the function its being used in, down between lines 37 and 38.

TeoTwawki:
@teschnei ***RE: tradeHas()**
Our entire trade system made of layers of turd, and this is a wrapper for 2 functions in said turd (checking count of an item and checking gil). We shouldn't even have to check for things that don't belong because instead of removing *whatever happens to be in the trade window* we should only be removing items that match the conditions we check for. In retail if you trade extra crap, the extra crap gets 100% ignored rather than failing the trade. If you trade things that meet 2 different sets of conditions **(example: crystal storing moogle, trade 2 types of crystal at once) one set will be ignored (same example: wind and earth crystal at same time, moogle will only take the wind ignoring earth). I'd like to see this eventually work the way retail does and tradeHas() drop into the abyss of crap we pretend never happened..

teschnei:
I made functions for that a long time ago, actually (something about
'confirm', if you grep that)

TeoTwawki:
@teschnei I never got them to behave like retail. I did tell you that in irc at the time. maybe I was doing it wrong - you could possibly convert some of the existing trash to it if so, and then we'd have better examples to work off...

TeoTwawki:
@teschnei meanwhile, the other crap is still being duplicated.

teschnei:
Sure, that sounds good to me once I get my server set up at my new place
sometime this weekend

scripts/zones/Western_Adoulin/npcs/Ansegusele.lua
ibm2431: Alright, moved all the NPC shop stocks into their onTrigger events~!

takhlaq:
resolve conflicts and i'll merge

ibm2431:
Merge conflicts resolved.

PR #3322, opened 2016-08-19
Separating conquest/nation variables from quest/mission variables

ibm2431:
As I mentioned in #3292, a lot of conquest and nation-related NPCs use integer constants to represent nations. Since these constants are (currently) simply `SANDORIA`, `BASTOK`, and `WINDURST`, this prevents those variables from being repurposed into tables for quests and missions (or anything else for that matter), because doing so would break aforementioned NPCs.

So as an initial step in the goal of table-lizing quest/mission variables, I've separated references to nations in the context of Conquest and checking the player's home nation (variables now given an enum-esque prefix of `NATION_`) from references to areas in the context of quests and missions (left alone). Now the variables for those two contexts can be modified later without influencing each other.

### DIFF
--- a/scripts/globals/conquest.lua
+++ b/scripts/globals/conquest.lua
@@ -11,11 +11,11 @@ require("scripts/globals/missions");
 -- convenience constants
 -----------------------------------
 
-SANDORIA = 0;
-BASTOK   = 1;
-WINDURST = 2;
-BEASTMEN = 3;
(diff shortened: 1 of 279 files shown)


takhlaq:
resolve conflicts and i'll merge

ibm2431:
Conflicts resolved.

PR #3325, opened 2016-08-22
Correct the maximum value allowed in the aggro column of mob pools..

TeoTwawki:
..It was defined as tinyint which stops at 255 for unsigned values, but the maximum flag value currently defined is 256, using all flags at once even higher..Thus we need smallint here.

### DIFF
--- a/sql/mob_pools.sql
+++ b/sql/mob_pools.sql
@@ -33,7 +33,7 @@ CREATE TABLE `mob_pools` (
   `cmbDelay` smallint(3) unsigned NOT NULL DEFAULT '240',
   `cmbDmgMult` smallint(4) unsigned NOT NULL DEFAULT '100',
   `behavior` smallint(5) unsigned NOT NULL DEFAULT '0',
-  `aggro` tinyint(1) unsigned NOT NULL DEFAULT '0',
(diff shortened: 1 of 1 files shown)


teschnei:
isn't aggro a boolean?

CMobEntity::m_Aggro

TeoTwawki:
isn't that column holding the detect flags? it looked like it was.

TeoTwawki:
I'm so fucking annoyed right now...Give me a few to look into this more

TeoTwawki:
Ok, so if I am seeing the chain of bullshit correct, the thing that used to be in the aggro column of pools is now in an entirely different table and the column in pools does something entirely different form those flags now. I'm trying to update old databases. Plural. IF I've got it wrong, poke me on IRC

teschnei:
Detects column is the detection type, truedetection column is a bool for
true sight/sound, aggro is if they aggro or not, I think. If that's what
you were thinking then yes

TeoTwawki:
Was thrown off by what is basically a new column occupying the same placement and name (since what it used to be, is now in the other table with a better name). Would have realized sooner had both bools been added to the tail end instead of middle.

PR #3327, opened 2016-08-23
NPC ID updates for client ver30160728_0

TeoTwawki:
All zones except starter cities. Someone else can get those they'll require manual intervention I am not up for.
```
Edit config.properties (windows users, it uses the opposite slash you'd expect in its conf paths)
C:\pathToPolutils\massextractor.exe pathToDumpFolder
C:\pathToDarkstarUtils\NpcIdUpdate.bat
C:\pathToDarkstarUtils\LuaScriptNpcIdUpdate.bat
Then review and fix. Thats all I did here.
```

### DIFF
--- a/scripts/zones/Arrapago_Remnants/IDs.lua
+++ b/scripts/zones/Arrapago_Remnants/IDs.lua
@@ -56,7 +56,7 @@ Arrapago = {
                 DOOR1 = 17080970,
             },
             [2] = {
-                ARMORY_CRATE = 17080598,
+                ARMORY_CRATE = 17080595,
             }
         }
     }


thedax:
Hopefully someone can capture the cities, then once it's all up to date, it should be a lot easier to keep it that way (I hope).

TeoTwawki:
The cities *can* be fixed by hand without capping (excepting event/seasonal NPCs, which will get removed/replaced) I'm just not dealing with it myself as it gives me a headache to look at.

starlightknight:
Events are rough because SE likes to swap their locations around all the time, even when the events are inactive. Think of the the old "find a ball under hidden under one of 3 cups", and then they're shuffled a billion times first. They do that all the time with events, and sometimes the events flat out disappear from the file for an update or two and then come back. 

Coupled with a lot of these NPCs having large sections of blanks that don't have POLUtils names to compare against, the job doesn't know what to do for these so they've fallen out of sync. Cities can also be rough if they linger in areas that have giant patches of no-name or csnpc NPCs. In this case, the batch job does the best it can on these by looking at how other npcs in the zone shifted. Usually if the zone shifted, once the start point of the shift is found, the rest of the zone shifts by the same value (i.e, they insert a single new npc in the middle of the zone. That npc takes the id that used to belong to another npc. From that point on, every subsequent npc in the zone shifts by 1). 
As long as the start of the shift is a named NPC in POLUtils, the job can fix it fine. But sometimes this isn't the case. If the shift starts on a "blank" npc - and the previous and next npcs are also "blank" - how do you tell where the shift actually started? They appear the same to the job so it can't keep that portion up to date and it may fall out of sync.

TeoTwawki:
@starlightknight I want to eventually come up with a faked NPC solution similar to how battlefield NPC helpers are fake pet objects, and then effectively blacklist known seasonal NPC names from the scrape after that works.

gedads:
" that don't have POLUtils names to compare against" also there are gaps in polutils ID, how is your script handling those? is it corecting next encountered npc with the following ID or filling that gap?
also can't push the PR, you removed hidden npcs allowing to go through balls of darkness in reisenjima (_830,831,832).
and in SOA there's some new npcs. 

for NPC, a lot of those need updates i doubt they're renamed by the script: 
INSERT INTO `npc_list` VALUES (17826159,'NPC[15f]','' in the updated file PR
is now 17826160,'NPC[170]  (17826159 is now fleuricette)

starlightknight:
@gedads depends what happened to the rest of the zone, as described above.

Also guys, don't forget the project is open source, you can view the source in the utils repo, and feel free to try to improve it further.

gedads:
adding placeholder would help your software to update?
(it's a bit troublesome to keep track of missing npcs but if it improves updating that might be more interesting)

TeoTwawki:
> INSERT INTO npc_list VALUES (17826159,'NPC[15f]','' in the updated file PR
is now 17826160,'NPC[170 (17826159 is now fleuricette)

Look one line below, that didn't become fleuricette, its directly beneath fleuricette..

As for Reisenjima I've restored the 3 hidden NPC that went missing, they got lost in some copy paste when I had to manually correct in that zone. Its handled now.

Also taking this opportunity to point out a few zones have some mistakes the preexist this pr, and I don't care because nothing is implemented there in dsp yet (and likely won't be anytime soon, I almost straight up reverted Reisenjima instead of manually editing it) so they haven't become any more wrong than they already were. If anyone would like to do a better job, have at it. Or we can go *another* 4 months being out of date everywhere else.

gedads:
" the existed prior to this pr, and I don't care because nothing is implemented there in dsp yet"
Here i don't share your point of view, so you can just let soa areas and rov new zones out of the update so i can add missing npcs from those because I 'm sure people outside of dsp are using those either for custom content or mission scripting or at least having them up to date and matching retail can push people to use them to script missions or whatever and eventually contribute.

ibm2431:
Regarding SOA NPCs: I've actually run into a couple invisible NPCs in Adoulin during my mission to script the whole town. So (naturally) I personally feel that they're relevant. Take that opinion for whatever you feel it's worth.

On capturing: I apologize for my ignorance on the subject, but what would need to be done to "recapture" NPCs? I have a retail subscription I specifically kept for the purpose of DSP development, so if it's the kind of thing you can simply point a somewhat competent person at (with a bit of direction), I wouldn't be adverse to doing so.

gedads:
Hi ibm2431, for capturing i use the old ashita2 and plogger method, there's another one i never used with windower probably with packets that appears a bit more clearly to read from what i saw but i'm used to the old plugin.
Just download ashitav2 and plogger (posted several times but if you don't find it just ask for it on irc i'll send it to you).
once ingame just type @addon load plogger and you'll see an incoming txt file from everything you'll encounter.

invisible npcs are probably there because of recent updates but it can also be an error in flags, could you tell me their names and when it occurs (in cs or when walking by) ?

ibm2431:
On Western Adoulin NPCs, I've been keeping a list of various NPC issues I intended to fix but have yet to get around to. Note: This was as of the previous NPC update, and last month's client version. I don't know if any of these would now be fixed by this PR, or if the NPC IDs the server yells about would have changed with August update.

- Missing/Invisible around town: Ornery Dhole, Gaddius, Louisareux, Ivoh Haalameb, Octavernost
- Positioning Issues: "Lalla-Dalla needs to be placed on top of his box", "Prah and Herchambaut are right on top of each other"
- Model Issues: Bilp's model is very screwy, wrong race, outfit, and distorts in cutscenes.
- A few NPCs have weapons at their feet: Quam Jitahr, Ishvad. Potentially others, I just now noticed this recently.
- Name Issues: Tuffle-Buffle's name keeps showing up in all Tarutaru Sauce quest CSes

Invisible NPCs in CSes:
- CS 121: Server claims need NPCs: 17826000, 17826001, 17826002.  Araustoix is probably one of them, Nashu another.
- CS 184: Araustoix and Nashu are invisible.
- CS 209: Server yells about needing NPCs 17826120 and 17826121, Zurko-Bazurko and Mikolulu have speaking lines but don't appear to be visible
- CS 5023: Fabioso and Celestin are invisible, server yells about needing NPC 17826021 and 17826022

gedads:
thanks , good infos, so far 3 errors in models one bilp isn't correct compared to others, and i made a mistake in the end of mid for npcs you indicates. pos of Lalla dalla is good, need subanim 1, for the rest it should be solved by update and addition of missing npcs. Tuffle buffle was missing namevis 120.

Glad to see someone interested in adoulin. If you see anything else don't hesitate to post it.

TeoTwawki:
> " the existed prior to this pr, and I don't care because nothing is implemented there in dsp yet"
Here i don't share your point of view, so you can just let soa areas and rov new zones out of the update so i can add missing npcs from those because I 'm sure people outside of dsp are using those either for custom content or mission scripting or at least having them up to date and matching retail can push people to use them to script missions or whatever and eventually contribute.

I'm saying I didn't change them...They ARE left out.

PR #3330, opened 2016-08-25
Update getMagicHitRate to be SJ aware

Kushdr:
Currently uses main level regardless of whether main has the skill or not. This change will take into account if it is from a sub and use the appropriate level. Ie 75SAM/37BLM should not have the same resist rate based on level differential as 75BLM/37RDM
Logic pulled from Hozu's commit for Spell Interrupt
Yes, I'm aware there's an odd space. i think the old one was a tab I forgot to fix. Sorry

### DIFF
--- a/scripts/globals/magic.lua
+++ b/scripts/globals/magic.lua
@@ -403,7 +403,10 @@ function getMagicHitRate(caster, target, skillType, element, percentBonus, bonus
 
     magicacc = magicacc + bonusAcc;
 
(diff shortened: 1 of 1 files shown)


Hozu:
Not my fix, I merely made the issue about it, heh.

TeoTwawki:
@LocoFFEra  @Kushdr  because it was also missing and "end" before the return.

TeoTwawki:
er, after.

TeoTwawki:
@LocoFFEra  @Kushdr  because it was also missing an "end" to close the if.

Edited to word better.

takhlaq:
make a new pull once you test and it works

PR #3333, opened 2016-08-28
GoV update for platform independence

abriasffxi:
WARNING:
Trying to set up a test server this weekend.  This code is untested - if someone could do it real quick I promise to do so in the future.

I just took the changes kj made to fields and applied them to GoV.  

I'm nervous about the repeat_regime options as I have no clue what they are used for.

### DIFF
--- a/scripts/globals/groundsofvalor.lua
+++ b/scripts/globals/groundsofvalor.lua
@@ -241,6 +241,14 @@ function finishGov(player,csid,option,r1,r2,r3,r4,r5,r6,r7,r8,r9,r10,msg_offset)
     local tabs = player:getCurrency("valor_point");
     local HAS_FOOD = player:hasStatusEffect(EFFECT_FOOD);
     local HAS_SUPPORT_FOOD = player:hasStatusEffect(EFFECT_FIELD_SUPPORT_FOOD);
+    local fov_repeat = bit.band(option, 0x80000000);
+    
+    if (fov_repeat ~= 0) then
+        fov_repeat = 1;
+    end
+    
+    option = bit.band(option, 0x7FFFFFFF);
+    
 -- ================= FIELD SUPPORT ============================================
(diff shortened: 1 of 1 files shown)


abriasffxi:
Alright- I had the opportunity to restart this morning and it looks like everything works right.

scripts/globals/groundsofvalor.lua
```diff
@@ -403,46 +411,57 @@ function finishGov(player,csid,option,r1,r2,r3,r4,r5,r6,r7,r8,r9,r10,msg_offset)
...
+        
     elseif (option == GOV_MENU_PAGE_9) then -- Page 9
-        writeRegime(player,r9,msg_accept,msg_jobs,0);
+        writeRegime(player,r9,msg_accept,msg_jobs, fov_repeat);
+        
     elseif (option == GOV_MENU_PAGE_10) then -- Page 10
-        writeRegime(player,r10,msg_accept,msg_jobs,1);
+        writeRegime(player,r10,msg_accept,msg_jobs, fov_repeat);
+        
     elseif (option == GOV_MENU_REPEAT_REGIME1) then -- Page 1 Repeat
```
teschnei: you can get rid of all of the REPEAT choices, since they are all no longer possible (and you already put it in fov_repeat)

TeoTwawki:
In the future make sure your commit messages describe the changes rather than just say "updated <filename>". If you used either the website interface or the github app to do the work, I recommend using a real git client instead next time (tortoisegit or command line git or both good choices).

teschnei:
actually, I will get you to squash those commits and give it a real commit message before I merge

abriasffxi:
Yeah, using the command line just didn't realize the message would be so prominent when it asked for it.  Not a problem.  Trial by fire!

teschnei:
still need to squash them together

abriasffxi:
done

PR #3334, opened 2016-08-28
Update Condition for New Players to flag ZM1

nyczducky:
Previously this caused ZM1 not to be flagged. Changed the condition to add ZM1 to a new player. Having no mission has a value of 255.


ibm2431:
Is it certain that the current mission for an [untouched Ziliart log isn't 0xFFFF](https://github.com/DarkstarProject/darkstar/blob/master/src/map/entities/charentity.cpp#L138)? Either way, static values in scripts should be avoided. Something that'd probably work instead:
`if ((not player:hasCompletedMission(ZILART, THE_NEW_FRONTIER)) and (player:getCurrentMission(ZILART) ~= THE_NEW_FRONTIER)) then`

PR #3337, opened 2016-08-29
Change condition to add ZM1

nyczducky:
Previously the condition checked if current Zilart mission was less than ZM1. Player having no mission had a value of 255 and failed this condition and ZM1 wasn't added.

The above condition will check if the player has not completed the mission and the current Zilart mission isn't THE_NEW_FRONTIER.

Thanks to ibm2431 for the suggestion.

### DIFF
--- a/scripts/zones/Throne_Room/bcnms/shadow_lord_battle.lua
+++ b/scripts/zones/Throne_Room/bcnms/shadow_lord_battle.lua
@@ -53,7 +53,7 @@ function onEventFinish(player,csid,option)
     
     if (csid == 0x7d01) then
         if (player:getCurrentMission(player:getNation()) == 15 and player:getVar("MissionStatus") == 3) then
-            if (player:getCurrentMission(ZILART) < THE_NEW_FRONTIER) then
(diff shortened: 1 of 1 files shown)


TeoTwawki:
Them dang missions and their client side indices that start with zero for the first mission..
(This was my derp when I fixed this fight constantly erases everyone zilart progress - fix a bug, make a bug!)

scripts/zones/Throne_Room/bcnms/shadow_lord_battle.lua
```diff
@@ -53,7 +53,7 @@ function onEventFinish(player,csid,option)
     
     if (csid == 0x7d01) then
         if (player:getCurrentMission(player:getNation()) == 15 and player:getVar("MissionStatus") == 3) then
-            if (player:getCurrentMission(ZILART) < THE_NEW_FRONTIER) then
+            if ((not player:hasCompletedMission(ZILART, THE_NEW_FRONTIER) and (player:getCurrentMission(ZILART) ~= THE_NEW_FRONTIER) then
```
ibm2431: It would appear that some parenthesis were lost.

TeoTwawki: @nyczducky ruh oh!
@takhlaq code review fail +1 :smirk:

PR #3342, opened 2016-08-29
Fix Nightfall's Regain

Fizztastic:
The Regain effect on Nightfall should be 10TP/tick instead of the current 100TP/tick as per http://ffxiclopedia.wikia.com/wiki/Nightfall.


takhlaq:
is that before or after max tp changed from 300 to 3000?

Fizztastic:
The current TP Regain should be 10/tick (in 3000 scale). SE clarified the text on the sword by adding "Regain +10" instead of just saying "Regain" like it used to.
img_20160829_210238

PR #3352, opened 2016-09-05
Bunch of small fixes

ymkc:
- Peryton's are not NM's, just regular mobs (NM version not accessable atm)
- Remove some duplicated lines from Aern MNK & DRG (same code in onMobDespawn and onMobDeath)
- Remove unneccessary NPC moves from Rolanberry Zone.lua (not sure why it's there, as the pos' seem to be wrong and it could be in the db)
- Move SetDropRate to onMobSpawn


scripts/zones/Grand_Palace_of_HuXzoi/mobs/Ix_aern_mnk.lua
```diff
@@ -79,9 +79,6 @@ end;
 -----------------------------------
 
 function onMobDespawn(mob)
-    -- Despawn his minions if they are alive (Qn'aern)
-    DespawnMob(mob:getID()+1);
```
TeoTwawki: If he depops rather than is killed his helpers still need removed.

scripts/zones/The_Garden_of_RuHmet/mobs/Ix_aern_drg.lua
```diff
@@ -56,20 +56,12 @@ function onMobDeath(mob, player, isKiller)
...
-    -- Pick a new for PH Ix'Aern (DRG)
-    SetServerVariable("[SEA]IxAernDRG_PH", AwAernDRGGroups[math.random(1, #AwAernDRGGroups)] + math.random(0, 2));
 end;
 
 -----------------------------------
 -- OnMobDespawn
 -----------------------------------
 function onMobDespawn( mob )
-    -- Despawn pets.
-    DespawnMob(wynavA);
```
TeoTwawki: same here

TeoTwawki: The onMobDespawn willl always run after death, so that one could be left and the duplicate code in oMobDeath taken out but then there's a delay before the helpers will begin their depop (though that might be retail, I dunno).

ymkc: @TeoTwawki Thanks, I see what you mean

PR #3355, opened 2016-09-08
zlib: Fix arm and big-endian archs

Cloudef:
Make "zlib" code more portable, by not violating strict-aliasing,
using the correct types when reading the data, and handling endianess
conversions for the 1% platforms out there.

Standalone C99 version can be found here:
https://gist.github.com/Cloudef/0d22235b87b8f53d65f8828cb7b66db7

NOTE: cbasetypes.h is retarded and very wrong, we should remove it
and just use cstdint instead. Forcing -fsigned-char for unix now as
the codebase relies on chars (and int8) being signed.

### DIFF
--- a/src/common/zlib.cpp
+++ b/src/common/zlib.cpp
@@ -1,118 +1,234 @@
 #include "../common/zlib.h"
 #include "../common/showmsg.h"
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
+#include <vector>
+#include <string>
+#include <cstring>
+#include <cassert>
 
-uint32 zlib_compress_table[512];
-uintptr zlib_decompress_table[2556];
+#if (defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__) || \
+    (defined(__BYTE_ORDER) && __BYTE_ORDER == __BIG_ENDIAN) || \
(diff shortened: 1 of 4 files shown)


Cloudef:
It fails on the non-standard static_assert, will fix that when home.

src/common/zlib.cpp
```diff
@@ -1,118 +1,215 @@
...
+    // Base address of dec table, if we substract pointer in dec table, we should should be
+    // able to normalize them to offsets starting from 0.
+    const uint32 base = dec[0] - sizeof(uint32);
+
+    for (size_t i = 0; i < dec.size(); ++i) {
+        if (dec[i] > 0xff) {
+            // Everything over 0xff are pointers.
+            // These pointers will be traversed until we hit data.
+            jump[i].ptr = jump.data() + (dec[i] - base) / sizeof(base);
+        } else {
```
TeoTwawki: *pssst*, @Cloudef, proj is run by brace-on-newline-people. :wink:

TeoTwawki: just letting you know

Cloudef: Thanks, will fix tomorrow. I recall there being style guide now too, I'll go read that _again_ :)

Cloudef:
Modified the code style to match current dsp codebase more.

Issue #3002, opened 2016-04-06
No EXP gain when high level player outside party assists with healing.


dazusu:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160203_0

**_Server Version_** (type `@revision` in game) **:**
5ca0bfd

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
If a player is fighting a monster, then a player outside the PT (who is higher level) heals that player, that player will **not** receive EXP from the monster, even though the outside player did not attack the monster.

Hozu:
I've only noticed it happening if the mob was to take a swing at the high level player (or at least the high level player reaches the top of the hate list). You sure it happens regardless?

dazusu:
Even if the mob takes a swing at the high level player (or the player reaches the top of the hate list), it still should **not** affect the EXP of the player offensively attacking the mob.

In my testing, it's happening regardless, though :+1:

Hozu:
Well it's not supposed to, but knowing how it's being applied helps pinpoint the issue.

Though I do have to say I like it this way.

dazusu:
Open to abuse; that's probably why it's not that way in retail.

It seems the issue is in the DistributeExperience method.

helixhamin:
Correct me if I am wrong, but didn't SE implement something similar? I remember everyone used to let the high level mage take hate, then the dunes parties could slowly kill the mob with little worries. But, I also remember something about the mage using spikes, so was that the only way the exp was killed in retail? Sorry, it has been a while.

dazusu:
Like we have discussed here, if the high level player acts offensively, they will then be registered as the highest level player to attack the monster; and EXP will be based on that players level.

So yes, if a Mage is using a 'spikes' spell, that counts as acting offensively on the monster; so EXP would be ruined.

However if the Mage "tanks" the monster, but doesn't deal damage - then EXP is still awarded.

maxtherabbit:
is this still a problem?

Hozu:
There wasn't a commit specifically to fix this, so I'd say most likely.

Lagosaurus:
This is still an issue.

eraffxi:
[...]

Issue #2698, opened 2016-01-30
FoV/GoV: Attacking mob with BLM nukes only gives no credit


xipies:
Used single target nukes (no -aga spells) and got no credit. Not a one-hit K.O. Tested on Decent Challenge mob.

Revision: d06c77e
Server: Classic

PR #3356, opened 2016-09-08
Party exp outside heals

Manticmage:
This resolves an issue where healing magic from outside the party, alliance, or solo player will result in no experience points or treasure being awarded.
It does so by checking if a PC with enmity is a party member before assigning that player's level to the highest level attacker of the mob.

Resolves issue #3002 
This issue was introduced with commit d3fee59d0d47d57760db06b9d277d50fb7d65ee6 which resolved issue #2698 wherein attacking a mob with only magic gives no FoV credit.

I have done extensive testing and all have passed: The correct EXP / FoV credit is awarded in all of the following scenarios:
solo
solo FoV
solo with PL
solo FoV with PL
solo magic only kill
solo FoV magic only kill
solo magic only kill with PL
solo FoV magic only kill with PL
party
party FoV
party with PL
party FoV with PL
party magic only kill
party FoV magic only kill
party magic only kill with PL
party FoV magic only kill with PL
alliance
alliance FoV
alliance with PL
alliance FoV with PL
alliance magic only kill
alliance FoV magic only kill
alliance magic only kill with PL
alliance FoV magic only kill with PL

### DIFF
--- a/src/map/enmity_container.cpp
+++ b/src/map/enmity_container.cpp
@@ -123,10 +123,21 @@ void CEnmityContainer::UpdateEnmity(CBattleEntity* PEntity, int16 CE, int16 VE, 
         CE = 0;
         VE = 0;
     }
-
-    auto PMob = dynamic_cast<CMobEntity*>(m_EnmityHolder);
-    if (PMob && PMob->m_HiPCLvl < PEntity->GetMLevel())
-        PMob->m_HiPCLvl = PEntity->GetMLevel();
+	
+	// Set mob highest level PC attacker to that of the person with enmity if they are solo, a party member, or alliance member (Fixes spell only kills not getting FoV/GoV credit)
+	auto PChar = static_cast<CCharEntity*>(PEntity);
+	auto PMob = dynamic_cast<CMobEntity*>(m_EnmityHolder);
+	std::vector<CCharEntity*> members;
+	// Get a list of all valid party members
+	PChar->ForAlliance([PMob, &members](CBattleEntity* PPartyMember)
+	{
+		// Check to see if they are in the same zone as the mob and within 100 yalms
+		if (PPartyMember->getZone() == PMob->getZone() && distance(PPartyMember->loc.p, PMob->loc.p) < 100)
+		{
+			// Set the mobs highest level PC attacker to this PC
+			PMob->m_HiPCLvl = PPartyMember->GetMLevel();
+		}
+	});
 
     auto enmity_obj = m_EnmityList.find(PEntity->id);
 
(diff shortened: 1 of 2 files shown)


src/map/enmity_container.cpp
```diff
@@ -123,10 +123,21 @@ void CEnmityContainer::UpdateEnmity(CBattleEntity* PEntity, int16 CE, int16 VE,
...
     }
-
-    auto PMob = dynamic_cast<CMobEntity*>(m_EnmityHolder);
-    if (PMob && PMob->m_HiPCLvl < PEntity->GetMLevel())
-        PMob->m_HiPCLvl = PEntity->GetMLevel();
+	
+	// Set mob highest level PC attacker to that of the person with enmity if they are solo, a party member, or alliance member (Fixes spell only kills not getting FoV/GoV credit)
+	auto PChar = static_cast<CCharEntity*>(PEntity);
+	auto PMob = dynamic_cast<CMobEntity*>(m_EnmityHolder);
+	std::vector<CCharEntity*> members;
```
takhlaq: this doesnt do anything

src/map/enmity_container.cpp
```diff
@@ -123,10 +123,21 @@ void CEnmityContainer::UpdateEnmity(CBattleEntity* PEntity, int16 CE, int16 VE,
...
+	// Get a list of all valid party members
+	PChar->ForAlliance([PMob, &members](CBattleEntity* PPartyMember)
```
takhlaq: members vector isnt needed here either

version.info
```diff
@@ -1,7 +1,7 @@
 #DarkStar Version Info
 
 #Expected Client version (wrong version cannot log in)
-CLIENT_VER: 30160728_0
+CLIENT_VER: 30160831_0
```
teschnei: no, we do not update this unless the npcs/textids have been updated

src/map/enmity_container.cpp
```diff
@@ -123,10 +123,21 @@ void CEnmityContainer::UpdateEnmity(CBattleEntity* PEntity, int16 CE, int16 VE,
...
+	{
+		// Check to see if they are in the same zone as the mob and within 100 yalms
+		if (PPartyMember->getZone() == PMob->getZone() && distance(PPartyMember->loc.p, PMob->loc.p) < 100)
+		{
+			// Set the mobs highest level PC attacker to this PC
+			PMob->m_HiPCLvl = PPartyMember->GetMLevel();
```
teschnei: this is a nullptr crash if PMob's dynamic_cast failed

teschnei:
I don't think very highly of iterating over the entire alliance every time UpdateEnmity is called.  The issue is that HiPCLvl is being updated for things like cures when it should just be when offensive actions are taken - so we should be fixing it to only update HiPCLvl when an offensive action is taken (I am pretty sure that cures go through this code path)

Manticmage:
Thanks for the feedback! This was an issue that has been affecting me and it hadn't seen any movement so I thought I'd give it a shot.

I removed the unused CCharEntity members variable as was mentioned.

As to the comment @teschnei made. I definitely see what you mean. Setting the m_HiPCLvl inside the UpdateEnmity method seems weird anyway. Im sure there is a better solution, perhaps as @takhlaq  mentioned seperating offensive magic / regular magic enmity for m_HiPCLvl calculation.

Thank you for your time looking at my Pull Request, we can put this PR on hold until until a better solution is found.

Manticmage:
This has been since resolved in a different way by @takhlaq in  commit 9dfcfe81110648161481af1d29ded3d736c77301

Closing Pull Request!

PR #3357, opened 2016-09-08
Fix Rolanberry Zone.lua for non-abyssea users

abriasffxi:
Adds an if statement for setting the rift position.  If required expansion does not include abyssea, rift was returning nil and crashing the script, rendering much of the content in the zone useless.


TeoTwawki:
A fix for this was already included in #3352 (there shouldn't be any setPos happening at all, the pos for that NPC should be being set in sql)

abriasffxi:
Even better.  Close that sheet!

TeoTwawki:
I don't have close permissions, users can close their own though.

takhlaq:
fixed in #3352

abriasffxi:
My fault, I'll get it next time.

PR #3359, opened 2016-09-10
Fixed useMobAbility with self-targeting abilities

ymkc:
Not sure it's in the best place, but it works well and fixes scripted 2hour's and other self-targeting abilities called via useMobAbilitiy


teschnei:
that changes the functionality of the call (previously: passing no target would let the mob use it on its current target, now it will try to use it on itself).  You want to check the target type of the mobskill inside the action that gets queued and decide there whether to use GetBattleTargetID or itself

ymkc:
moved it to the queueAction call

ymkc:
@teschnei what do you think about having a distance check in there aswell?

teschnei:
that's not what it's for

teschnei:
ed2e95e10bf5cab63b8be0e522f652254651802e

PR #3360, opened 2016-09-10
NPC recapture for starting cities and Rabao (2016/08)

ibm2431:
This is a full NPC recapture from retail of (only) the starting cities and Rabao. I didn't touch any other zones.

It's based off of two captures I made, one during the Sunbreeze Festival, and one after SE shut it off, but before the next version update. This let me definitively mark what's Sunbreeze and what isn't.

**Summary:**

- **The NPCs are now properly sorted by their ID!**
- Many missing NPC IDs were added.
- Quotes around integers and floats have been removed.
- Floats for NPCs that I added or whose position I modified are now to three decimal places.
- Home Point 1s in Port San d'Oria and Port Windurst were moved. I've already adjusted them in homepoint.lua
- I strangely did not get ROV NPCs on either capture for Bastok Markets, despite getting them for every other zone. I intend to do yet another recapture after updating client, and will see then.
- A couple Door:House's in Bastok Mines swapped IDs. Don't ask me why; but now they're sequential with their ID vs raw name.
- Yes, the AH counters in Port San d'Oria swapped IDs with each other. Another mystery.
- The AH counters in Windurst Woods also shuffled around.
- Nanaa_Mihgo in Windurst Woods switched two of her NPC IDs, and dropped the third.

I was tempted to remove quotes around ints/floats, and do three-decimal floats for the entire file, but those reg-exes would have made the actual NPC changes more difficult to review. So maybe in a later PR.

**To readers from the future who found this PR looking for information on implementing Sunbreeze:** Here's everything I got:
[NPC Recapture Lua Tables + Raw Logs](https://drive.google.com/file/d/0B6J1HCKIjYoTUDdKYzNVNkVyQ2M/view?usp=sharing)
_Note: The values for 'flag' and 'flags' in the Lua table are incorrect, as at the time they were written my addon wasn't taking endianness into account. The values in the raw packets are correct - you just need to pull them yourself._

One thing I'll note about Mumor and Uka: They have two IDs, one for when they're standing there slinging fireworks, another for when they're doing the stage show. There was some weirdness going on with their models (set to almost empty) when that ID wasn't currently in-use (the same goes for the cheerleaders). For the complete "during Sunbreeze" SQL list, I manually tweaked it so all related models appear in Bastok Markets save for the battle Mumor and Uka, which appear in Northern San d'Oria.


starlightknight:
the hero we all needed arrives..... :-)

PR #3363, opened 2016-09-12
Cosmetic fixes to Land King pops

ymkc:
- Do not spawn HNM's unconditionaly (honor spawn condition in settings.lua)
- do not overlap HNM and QM (hide QM on trade and on HNM spawn [in case only NQ or HQ is force popped])

### DIFF
--- a/scripts/zones/Behemoths_Dominion/mobs/Behemoth.lua
+++ b/scripts/zones/Behemoths_Dominion/mobs/Behemoth.lua
@@ -15,6 +15,16 @@ function onMobInitialize(mob)
 end;
 
 -----------------------------------
+-- onMobSpawn
+-----------------------------------
+
+function onMobSpawn(mob)
+    if (LandKingSystem_NQ > 0 or LandKingSystem_HQ > 0) then
+        GetNPCByID(17297459):setStatus(STATUS_DISAPPEAR);
+    end
+end;
+
+-----------------------------------
 -- onMobDeath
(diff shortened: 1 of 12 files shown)


scripts/zones/Behemoths_Dominion/mobs/Behemoth.lua
```diff
@@ -15,6 +15,16 @@ function onMobInitialize(mob)
 end;
 
 -----------------------------------
+-- onMobSpawn
+-----------------------------------
+
+function onMobSpawn(mob)
+    if (LandKingSystem_NQ > 0 or LandKingSystem_HQ > 0) then
+        GetNPCByID(17297459):setStatus(STATUS_DISAPPEAR);
```
TeoTwawki: Would this be better as an "and" because of the oddball scenario of one being timed but the other being set force pop? Then again maybe requiring ppl to take down the opposite one before they force-spawn the other is better (dragons aery in particular gets kinda crowded with 2 dragons :wink: ) I'm not complaining either way, just a thought because back when I ran a server I did that and I'm not sure which people would want.

ymkc: I wasn't sure either, but thought since there are 2 different options in the settings.lua I would just separate them. But can change it anyway. The only "important" part is not to spawn the mobs when the settings.lua says they should be force popped

PR #3370, opened 2016-09-15
Same return I just fixed, in the other Seiryu..

TeoTwawki:
didn't notice it till after the previous one was merged


TeoTwawki:
Actually...does the version assisting Kirin even *have* a 2hr??

PR #3373, opened 2016-09-18
Changed missing script message type to warning so it can be filtered without losing more serious error messages. 

TeoTwawki:
Some entities never need a script (primarily doors in towns) and many others are unimplemented, rather than broken/lost/whatever. I figure that's good enough for unimplemented packets, so its good enough for unimplemented scripts too.


TeoTwawki:
I figure that's good enough for unimplemented packets, so its good enough for unimplemented scripts too.

Issue #2731, opened 2016-02-07
Shield block should not be possible under Terror status effect


Deadwing888:
It currently is. Most easily recreated at Fafnir or Nidhogg.

Deadwing888:
Furthermore I noticed I got a parry check while asleep a moment ago. As I recall this is also non retail? But I'm not 100% sure, could anyone weigh in?

Topace1:
I think parrying can be done while under the effects of sleep. All you need is to be facing the mob,

aizenmyou:
Parry definitely is available in retail while asleep. We used to make jokes about it years ago. (That and how characters sleep standing up)

Scavenge:
you also need to have your weapon drawn in order to parry, not just be facing the mob.

Deadwing888:
Yeah, that's coded correctly scavenge

TeoTwawki:
Was the damage reduced, shield skillup obtained, or both?

Deadwing888:
Damage reduced. Unknown if skillup was obtained.

TeoTwawki:
186cb73ace59a58d7269426e3b43f1523b1142e8

PR #3376, opened 2016-09-19
Do not allow shield blocking while terrorized

ymkc:
Stumbled across this and was marked as easy fix :)

and btw what about parrying while being terrorized, is it possible on retail?

#2731

### DIFF
--- a/src/map/utils/attackutils.cpp
+++ b/src/map/utils/attackutils.cpp
@@ -26,6 +26,7 @@
 #include "battleutils.h"
 #include "../attack.h"
 #include "../items/item_weapon.h"
+#include "../status_effect_container.h"
 
 namespace attackutils
 {
@@ -132,7 +133,7 @@ namespace attackutils
     ************************************************************************/
     bool IsBlocked(CBattleEntity* PAttacker, CBattleEntity* PDefender)
     {
-        if (isFaceing(PDefender->loc.p, PAttacker->loc.p, 40))
+        if (isFaceing(PDefender->loc.p, PAttacker->loc.p, 40) && !PDefender->StatusEffectContainer->HasPreventActionEffect())
         {
             return(dsprand::GetRandomNumber(100) < battleutils::GetBlockRate(PAttacker, PDefender));
         }


Deadwing888:
I think yes parry while asleep, no while terrorized.

Not 100% sure.

teschnei:
what about sleep? stun?

Deadwing888:
I believe shield block cannot proc during petrify stun or sleep

takhlaq:
hmm is it only terror or other effects such as stun?
if other effects prevent then https://github.com/DarkstarProject/darkstar/blob/master/src/map/status_effect_container.cpp#L1444

bendangelo:
If there are more status effects, might as well add a flag `no_shield_proc` and add it to those status effects. Then in that condition you can just check if any status effects have the flag.

teschnei:
an entire flag for one spot...?

Deadwing888:
Is it still the case that shield block does not work VS mob's physical ws on dsp?
I have not tested recently, but this was the case for quite some time

TeoTwawki:
Lets start by verifying what retail does for each situation *exactly* since I'm not the only one who's commenting with uncertainty here.

I'm fairly sure I've gotten retail shield procs while slept, but not only was that a long time ago and my memory hazy, but entirely possible things could have changed (though I see nothing about any changes on either wiki, nor any mention of being able to proc or not).

Lets just make sure before we fix more of it, is all mean to say.

teschnei:
yes, Deadwing

Deadwing888:
30/100 shield blocks while awake
0/100 shield blocks while asleep

I wasn't 100% sure myself, but now we know. I am much more confident about petrify and stun, but am still happy to check if there are doubts.

TeoTwawki:
Thank you Deadwing.

Deadwing888:
In regards to the mob's physical weaponskills, shield block on retail has a chance to proc vs those. Would someone here like to look at it or should I write it up as a separate issue?

teschnei:
Separate issue if there isn't one already (and please include that additional effects can't proc on blocked mob weaponskills, in case the implementer does not know about that)

teschnei:
As for this PR, that should be good enough reason to use the function @takhlaq linked instead of a specific effect check

ymkc:
Updated PR

Is it safe to assume that neihter guarding nor parrying is possible while being terrorized?
Open a seperate PR for those two and see if some can verify?

Issue #2492, opened 2015-12-15
Changeing jobs in Al Zahbi MH (From DRG)


Leelink1:
If you enter Al Zahbi mog house with your Wyvern out as a DRG he appears inside your Mog House.

If you change job and exit the mog house the server crashes. When you log back in it continues to crash.

Hozu:
On retail, when you change main or sub job, does it always despawn your pet? I think that would be the best way to go about fixing this.

Leelink1:
It does for every other zone. Normally you don't even see your Pet in the moghouse. I guess since Al Zahbi is a Battle zone (due to Besieged) it allows them to appear.

teschnei:
I don't think pets spawn in your mog house either way

teschnei:
at any rate, I think it should be possible to extend the "is pet allowed in this zone" check to check the zoneflag and the moghouse instead of just the zoneflag

TeoTwawki:
Fixed by https://github.com/DarkstarProject/darkstar/commit/55fab7ae1a23207623c07b1a0bfb45130c522197 and we forgot to close

PR #3381, opened 2016-09-21
Adjusted zone check for pets

ymkc:
fix for #2492

tried to follow @teschnei suggestion and added it to the lua pet call check aswell

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -4401,8 +4401,15 @@ inline int32 CLuaBaseEntity::canUseChocobo(lua_State *L)
 inline int32 CLuaBaseEntity::canUsePet(lua_State *L)
 {
     DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
 
-    lua_pushboolean(L, m_PBaseEntity->loc.zone->CanUseMisc(MISC_PET));
+    auto PChar = static_cast<CCharEntity*>(m_PBaseEntity);
+
+    if (PChar->loc.zone->CanUseMisc(MISC_PET) && !PChar->m_moghouseID)
+        lua_pushboolean(L, true);
+    else
+        lua_pushboolean(L, false);
+
     return 1;
 }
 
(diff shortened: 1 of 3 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -4402,7 +4402,11 @@ inline int32 CLuaBaseEntity::canUsePet(lua_State *L)
 {
     DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
 
-    lua_pushboolean(L, m_PBaseEntity->loc.zone->CanUseMisc(MISC_PET));
+    if (m_PBaseEntity->loc.zone->CanUseMisc(MISC_PET) && ((CCharEntity*)m_PBaseEntity)->m_moghouseID == 0)
+        lua_pushboolean(L, true);
+    else
+        lua_pushboolean(L, false);
+
```
takhlaq: you'll want to check if the entity is actually a player (in case mobs can call this too) so `dynamic_cast<CCharEntity*>(m_PBaseEntity)`
if the cast doesnt return nullptr you know its a player, so can check `PChar->loc.zone->CanUseMisc(MISC_PET) && PChar->m_moghouseID == 0` otherwise only need to check the zone's misc pet flag

teschnei:
Also, a mob can never use pets like players. So return false if it's not a
player

ymkc:
ok I hope I got this right

src/map/lua/lua_baseentity.cpp
```diff
@@ -4402,7 +4402,17 @@ inline int32 CLuaBaseEntity::canUsePet(lua_State *L)
 {
     DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
 
-    lua_pushboolean(L, m_PBaseEntity->loc.zone->CanUseMisc(MISC_PET));
+    if (dynamic_cast<CCharEntity*>(m_PBaseEntity) == nullptr)
```
takhlaq: almost! can tidy up a bit by doing this instead
```
auto PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity);
if (PChar && PChar->loc.zone->CanUseMisc(MISC_PET) && !PChar->m_moghouseID)
    lua_pushboolean(L, true);
else
    lua_pushboolean(L, false);

return 1;
```
if it fails on casting to CCharEntity* we know its not a player, and this check is always false for mobs so return false
if the cast is successful, we know its a player so we check if the zone allows pets and they player is not in mog house

teschnei: I think it's better to just DSP_DEBUG_BREAK if it's not a char and static_cast, since mobs cannot call a pet ever anyways using the player pet functions

ymkc: @takhlaq Thanks, didn't see that!

ymkc: should have reloaded the page first :D 
move it back to DSP_DEBUG_BREAK?

takhlaq: yea what @teschnei said

ymkc: updated pr

src/map/lua/lua_baseentity.cpp
```diff
@@ -4401,8 +4401,15 @@ inline int32 CLuaBaseEntity::canUseChocobo(lua_State *L)
 inline int32 CLuaBaseEntity::canUsePet(lua_State *L)
 {
     DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(static_cast<CCharEntity*>(m_PBaseEntity) == nullptr);
```
ymkc: @takhlaq I see, thanks. slowly getting there :D

PR #3385, opened 2016-09-23
Fix derpy Moongate in Ro'Maeve

TeoTwawki:
These weren't opening with the moongate pass KI. So you could only get through by waiting on moonphase or wallhacking. According to the git log its been that way since the files first existed.
https://github.com/DarkstarProject/darkstar/pull/3385/files?w=1


TeoTwawki:
https://github.com/DarkstarProject/darkstar/pull/3385/files?w=1

PR #3396, opened 2016-09-28
Fixed Jeanvirgaud (outpost warp NPC) Location.

chaosdrop:
Hope this is how this works first time using Github... 

Fixed the location for the sandy outpost warp NPC Jeanvirgaud.


takhlaq:
iirc a retail update moved him

TeoTwawki:
Also need to consider the pos that the return warp from an outpost to home send you to. Both need to be current retail if they aren't already. I thought someone already updated this in darkstar /shrug.

chaosdrop:
Didn't realize he was moved retail. I used the pos from the lua file after a player on my server said they couldn't find him, and when I looked in the db his pos didn't match the lua info.

PR #3401, opened 2016-10-03
Found another missing onMobDeath..

TeoTwawki:
While I'm here, lets make that message a bit less useless shall we?
https://github.com/DarkstarProject/darkstar/commit/5713bdb980ee122f780902db78490e2c202ec082?w=1#diff-1905a2168fa2f33669c619683ed327ffR2768


src/map/lua/luautils.cpp
```diff
@@ -2765,7 +2765,7 @@ namespace luautils
                     lua_getglobal(LuaHandle, "onMobDeath");
                     if (lua_isnil(LuaHandle, -1))
                     {
-                        ShowError("luautils::onMobDeath: undefined procedure onMobDeath\n");
+                        ShowError("luautils::onMobDeath: undefined procedure onMobDeath in %s/mobs/%s.lua\n", PMob->loc.zone->GetName(), PMob->GetName());
```
teschnei: the var File should have the filename used (less copy/paste)

TeoTwawki: Was trying to clip `scripts\zones\` off to shorten the message, but have amended it now to use the var.

PR #3406, opened 2016-10-05
Fix Svarbhanu using same mob group ID as Promathia..

TeoTwawki:
Didn't create the mob/fill in all its details, just separating it from the other so 2 mobs aren't using same ID by mistake.

This error that rode in with 05c8bbaafad9f8f9d8dbfab96297b157a4aa0d0a way back in 2013.


takhlaq:
https://travis-ci.org/DarkstarProject/darkstar/builds/165142688#L2421

```
Importing mob_groups.sql into the database...ERROR 1062 (23000) at line 12353: Duplicate entry '14545' for key 'PRIMARY'
```

TeoTwawki:
You would notice right as I'm amending the dang thing

TeoTwawki:
Its been a really shitty day

TeoTwawki:
re-fixed and sql execs now

takhlaq:
i always notice before you get to amend

PR #3407, opened 2016-10-05
Fix incorrect flags (stealth) on Amhuluk (force pop abyssea NM)

TeoTwawki:
Assuming this was copy paste from all the other mobs of same family, which fly hide their name and are untargetable till they land. This guy doesn't spawn in that state.


gedads:
some amphipteres pop on the ground and targetable like this one, the one from rolanberry vwnm i think the others in woe , legion, reisenjima

TeoTwawki:
Your right I forget size is in that field sicne DSP doesn't have it all enum'd yet. derp. amending

TeoTwawki:
p.s.
DSP seems to presently do some stuff with the flags that doesn't match retail for that same value :sad: like someone mixed up their bytes.

TeoTwawki:
Amended. Only specifically removing the one flag now. It'd help if we didn't have "flag" and "flags" next to each other in the same table when I try to follow this stuff back through the core..I'll see about pull requesting better naming for the fields sometime soon.

PR #3409, opened 2016-10-05
Basic spellcasting for SMN spirits

ymkc:
Added all the spells according to wikia and the very minimum so the spirits can cast their spells

I set the mobmod_magic_cool to 48s, that is the base value without modifications.
The mobmod_magic_delay is just an estimation and is probably too low once modifications are coded


ymkc:
turned mob_pool_mods into defaultMobMods

takhlaq:
sorry completely forgot to merge this (was trying to think of another way to do this but ended up not thinking at all)
rebase and i'll merge

ymkc:
@takhlaq np, I will get back to it by tomorrow

PR #3410, opened 2016-10-05
Make Maat spawn as the correct job, and lower his H2H attack speed.

wrenffxi:
This PR contains two fixes.  First, I lowered Maat's attack speed from 240 to 360, which based on videos is how fast he attacks on retail.  Second, I delayed spawning until Maat is initialized.  This prevents Maat from frequently spawning as a monk against non-monk players.


src/map/utils/mobutils.cpp
```diff
@@ -928,8 +928,10 @@ void SetupNMMob(CMobEntity* PMob)
 
 void SetupMaat(CMobEntity* PMob)
 {
-    // Should swing normally when not a mnk
-    PMob->m_Weapons[SLOT_MAIN]->setDelay((240*1000)/60);
+    // I arrived at this attack speed value by comparing his attacks to
+    // videos of retail Maat.  I believe it's base attack rate of 480
+    // minus six levels of martial arts (-20 each)
```
TeoTwawki: According to the original comment this is for non mnk Maat so I'm betting mnk Maat uses the 300 in the db.

From my comment in the open issue about maat:
> Here's the actual line screwing with Maats melee delay. 
`PMob->m_Weapons[SLOT_MAIN]->setDelay((240*1000)/60);`
Prior to that in the DB he's set to exactly 300.
```INSERT INTO `mob_pools` VALUES (2460,'Maat',335,0x0000FA0200000000000000000000000000000000,2,2,1,300,100,0,0,0,0,18,0,0,0,3,0,0,0,1,0,335);```
The majority of monk mobs are set to 360. A few are different values. No idea if they should all have the same delay or not. If there are mistakes now, we can bet on people copy and pasting those mistakes to new mobs later. 
>
> If there are "tiers" for mob h2h delay there should probably be some guidance added to the wiki about it.  (perhaps it was because of MA traits on mnk mobs? Does that even apply with mobs? If so I'd wager that being the reason for any having wrong melee speed when its not the zero delay problem) 

So I suggest editing that sql line as well.

Comment (wrenffxi) [inline src/map/utils/mobutils.cpp]: Done.

File: src/map/utils/mobutils.cpp [hunk 2]
```diff
@@ -928,8 +928,10 @@ void SetupNMMob(CMobEntity* PMob)
...
+    PMob->m_Weapons[SLOT_MAIN]->setDelay((360*1000)/60);
```
takhlaq: hmmm is this even needed (shouldnt it just use the value in mob_pools)?

wrenffxi: Yeah seems kind of redundant now.

TeoTwawki: The idea behind it was whoever did Maat before thought his delay should be lower when he wasn't both-hand punching you.

This whole on the fly set Maats job etc is non retail. If you look in retail each jobs Maat fight spawns a different Maat (they have diff IDs, and you can see multiples of him in the clients dat file via pol utils). But lets get this merged so it goes back to working the way it did before and worry about anything else later.

PR #3416, opened 2016-10-07
More duration bonuses as mods (rampart, flee)

wrenffxi:
Added takhlaq's new MOD_RAMPART_DURATION to Caballarius Coronet and its upgrade.
Created flee duration mod and added it to Rogues and Pillagers Poulaines and their upgrades.
Flee ability now uses flee mod rather than checking equipment directly.


takhlaq:
thanks for reordering the mods (completely forgot to move rampart to where the rest of paladin's mods are)

PR #3420, opened 2016-10-08
Fix some issues with JoL fight (untargetable Ru'Phuabo, mobs depopping when they should not)

TeoTwawki:
We could actually get rid of that stealth roamflag entirely now and use a mixin with an onMobRoam listener for that, couldn't we?


takhlaq:
i'll leave this for @teschnei to take a look at

TeoTwawki:
What I removed was roamflag stealth that that mob should not have, and the param that depops mobs on a timer from when they spawned (which is not what they should be doing)

TeoTwawki:
the question I posed is relate to every mob I didn't touch, that still has its roamflag intact.

PR #3430, opened 2016-10-10
Fix incorrect info for Ironclad Observer

TeoTwawki:


TeoTwawki:
Scuze my amends. Its correct now..

PR #3432, opened 2016-10-10
Sublimation is now removed upon job change

wrenffxi:
Added EFFECTFLAG_ON_JOB_CHANGE status flag.
Added this flag to Sublimation, to remove the effect upon changing jobs.


Hozu:
There are probably other effects that are lost on job change that can't be removed by dispel - perhaps Velocity Shot. Someone would have to verify that on retail I suppose.

wrenffxi:
According to [bg-wiki](https://www.bg-wiki.com/bg/Velocity_Shot), Velocity shot is able to be dispelled and is lost when zoning.

PR #3433, opened 2016-10-10
Phomiuna Aqueducts Mob Spawn Location corrections

R3P0FFXI:
Corrected mob spawn locations in Phomiuna Aqueducts


Hozu:
What are servers that enable level capped CoP supposed to do about those spiders? Perhaps there should be some way of disabling the extra high level mobs?

TeoTwawki:
@hozu there is a setting in setting in settings.lua to cap them off but for server doing super old school with it as a 40 cap zone its really simply to run a simple REPLACE INTO statement to lower their min/max lv for one mob group. In the future after more of these are implemented we could run something based on the existing CoP zone cap setting, rather than adding a another (setMlevel and setSlevel do work on mobs it'd just be another for loop in zone init, like FoV books).

TeoTwawki:
```sql
REPLACE INTO `mob_groups` VALUES (10849,344,27,330,0,0,0,0,41,44,0);
```
Bam all of that spider lv 41-44

TeoTwawki:
Also, R3 did a pr without me! My falling, must hide in mountains.

TeoTwawki:
*\sky not my

PR #3441, opened 2016-10-12
Merge duplicate GM commands for deleting KeyItems.

TeoTwawki:
Same deal as the last one I merged.  Tested working, and no need for 2 of'em.


TeoTwawki:
Pardon me just derped something..amending already

TeoTwawki:
There all better. Realized it was getting a name from an object to get and object from the name to get a name from the object ..or something like that..and my brain went WAIT WHAT.  Also fixed the command I copied that from to not do that, in a way that doesn't generate that nil error I fixed in the wallhack script yesterday.

PR #3448, opened 2016-10-13
Command to check keyitems on players..

TeoTwawki:
Because some players are very unreliable in their story telling about whats "broken".

Input can be the ~~all caps~~ string from the global, the decimal number, or even hex.


TeoTwawki:
Amended just now to make it not care if the input was all caps or not.

PR #3453, opened 2016-10-14
Use JOBS table rather than hard-coded job numbers

wrenffxi:
Replaced as many hard-coded job numbers as I could find.


TeoTwawki:
Thank you very much.

PR #3455, opened 2016-10-14
Created mods for gear related to player abilities

wrenffxi:
I went through all the player abilities and replaced hard-coded gear checks with new mods.


scripts/globals/weaponskills.lua
```diff
@@ -263,7 +263,7 @@ function souleaterBonus(attacker, numhits)
         if attacker:getMainJob() ~= 8 then
             percent = percent / 2;
         end
-        if attacker:getEquipID(SLOT_HEAD) == 12516 or attacker:getEquipID(SLOT_HEAD) == 15232 or attacker:getEquipID(SLOT_BODY) == 14409 or attacker:getEquipID(SLOT_LEGS) == 15370 then
+        if attacker:getMod(MOD_SOULEATER_EFFECT) >= 2 then
```
teschnei: so it only works if its >=2?

what happens if there's a souleater bonus of 1%?

wrenffxi: New commit contains more elegant handling for souleater bonus cap.

PR #3456, opened 2016-10-14
Update effects lists (converted insomnia into work)

TeoTwawki:
Obviously these don't work yet, just adding to the list so people can implement them someday.

Input some missing data such as element on some existing ones, and corrected a mistake or 2 from previous updates (like magic_evasion_boost_ii).


teschnei:
hehehe, clod spikes

TeoTwawki:
I thought it was funny to. I was expecting maybe "dust spikes" or something when they finally used that animation, but nope, "clod".

scripts/globals/status.lua
```diff
@@ -698,7 +699,34 @@ EFFECT_COMMITMENT               = 579
 EFFECT_HASTE_II                 = 580
 EFFECT_FLURRY_II                = 581
 EFFECT_APOGEE                   = 583
-
+EFFECT_ENTRUST                  = 584
+EFFECT_COSTUME_II               = 585
+EFFECT_CURING_CONDUIT           = 586
+EFFECT_TP_BONUS                 = 587
+EFFECT_FINISHING_MOVE_6         = 588
+EFFECT_FIRESTORM                = 589
```
Hozu: Was this supposed to be EFFECT_FIRESTORM_II ?

TeoTwawki: Indeed it was. Nice catch, thanks!

PR #3458, opened 2016-10-14
October 2016 Item_Basic Update

Ninjistix:
- corrected some flags
- corrected some spelling
- added new/missing items

before you guys get touchy the number changes... I haven't done item updates in awhile... so yeah...

Edit: technically minus the name/flag corrections its only about 123 new items 👍


TeoTwawki:
Nobody is "getting touchy". You had previously been asked, by me, to break up your big changes to fit githubs limitation so code reviewers could actually *see* the changes without having to download your branch locally. That got ignored several times while someone merged your changes anyway and the only think I responded with on subsequent repeats was a sadface emoji. Doesn't mean I'm gonna let you cop an attitude about it though.

Ninjistix:
Lol I wasn't copin an attitude I was just saying xD

R3P0FFXI:
@Ninjistix can you break this down please so we can see what was actually changed?

Ninjistix:
I have that saved on my computer at work, so can't access it until Monday. plus I'm not sure how to split the changed w/o missing anything since I dumped my SQL table and copied over the fields. if you know a way to limit or split the commits to whatever lines you guys want let me know.

TeoTwawki:
Its done by making multiple smaller commits in the first place. To change it after the fact would require a history reset then new commits each adding a chunk of the changes.

Just be sure your flag values mean what the DSP core says they do, rather than whatever they polutils or windower's resource says they do (I never remember what flag is which number without looking, all 3 sources do *not* match and we have a many of them wrong right now from people not actually reviewing this). The correct values are in `\src\map\items\item.h` and the core uses hex while the use SQL file uses decimal. 

Usually the project actually asks people to collapse down to one commit, its just very helpful to actually be able to see what changed in cases like this especially where we've had problem areas in the past.

Ninjistix:
I don't have the means/times to do smaller commits atm,

https://www.diffchecker.com/JWmlyhUt

R3P0FFXI:
@Ninjistix im not going thought the entire file however there are errors in this file.

INSERT INTO `item_basic` VALUES (4604,0,'rogue_rice_ball','rogue_rice_ball',12,1548,56,0,153);

INSERT INTO `item_basic` VALUES (4604,0,'rogue_rice_ball','rogue_r.ball',12,1548,56,0,153);

your shortname column is wrong on a lot of these.

Ninjistix:
I'll redo them in smaller batches sometime next week when I have time, unless someone updates it first

PR #3465, opened 2016-10-16
Re-write Virelai spell script, re-indent paralyze/paralyze2 scripts.

TeoTwawki:
In addition to someone adding that broken casting check, the spell never worked even before that anyway. It simply set an (incorrect) message and returned the effect type without ever doing anything.

I've changed it to check the caster as intended, and also made it you know, actually try to charm the target.. 

**Tested both mob and player using this, I used `@exec` for force a mob to cast on me.** It's probably much easier to resist this spell than retail. Someone with better knowledge of that will have to tweak it later if so.

### DIFF
--- a/scripts/globals/spells/maidens_virelai.lua
+++ b/scripts/globals/spells/maidens_virelai.lua
@@ -2,8 +2,9 @@
 -- Spell: Maiden's Virelai
 -- Charms pet
 -----------------------------------------
-
 require("scripts/globals/status");
+require("scripts/globals/magic");
+require("scripts/globals/pets");
 
 -----------------------------------------
 -- OnSpellCast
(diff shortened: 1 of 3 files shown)


scripts/globals/spells/maidens_virelai.lua
```diff
@@ -2,27 +2,52 @@
...
 -----------------------------------------
 -- OnSpellCast
 -----------------------------------------
 
 function onMagicCastingCheck(caster,target,spell)
     -- Ix'Aern DRG pets are Wyverns that 2hour.
-    if (mob:getID() >= 16921023 and mob:getID() <= 16921025) then
-        if (caster:getStatusEffect(EFFECT_SOUL_VOICE)  == nil) then
+    if (caster:getID() >= 16921023 and caster:getID() <= 16921025) then
+        -- This would probably be better to handle in the mobs script but eh w/e..
```
teschnei: this would indeed be better handled in the mobs script! that's why you are going to remove it from here

PR #3466, opened 2016-10-16
Fix multiple overlapping modifier IDs

TeoTwawki:
Someone used an existing modifier ID for a new one.
QD prolly came first I dunno, but it was easier to spot in item_mods.sql so that's the one I changed.

### DIFF
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -1189,9 +1189,9 @@ MOD_GRIMOIRE_INSTANT_CAST    = 419 -- Spells that match your current Arts will o
 MOD_COUNTERSTANCE_EFFECT     = 543 -- Counterstance effect in percents
 MOD_DODGE_EFFECT             = 552 -- Dodge effect in percents
 MOD_FOCUS_EFFECT             = 561 -- Focus effect in percents
-MOD_MUG_EFFECT               = 571 -- Mug effect as multiplier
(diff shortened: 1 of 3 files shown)


TeoTwawki:
p.s. someone may wanna check the amounts of the quick draw mod being used in the item mods file..looks fishy to me.. That is outside the scope of this pull request though.

TeoTwawki:
I just added a 2nd commit to remove over 82 liens of modifier that do not even exist in either status.lua or modifier.h 😠

TeoTwawki:
Refix'd PR. Some were wrong because the previous editor thought IDs were spare because they were not enum'd. Because they are calculated as base+(some other ID) instead.

wrenffxi:
Noted, and sorry about that.

TeoTwawki:
np, I was confused and thought we had a bunch of non existent mods in the sql table for a few min, Hopefully its clearer now for the next dude after us.

Hozu:
Oh... oh god I think I'm responsible for this one. Damn it.

R3P0FFXI:
@Hozu its all good man its what caused the confusion on my initial PR for the mods.

TeoTwawki:
Hozu if its any consolation it let me to noticing the other 3 lol

teschnei:
when I added the WS dmg% modifiers, I made sure the MOD_SPARE were right...

PR #3470, opened 2016-10-17
Add spellflag so virelai can be done right ..

TeoTwawki:
.. and check bard level when bard is mainjob
(virelai is 75 you won't have it from sub, so don't need worry about whm/brd or something trying to land charm)

Since I was brd/bst yesterday (derp) I failed to notice that virelai had a 100% chance to fail if your bst level was zero. Thats fixed now, in a passable way (other than being CHR based, I don't know that level should be involved at all. wiki doesn't say, I just presumed it calculates the way bst charm would, but as a bard). Better than the nothing we started with.

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -4042,13 +4042,13 @@ namespace battleutils
     float GetCharmChance(CBattleEntity* PCharmer, CBattleEntity* PTarget, bool includeCharmAffinityAndChanceMods)
     {
         //---------------------------------------------------------
-        //	chance of charm is based on:
-        //	-CHR - both entities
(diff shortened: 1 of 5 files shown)


src/map/utils/battleutils.cpp
```diff
@@ -4095,11 +4095,20 @@ namespace battleutils
             base = 10;
 
         uint8 charmerBSTlevel = 0;
+        uint8 charmerBRDlevel = static_cast<CCharEntity*>(PCharmer)->jobs.job[JOB_BRD];
```
takhlaq: you'll want this under the player specific check

TeoTwawki:
Amended to fix that derp you spotted

PR #3471, opened 2016-10-17
Ix'Aern Wynav adjustments.

TeoTwawki:
Hey look no ~~hands~~ excess magic number static IDs or extraneous mobmods.

And they soul voice and charm you. And no longer infinitely try to spawn every tick of onMobFight resulting in invisible wynavs (the code was trying to spawn right as the body does fade out) that you can't kill.

In other words, its tested for real.

testdummy_2016 10 17_000012


Hozu:
Do the pets respawn after being killed? That behaviour was intentional, as they are summoned infinitely on retail.

TeoTwawki:
Nothing on the wiki suggested they repopped. I can make them repop real quick before bed, but they should not be doing so even while already up and while dying. there has to be some sort of pause.

Hozu:
I was under the impression that they will respawn, as the one time I fought the thing on retail we didn't kill any of the pets.

TeoTwawki:
I'm changing it, I think your are right and I just missed where it says it. I'm really tired.

TeoTwawki:
Amended. The timing can be adjusted if anybody knows exactly how fast they respawn. The main issue there was that onMobFight cycles so quickly the action check can't prevent it from breaking them,

PR #3473, opened 2016-10-17
Fix San d'Oria mission 5-1 missing KI and var change.

TeoTwawki:
Someone confused the beginning cs of 5-1 for the end cs of 5-1


TeoTwawki:
For players that already saw the cs event:
`addkeyitem NEW_FEIYIN_SEAL <name>`
`setplayervar <name> MissionStatus 10`

PR #3475, opened 2016-10-18
Regen spells and mods fixed

wrenffxi:
Added MOD_REGEN_MULTIPLIER for gear that affects hp/tick of Regen.
MOD_REGEN_EFFECT is flat bonus hp from Light Arts / Tabula Rasa only, not gear.
Updated item_mods.sql to reflect this, and fixed a couple other mistakes:
hes._cyclas (and +1) now has auto-regen mod instead of MOD_REGEN_EFFECT
marduks tiara +1 no longer has fast cast mod, as this is supposed to be a set bonus


wrenffxi:
I made a Google Sheet to arrive at the multiplier mod values.  Shared here:
https://docs.google.com/spreadsheets/d/1nzsVgsEJay9OPMzA86XInsAO9QBuVnxv60fKgfwKvHI/edit?usp=sharing

teschnei:
can you explain the reason for "regen multiplier"?

teschnei:
nevermind, bolelabunga explicitly states regen bonus is a percentage

teschnei:
the next question is: are you sure that it's a ceiling function? ie. does bolelabunga on regen II add 1 or 2 regen?

wrenffxi:
Gear that enhances regen does so as a percent of spell's base.  For example, Piety Briault +1 has a 36% multiplier (as seen in its item text).  If you cast Regen III (base 20 hp / tick) with the Piety Briault +1 on, you'll get math.ceil(20 * 1.36) HP/tick.

After this multiplier is calculated, bonus HP from Merits, Light Arts, and Tabula Rasa are added.

teschnei:
using your example - on bg-wiki, it states that it adds 7 HP to regen 3.  regen 3 is base 20, 20*1.36 is 27.2 (added 7.2).  that's being floored, not ceilinged

wrenffxi:
All tests done here show observed values to be ceilings of predicted values.  I suspect that 7 on bg-wiki may be incorrect, though I cannot test retail myself.  Anyone able to help?
https://www.bluegartr.com/threads/109412-Regen-Spells-and-their-Enhancements?p=6065273&viewfull=1#post6065273

teschnei:
ok, that seems right then

can you rename the modifier for the lights arts bonus? if the only source of that mod is light arts, then it should be named appropriately

wrenffxi:
I don't see an example on other light/dark arts-boosted mods.  How's MOD_REGEN_ADD_HP ?

teschnei:
sorry, i was thinking something that made it extremely obvious that it shouldn't be used for anything except light arts (like MOD_LIGHT_ARTS_REGEN)

PR #3476, opened 2016-10-18
Final missing onMobDeath

wrenffxi:
This is the only remaining mob .lua missing onMobDeath.  All mob scripts now have one.


TeoTwawki:
meme image

teschnei:
probably a bit late, but does anyone know if there's a reason every mob
needs onMobDeath? we probably could have just disabled the error message

TeoTwawki:
Global thing that piggy backs same function I think was why. 

My memory isn't really clear on the details but when I was messing it before you made sure I kept the messaging the way it was before and certain things might possibly silently error without it?

PR #3485, opened 2016-10-23
Added additional effect modifier to Tizona for lv 75 - 119.

nyczducky:
To enable the damage converted to MP additional effect.

Additional Effect - scripts\globals\items\tizona.lua

### DIFF
--- a/scripts/globals/items/tizona.lua
+++ b/scripts/globals/items/tizona.lua
@@ -12,13 +12,27 @@ require("scripts/globals/status");
 
 function onAdditionalEffect(player,target,damage)
     
-    local chance = 10;
+    local chance = 0;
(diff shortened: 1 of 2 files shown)


teschnei:
need to do something for the proc rate (it goes up to 30% at 99)

I don't think we use the actual mod value for anything, so it might be possible for me to have it exposed in some way to the lua function (a param, maybe?), then for this weapon we could store the proc chance in the mod (instead of just 1), and we wouldn't need to do any weird item level checks in the script

or we could do weird item level checks in the script, I haven't decided yet which is better

nyczducky:
Until you decide, I added the ugly item level checks to the script in the meantime.

PR #3488, opened 2016-10-23
A Complete, End-to-End Playable ToAU Storyline

starlightknight:
Hi guys,
It's been awhile :-) 

It's been a little over 10 years now since ToAU radically changed the way we played this game. Along with tons of new and different types of content- besieged, assault, salvage etc- it also brought about a nice new storyline. Today, I offer that storyline preserved for future generations of players ;-)

First off, a huge special thanks to @gedads for gathering a ton of good data and posting it on the forums. After he previously posted some WoTG data, I asked about ToAU and he posted that as well. I had some time off this past week, so I decided to take a shot at implementing all of this, and here it is :-)

This pull request offers a fully playable end-to-end ToAU main storyline, including all mission battles, and some supporting core functionality for them. Pre-existing missions 1-14 were also reviewed and playtested, and bugs were fixed there as well. Each mission is its own commit that can be individually reviewed, and each core enhancement is also its own commit that can be individually reviewed. However, each commit is dependent on the last, especially since there are many shared npcs like Naja. 

As for the current state of this implementation - I believe it is in a state of pretty good quality for a first pass. All major unique behaviors of the mobs in mission battles are implemented. End-to-end testing was done on a Linux Mint 18 (native server & client w/ Wine 1.9). All battles were playtested solo in a variety of scenarios. They should theoretically of course work in a party as well, but I haven't tested them that way yet. Volunteers welcome ;-) 

There are, however, a few known issues. I'll leave some notes below.

**Notes for people:**

@teschnei - I made some changes to core to support the mission battles, mostly around fixing instance or bcnm related issues, as well as making it easier to control ally behavior, since these battles have a handful of npc allies. I have left more specific notes in each commit message for core. I would be happy to hear any feedback you may have - everything can always be improved, and I'm sure my implementations aren't perfect :-) 

I also have two hackfixes I'd love to get rid of, but wasn't able to figure out proper solutions for. Maybe you know the answer to these. One was an issue where-in the server would crash in the C++ STL if a local mob var was set while the mob was dead-- I'll file an issue with a backtrace for you. Another is, the event that enters Nyzul Isle seems to spam event updates uncontrollably, which was preventing proper entry into the zone. The other instanced areas did not have this issue. For the mission battle, I used a double lock in the runic seal script to trap and ignore the spam while the instance started up, which makes it work properly but is of course a bit ugly. I'd love to understand why the event spams so that I can get rid of this, but I failed to figure it out :-(

@gedads - I need some npcs capped to finish things up here. In particular, the Jade Sepulcher npcs for Puppet in Peril. I was able to get the cutscene in a working state using POLUtils data + stealing mob looks from similar npcs, and it plays with minor glitches. Some of the accessories, weapons, or colors are not quite right which cause minor glitches, and I don't have the look for the mount. If you could cap these to get the right data that would be great. I also put a comment "-- CAP ME PLEASE" next to other misc NPCs that need to be capped. For any I was missing, I stubbed them to avoid client crashes. Most of the missing ones are "blanks".

@TeoTwawki - I haven't verified that the skill lists for these battles are all accurate. All the mob pools already existed in the database and were attached to one. They seem reasonable but not quite right either. If you and any of your minions want to help by looking at these (i see you guys committing a lot of this type of stuff), that would help ;p If not I'll get to it at some point...

**Known issues for battle missions:**

**TOAU-15 Black Coffin**
- If a player logs out on the Cutter before entering the battlefield, when they log back in they'll be standing on invisible ground. The cutter will be moving in automatically but is not there yet. 
- Additionally, if the player is on the docks while the cutter is arriving, they can walk into the air without it being there. I'm assuming there is either an npc thats supposed to be a gate like with other transports (and the player would be placed on the dock on login if they were in the region), or, that there's a way to make the cutter just be there since it has no set schedule - but I don't know which is right or how this is controlled.
- These issues have no effect on the battle, and if you enter the zone normally the cutter will have arrived by the time you get to it.

**TOAU-22 Shield of Diplomacy**
- Battlefields 2 & 3 do not have NPCs/mobs available

**TOAU-29 Puppet in Peril**
- Some minor cutscene glitches that need NPCs capped
- Battlefields 2 & 3 do not have NPCs/mobs available
- The BCNM door on Battlefield one needs to be capped, preventing the playing from running away (though that's broken by another bug anyways)

**TOAU-31 Shades of Vengeance**
- The instance ID is definitely wrong, but the instance is not on the instance list. I used an available
id which appears to work, but it thinks its Better than One at the beginning. This does not effect the
battle other than some incorrect text at the beginning. I'm not sure how this is supposed to work -
maybe there are some special params that make it share an instance.
- The starting position for this mission needs to be capped. I made a guess off of watching a few videos for now, as I did not have this data.

**TOAU-35 Legacy of the Lost** 
- Three missing npcs that need captured that don't appear to cause any visual defects that I can tell

**TOAU-42 Path of Darkness**
- Entry/exit transitions do not black out, making them look clunky. I could not find the right event.
- Hack needed to enter properly. See commit notes for details.

**TOAU-44 Nashimera's Plea**
- Entry/exit transitions do not black out, making them look clunky. I could not find the right event.
- Hack needed to enter properly. See commit notes for details.
- AFAICT we don't have the right damage tracking available to do the immunity effect properly, so
I scripted a guess using the combination of jobs in the party, so that an effect is chosen for now

**Known general issues:**
- Cutscenes at the imperial whitegate used an npc with a non-ANSI character as the name. The server occasionally loses track of this NPC (properly a bug related to it being a non-ANSI name), but it HAS to be named this or you get a client crash. Any cutscenes that play that require this npc after
the server has lost track of the npc will cause a client crash. If the server loses track of the npc, you'll see a Server Need NPC message in the log. The only fix is to restart the server.

Looking forward to thoughts and feedback :-)


gedads:
Thanks a lot! that's amazing not only for the script but because you actually added a way to finally have mobs assisting players and fighting randomly pick a mob in those assault areas. awesome!. A big thank you man.

starlightknight:
@gedads Thanks to you for making this possible ;-) I'd never have been able to do this without your data. 

And yes, I tried to make directing ally targeting easy to trigger from a script, as that is increasingly necessary in these (and later) missions.

gedads:
thanks, 
just a few details (for toaum15) gessho uses text message 7563 and 7564 when it's about to die and when it dies, the captain uses 7569 before jump then 7570 on land (something like "you overpowered my crew"...).
Also i think we miss the right anim for vulcan shot. and gessho's moves, he uses teleport (anim 1301-1302 and uses the double kick on exit) . those are small things compared to the job you did and can totally be added later

TeoTwawki:
@starlightknight wow, huge pr. Surprised to see so much at once. Thank you for breaking that down in multiplier commits. Will look into those skill lists when I get a chance, got a guy that wants to learn that anyway so we can tackle those together.

teschnei:
I will probably only be able to look at the core changes and the bug reports you added once I have time

starlightknight:
@gedads Thanks. Those are the kind of details never captured in wikis. I was able to easily find Karababa's text ids for Shield of Diplomacy, but the rest I couldn't tell what is what. I'd be happy to add those soon, and if you know any more please let me know :-)

@teschnei Thanks. no problem. Hopefully some people will merge this in and playtest in the meantime

@TeoTwawki Well, it kind of happened. I had a week off, so I started on Monday and it just kind of came together. I was hoping to post this Friday but I lost a lot of time trying to figure out the the two core issues I gave up on, as well as the nyzul entry event spam issue. I made sure to keep the commits as fine grained as possible so that it is well broken down for review :-)

TeoTwawki:
> AFAICT we don't have the right damage tracking available to do the immunity effect properly, so I scripted a guess using the combination of jobs in the party, so that an effect is chosen for now

Need a listener then can use a setLocalVar to track each type. On the old AI I had something similar on a custom mob on Annwn but unfortunately that code wouldn't be retail behavior here. I had made up extra `onMobHit[type](arg,arg,arg)` things (see [onCriticalHit](https://github.com/DarkstarProject/darkstar/blob/0889d2359c897166a37f81234ae5aa6ef03d833e/src/map/entities/battleentity.cpp#L1463)) and was applying the modifiers for dmg resistance on a per hit basis, I think it'd have to apply before he stands back up here (or possibly just on the mobs engage). A listener would be much better than my ratty old code.

nyczducky:
I will definitely load this up. 

I was starting to tackle ToAU but you were way ahead. I'm open to learning as well. I'd like to help out. Getting retail access soon and I'll be able to get some good info. Just need some guidance as to what.

Hozu:
re: Alexander, it is to use Divine Judgment at some point around 50% (I've never seen it above 60 or below 40, and I failed the damn mission 43 times on retail -.-). When it uses it, it uses the line of text that is solely the character omega. I think there may be another text when it's about to use it, but I can't remember.

src/map/instance.cpp
```diff
@@ -257,3 +259,130 @@ bool CInstance::CheckFirstEntry(uint32 id)
...
+            PCurrentMob->PEnmityContainer->Clear();
+            PCurrentMob->PAI->Disengage();
+            PCurrentMob->PAI->Inactive(std::chrono::milliseconds(10000), false);
+        }
+    }
+}
+
+void CInstance::StartAllyAssist(ALLY_ASSIST_MODE mode)
+{
+    // If there's no mobs, we can't assist. Don't waste time.
```
teschnei: why did you make this a core function?  instead of a mixin, or a generic lua function?

src/map/instance.cpp
```diff
@@ -257,3 +259,130 @@ bool CInstance::CheckFirstEntry(uint32 id)
     //insert returns a pair (iterator,inserted)
     return m_enteredChars.insert(id).second;
 }
+
+void CInstance::DisengageAll()
```
teschnei: i don't really understand the usefulness of this - how many instances are going to use this? an instance isn't necessarily a bcnm - was anything preventing you from doing it in script?

src/map/battlefield.cpp
```diff
@@ -453,11 +540,13 @@ void CBattlefield::OpenChestinBcnm() {
 
 bool CBattlefield::loseBcnm() {
     beforeCleanup();
-    for (auto&& PChar : m_PlayerList)
+    for (int i = 0; i < m_PlayerList.size(); i++)
```
teschnei: I don't think this is right, but it's something I can look at (it probably should have just had its own issue though)

src/map/instance.cpp
teschnei: worse, it's a copy/paste from the bcnm version..

src/map/zone_entities.h
```diff
@@ -71,6 +72,7 @@ class CZoneEntities
     bool			CharListEmpty();
     uint16			GetNewTargID();
 
+    EntityList_t	m_allyList;
```
teschnei: probably better to define this in CInstance - it has no relevance to a regular zone (and just adds confusion with the bcnms ally list vs the zones, if it's visible)

src/map/lua/luautils.cpp
```diff
@@ -1517,27 +1517,61 @@ namespace luautils
         lua_pushnil(LuaHandle);
         lua_setglobal(LuaHandle, "onEventUpdate");
 
+        bool luaFileFound = false;
```
teschnei: this kind of turned into a giant shitshow - maybe you can break out the repeated logic into a smaller function?
I can handle that if you don't know how to do it

starlightknight: I can do that. I should have to begin with, but the previous logic was itself copy/paste and I was getting things working first (bad excuse). I'll clean it up today.

src/map/instance.cpp
starlightknight: I was using it in most of them, but I'm not sure its needed now. I had to do this manually before i fixed the event issues below. Now I notice some of the events cause this to happen as a side effect. if that's the case I may be able to drop this entirely. I had already started removing it where it wasn't strictly necessary. As for doing it in a script, there are no bindings to directly engage or disengage a mob.

src/map/battlefield.cpp
starlightknight: I'm not entirely sure it is either, but it works. I hit this issue repeatedly when testing the lose containers on Karabara leaving the battlefield. An event would flood the player making the leave get interrupted an never happen. I got rid of the auto and went back to the old school style because I'm unfamiliar with modern C++ and kept getting a server crash related to the dynamic typing, and the rest of the file is using the old style anyways

src/map/instance.cpp
starlightknight: I added both versions, and indeed it is largely copy/paste :-( 

The data structures aren't the same - instances are using std::maps everywhere, and battlefields are using std::vectors everywhere, meaning I couldn't share the code without refactoring or repeatedly converting all data sets at runtime. 

This is ultimately triggered by a generic lua function at the appropriate time in the scripts. I had bound it to this in core for now. For doing more than this directly in lua, more bindings would have to be added, which may not necessarily be a bad thing.  There is no command to directly engage a given mob in lua for example. When I looked at the Dawn battlefield with Prishe assisting, this was done by adding enmity to her on the mob. However, that doesn't work in the ToAU battlefields. Prishe is technically classified as an NPC, and these are mobs with allegiance flag = player. That trick doesn't seem to work in this case - the mobs will never engage initially no matter how much enmity you dump on them. It was also more scripting than I wanted for repeated use, so I tried to make a single trigger here to minimize what someone would have to do to get it going. 

We could definitely push more of it out to lua if we added more bindings, but I wasn't sure of where the line should be as to how much of core we expose to lua, in that once scripts start using things then refactoring the previously private behavior now has script breaking consequences.

src/map/zone_entities.h
starlightknight: Ironically I had done that at first, then changed it. I was thinking that it may be useful in more cases outside of instances eventually - like besieged or campaign for example. But then again, my targeting system is tied to the other two. *sigh* refactoring needed either way. Thoughts on that?

starlightknight:
@teschnei Thanks for feedback, I'll get started on the cleanup :-) In the first pass I was primarily focusing on getting things working in the old mantra of, make it work, make it right, then make it fast (if needed). Stage 1 complete, starting stage 2.... ;-)

src/map/battlefield.cpp
Cloudef: If this modifies the m_PlayerList then both ways are wrong.

R3P0FFXI:
@starlightknight let me know which mobs you need the correct skills/spells for and I will add it in as I am adding more abyssea skills/spells

starlightknight:
@R3P0FFXI if you look at mob_pools.sql in the diff it will show you - its small enough to show up in github. You can ignore the random snipper, there was a trailing whitespace my editor removed. The rest are the mission mobs. Thanks!

src/map/battlefield.cpp
teschnei: ah right - the previous code didn't remove them, though, which is why I think this is wrong (and the analogous winBcnm doesn't either) - probably needs a check for bcnm->lost in onEventFinish to mirror how win works

src/map/instance.cpp
teschnei: disregard second half of previous comment - I think it's because allies in bcnms get added to the zones mobList, but the ones in instances don't (they get added to allyList, which probably doesn't get iterated over in the zone counter)

scripts/zones/Navukgo_Execution_Chamber/bcnms/shield_of_diplomacy.lua
```diff
@@ -0,0 +1,63 @@
+-----------------------------------
+-- Area: Navukgo Execution Chamber
+-- BCNM: TOAU-22 Shield of Diplomacy
+-----------------------------------
+package.loaded["scripts/zones/Navukgo_Execution_Chamber/TextIDs"] = nil;
+-----------------------------------
+
```
nyczducky: require("scripts/globals/missions");

src/map/lua/luautils.cpp
```diff
@@ -1629,7 +1597,7 @@ namespace luautils
     int32 OnEventFinish(CCharEntity* PChar, uint16 eventID, uint32 result)
     {
         //#TODO: move this to BCNM stuff when it's rewritten
-        if (PChar->PBCNM && PChar->PBCNM->won())
+        if (PChar->PBCNM && (PChar->PBCNM->won() || PChar->PBCNM->lost() || eventID == 32003)) // 32003 is the run-away event
```
teschnei: you can remove this, I fixed the bcnm loss message (and will be looking at run away shortly)

src/map/lua/lua_battlefield.cpp
```diff
@@ -187,7 +187,7 @@ inline int32 CLuaBattlefield::lose(lua_State* L)
 {
     DSP_DEBUG_BREAK_IF(m_PLuaBattlefield == nullptr);
 
-    m_PLuaBattlefield->lose();
+    m_PLuaBattlefield->loseBcnm();
```
teschnei: undo this change as well

src/map/zone_entities.h
teschnei: i'll think about what to do about this

ibm2431:
@Hozu Re: Alexander text. I was able to pull up my log from when I cleared the fight for my mule two years ago. I intentionally dragged the fight out a bit to make it feel more epic than it was (fighting at ilvl). Has text messages, TP moves, and a rough estimate for how much HP it has if someone feels like adding up all the damage: https://gist.github.com/ibm2431/0942c47423128acf49f54532d974c5ce

Apologies for the length of the log; this was before I started using Battlemod. I have a handful of screenshots for eyeballing HP%, but the screenshot of the Omega line does appear to be between 40%-50%. When it uses Radiant Sacrament for 165 near the end, its HP% appears to be around 5-10%.

starlightknight:
@ibm2431 thanks! im working on textid stuff for the battles now, this will help

nyczducky:
And here I am working on this and you got it all done. Awesome job man.

@gedads teach meh. I'll be getting a retail sub to pull cs id's and such.
Will help with WoTG etc

teschnei:
for the black coffin, are you just using local vars to signal an action to the other mobs?  if you are, you could just replace that with your own listener and have them trigger it - it's clearer, doesn't rely on the later tick for the other mobs, and probably works better

src/map/lua/luautils.cpp
```diff
@@ -4572,4 +4566,70 @@ namespace luautils
...
+            luaFileFound = !(luaL_loadfile(LuaHandle, luaFile) || lua_pcall(LuaHandle, 0, 0, 0));
+
+            if(luaFileFound)
+            {
+                lua_getglobal(LuaHandle, functionName);
+                luaFunctionFound = !(lua_isnil(LuaHandle, -1));
+            }
+        };
+
+        auto resetState = [&luaFile]() mutable {
```
takhlaq: dont think this needs to be mutable since youre capturing by reference

starlightknight:
@teschnei yes, that's exactly what i'm doing. Are there any examples I can look at? I did clean up Black Coffin to where it won't trigger the crash anymore, but if there's a better way to do things I'd still like to learn :-) Any other feedback on the new state? If not, I'll clean up this last round and hopefully we can merge this as a starting point :-)

src/map/lua/luautils.cpp
```diff
@@ -2286,6 +2239,47 @@ namespace luautils
         return retVal;
     }
 
+    int32 OnMonsterSkillPrepare(CBattleEntity* PMob, uint16 skillId)
+    {
+        DSP_DEBUG_BREAK_IF(PCaster == nullptr || PTarget == nullptr);
```
starlightknight: Wut... how did that even compile

teschnei:
if you can explain to me exactly what you are trying to accomplish in the script, I can think about it

I have not looked at all the changes since yesterday yet, probably something to do later tonight

starlightknight:
@teschnei Which one? The ally assist? In that one, just provide two standard profiles for picking a target for the ally, and make it an easy one liner without a bunch of code to manage, as this will be increasingly common going forward. 

The first one attempts to target the same mob that the player does, assuming the player has a target. The 2nd one will pick a random mob to attack. Some allies actively try to assist the player. Other ones engage random mobs. The point of the script is a reusable piece to analyze the available mobs in the instance or BCNM, and then pick according to that + the profile.

At this point, I did put a lot of effort to move it to lua as you asked in the first review on it, and I retested everything. I'd prefer not to reinvent it a third time - I'm running out of time to spend completely rewriting stuff.

teschnei:
whatever you were using local vars for - i didn't look at the disengage stuff yet

you could have saved a lot of time asking for advice on the more esoteric things in IRC :)

Hozu:
@ibm2431 looking at your log there, it was doing regular melee attacks? Huh... Actually now that I think about it, when it was attacking people at range with the mobskill, the skill was in the log every time, so it's not exactly a replaced attack. Rather, it's just forced. I wonder what the delay between uses are then.

TeoTwawki:
@hozu basically thats its ranged attack, and if I remember right its about (maybe a bit quicker) the time between a gigas's boulder throws. If your tank stands back and nobody else pulls hate you'll never see its regular melee, since it doesn't move. When I cleared we used a -mdt setup and let the tank eat those from a distance till after Divine judgement, then zerged down its remaining health after recovering from that.

Hozu:
Oh duh, a ranged attack. I don't know why I never made that connection before.

teschnei:
I merged this to a new branch (toau) - anyone can make PRs to that branch now for updating this

Issue #3477, opened 2016-10-18
Headstone Pilgrimage Opo Opos reappear if you click the stone too early.


wrenffxi:
**_I have:_**
- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160831_1

**_Server Version_** (type `@revision` in game) **:**
7a7a317c1a1e2052cf95237959cc6eafce2268f

**_Source Branch_** (master/stable) **:**

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

Be on Zilart mission HEADSTONE_PILGRIMAGE.
Spawn the two Opo Opos at the Cermet Headstone in Yuhtunga Jungle.
After they die, but before they fade out, click on the headstone to attempt to get Fire Fragment.
Instead of receiving the KI, the Opo Opos reappear.

phlare:
I thought I had the same thing happen with dark spark the other day, but I also may have been clicking the wrong torch after the fight. Will try to confirm.

TeoTwawki:
@teschnei sounds related to what I was telling you last night about the death/fadout/repop

teschnei:
i think you're right there - after death but before despawn they are
considered unpopped, and perhaps the complete var is not set until
onmobdespawn - so the check if theyre still spawned returns false before
they fade out, and it shows you having not killed them yet

i might be able to test the changes I made last night tonight

TeoTwawki:
WHT...Why is this even a server var..
```lua

    if (csid == 0x00C8 and option == 1) then
        SpawnMob(17281031):updateClaim(player); -- Carthi
        SpawnMob(17281030):updateClaim(player); -- Tipha
        SetServerVariable("[ZM4]Fire_Headstone_Active",0);
```
```lua
    if (GetServerVariable("[ZM4]Fire_Headstone_Active") == 0) then
        SetServerVariable("[ZM4]Fire_Headstone_Active",1);
    elseif (GetServerVariable("[ZM4]Fire_Headstone_Active") == 1) then
        SetServerVariable("[ZM4]Fire_Headstone_Active",os.time()+ 900);
    end
```

Kushdr:
It is most certainly that the variable is set in onMobDespawn.
Without changing anything, the easy answer is just wait until they depop.
[...]

PR #3492, opened 2016-10-24
Quick fix for issue #3477

TeoTwawki:
I also re-indented the headstone and added a check for if the NMs were already up. If they are already up it just doesn't run the event.

### DIFF
--- a/scripts/zones/Yuhtunga_Jungle/mobs/Carthi.lua
+++ b/scripts/zones/Yuhtunga_Jungle/mobs/Carthi.lua
@@ -15,6 +15,13 @@ end;
 -----------------------------------
 
 function onMobDeath(mob, player, isKiller)
+    if (isKiller) then -- Causes this to only run once per Carthi kill
+        if (GetServerVariable("[ZM4]Fire_Headstone_Active") == 0) then
+            SetServerVariable("[ZM4]Fire_Headstone_Active",1);
+        elseif (GetServerVariable("[ZM4]Fire_Headstone_Active") == 1) then
+            SetServerVariable("[ZM4]Fire_Headstone_Active",os.time()+ 900);
+        end
+    end
 end;
 
 -----------------------------------
@@ -22,11 +29,4 @@ end;
 -----------------------------------
 
 function onMobDespawn(mob)
-
-    if (GetServerVariable("[ZM4]Fire_Headstone_Active") == 0) then
-        SetServerVariable("[ZM4]Fire_Headstone_Active",1);
-    elseif (GetServerVariable("[ZM4]Fire_Headstone_Active") == 1) then
-        SetServerVariable("[ZM4]Fire_Headstone_Active",os.time()+ 900);
-    end
-
 end;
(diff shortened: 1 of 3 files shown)


scripts/zones/Yuhtunga_Jungle/mobs/Carthi.lua
```diff
@@ -15,18 +15,18 @@ end;
 -----------------------------------
 
 function onMobDeath(mob, player, isKiller)
+    if (isKiller) then -- Causes this to only run once per Carthi kill
```
Hozu: My concern with is approach is, if it dies to DoT while unclaimed then this won't run. Does it have to die claimed on retail to work or no?

teschnei: if isKiller or !player then
?

Hozu: isKiller returns true or false on the player that landed the killing blow. I'm 99% sure it will return false if it was killed by DoT while unclaimed.

TeoTwawki: There is still a killer when its a dot kill, unless someone broke that.

TeoTwawki: And unclaimed would cause probs on retail yes.

Hozu: I'm assuming there's no killer because of this: https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Sauromugue_Champaign/mobs/Old_Sabertooth.lua#L25

...which needs to be fixed.

TeoTwawki: Old saber has an altogether other problem..Player gonna be nil. You aren't supposed to even engage it. So when it dies its not going to have a valid "owner" in the core.

Also see edit on my previous comment.

TeoTwawki: I got to do my fight twice on retail coz I died and reraised, and before I stood up bio killed the 2nd chimp. I was a little po'd at the time. Woudn up having to come back later with help. NO SORO. >:(

TeoTwawki: One thing I'm not so sure of is the way this is requiring both Nm kills, I'd wager retail lets you depop one and then when the 2nd dies it only cares that both are now dead. But I only set out to quick get this back to working near as it had been as fast as I could, not redo whatever the previous implementation might have been wrong about. If anyone wants to PR a better implementation (maybe ditch the server vars..) I'll happily close this PR in favor of theirs.

teschnei:
There's a message for clicking when they are still up

TeoTwawki:
> There's a message for clicking when they are still up

is it the same cannot remove fragment message already in there? I can re-arrange check so it displays during the fight too, as well as when you just don't meet the ki get requirements.

TeoTwawki:
I added the default message to it in that spot, rather than combine the 2 conditionals because I think its more readable this way. If thats not the right message let me know.

PR #3494, opened 2016-10-24
Fix variable name in onMobDeath of Old Sabertooth

Hozu:
It wasn't changed when onMobDeath was modified.

### DIFF
--- a/scripts/zones/Sauromugue_Champaign/mobs/Old_Sabertooth.lua
+++ b/scripts/zones/Sauromugue_Champaign/mobs/Old_Sabertooth.lua
@@ -22,7 +22,7 @@ end;
 
 function onMobDeath(mob, player, isKiller)
 
-    if (killer == nil) then
+    if (player == nil) then
     
         local players = mob:getZone():getPlayers();
         for i, player in pairs(players) do


TeoTwawki:
This and the NPC need redone. Player should indeed be nil, because you aren't supposed to fight it. But you can't setvar on a nil entity.

Hozu:
Oh it's using "player" as a new variable in the for loop too... ugh I guess I'll have to close this request because I'm stuck using the stupid github editor that can't merge commits, and redo it.

TeoTwawki:
server var might actually be cleanest option here, due to there only ver being one old tiger at a time. If it also requires all members of a party to each pop one (I don't recall/don't think I ever did this in a group) then set/getlocalvar can be used to see who popped the kitty too.

TeoTwawki:
found it!
so by looking at the ole code that this removed theres a way to do it without server vars, sorta. https://github.com/darkstarproject/darkstar/commit/4863f9d3070fbb5c76821c6dea3dd3f46ab48b78#diff-d3138280cd5272931b563454302332e3L25?w=1

Hozu:
Are you looking at the removed lines or the added ones from that commit?

gedads:
someone should go capture the path, the difficult part of this quest is that the mob is moving inside the cave between young tigers that aggro @ level 30

TeoTwawki:
@Hozu the removed ones in that one and prior to even that, it was just setting a var that equaled failing on kill (and then the NPC counted it as a **win** if you did **not** have this var and the tiger was also down after being popped) which seems even more simple that iterating ppl in the zone to see who is at the tiger area. Didn't see the reason it was changed, prolly buried in the pr comments.

TeoTwawki:
@gedads the tigers are sneakable and are not at the back part of cave at all. A difficult part of this doesn't exist. :)

Hozu:
Oh, well I made a new pull already, but if it needs to be changed beyond what I did someone else will have to do it. Sorry.

TeoTwawki:
its fine, I think it works as you have it now on your new one.

gedads:
the old tiger spawns outside the cave and walk all the way back to the cave where it dies alone

TeoTwawki:
We already know this. 

If its path doesn't begin at the correct spot thats a totally separate issue. I was making a wisecrack about how it doesn't really create any new challenge, (everyone just slips past those all undetected and sits at the back of cave till it dies).

gedads:
that was challenging back then

TeoTwawki:
If you had no whm to main or sub had no friends and were also too cheap for oil/powder on everything you did, sure.

Hozu:
Can't you just ride a chocobo into the back of the cave anyway? lol.

TeoTwawki:
Yes. you can 😆

gedads:
that's if -> you have a chocobo licence, got the grass to do the quest (having no clue where to find it of course) have money as a lev 30 war to buy oil? (how? @ 30K a stack of oils and 30 k a stack of powders etc). not everyone choose thief or whm/blm and back then oils would last from 20 seconds to 90 seconds not a reliable 3-4 minutes likes it's now or 8 minutes for sneak.

Hozu:
How is that any different from retail though? Well aside from the improper path of Old Sabertooth itself I guess.

If sneak/invis items have a shorter duration than retail, they should be updated.

TeoTwawki:
I had my choco license at 20. I was able to afford oils and powers before I even unlocked my first advanced job. Its not hard. And non of it has a thing to do with the actual issue he's addressing. I know you care about the details. thats great, really. But don't you think a better way to get attention to the thing you want to see corrected is to give it its own issue?

takhlaq:
discussions on forums or irc
everyone following this project emails for every comment and its kinda annoying when its just arguing "difficulty" instead of anything productive

teschnei:
Is this email conversation really necessary? You're both in IRC, nobody
else needs to read this

Issue #3258, opened 2016-07-02
Radiant Breath effects wear instantly


xipies:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160329_1

**_Server Version_** (type `@revision` in game) **:**
790e080

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
Both Radiant Breath effects (slow, silence) wear instantly (no delay in chat log between spell cast finishing and effects wearing).

Chaotic Eye (silence effect) works as expected, on same mob. Effect lasts expected amount of time (or at least doesn't wear instantly).

This behavior feels like #2009 did. Radiant Breath does light based damage in addition to the effects.

PR #3500, opened 2016-10-27
Added EFFECT_SLOW and EFFECT_SILENCE to getBlueEffectDuration

nyczducky:
Max durations will need retail confirmation along with resist values but this will cap duration based on white magic equivalents of slow and silence.

Resolved issue #3258


takhlaq:
since the login campaign is up, i'll wait til someone with a retail char tests and gets data so we can get this closer to retail

PR #3509, opened 2016-10-29
Brothers ENM added

Fiocitrine:
Gave functionality to NPCs involved for the ENM
Droplist ID for ENM
Mob spell list added
Among other things!


Hozu:
I have a question - can a player actually reach the specific battlefield entry point? Last time I checked in-game the waterfall would not disappear when there was no ice weather.

Fiocitrine:
As far as I know, you can reach the battlefield. It's the same battlefield as the CoP mission Flames for the Dead (Snoll fight) and we were able to finish the mission. It's basically: fall down a hole in Ulegarand Range, head to the zone, head up to the Whirling Vortex and click on it with the KI.

Hozu:
That's not the correct location. http://ffxiclopedia.wikia.com/wiki/Brothers

> - Climb to the top of Uleguerand Range to H-8. During clear weather, an ice wall will open up to allow passage to the east. Follow this tunnel to I-7 to zone in to Bearclaw Pinnacle. Examine the Wind Pillar to enter Brothers ENM.
- The ice wall is the small section of the large ice-flow that is a markedly different color- it is more white than the rest of the ice flow, and looks slightly out of place. 

I assume the frozen waterfall is treated as a door - the hard part is finding the correct NPCID. You would use the onWeatherChange function in the zone.lua file for the zone, and have it "close" for ice/double ice, and open for any other weather.

gedads:
hi, the npc is 16798112.

Fiocitrine:
I see what you're saying. I'm just learning lua and not incredibly familiar with this. It may have to be someone else, with more experience, who scripts the waterfall to open and close.

Hozu:
http://pastebin.com/D2cUsf3y

Add that to the zone.lua file for Ulegurand Range and test it for me. There's a GM command to change the weather which *should* work, but I do know that command does not affect weather-based latent effects. So you may have to sit around the waterfall and wait for a weather change.

I did some searching and it looks like there's no need to put this check in onZoneInitialize because the function is triggered for any weather change, including when it gets set on server boot.

takhlaq:
most the logic for enm npcs was already done on my fork a couple years ago https://github.com/takhlaq/darkstar/commit/91f2519cffb9a493a8bd1ec1dbd2707255a919b5

though i dont know if the quests were actual quests in quest log
also i lost like everything on that branch after force push so some stuff might be missing

Fiocitrine:
Gonna try that script now.

> Edit: You'll also have to make it so the only valid entry point for the fight is the proper location, if you haven't already done so already.

do  you mean the whirlwind?

Hozu:
Oh the NPC isn't placed in the DB? Huh... I wonder if it's necessary to place it? If it's not working in the test, try moving it to 1,1,1 and see if that works. If that doesn't either, then it will need an actual position.

Fiocitrine:
Alright I've give it a pos and try it from there. 

Yeah, Wind_Pillar should be good.

Fiocitrine:
So I gave the door a pos 1 1 1 just to test and it's not spawning let alone opening or closing during the respective weather.

Hozu:
Yeah looks like it will probably need a proper pos. I've checked the rising bridge NPCs in Ro'Maeve in the npc_list table and they have a pos set.

lichtl:
Sure about it not being in the DB?

INSERT INTO `npc_list` VALUES (16798112,'_058','',0,4.874,-180.593,67.569,1,40,40,9,0,0,0,6147,0x0200000000000000000000000000000000000000,0,'COP',0);

TeoTwawki:
Special NPCs often have special names. So it being _058 instead of Wind_Pillar is not surprising, and probably why they didn't see it.

Fiocitrine:
ah I see! thanks guys! Still haven't been able to get this to work :/

lichtl:
Got it working.

https://github.com/lichtl/darkstar/commit/80fc13616aa391b6a3c5a4b9bc607f91dfb007c5

Fiocitrine:
Sweet thanks! push that shit so they will get merged!

Hozu:
Ooh looks like my thing didn't work because I forgot the include.

lichtl:
@Hozu yes thanks. It gave me a good idea. Just the combined if wasnt working properly. 

I also just noticed that when you zone through the waterfall zone exit you end up at 0,0,0 in Bearclaw Pinnacle. Is anyone able to provide the expected point where to zone in?

Fiocitrine:
Should be -680 18.5 -540 when I use @zone

lichtl:
@Fiocitrine did you zoned in through the Waterfall? I do appear at 200, 660 -195.490 which is in the nowhere. Looks like this zoneline is wrong. However I recall that you would appear somewhere else (as much as I remember the ENM battlefield was at a different location than the COP ones

INSERT INTO `zonelines` VALUES (943009914,5,6,200.000,-195.490,660.000,251);
ffxi_2016 11 09_00 27 22

Think I finally got it. Should be 121, -193, 660

Is someone able to confirm this with retail?

Fiocitrine:
when I zone in from Uleguerand into Bearclaw that's the pos I get too, according to DSP. I didn't think the waterfall was anything but a "wall" of sorts that opens during clear weather. Or am I wrong?

lichtl:
I think you mean the zone used for the COP BC. I recall the ENM's only be reachable through the waterfall.

http://ffxi.allakhazam.com/db/quests.html?fquest=613

Once you hit the correct ledge follow the cave and trade your pouch for a key item: Zephyr Fan. Then climb the mountain again. When you get to G-8 about were Jormungand Spawns take the East path that leads to the frozen waterfall wait for the weather to clear and the waterfall will thaw out and an opening will appear follow that path to get to the ENM.

I've adjusted my PR including the zone lines being fixed.

The Wind Pillar in the zone (via Waterfall) is 16801888. I've just checked the DB (10 entries) so I think there depends on which one you reach they do different events (one being COP missin BC) other ENM.

lichtl:
I have to correct myself. The right ID for the Wind Pillar is 16801894 .I also noted above the comment from Hozu regarding reaching the zone.

Fiocitrine:
Ah ok I'm sorry. I wasn't getting what you were saying. When I zone into Bearclaw via the waterfall, it sends me to the same place you are.

scripts/globals/bcnm.lua
```diff
@@ -398,6 +398,9 @@ function checkNonTradeBCNM(player, npc)
            if (player:getCurrentMission(COP) == THREE_PATHS  and  player:getVar("COP_Ulmia_s_Path") == 6) then -- flames_for_the_dead
              mask = GetBattleBitmask(640, Zone, 1);
              player:setVar("trade_bcnmid", 640);
+	    elseif (player:hasKeyItem(ZEPHYR_FAN)==true) then -- Brothers ENM
```
takhlaq: no need for the == true

scripts/zones/Uleguerand_Range/npcs/Zebada.lua
```diff
@@ -12,14 +12,35 @@ package.loaded["scripts/zones/Uleguerand_Range/TextIDs"] = nil;
...
+    local ZephyrFanCD = player:getVar("[ENM]ZephyrFan");
+    
+    if (player:hasKeyItem(ZEPHYR_FAN)) then
+        player:startEvent(12);
+    else
+        if (ZephyrFanCD >= os.time(t)) then
+            -- Both Vanadiel time and unix timestamps are based on seconds. Add the difference to the event.
+            player:startEvent(15, VanadielTime()+(ZephyrFanCD-os.time(t)));
+        else
+            if (player:hasItem(1780)==true or player:hasItem(1779)==true) then -- Chamnaet Ice -- Cotton Pouch
```
takhlaq: same here

PR #3517, opened 2016-11-02
Absorb and Nullify damage mods

wrenffxi:


wrenffxi:
While doing this QC I may have found a related issue, and I'm hoping someone more familiar with the code can weigh in.  It looks to me like elemental magic null and absorb mods checks occur twice, once at battleutils.cpp:4239-4244, and again at magic.lua:676-681.  Does this double the proc chance for gear with, say, MOD_FIRE_NULL?

teschnei:
there's at least one code path where that's the case, it seems - make an
issue report for it

PR #3523, opened 2016-11-03
Some SQL and Script adjustments for ToAU branch. And some required fixes to non ToAU scripts (textIDs issue).

TeoTwawki:
My apologies for not breaking this down better and having bad organizational skills.

Includes skill list and spell list adjustments for:
 - Lancelord Gaheel Ja
 - Khimaira 13
 - Amnaf (blue mage form)
 - Amnaf (flayer form)
 - Naja Salaheem
 - Raubahn
 - Razfahd
 - Alexander

There are also some previously missing/wrong skill animations added to mob_skills.sql

**Notes:**

 - Git thinks Savage_blade was renamed into Flat_Blade..What I did was correct the casing on the all the "ws as mobskill" scripts, and yes I used the correct method for this not the OS's rename. Git is just the electronic equivalent of drunk.

 - Mobskill versions of Elemental WS were doing physical instead of elemental damage. Corrected to elemental now.

 - Those uses of messageBasic in the skillscripts were already present, I just moved them while fixing an issue where the scripts could call invalid textIDs..

 - Amnaf (flayer)'s Sleepga has a hate reset. I don't like what I had to do here, but comments in file explain the reason it is this way.

 - I tested in as far as making sure I didn't break anything that was already working, and that my additions are as correct as can currently be made. Preexisting problems interfere with some parts, and I have left comments all over related to those.


gedads:
from old mobskill db, zantetsuken was 1447. Also there's gessho teleport in/out, 1301 and 1302 (no idea about how to do it , tried to check AATT but i don't get it... if anyone figures it out you're welcome to finish windurst 9-2).

scripts/globals/spells/sleepga.lua
```diff
@@ -17,8 +17,20 @@ function onSpellCast(caster,target,spell)
...
+        if (caster:getPool() == 5310) then -- Amnaf (Flayer)
+            caster:resetEnmity(target);
+        end
+        -- Todo: get rid of this whole block by handling it in the mob script
+        -- this requires a multi target enmity without specifying a target (have to get hate list from mob)
+        -- OR by altering onSpellPrecast to have a target param..
+        -- onMonsterMagicPrepare is not a realistic option.
+        -- You'd have to script the use of every individual spell in Amnaf's list..
+    end
+
```
Hozu: Is it not possible to add the onSpellCast function into the individual mob script, then do a check for spell ID being sleepga? I did something similar for Jailer of Hope and Plasma Charge, and it worked.

https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/AlTaieu/mobs/Jailer_of_Hope.lua#L40 (though looking back at it I'd have done the resulting action differently now, heh).

TeoTwawki: To my knowledge that doesn't work with onSpellCast like it does with onMobWeaponSkill. We had onMonsterMagicPrepare (which lacks a spellID/spell argument) and onSpellPrecast (which lacks a target argument). I think the best solution will be to add anotehr form of resetEmnity that doesn't require target input so I can just say `mob:resetAllEnmity();` in onSpellPrecast in the mobs script. But that doesn't exist yet.

TeoTwawki: If someone wants to either make onSpellCast work like that or gimme my multi target hate reset function (either/both of which I am certain would see some other ruses in the future), I will happily nuke that blight out of sleepga's script (hint hint).

scripts/globals/mobskills/Flat_Blade.lua
```diff
@@ -1,36 +1,39 @@
...
+        mob:messageBasic(43, 0, 35);
+    end
+
     return 0;
 end;
 
 function onMobWeaponSkill(target, mob, skill)
-skill:setMsg(0)
-target:showText(mob,SAVAGE_LAND);
+    if (mob:getPool() == 4006) then -- Trion@Qubia_Arena only
```
takhlaq: why not just check zoneid?

takhlaq: derp
gudpoint

scripts/globals/spells/sleepga.lua
starlightknight: i had a mob:clearAllEnmity() in this branch, but it was reverted a few commits back with the disengage changes

PR #3528, opened 2016-11-07
Fix waterfall in Uleguerand Range

lichtl:


scripts/zones/Uleguerand_Range/Zone.lua
```diff
@@ -77,4 +78,23 @@ function onEventFinish(player,csid,option)
...
+
+function onZoneWeatherChange(weather)
+
+    local waterfall = GetNPCByID(16798112);
+
+    if (weather == WEATHER_SNOW or weather == WEATHER_BLIZZARDS) then
+        if (waterfall:getAnimation() ~= 9) then
+            waterfall:setAnimation(9);
+        end
+    elseif (weather ~= WEATHER_SNOW or weather ~= WEATHER_BLIZZARDS) then
```
lichtl: Thanks. Just tested. Indeed works without else. Pushing fix shortly

PR #3529, opened 2016-11-08
Added fix for pets to not respawn on zone if on chocobo

thedraked:
SMN pet behavior update to retail behavior.  Pet's automatically attack whatever aggro's Master if not already attacking a mob.

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -343,8 +343,8 @@ void SmallPacket0x00C(map_session_data_t* session, CCharEntity* PChar, CBasicPac
     // respawn any pets from last zone
     if (PChar->petZoningInfo.respawnPet == true)
     {
-        // only repawn pet in valid zones
-        if (PChar->loc.zone->CanUseMisc(MISC_PET) && !PChar->m_moghouseID)
+        // only repawn pet in valid zones and if not on chocobo
+        if (PChar->loc.zone->CanUseMisc(MISC_PET) && !PChar->m_moghouseID && !PChar->StatusEffectContainer->HasStatusEffect(EFFECT_CHOCOBO))
         {
             switch (PChar->petZoningInfo.petType)
             {


teschnei:
why mob_controller and not pet_controller? it's a pet specific behaviour

teschnei:
also, this did at one time work, so it might be worth looking around to see if there's already an implementation of it that broke for some reason

teschnei:
actually the better place probably is when the master gets aggroed - check only once for an avatar rather than having to check the master every tick

thedraked:
Thats where I got the code from, original implementation.  It didnt work and was in wrong spot, at least to function as retail behavior.  

I had it in the initial aggro originally, but then it would only work on initial aggro, not after pet dies and you resummon the pet would just sit there do nothing.  I did this many times in retail: have pet attack, run to safe distance for recasting after pet died, pet would die, recast then pet would immediately go after mob, this way the master never got hit.

Its under mob_controller as I needed to know when mob started aggro'ing.

thedraked:
If you can find a place to put it in pet_controller that behaves the same, I would love to put it there.

teschnei:
Pet controller inherits from mob controller.. so it would be the same
place.  Also, if this fix is intended to fix the existing implementation,
you should probably remove the old one so we don't have both

thedraked:
Might try it again later, but that was the only place I could put it where it gave the behavior of retail.

thedraked:
I thought I did remove the original implementation?  bleh, maybe not.  My first commit so sue me :P

teschnei:
disregard what i said about pet_controller.. I was reading it backwards

thedraked:
Bah, I wanted to make a seperate commit, but it got rid of the original file.  Do I have to wait until this one is closed before I can do another pull request?

takhlaq:
commit to new branch (based off dsp master)

PR #3537, opened 2016-11-14
Implements modifiers for macaron foods.

xipies:
Implements synthesis success and synthesis skill gain from Cherry Macaron, Coffee Macaron, and Kitron Macaron.

Clamped success rate to 99% since people were still reporting breaks even with Kitron Macaron (+7% success rate). References cited in source code comments.

### DIFF
--- a/scripts/globals/items/cherry_macaron.lua
+++ b/scripts/globals/items/cherry_macaron.lua
@@ -14,11 +14,11 @@ require("scripts/globals/status");
 -----------------------------------------
 
 function onItemCheck(target)
-local result = 0;
+    local result = 0;
     if (target:hasStatusEffect(EFFECT_FOOD) == true or target:hasStatusEffect(EFFECT_FIELD_SUPPORT_FOOD) == true) then
         result = 246;
     end
-return result;
+    return result;
 end;
 
 -----------------------------------------
--- a/scripts/globals/items/cherry_macaron.lua
+++ b/scripts/globals/items/cherry_macaron.lua
@@ -1,0 +1,48 @@
+-----------------------------------------
+-- ID: 5779
+-- Item: cherry_macaron
+-- Food Effect: 30Min, All Races
+-----------------------------------------
+-- Increases rate of synthesis success +3%
+-- Increases synthesis skill gain rate +3% 
+-----------------------------------------
+
+require("scripts/globals/status");
+
+-----------------------------------------
+-- OnItemCheck
+-----------------------------------------
+
+function onItemCheck(target)
+local result = 0;
+    if (target:hasStatusEffect(EFFECT_FOOD) == true or target:hasStatusEffect(EFFECT_FIELD_SUPPORT_FOOD) == true) then
+        result = 246;
+    end
+return result;
+end;
+
+-----------------------------------------
+-- OnItemUse
+-----------------------------------------
+
+function onItemUse(target)
+    target:addStatusEffect(EFFECT_FOOD,0,0,1800,5779);
+end;
+
+-----------------------------------
+-- onEffectGain Action
+-----------------------------------
+
+function onEffectGain(target,effect)
+    target:addMod(MOD_SYNTH_SUCCESS, 3);
+    target:addMod(MOD_SYNTH_SKILL_GAIN, 3);
+end;
+
+-----------------------------------------
+-- onEffectLose Action
+-----------------------------------------
+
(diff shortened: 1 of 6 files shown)


scripts/globals/items/cherry_macaron.lua
```diff
@@ -0,0 +1,48 @@
...
+-----------------------------------------
+
+require("scripts/globals/status");
+
+-----------------------------------------
+-- OnItemCheck
+-----------------------------------------
+
+function onItemCheck(target)
+local result = 0;
```
TeoTwawki: Indention, please.
```lua
function onItemCheck(target)
    local result = 0;
    if (target:hasStatusEffect(EFFECT_FOOD) == true or target:hasStatusEffect(EFFECT_FIELD_SUPPORT_FOOD) == true) then
        result = 246;
    end
    return result;
end;
```

xipies: Sorry, I had cloned melon_pie.lua. :) Committed update.

Hozu:
Are these foods meant to increase the HQ rate as well? Where you put it in synthutils I'm pretty sure the food will be able to increase the HQ tier.

scripts/globals/items/cherry_macaron.lua
```diff
@@ -0,0 +1,48 @@
...
+
+require("scripts/globals/status");
+
+-----------------------------------------
+-- OnItemCheck
+-----------------------------------------
+
+function onItemCheck(target)
+    local result = 0;
+    if (target:hasStatusEffect(EFFECT_FOOD) == true or target:hasStatusEffect(EFFECT_FIELD_SUPPORT_FOOD) == true) then
```
takhlaq: no need for == true, can just check if they have food effect or support effect

scripts/globals/items/coffee_macaron.lua
```diff
@@ -0,0 +1,48 @@
...
+
+require("scripts/globals/status");
+
+-----------------------------------------
+-- OnItemCheck
+-----------------------------------------
+
+function onItemCheck(target)
+    local result = 0;
+    if (target:hasStatusEffect(EFFECT_FOOD) == true or target:hasStatusEffect(EFFECT_FIELD_SUPPORT_FOOD) == true) then
```
takhlaq: same here

scripts/globals/items/kitron_macaron.lua
```diff
@@ -0,0 +1,48 @@
...
+
+require("scripts/globals/status");
+
+-----------------------------------------
+-- OnItemCheck
+-----------------------------------------
+
+function onItemCheck(target)
+    local result = 0;
+    if (target:hasStatusEffect(EFFECT_FOOD) == true or target:hasStatusEffect(EFFECT_FIELD_SUPPORT_FOOD) == true) then
```
takhlaq: and here

TeoTwawki: @xipies past me apologizes for giving you bad examples in dozens of food scripts.

Also for the crap way I handled the subid thing with that fov food effect instead of changing the thing that would have let us not check 2 effects for that. >.>;;

TeoTwawki: (food uses the item id to link it up with the item script, which fov food doesn't have)

xipies:
@Hozu, from simply eyeballing it, "synthDiff" affects "hqtier", and "hqtier" affects "chance" (HQ chance), which gets compared to "random". I am not seeing where "success" is part of the HQ check.

If I'm missing something please let me know.

Hozu:
Oh you're right I didn't expand above the change. Though, what's up with the hardcoded lightning element thing from before? I can't figure it out in the ~5 mins I had to look at it.

xipies:
Looks like if using lightning crystals, subtract 20% from success chance.

teschnei:
i don't think we have a way to differentiate synthesis from desynthesis in
the recipe table... which is not good, but that's trying to make desynths
harder (even though not all desynths are lightning nor vice versa)

Hozu:
I took a quick glance at it and I don't think there's an entry for that in the table. Well, there's a column named Type but it's not even looked at in the query: https://github.com/DarkstarProject/darkstar/blob/master/src/map/utils/synthutils.cpp#L72

I'll make an issue later today in case anyone has anything else they want to say about it by time I'm able to.

PR #3538, opened 2016-11-14
SMN Pet Bodyguard behavior fix

RunAwayDSP:
For thedraked


thedraked:
Haha, thanks RunAway :)  I was hoping this would be easier than it is, but alas it is beyond me atm.

RunAwayDSP:
Honestly thedraked I simply forked from here went to the file edited it through the website and compared/push 

When there is more than 1 file it can be easier to use git programs or cmd lines but when it comes to one file very easy to just do it on the git website

PR #3527, opened 2016-11-06
Aggro detection changes on some families

Brierre:
Pugils detect by sound, not sight
Imps detect true sight
Treants detect sound
Flan detect sight and job ability
Lizards detect sound

PR #3539, opened 2016-11-14
Aggro detection changes on some families

Brierre:
Recommit of #3527

Pugils detect by sound
Imps detect true sight
Treants detect sound
Flan detect sight and job ability
Lizards detect sound


takhlaq:
yep, thanks!

PR #3540, opened 2016-11-14
removed non player spells from the list

Ninjistix:
- http://pastebin.com/raw/BsfQxhzY (master spell list for all jobs)
- separated spells currently disabled in spell_list.sql
- http://pastebin.com/raw/fkZwEwkW (current spells missing from spell_list.sql)


takhlaq:
mmm if the spells removed were just the ones with blank names then yeah, but i don't really see the need to remove mob spells if they can be used for testing

Ninjistix:
would it be better to split the cmds? into something like @addplayerspells or something

TeoTwawki:
I have a compromise suggestion: leave it split like you have it, but default uncommented. servers that don't want addallspells to mean ALL spells can just comment those lines

TeoTwawki:
easier than picking out the spell IDs in the table

Ninjistix:
actually the ones that i have commented out are actually commented out from the current spell_list, but there's a set of spells that are missing form the list too, in that second pastbin, has the names/jobs

Ninjistix:
if anything i can split the lines between players and mobs, to make it easier for servers to disable mob spells...

TeoTwawki:
> if anything i can split the lines between players and mobs, to make it easier for servers to disable mob spells...

sounds fine to me.

As for missing, those do no harm and work once someone actually implements them. Command was rework awhile back with something about that along with making it not spam packets, iirc.

ibm2431:
What about a flag when calling the command? Like `@addallspells player_only` or `@addallspells include_mob_spells`.

PR #3543, opened 2016-11-15
Fixed cutscene ID for exiting to Caedarva via Nasheefa.

wrenffxi:
Old CSID (0x00be) moves the NPC to the side but locks the player in place until they hit escape, making it impossible to pass.

New CSID (0x00b8) brings up the "Do you wish to leave the ruins?" menu, and allows the player to pass.

For easy testing:
Nasheefa is @pos -443.524 4 -739.546 79
Imperial Silver Piece is @additem 2185


takhlaq:
remember to three backticks before and after code (especially gm commands since we end up sending notifications to users unfortunate enough to have the same usernames as our commands otherwise)

PR #3549, opened 2016-11-19
Fix NPCs expansions

ymkc:
According to wikia/bgwiki


sql/npc_list.sql
```diff
@@ -1554,7 +1554,7 @@ INSERT INTO `npc_list` VALUES (16879988,'Riftworn_Pyxis','Riftworn Pyxis',0,443.
 INSERT INTO `npc_list` VALUES (16879989,'Riftworn_Pyxis','Riftworn Pyxis',0,267.1,-15,222,1,50,50,0,4,100,2,3,0x0000C90300000000000000000000000000000000,2,'ABYSSEA',1);
 INSERT INTO `npc_list` VALUES (16879990,'Riftworn_Pyxis','Riftworn Pyxis',0,517.099,15,-313,1,50,50,0,4,100,2,3,0x0000C90300000000000000000000000000000000,2,'ABYSSEA',1);
 INSERT INTO `npc_list` VALUES (16879991,'Survival_Guide','Survival Guide',64,-61,-30,261,1,50,50,0,4,100,2,3,0x0000C90300000000000000000000000000000000,2,null,1);
-INSERT INTO `npc_list` VALUES (16879995,'Undunlating_Confluence','Undulating Confluence',12,-48.977,-23.309,572.489,31,50,50,0,0,0,0,3,0x0000BA0900000000000000000000000000000000,0,null,1);
+INSERT INTO `npc_list` VALUES (16879995,'Undunlating_Confluence','Undulating Confluence',12,-48.977,-23.309,572.489,31,50,50,0,0,0,0,3,0x0000BA0900000000000000000000000000000000,0,'ABYSSEA',1);
```
TeoTwawki: For ROV but you put "ABYSSEA"? There is no bit in the lobbies expansions mask which is why it was null, since these aren't tied to abyssea (or any other expansion). More accurate way to handle it would be to default them to not spawned, and then have lua script check the ENABLE_ROV in settings.lua to spawn them (similar to what we do for FoV, GoV, and more recently, Voidwatch NPCs).

TeoTwawki: Not really important since these don't do anything yet, so can stay like it is unless yer bored :P

TeoTwawki: Oh nope, I'm completely wrong, and you can just put "ROV" where you have "Abyssea"
https://github.com/DarkstarProject/darkstar/blob/8a19f0b716ed49432bfb417e7e26e577a220c64e/src/map/lua/luautils.cpp#L1169
its actually reading settings.lua and trimming "ENABLE_" off of the var.

teschnei:
the typo in the script name for the confluences bothers me

TeoTwawki:
> the typo in the script name for the confluences bothers me

I've infected you.

ymkc:
Not sure how i managed to do that, but i actually used ROV like in the commit message... Might have cherry-picked from the wrong branch, will change it

PR #3555, opened 2016-11-22
Repair Maat fight win conditions.

TeoTwawki:
I did BLM Maat just 2 days ago, he gave up **_around_** 20%. We used to have a 10% check in the core, and the only online source I could find says 25% but isn't all that trust worthy, so I'm going with 20% unless someone has solid evidence to the contrary.

I have not tested this. If you want to wait, it may be a few weeks before I get around to doing so. If someone else gives this a shot in the meantime, please comment.

edit: if this doesn't work, I expect Ouryu to also not work right.

### DIFF
--- a/scripts/zones/Balgas_Dais/mobs/Maat.lua
+++ b/scripts/zones/Balgas_Dais/mobs/Maat.lua
@@ -29,6 +29,23 @@ function onMobEngaged(mob,target)
 end;
 
 -----------------------------------
+-- onMobFight Action
+-----------------------------------
+
+function onMobFight(mob, target)
+    local bf = mob:getBattlefield();
+
+    if (mob:getHPP() <20) then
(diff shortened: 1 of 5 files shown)


TeoTwawki:
helps if I remember to define the var for the battlefield we are winning, huh? 😊

TeoTwawki:
Amended

nyczducky:
Confirmed working. WHM gives up after 5 and THF steal works.

TeoTwawki:
Thanks!

Fiocitrine:
So weird, both WHM 5 minutes and THF steal method worked over a year ago.

TeoTwawki:
Special win conditions were removed from core and needed redone in script fio.

Fiocitrine:
ah I see

Issue #3438, opened 2016-10-11
Mining Errors


R3P0FFXI:
**_I have:_**
- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
301601831_1

**_Server Version_** (type `@revision` in game) **:**
50868e308882da87ca59798352f348186b000f2f

**_Source Branch_** (master/stable) **:**
Master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

Anytime someone mines the following error appears in the log

[11/Oct] [09:30:46][Error] luautils::onTrade: .\scripts/globals/mining.lua:445:attempt to index local 'currentnpc' (a nil value)

PR #3559, opened 2016-11-23
Update mining.lua

EulerPath:
updating mining points for Palborough mines to match what is in sql file.
I play on Official DarkStar Classic Server and mining points in Palborough mines were not despawning after mining them. The issue seems to be that the mining point indexes were not correct to what was in the sql file.  I have tested this on my test server and this seems to work.

this is my first pull request so let me know if I need to do anything different next time.

This pull request may or may not solve this issue:
Mining Errors #3438


TeoTwawki:
At anybody reading:

Probably similar ID offset issues in other mining zones too, if anybody wants to check.

Not sure if the IDs in script have been getting updated when the SQL batch was run (there's a tool for handling _most_ of the globals, that has to run after the SQL tool is finished) or if these were already wrong before we ran npc ID updates the last few times.

starlightknight:
Unless someone forgot to run the lua id job afterward, they should have been right.

TeoTwawki:
thats what I was saying, is it might not have been

PR #3565, opened 2016-12-04
Update version.info

tank19041:
Updating version


TeoTwawki:
We don't update the version info till we are updated to actually use that version. Right now we are not.

PR #3573, opened 2016-12-10
Improve treasure binding and fix bug where client says you dropped item to yourself.

TeoTwawki:
So, how do you feel about moving BCNM loot from SQL to script? It'd be fairly easy to do and I could fix the longstanding bug where the last item gets force dropped at the same time.

### DIFF
--- a/scripts/commands/addtreasure.lua
+++ b/scripts/commands/addtreasure.lua
@@ -6,10 +6,10 @@
 cmdprops =
 {
     permission = 1,
-    parameters = "is"
+    parameters = "isi"
 };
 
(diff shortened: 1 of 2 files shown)


teschnei:
that's something I do want to see done, yeah

TeoTwawki:
I think I can accomplish it fairly quickly next time I get a day to myself.

takhlaq:
if you feel like finishing off rewrite im not even using sql there (and i'd prefer there not to be conflicts when it is finally ready to merge)

TeoTwawki:
@takhlaq I can just do it in that branch and it can land in master when the branch gets merged

PR #3578, opened 2016-12-12
Version update 2016/11/10 (client: 30161103_1)

ibm2431:
Updating for the version update in November 2016. Notes:

- Batch tool picked up massive ID shifts in the outdoor 3 nation event areas. It seemed pretty positive and I spot-checked a couple from retail, they appear to be accurate.
- I captured and added the new Jingle Baas event NPCs from retail. Did a capture during the event and tried to keep the (squashed) changes from During to After in the commit log, but screwed up my rebasing. This NPC list is for after the event, so none of the event NPCs will appear, but I _do_ have data from during the event if anyone needs it.
- There are a couple of free spaces between the new event NPCs and the Treasure Coffers. They might be used for future events, or SE could be reorganizing event NPCs.
- My NPC logger picked up a "moved" Riftworn Pyxis in South Gustaberg, but apparently the SQL line was a copy/paste of another Pyxis. There was a second copy/paste Pyxis, so I set its position to that of the third Planar Rift which was missing a Pyxis.
- Judging by position and Look, South Gustaberg's Fireworks were moved from 17216253 to 17216224.
- Slight re-ordering of some Moogles in South Gustaberg. Probably got put in wrong order by the batch tool at some point, rather than a deliberate change by SE.
- Added a couple (currently inactive) event NPCs that were missing from South Gustaberg and West Sarutabaruta which SE was kind enough to send me packets for.
- Batch tool seemed to indicate there was a either a new, or previously missing, Shadow Lord in Throne Room [S]. The ID in relation to others in the zone seemed to indicate that it was a CS NPC, so I gave it the properties of all the other CS Shadow Lords nearby.
- There were no Text ID changes.


helixhamin:
Hopefully this will be accepted xD
So annoying that I keep having to change the version number in the code, and that things keep breaking because of an out of sync client.

Thank you for this @ibm2431

Ninjistix:
@ibm2431 sorry to be a bother, didn't have another way to contact you, is it possible to check the December client version that was just released for Client: **30161206_3**

takhlaq:
ya you can update server with https://github.com/DarkstarProject/darkstar-utils/tree/master/id-update-batch-jobs/binaries

tagban:
Yea I managed to build the entire thing, followed the readmes to the letter, (Although the one pointing to POLUtils is wrong so I had to find the most recent version of that from windower). And still haven't been able to get it to work. It also required JDK 7 which doesn't exist anymore, so  hopefully JDK8 is good enough. I did the setup with maven inside of Eclipse, and everything 'appears' to work. But it didn't generate the necessary folders, etc. I'll post a bug report on that tool though instead.

PR #3580, opened 2016-12-13
SAM AF3 demons should be immune to sleep and silence

wrenffxi:
Gaki, Onki, and Rasetsu immune to sleep and silence


whasf:
What is your source?

wrenffxi:
All sources I can find agree.  Common strat seems to be sleep only the SMN's pet, because all three demons are immune to sleep and silence.

http://ffxiclopedia.wikia.com/wiki/A_Thief_in_Norg!%3F
http://ffxi.jento.mobi/2008/12/samurai-af-guide.html
https://translate.google.com/translate?hl=en&sl=zh-CN&u=http://www.ffxisf.com/thread-796-1-1.html&prev=search
http://ffxi.allakhazam.com/db/quests.html?fquest=359#m113905654610922
http://www.ign.com/wikis/final-fantasy-xi/Samurai_AF1-3:_A_Thief_in_Norg

PR #3584, opened 2016-12-15
Rampart fix

eraffxi:
https://www.bg-wiki.com/bg/Rampart

gedads will check tomorrow to verify if this is a phalanx or stoneskin type of effect, it will give a magic shield effect on the party at the power of vit*2 and magic shield will wear off after magic dmg exceeds or duration expires.

placed a var set and check to make sure it doesn't interfere with how magic shield functions with mobs. tested and working correctly in a party. this does not take into effect the def bonus, someone can add that in later if we can actually verify exactly what levels the def bonus changes at.

### DIFF
--- a/scripts/globals/abilities/rampart.lua
+++ b/scripts/globals/abilities/rampart.lua
@@ -23,5 +23,9 @@ end;
 
 function onUseAbility(player,target,ability)
     local duration = 30 + player:getMod(MOD_RAMPART_DURATION);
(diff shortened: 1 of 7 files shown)


whasf:
Waiting on gedads..

eraffxi:
no need yet, still having some trouble after changing it to use subpower instead at kj's suggestion. was told just not to close it for the time being.

takhlaq:
reopen when you get round to updating (and conflicts are resolved)

PR #3587, opened 2016-12-17
Abyssea Resistance Fighter models fixup

ibm2431:
@gedads noticed my NPC Logger didn't properly output display models for Resistance Fighters. This should fix them right up.


TeoTwawki:
ERROR 1064 (42000) at line 1000: You have an error in your SQL syntax; 
check the manual that corresponds to your MySQL server version for the right syntax to use near '0x0100050100100D20163005400550836083700000,0,'ABYSSEA',0)' at line 1

ibm2431:
How embarrassing. Remedied.

PR #3588, opened 2016-12-18
Workaround for antlions that have ambush ability, but are non-aggresi…

xipies:
…ve, getting stuck underground after disengage, and upon roaming after returning to spawn point. This is especially for Alastor Antlion and Executioner Antlion.

Alastor Antlion and Executioner Antlion are the only antlions with ambush that are non-aggressive in mob_pools.

Alastor Antlion has ambush animation when popped [1], but does not burrow when going unclaimed [2]
[1] Alastor Antlion https://www.youtube.com/watch?v=Q3L0SAxiK54
[2] Alastor Antlion duo part 2 of 5 https://www.youtube.com/watch?v=-hAyx56fVnE

### DIFF
--- a/scripts/mixins/families/antlion_ambush_noaggro.lua
+++ b/scripts/mixins/families/antlion_ambush_noaggro.lua
@@ -1,0 +1,24 @@
+-- Antlion family mixin (for ones that ambush and are non-aggressive)
+-- Not hiding upon DISENGAGE event, since mob would be spawned and no longer be accessible
+-- Example: Alastor Antlion has ambush animation when popped [1], but does not burrow when going unclaimed [2]
+-- [1] Alastor Antlion https://www.youtube.com/watch?v=Q3L0SAxiK54
+-- [2] Alastor Antlion duo part 2 of 5 https://www.youtube.com/watch?v=-hAyx56fVnE
+
+require("scripts/globals/mixins")
+
+g_mixins = g_mixins or {}
+g_mixins.families = g_mixins.families or {}
+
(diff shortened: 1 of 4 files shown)


Fiocitrine:
Did you test this out, Ash? If so, and it works, awesome job! Vanyel will be excited!

Hozu:
I mentioned this in the issue already, but I'm pretty sure you can remove the entire conditional check for the ambush roamflag in the core and the aggressive antlions should be unchanged (ie. I forgot to remove that when I fixed their behaviour a while back). Or did you try that and it didn't work? Ideally, there shouldn't be anything in the core for the (un)hiding upon aggro/deaggro for antlions due to the mixins. But the roamflag itself needs to remain because it's used to restrict their movement range from spawn point somewhere else in the core.

xipies:
Yes, I tested locally.

I did not try to remove the entire ambush conditional in the core (and test with aggressive ambush antlions). I can check on that.

I did not remove the ambush roamflag.

Hozu:
Ok, I'm fairly certain the existing mixin would handle the aggressive ones just fine if the stuff in the core were entirely removed. The idea was to completely remove that when I fixed it, but I apparently forgot. Sorry. :/

xipies:
I just tested Cave Antlion with the entire conditional removed. It appears to be working fine. Tested with "too far from home" [1] and player dying (and mob going unclaimed) right next to spawn [2]. I'll make the change and commit.

[1] Resulted in burrow -> despawn.
[2] Resulted in burrow -> stay spawned (at spawn point).

Issue #3590, opened 2016-12-21
The Competition quest - requires trading 10001 fish


TheLoneUs:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** **30161206_3**

**_Server Version_** **718623696ae4431bdbb9613bedf2d3c75c56a7de**

**_Source Branch_** **master**

**_Additional Information_**
When trading fish in **The Competition** if the quest is set to 9999 fish and you trade one fish, you are not awarded the fishing rod. Suspect it is checking the total before trading instead of after trading. I actually had to trade 10001 fish.

TeoTwawki:
#3591

PR #3591, opened 2016-12-21
Fix issue #3590

TeoTwawki:
Someone double check that I didn't derp something..


TeoTwawki:
Meh if theres a prob I'll either fix or revert what I did here.

PR #3592, opened 2016-12-21
Create separate NM timers so server operators can customize the NM sp…

helixhamin:
… and the respawns.
This allows users to customize when the timed spawns respawn.
I will work on completing the additional timed spawns when I have additional time.

Currently checking this pull for feedback. I still need to test to make sure I created my lua arrays correctly (Unless someone can confirm that I did, since this is the first time I have created lua arrays)

### DIFF
--- a/scripts/globals/timedspawns.lua
+++ b/scripts/globals/timedspawns.lua
@@ -1,0 +1,85 @@
+------------------------------------
+--
+-- Contains timed spawn data
+-- for both specific zones and mobs
+-- 
+------------------------------------
+
+-- Landking timer is set up as {initial spawn timer after server startup, respawn timer, the window timer, the king timer once ph is killed}
+behemoth_timer = {900, 75600, 10800, 72 * 3600} -- Timer defaults are: {900, 75600, 10800, 72 * 3600}
+fafnir_timer = {900, 75600, 10800, 72 * 3600} -- Timer defaults are: {900, 75600, 10800, 72 * 3600}
+adamantoise_timer = {900, 75600, 10800, 72 * 3600} -- Timer defaults are: {900, 75600, 10800, 72 * 3600}
+
+-- Wyrm timer is set up as as {initial spawn timer after server startup, last window after server start/first window after last kill, last window after last kill}
+tiamat_timer = {86400, 259200, 432000} -- Timer defaults are: {86400, 259200, 432000}
+vrtra_timer = {86400, 259200, 432000} -- Timer defaults are: {86400, 259200, 432000}
+jormungand_timer = {86400, 259200, 432000} -- Timer defaults are: {86400, 259200, 432000}
+
+-- Abyss Sahagin Sea Serpent Grotto
+-- Ahtu Batallia Downs
+-- Amanita West Ronfaure
+-- Ansherekh Abyssea - Altepa
+-- Argus Maze of Shakhrami
+-- Ash Dragon Ifrit's Cauldron
+-- Bakka Abyssea - Konschtat
+-- Balaur Abyssea - Konschtat
+-- Bedrock Barry North Gustaberg
+-- Bendigeit Vran Konschtat Highlands
+-- Bo'Who Warmonger Pashhow Marshlands
+-- Bloodsucker (NM) Bostaunieux Oubliette
+-- Bright-handed Kunberry Yhoator Jungle
+-- Calcabrina Beaucedine Glacier
+-- Capricious Cassie Fei'Yin
+-- Carmine Dobsonfly Riverne - Site A01
+-- Centurio XII-I Eastern Altepa Desert
+-- Chonchon Meriphataud Mountains
+-- Citipati Attohwa Chasm
+-- Coo Keja the Unseen Meriphataud Mountains
+-- Coral Sahagin Sea Serpent Grotto
+-- Drexerion the Condemned Bostaunieux Oubliette
+-- Duke Decapod East Sarutabaruta
+-- Eba Phomiuna Aqueducts
+-- Fistule Abyssea - Konschtat
+-- Fraelissa Jugner Forest
+-- Gangly Gean Abyssea - Konschtat
+-- Ghillie Dhu Konschtat Highlands
+-- Haty Konschtat Highlands
+-- Highlander Lizard Konschtat Highlands
+-- Hippomaritimus Valkurm Dunes
+-- Khalamari Abyssea - Konschtat
+-- King Arthro Jugner Forest
+-- King Vinegarroon Western Altepa Desert
+-- Koropokkur Yuhtunga Jungle
+-- Leech King Maze of Shakhrami
+-- Mahisha Phomiuna Aqueducts
+-- Metal Shears Valkurm Dunes
+-- Meteormauler Zhagtegg Jugner Forest
+-- Meww the Turtlerider Yuhtunga Jungle
+-- Morbolger Ordelle's Caves
+-- Mysticmaker Profblix Labyrinth of Onzozo
+-- Nandi Eastern Altepa Desert
+-- Novv the Whitehearted Sea Serpent Grotto
+-- Numbing Norman West Sarutabaruta
+-- Ocean Sahagin Sea Serpent Grotto
+-- Phanduron the Condemned Bostaunieux Oubliette
+-- Qu'Vho Deathhurler Palborough Mines
+-- Rambukk East Ronfaure
+-- Roc Sauromugue Champaign
+-- Rose Garden Yuhtunga Jungle
+-- Sappy Sycamore Jugner Forest
+-- Sekhmet Attohwa Chasm
+-- Shikigami Weapon Ro'Maeve
+-- Silk Caterpillar Rolanberry Fields
+-- Simurgh Rolanberry Fields
+-- Skirling Liger Batallia Downs
+-- Slumbering Samwell La Theine Plateau
+-- Tococo South Gustaberg
+-- Tonberry Decapitator Den of Rancor
+-- Tonberry Pontifex Den of Rancor
+-- Tonberry Tracker Den of Rancor
+-- Toxic Tamlyn Pashhow Marshlands
+-- Tres Duendes Phomiuna Aqueducts
+-- Voluptuous Vilma Yuhtunga Jungle
+-- Waraxe Beak Meriphataud Mountains
+-- Weeping Willow Batallia Downs
+-- Xolotl Attohwa Chasm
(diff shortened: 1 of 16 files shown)


teschnei:
Do you think it is more obvious that the hnm timers are in a global called
timedspawns rather than their own scripts? What about the other timed spawn
NMs?  Is every timed spawn NM going to be in there?

ibm2431:
Would timed NM respawn timers being looked up from SQL or Lua global table be something worth pursuing? Taking out the hard-coded times in individual scripts altogether and making it more automatic somehow (probably requiring changes to `setRespawnTime`, or adding a new `setDefaultRespawn`).

teschnei:
Do you think that is more obvious to someone who does not know anything
about the project's architecture?

TeoTwawki:
We used to have the bits that are in `scripts\zones\zonename\zone.lua` in a global zone.lua back before multiple map servers were a thing. I personally don't think it'd be terrible to globalize that bit of lua code in some way or another if its sufficiently readable (and if it doesn't add even more duplicate code) and said so on the forum, but I did also suggest consulting @teschnei or the whole team on irc *before* doing the work so there wouldn't be a bunch of wasted effort for nothing.

Deadwing888:
Minimum re-spawn is fine, if you want to do this properly though you need to have fields for "window length" and "number of windows." The default options for this should be 7 and 3600. Then have your respawn code add minimum respawn (21hrs) + mathrandom(0, (number of windows -1) * 3600)

The onmobinitalize similarly needs to be using windows which is accomplished by doing the same math in a local variable and passing the respawn function that time for the first value, and then that time +1 second for the second value.

Tia jorm and vrtra also have windows. The default options there should be 97 and 3600.

Another annoying aspect of this respawn code is, if you kill nq's for 3 days straight odds are you killed the third one slightly less than the 72hr mark for the HQ to be open. Now on retail going into day 4's camp you would expect an HQ to be a chance. Here though it still isn't, it is not until it runs on the onmobdespawn of day 4 that it actually has a chance to pop HQ for the first time on day 5. Also retail kings could go to day 12+ so the "or kills > 6" bit is wrong.

I have mixed feelings on whether creating the timedspawns.lua is a good idea at all, but I suppose it's easier to rip it out to rewrite your own than to put it in if you maybe can't navigate the respawn code that well, so I kind of get it. In either event nothing is handling windows correctly right now, so you might as well put it in now so someone else doesn't have to come in behind you and put it in later. Lots of things have windows (cerb/khim/hydra 1hr, King Arthro 30mins, Dark Ixion 15minutes etc)

helixhamin:
@teschnei For your questions:
1. My goal is to include all, yes. As @TeoTwawki mentioned, right now I am just putting out what I have done for my own personal build to get a sense of this idea, and see if the project team finds it useful.

2. It may not make it that much more obvious for someone with no knowledge of how to do more than follow the setup guide, but it would help those who wish to adjust the timers go to a central source instead of trying to modify them script by script.

I was just hoping to simplify the overall process for those who would prefer to adjust the spawn timers.

@ibm2431 The global lookup table is what I am attempting here. Sadly, I am not yet good enough to be able to do an sql modification. I have tried to think of a way to turn them into more simple formulas, rather than having to code each, but am unsure how to do that given the wide range of spawn timers. For just "land kings" or "wyrms", it would be easy enough. I could even just merge faf/adam/behe together with a single timer, and do the same with jorm/vrtra/tia, but my thought process was what if someone wanted a specific timer for one that is not the same as the others. Trying to find a way to allow for customization for those who want such a thing. But, if it would be better to merge those with the same timers, then I could go that way, since I just set all land kings the same value and wyrms the same value anyways.

@TeoTwawki Yes, you had. But, I already had the scripts done to run on my own server, so thought I could show how it works to get feedback and see if it was worth sharing further, or if I should just do my own modifications on my local source.

helixhamin:
@Deadwing888

Sorry, I am confused. Do you mean there is an issue with what I had done, or something else? Because I had not adjusted the 
`local popNow = (math.random(1,5) == 3 or kills > 6);`

Or is this a problem no one has noticed with the way these have been working before?

teschnei:
I don't think you're going to make hundreds of globals for every single
timed NM spawn?

I'm not convinced.. it seems to make it more complex for relatively little
gain, but I'd like to hear more opinions on it.

Deadwing888:
@helixhamin Nothing is wrong with how you moved variables to the centralized source based on the code that was written previously.

The code that was written previously has several problems that it would be very nice to correct now. Putting the window stuff in now avoids it have to be rewritten later when someone finally makes these things function correctly. It could also more smoothly pave the way for future window based respawns to be added.

The other stuff I was just putting out there in hopes you or someone who cared about proper respawns would take another look at it.

@teschnei my meaningless opinion is in agreement with you, I think it was better the way it was. If it did get implemented though I can't say I'd be terribly bothered either, it is easy to rip out after all.

helixhamin:
@Deadwing888 Okay, I understand. I am still a novice, and have only been able to contribute in a limited way. This is probably the biggest contribution I am currently (trying) to work on. Unfortunately, I am unsure how to fix these myself. I am more than open to trying to though if this survives, and if you could explain how I would do it put it separately if this does not survive. I am at the level where I can follow the steps, but not where I can do that great at rewriting the actual logic...

@teschnei I think I get a couple of the thoughts here. For the NMs, I am trying to target the primary NMs that have a built in timer, not all timed mobs.  Go through the zones one by one and grab them as I can. This would only be zone specific (Those hardcoded in the zone lua files)

ibm2431:
@teschnei Just re: the hundred of globals thing, one could use just a single Lua table indexed by NM name or (potentially better) mob ID

```
NM_Timers = {
    ["Behemoth"] = {900, 75600, 10800, 72 * 3600},
    ["Fafnir"] = {900, 75600, 10800, 72 * 3600}
}
```

If it were indexed by ID you could probably do some shennigans with a new/modified `setRespawn` which fetches the mob's ID from the "calling" mob and pulls its own timers from the table.

As for whether that'd be simpler for people to edit, I'm not certain. I like the idea of a centralized location for these, but someone would have to know where to look if they wanted to edit it. But your question over ease of use makes me think that a SQL table is probably not the way to go with this.

TeoTwawki:
yet anther sql table for something easily scripted is something I am against.

helixhamin:
@ibm2431 The mob ids are in the zone. I am trying to match the name so I can check the name list (tiamat_timer is easier to understand than 16806227) so that I, as a player on my own solo server, can just go "Hey, I want Tiamat to pop quicker. So, I will just decrease these numbers". Maybe more detailed explanation in the file? I am trying to make sure it is clear which numbers to edit.

With your NM timer, how would I do it? It creates a single variable, but how would I get the lua scripts to link it? I mean, I like the idea, but just not sure how to do it.

ibm2431:
@helixhamin You can pull values from a multi-dimensional table/array with moar brackets:
`behemoth_respawn = NM_Timers["Behemoth"][2]`
Or you can skip the variable definition and just use it directly:
`mob:setRespawnTime(math.random(NM_Timers["Behemoth"][2], NM_Timers["Behemoth"][2]+NM_Timers["Behemoth"][3]))`
Optionally with a simple timer definition for better readability:
```
local behemoth_timer = NM_Timers["Behemoth"]
mob:setRespawnTime(math.random(behemoth_timer[2],behemoth_timer[2]+behemoth_timer[3]))
```

**At everyone:**
I'm getting ready for bed, so not checking exactly how everything currently works - pretend this rough pseudocode is remotely valid and potentially relevant.

nm_timers.lua:
```
NM_Timers = {
    [behemoth_id] = {900, 75600, 10800, 72 * 3600}, -- Behemoth
    [fafnir_id] = {900, 75600, 10800, 72 * 3600} -- Fafnir
}
```
In behemoth.lua:
```
UpdateNMSpawnPoint(Behemoth);
mob:setNMRespawnTime();
SetServerVariable("[PH]King_Behemoth", kills + 1);
```

Wherever this mob function would go:
```
function setNMRespawnTime()
    local nm_timer = NM_Timers[mob:getID()]
    mob:setRespawnTime(math.random(nm_timer[2], nm_timer[2]+nm_timer[3]));
end
```
Extend with pop windows, PHs, etc. But basically pull out the hard set numbers spread across all the NM scripts, and have them call some mob functions instead which pulls values from a single global table/file. Person who wants to edit NM timers can go to that file and not have to worry about (or be intimidated by, if they lack programming background) all the function calls and math.randoms.

helixhamin:
@ibm2431 Okay, I could do the first. Seems pretty simple to change, and I think the definition looks better in the code.

Can you help explain more on the at everyone part? Sorry, it is not that I am not paying attention, just that I am not good at programming logic, so not sure what to do.

You mention:
`[behemoth_id]`

Would this be the ID number from the individual files, or an actual variable called "behemoth_id"? Would this remove the id numbers hardcoded in the zone and monster files, like:
`SetRespawnTime(16806227, 86400, 259200);`
with...sorry, not sure. I get lost on this:
`local nm_timer = NM_Timers[mob:getID()]`

Not sure what would be there for the NM_Timers definition.

If it is cleaner to do it this way, then should it be something like:
`SetRespawnTime(tiamat_timer[1], tiamat_timer[2], tiamat_timer[3]);`
And add the first as the mob ID so the timedspawns.lua would show Tiamat as:
`tiamat_timer = {16806227, 86400, 259200, 432000}`
The first now being the mob ID?

Seems like it might be helpful doing that when SE starts shifting IDs around. Would it actually be better to try to centralize all lua based IDs?

Or am I just reading too much into it, and showing my ignorance of advanced coding?

ibm2431:
I was too lazy/tired at the time to look up Behemoth's actual ID that it'd be indexed by.

As far as shifting hardcoded IDs go, that really shouldn't be a problem for mobs. It's typically just "NPCs" that get moved around, and the batch tool already detects and fixes those (when it gets run).

You probably wouldn't need to include the ID of the now-killed NM itself as a value in any table (except for as a key), since mobs should already know their ID. But there might be value to storing the PH ID (in a king's table), or the king/NM ID for a PH table, so that you can use them for any DeterMob you might need. I guess that's one thing that should be considered if one were to design such a table.

helixhamin:
As per @ibm2431 recommendation on IRC, I have started a data dump of all SetRespawnTime monsters.
https://gist.github.com/helixhamin/315b9cb18e7d9511f08a626e3888b38f

I will try to break out the data further, but if anyone sees a pattern for now that might be helpful, please let me know if there might be a better way to do this.

Scavenge:
i skipped ahead but throwing my simple mind in here, isn't the system for players that wait on their relics to be finished just like what you're after here?  using epoch values as a conditional check seems like it would be enough, it would be a lot of them though, but the char vars are also numerous so it seems like it's already an accepted method

helixhamin:
While I hoped to keep this open so I could look it up easily once I had some time to see if I could redo it, I doubt I will have time for a while. Closing this.

PR #3597, opened 2016-12-24
Version update 2016/12/12 (client: 30161206_3)

ibm2431:
So! SE, for reasons known only to them, decided to add four (presumably debug) Moogles to every single zone (minus two) in the game. But not at the end - right smack dab in the middle of the zone list! The NPC ID batch did not like this at all, so I spent the past week (at least 30 hours!) going through all the IDs, verifying correct changes, and correcting the mishaps.

Here are the notes from my notepad... notes:

- I went ahead and added the four new Moogles to each zone's NPC list, so hopefully the tool won't trip up as badly next time. These Moogles were always added in a pack right after four Moogles which had already existed in the zone, and I've confirmed their existence and IDs via POLUtils.
- While I was dealing with these debug Moogles, I went ahead and made them stop showing up on Widescan.
- Removed the Widescan attribute from Achieve and Unity Masters as well.
- Funnily enough, the two zones which didn't see the Moogle additions were GM Home, and strangely, Mount Kamhir. The Moogles' purpose might have something to do only with areas where players can walk around in.
- **The batch tool did nothing to (and may not be configured properly for) the following zones: Crawlers' Nest, Crawlers' Nest [S], Ceizak Battlegrounds, Inner Ra'Kaznar, and Ra'Kaznar Turris**. The last two make sense since my config didn't even have their names listed, but I'm not certain why it ignored Crawlers' Nest of all things.
- SE completely removed the Jingle Baas NPCs they added for the event just last month. The resulting lists for the affected zones required quite a bit of fixing.
- I popped a couple CS NPCs in spots where I noticed they were missing: ROV's Tenzen in Misareux Coast, a missing Promathia in Empyreal Paradox, a Lehko Habhoka in Jugner Forest [S], etc.
- If I spotted an NPC missing from the list (according to POLUtils) and I couldn't pop it in based off of other copies of that NPC, I sometimes added a comment stating which it was (mainly when the gap and result of the tool tripped me up, so I added a note for the next person to make sure they're aware).
- If I saw something wrong (which there is still a lot of) which I felt like fixing at the time, I went ahead an fixed it. Certain NPCs not having their sensible required expansion listed (some Orcish_DRG1s in Jugner Forest [S] being null instead of 'WOTG'), expansions of 'Null' instead of null, NPCs listed out of order, Widescan working for GoalPoints, etc.
- Not matter how much comment finagling I did, I couldn't get the Text ID batch to pick up the shifted IDs. They were all shifted just one slot back, so I just did them manually until it stopped tagging them with FIXMEs. I don't know why it couldn't find the new IDs automatically.


tagban:
Great work man, even if there are issues this will fix the majority of them. Thanks for your dedication and hard work!!

wrenffxi:
Thanks for your work on this!

TeoTwawki:
> The batch tool did nothing to (and may not be configured properly for) the following zones: Crawlers' Nest, Crawlers' Nest [S], Ceizak Battlegrounds, Inner Ra'Kaznar, and Ra'Kaznar Turris. The last two make sense since my config didn't even have their names listed, but I'm not certain why it ignored Crawlers' Nest of all things.

I just double checked and there were no changes to Crawlers' Nest (Zone 197) so was nothing for the tool to adjust there.

ibm2431:
POLUtils lists Olavia as now being 0x010C515E (17584478) when we had her as 17584474, first Planar Rift as 0x010C5172 (17584498) when 17584494 before, etc.

Shift started near the end of the list for that zone, 0x010C5157+

TeoTwawki:
odd. My client is up to date and I compared my polutils list with your file when I posted.

TeoTwawki:
oh, derp. you had already manually changed it, didn't you? I thought your comments meant you left it alone

nyczducky:
I was working on this too and this shit made my head hurt. Guess I picked the wrong patch to learn ahah. Awesome job again @ibm2431 and thanks for showing me the stupid issues with the batch tool.

PR #3608, opened 2016-12-30
Reduce cast range of mobs to match retail.

nyczducky:
Cast range was previously set to 40. 

Retail is 26.8. Rounded up to 27. http://ffxiclopedia.wikia.com/wiki/Distance

Tested on worms to see if mobs correctly get interrupted if you run past 27.

### DIFF
--- a/src/map/ai/states/magic_state.cpp
+++ b/src/map/ai/states/magic_state.cpp
@@ -182,7 +182,7 @@ bool CMagicState::CanCastSpell(CBattleEntity* PTarget)
     {
         return false;
     }
-    if (distance(m_PEntity->loc.p, PTarget->loc.p) > 40)
+    if (distance(m_PEntity->loc.p, PTarget->loc.p) > 27)
     {
         m_errorMsg = std::make_unique<CMessageBasicPacket>(m_PEntity, PTarget, m_PSpell->getID(), 0, MSGBASIC_TOO_FAR_AWAY);
         return false;


src/map/ai/states/magic_state.cpp
```diff
@@ -182,7 +182,7 @@ bool CMagicState::CanCastSpell(CBattleEntity* PTarget)
     {
         return false;
     }
-    if (distance(m_PEntity->loc.p, PTarget->loc.p) > 40)
```
TeoTwawki: The intention is to alter only the mobs casting range right?
https://github.com/DarkstarProject/darkstar/pull/3608/files#diff-7fd2c48b6004c4fc5a400cdb8c98baa5R195 appears to be the part for mobs. Currently set to 28.5

wrenffxi: 28.5 is the range at which an in-progress mob spell should be canceled, and current code seems to work as expected.  But the maximum range at which a mob will begin to cast should be 26.8, and in current code because of the 40, ends up falling back on the 28.5 value also.  So I think ducky's right to change the 40 to a 26.8, but misleading in his comment that he tested to see if spells were interrupted past 27.

TeoTwawki: k. Guess I didn't understand what the existing code did. Paging @teschnei

teschnei: do not change this line - it is for the "too far away message" (as opposed to the "out of range" message, which is the next check).

teschnei: At any rate, Teo is right in this regard - the mob range check is right after the spell range check.

PR #3619, opened 2017-01-03
Cleanup the extra ()s in some of the setRespawnTime math functions.

helixhamin:
Also added in respawn time for Martinet.lua as per the 2 hour wiki respawn:
http://ffxiclopedia.wikia.com/wiki/Martinet -- 2hr Respawn

Also fixed Cerberus so the math function is the same as, well, every other timed mob...


scripts/zones/Mount_Zhayolm/mobs/Cerberus.lua
```diff
@@ -37,5 +37,5 @@ end;
 -----------------------------------
 
 function onMobDespawn(mob)
-    mob:setRespawnTime((math.random(0,24)*3600)+172800); -- 48-72 hours proper 1 hour windows
+    mob:setRespawnTime(math.random(172800,259200)); -- 48-72 hours proper 1 hour windows
```
TeoTwawki: There is a reason it wasn't like all the others. You're breaking the window timing which will make deadwing very sad.

helixhamin: lol, easier than trying to rewrite every other one to keep the same consistency xD
Also easier for my master plan, to take over the world!
...I mean, to create a separate timed spawn lua...

TeoTwawki: This was technically correct as it was though.

(and nearly every other one isn't - its just hardly anybody cares about those exact windows so nobody ever pull requested to change them)

helixhamin: Hmm, bleh, choices...
1. Change this one back, and leave them mismatched (Well, this and the zone)
2. To see if anyone wants to merge this as consistent
3. To change all others so they are consistent with windowing principles...

Meh, I guess if we wish to go with more authentic, I can window all mob:setRespawnTime instances. Fortunately, there are just a bit over 50, so it would be a little timely, but not overly.

Preference?

TeoTwawki: 1 or 3. But 3 would mean you get to look and see what each NM's mini-window sizes are and how many windows they get. Which is why 1 is an option. As much as my OCD loves consistency to death we can't take something that is correct and make it wrong so 2 is out.

(Back when I retail HNM'd, we called 48-72 "the window" and every 1 was the "mini window" it would actually pop at during that. I dunno what terms the kids use today)

helixhamin: Bleh, okay, I prefer consistence, and that gives me a better way to do the nm timers (adjust the window time, rather than the raw time), so I will work on that. Fortunately, work is dead at the moment xD

TeoTwawki: o.O
:astonished: 

@Deadwing888 merry late Xmas deadwing, looks like someone is actually gonna do it!

helixhamin: If I can get the pull accepted

scripts/zones/Batallia_Downs/Zone.lua
```diff
@@ -63,7 +63,7 @@ function onInitialize(zone)
     SetVoidwatchNPC(vwnpc);
 
     -- Ahtu
-    SetRespawnTime(17207657, 900, 10800);
+    GetMobByID(17207657):setRespawnTime((math.random(0,4) * 1800) + 900); -- Window is 0-2 hours, since no source shows the window intervals, setting to 30 mins
```
TeoTwawki: Is it for sure even that type of pop? not all of them are.

TeoTwawki: In general if we don't have a source and don't know, its usually better to leave well enough alone.

helixhamin: @TeoTwawki No, I am not sure. At this point, I am only reformatting what exists already and putting it in the proper windowed format. No documentation shows mobs like Ahtu with a different maint pop window. It could be this was an old insert that was never relevant, but I am unsure what ones should be removed to clean it up.

helixhamin:
Closing this one. I will try first to window mobs directly in their lua and do it in smaller batches.

PR #3621, opened 2017-01-03
More food .lua edits

wrenffxi:


TeoTwawki:
I'm skeptical on the anglers_cassoulet script, no other food works that way and its a bad idea to have 1 onEffectGain set off a second effect script. I think when you changed it the first time it made more sense. If it really does do 2 status effects, they should both be applied in the same place. Have you seen its effects on retail? I couldn't locate a reliable source online but a comment on lolZam implies it is auto-regen. http://ffxi.allakhazam.com/db/item.html?fitem=12759

wrenffxi:
Perfect, that's exactly what I was looking for.  Updated the duration too.  I wish I could verify some things on retail, but I haven't got an account.

PR #3622, opened 2017-01-04
Fixed issue where it would loop you back into MH when trying to zone out

LegacyFFXI:
Not sure if removing "AND pos_zone = %u" has implications elsewhere but this resolved the MH zoning issue (looped me back into MH when trying to exit).

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -4754,7 +4754,7 @@ namespace charutils
                 "pos_y = %.3f,"
                 "pos_z = %.3f,"
                 "boundary = %u "
-                "WHERE charid = %u AND pos_zone = %u;";
+                "WHERE charid = %u;";
 
             Sql_Query(SqlHandle, Query,
                 PChar->loc.destination,


src/map/utils/charutils.cpp
```diff
@@ -4754,7 +4754,7 @@ namespace charutils
                 "pos_y = %.3f,"
                 "pos_z = %.3f,"
                 "boundary = %u "
-                "WHERE charid = %u AND pos_zone = %u;";
```
ibm2431: What happens to the corresponding currZone it was passing into that query on line 4768?

TeoTwawki:
Only time I have ever seen people get looped back to the mog house on exiting the mog house was after they had done something to cause their data to be wrong - like warping out of their mog house using a GM command to avoid the "you should set you homepoint" message after job change.

I've never seen this issue ever during normal gameplay. How is it triggered?

LegacyFFXI:
Hey, Apologies if i say anything stupid :/ this is all a bit new to me. This is related to the following:
https://forums.dspt.info/viewtopic.php?f=14&t=21198
That zoning issue happens to me on a completely freshly built version of the master branch of the server, i noticed that there was a branch called zoning and a few others when trying to look into resolving this so I then merged zoning and fishing into master and built the changes using VS and at that point we get the MH loop issue. 
Only other changes made are usual config files and a couple drops added to some mobs. we get this zoning in as normal and then zoning out using the door as normal, experienced by a couple non gm's testing on the server atm and myself to replicate with gm account.

From the logic, and my very limited understanding of it... it either looks at the new zone or sets value to MH, then the set query is setting the currzone but only if currzone is not MH so then drops the SET to the new zone as the WHERE isn't met. Removing this second check on the WHERE has on my test server at least stopped this behaviour.

Sorry ibm2431, which bit are you referring to, i have that line in the next function as: 
4768     if (PWeapon && PWeapon->isUnlockable() && !PWeapon->isUnlocked())

Again, I am very new to this but would love to contribute in a positive way. I was very excited when I thought i have found something that i could actually pass back to you guys. I am a little confused though about the methodology as the set up guide says to use stable but that is a long long way behind master and it looks like other branches are sometimes finished (not been worked on for a long time) but still listed and not merged back into master/stable. I am now thinking I might be getting this zoning issue as I am doing something wrong in the setup. Could you confirm for me which branches, if any (apart from stable) should i be using/merging before doing any further testing. I really don't want to make any unnecessary work for you guys due to any silly mistakes on my part! :(

I love that this project exists, love FFXI and love the hard work you are all doing :D

helixhamin:
@LegacyFFXI you need to pull master, not the other branches. See if it still occurs after that.

teschnei:
I have never seen this happen... How do you reproduce it?

LegacyFFXI:
@teschnei to reproduce the double char added to new zone and then time out issue, this one is a bit random and can be a day before i see it happen again, i build vanilla master branch with no changes, this happens infrequently and not to all char, for instance I myself have had the issue multiple times but another GM on the server has never seen it, for the people it happens to it can be on normal zone line, going in or out of MH or using @zone GM commands, nice and easy to troubleshoot >.>

For the MH loop issue, built master with zoning branch merged and the just go into MH normally and then try to come out using the door, screen goes black and then you reappear in MH.

teschnei:
zoning is @takhlaq 's branch, so he can look into it.  At a glance, seems like there is a bigger issue and this isn't the right fix, but I don't know what he did to that branch :)

As for your other issue, I have never heard that happen to anyone else, so it may be related to your setup.  But you can make another issue if you want to discuss it further.

LegacyFFXI:
@takhlaq @teschnei Cheers, will do. for this one thing though i will stay with master with zoning and my fix merge (my g/f was the one getting the most double char d/c before and hell hath no fury like a girl gamer in rage mode).

Issue #2082, opened 2015-09-16
Avatar Perpetuation Reduction Mods


Zenny89:
MOD_PERPETUATION_REDUCTION = 346, // stores the MP/tick reduction from gear
MOD_AVATAR_PERPETUATION = 371, // stores base cost of current avatar

In item_mods.sql, Adhara Gages and Augur's Brais are using 371 as the modifier, while every other piece of reduction gear is using 346 despite having the same "Avatar Perpetuation Cost -X" attribute. Which is correct?

teschnei:
Using 371 is not correct. Only avatars themselves set and use that
modifier. It either doesn't work or continuously drains mp when you equip it

Zenny89:
Got it, I'll change them to use the 346 mod.

PR #3631, opened 2017-01-07
Summoner mods QC and reordered item_mods_pet.sql

wrenffxi:
Fixed inconsistent sign on summoner mods
Added some missing summoner pet mods
Reordered item_mods_pet.sql by itemId
QC on Evokers, Convokers, Summoners, Glyphic, Callers sets (and their upgrades)


wrenffxi:
Also related to issue #2082

Hozu:
The modifiers for pets on equipment needs rework - they apply to all pets, and there's no way to limit that to anything specific.

teschnei:
Pet latent?

Hozu:
There is a latent that checks for the pet, yes, but the mods in item_mods_latent are applied to the master. item_mods_pet is currently the only table that applies mods to pets, but it can't do latents.

teschnei:
it's not hard to add a way for latents table to add to pets - i thought we
had done that already, but apparently not

TeoTwawki:
Or do what we did with mods and have 2 tables.

PR #3634, opened 2017-01-09
abilities using target:takeWeaponskillDamage() rather than takeWeaponskillDamage()

wrenffxi:
This PR started as a fix for crashy card shots.  It has evolved. :)

### DIFF
--- a/scripts/globals/abilities/earth_shot.lua
+++ b/scripts/globals/abilities/earth_shot.lua
@@ -38,8 +38,7 @@ function onUseAbility(player,target,ability)
     dmg = dmg * applyResistanceAbility(player,target,ELE_EARTH,SKILL_MRK, (player:getStat(MOD_AGI)/2) + player:getMerit(MERIT_QUICK_DRAW_ACCURACY));
     dmg = adjustForTarget(target,dmg,ELE_EARTH);
     
-    target:takeWeaponskillDamage(player, dmg, SLOT_RANGED, 1, 0, 0); -- targetTPMult is 0 because Quick Draw gives no TP to the mob
-    target:updateEnmityFromDamage(player,dmg);
+    dmg = takeWeaponskillDamage(target, player, {}, true, dmg, SLOT_RANGED, 1, 0, nil);
     
     local effects = {};
     local counter = 1;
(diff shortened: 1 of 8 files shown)


Hozu:
https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/weaponskills.lua#L881

Looking at that, either the binding or the function should be renamed so this confusion doesn't happen.

teschnei:
takeWeaponskillDamageDontCallThis?

nothing should be calling the binding except for the lua function (and I don't see a reason why the shots are bypassing the lua function)

Hozu:
I intended to use the function when I wrote it, but it looks like I did it wrong. I assumed that wasn't the case since it was working. Sorry about that.

TeoTwawki:
`weaponskillDamageCaughtInCeilingFan()`

The instructions, they were unclear again..

wrenffxi:
All right.  That should be better.

teschnei:
tp mult and ce/ve override don't need to be defined, just pass an empty table into the function

dmg = takeWeaponskillDamage(target, player, {}, true, dmg, SLOT_RANGED, 1, 0, nil);

wrenffxi:
Third time's the charm.

TeoTwawki:
> nothing should be calling the binding except for the lua function (and I don't see a reason why the shots are bypassing the lua function)

@teschnei  Totally slipped my mind this came up before. Now I recall being grumpy about the naming myself after: https://github.com/DarkstarProject/darkstar/blob/1b705c79a11724e2b977947fa23707b98867dca0/scripts/globals/weaponskills/spirits_within.lua#L58
I put that comment there to remember which one it was already using, so the fact of it already using the wrong one ain't my fault. Also happens in Atonement.lua 

So if no place should call that binding except the lua wrapper function the same name, and its already been done wrong in several situations, we need to either fix'em all and then keep a better eye on pr's using the wrong one, or just rename one of them already. `doNotUseTakeWeaponskillDamageYouPunkKids()`? I got nuthin.. 😃

teschnei:
you could skip the call if it didn't add enmity, but you could also call it with 0 enmity multiplier.. 

I say just make everything use the lua function.  

also, I dont think it needs changing, but you could have just omit the last argument entirely if it was just going to be nil (Lua will give you nil for a missing argument)

TeoTwawki:
> but you could have just omit the last argument entirely if it was just going to be nil (Lua will give you nil for a missing argument)

In my best archer voice: Do you want ~~ants~~ messages about nils in the console long? because thats how you get ~~ants~~ messages about nils in the console log. (it'll notify you about it even when it doesn't cause a problem)

teschnei:
i think that was just for tables that you were seeing those... would be no difference in this case as they'd both evaluate to nil

wrenffxi:
There we go.  No more calls to the binding.

Issue #3589, opened 2016-12-20
Various Emotes are Crashing Clients


nyczducky:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) 30161103_1

**_Server Version_** (type `@revision` in game) 7186236

**_Source Branch_** (master/stable) master

**_Additional Information_** (Steps to reproduce/Expected behavior)
Some emotes are crashing clients. Sometimes it's just the targeted player. Othertimes, it's to everyone.

/wave and /laugh are the usual culprits.

wrenffxi:
Anyone with retail access able to capture the incoming packet when you ...

/bow {to someone / to nobody} {with no weapons / with a one-handed weapon / while dual wielding}

Would like to compare this to the one created by DSP.

Load [PacketViewer](https://github.com/z16/Addons/tree/master/PacketViewer) and do: **//pv l chatlog i 0x05A**  Then when you emote it'll show the packet in your chatlog.

nyczducky:
Logged some packets but I couldn't reproduce the crashes reliably. The crashes did happen but after like... 100 emotes.

Players have noticed that doing emotes in towns are safe while doing this in areas outside of town causes more crashes. I don't know how that makes sense but just putting that observation here.

wrenffxi:
You logged incoming 0x05A packets from retail?  If so can you post a few?  Want to either confirm or eliminate outdated packet fields as a source for this bug.  Thanks!

nyczducky:
I just resubbed. I'll have it shortly.

nyczducky:
`Packet
        |  0  1  2  3  4  5  6  7  8  9  A  B  C  D  E  F      | 0123456789ABCDEF
    -----------------------------------------------------  ----------------------
      0 | 5A 1C 96 05 24 0C 02 00 B8 CF 00 00 56 05 89 04    0 | Z...$.......V...
      1 | 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00    1 | ................
      2 | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00    2 | ................
      3 | 00 00 00 00 00 00 00 00 -- -- -- -- -- -- -- --    3 | ........--------

Packet
[...]

PR #3635, opened 2017-01-10
Update to char emote packet to match retail.

wrenffxi:
From my limited testing this seems to stop the random client crashes caused by emotes.


wrenffxi:
For issue #3589.

Brierre:
This makes me /cry tears of /joy!

PR #3640, opened 2017-01-12
2 changes, one means player can get past lvl 4 fishing, the other stops it saying player levelled up at each skillup

LegacyFFXI:
Hey,

Feel free to ignore most of these changes as i changed some of the formatting for my personal pref and some of the skill up chances to make the skill up chance and amount var modifier effect it more than it did before. Main 2 changes were:

Changing the var type of skillupamount to int32 as with it previously being double it was using floating point arithmetic to do the compare and so not discarding the remainder if the " / 10" and so always executing the code as if you levelled up the fishing skill.

Changing PChar->WorkingSkills to PChar->RealSkills meant i could get past level 4 fishing, for some reason at level 4 fishing the working skill value was 128.. not really sure what is happening here.

Cheers :)

### DIFF
--- a/src/map/utils/fishingutils.cpp
+++ b/src/map/utils/fishingutils.cpp
@@ -66,7 +66,8 @@ namespace fishingutils
 
     void LoadFishingMessages()
     {
-        zoneutils::ForEachZone([](CZone* PZone) {
+        zoneutils::ForEachZone([](CZone* PZone)
+		{
             MessageOffset[PZone->GetID()] = luautils::GetTextIDVariable(PZone->GetID(), "FISHING_MESSAGE_OFFSET");
         });
     }
@@ -278,10 +279,11 @@ namespace fishingutils
         DSP_DEBUG_BREAK_IF(PChar->UContainer->GetType() != UCONTAINER_FISHING);
         DSP_DEBUG_BREAK_IF(PChar->UContainer->GetItem(0) == nullptr);
 
-        if (special < 20) {
+		if (special < 20)
+		{
             return 0;
         }
-
+		
         CItem* PFish = PChar->UContainer->GetItem(0);
         uint16 ModID = MOD_FISH;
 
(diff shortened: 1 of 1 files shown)


takhlaq:
tabs to 4 spaces

takhlaq:
> Commits should contain a descriptive name for what you are modifying

src/map/utils/fishingutils.cpp
```diff
@@ -587,7 +599,7 @@ namespace fishingutils
                         CItem* PFish = PChar->UContainer->GetItem(0);
                         uint8 rankSkill = PChar->RealSkills.rank[SKILL_FISHING];
                         uint16 rankSkillCap = (rankSkill + 1) * 100;
-                        int32 charSkill = PChar->WorkingSkills.skill[SKILL_FISHING];
+                        int32 charSkill = PChar->RealSkills.skill[SKILL_FISHING];
```
maxtherabbit: why did you change this?

LegacyFFXI: This is something that confused me too, I looked over the crafting cpp and it followed that same format as this fishing one, in that it used RealSkills when first querying your player skill level and WorkingSkills when looking if it was to do the skillups and that obviously works in that cpp. On my server though, after fishing skill level 4 all skillups stopped. I traced this to the fact that on line 641:

`                            if (((charSkill / 10) < fishMaxSkill) && ((charSkill / 10) < (rankSkillCap / 10)))`

charSkill, which was defined with WorkinSkills had a value of 128, so the code in the if statement would never run seeing that the player is supposedly above the skill cap (10), changing this to RealSkills fixed this and pulled the correct value of 40 for that player at the time and so allowed more skillups.

I am not sure where WorkingSkills is set and where it got that value of 128 from? As in, which skill is the working skill. But I guess this must be defined somewhere as the crafting cpp knows which skill it is currently working with. sorry if my change isn't the proper methodology for the function! muddling my way through tbh

Brierre:
So does this mean you can actually catch fish at all?  I can't get fishing to do a damn thing.

LegacyFFXI:
@Brierre Tt did, though i hadn't tested to max fishing level. Also i did a rollup to my fork of the last months or so pushed commits to dspt and merged my code again and it was broken. not gotten around to looking into why yet.

Issue #3639, opened 2017-01-12
Various ilvl weapons have skills listed in item_mods and sql giving incorrect stats in game.


R3P0FFXI:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30161227_1

**_Server Version_** (type `@revision` in game) **:**
https://github.com/DarkstarProject/darkstar/commit/f0b1f5f0795fa0b057ce0636f73bcb3ade1c4b8a

**_Source Branch_** (master/stable) **:**
Master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

Tons of ilvl gear has incorrect mods in item mods

INSERT INTO `item_mods` VALUES (20612,82,228);    -- Sword skill +228
INSERT INTO `item_mods` VALUES (20612,110,228);   -- Parrying skill +228

INSERT INTO `item_mods` VALUES (20647,82,242);    -- Sword skill +242
INSERT INTO `item_mods` VALUES (20647,110,242);   -- Parrying skill +242

Theres more not going to paste them all mods should be deleted and skills should be handed through item _weapon.sql in ilvl cats.

TeoTwawki:
rode in on #3478

teschnei:
What is the question?  No weapon should show item level skill in your
skills pane.  If they are, then they are probably not inserted properly
into the db and are using a raw skill item_mod instead of the item level one

TeoTwawki:
His question got answered and now its just pointing out the mistake happened.

I answered it by asking "are you sure you checked that the skill wasn't mistakenly put in item mods? coz its no supposed to be there".

And "awww f\*ck" followed :smile: and then I looked up which merge it was in.

So now just needs someone to pr moving those from mods to item levels, or just deleting if they already have the item level part.

wrenffxi:
I've been slowly working through item mods by itemId high to low.  Had no idea the iLvl weapon and parry skill weren't supposed to be mods.  I'll go back and fix the ones I've done, and keep this in mind going forward.

TeoTwawki:
@wrenffxi thanks much wren, we appreciate the pull request's.

teschnei:
the item level dependent skill, parry and magic accuracy all have their own
[...]

PR #3641, opened 2017-01-13
weapon iLvl mods fix

wrenffxi:
Removed iLvl-based weapon skill, parry, and MACC mods from item_mods.sql
Various iLvl mod QC
Added Abrasion Bolt lua

For issue #3639 .


wrenffxi:
For issue #3639 .

PR #3651, opened 2017-01-18
Fixed an issue causing certain 'cannot equip' type armor to not also unequip their remove slot when equipped

ThisIsPatrick:
Equipping armor with a 'cannot equip footgear' prefix would cause feet to not unequip.
This was due to the item_armor.sql rslot for feet being 256 and the uint8 max being 255.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -1683,7 +1683,7 @@ namespace charutils
         {
             auto removeSlotID = PItem->getRemoveSlotId();
 
-            for (auto i = 0; i < sizeof(removeSlotID) * 8; ++i)
+            for (auto i = 0; i < SLOT_FEET + 1; ++i)
             {
                 if (removeSlotID & (1 << i))
                 {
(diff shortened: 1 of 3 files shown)


src/map/utils/charutils.cpp
```diff
@@ -1683,7 +1683,7 @@ namespace charutils
         {
             auto removeSlotID = PItem->getRemoveSlotId();
 
-            for (auto i = 0; i < sizeof(removeSlotID) * 8; ++i)
+            for (auto i = 0; i < SLOT_FEET + 1; ++i)
```
teschnei: probably make that "i <= SLOT_FEET", makes a bit more sense than SLOT_FEET + 1

on the other hand, maybe don't change it at all, in case something ever removes belts or backs or something?

ThisIsPatrick: I'll take your second bit of advice on that, and since I don't know how to recommit this pull request with the changes I'll make a new pull request.

TeoTwawki:
Just so you know, you never have to close a pull request and open a new one. Any change you make to the same branch on the same fork automatically changes the pull request for you as long as you remember to send it to the remote.

PR #3654, opened 2017-01-19
Add MACC% mods for food scripts to use. 

TeoTwawki:
I need someone to verify I did this right. Am not entirely sure at what step retail checks for the % buff from food or that I should not be using on of the other values in this function to add this into. 

If its good as is just merge and then we can fix the other food items that need this mod.


scripts/globals/magic.lua
```diff
@@ -403,6 +403,10 @@ function getMagicHitRate(caster, target, skillType, element, percentBonus, bonus
 
     magicacc = magicacc + bonusAcc;
 
+    -- Add macc% from food
+    local maccFood = caster:getMod(MOD_FOOD_MACCP)
+    magicacc = magicacc + utils.clamp(maccFood, 0, MOD_FOOD_MACC_CAP);
```
wrenffxi: If MACCP is 21%, isn't this just adding 21 instead of multiplying by 1.21?

TeoTwawki: somehow I did not push "save" in my editor before making the commit

TeoTwawki: that line with the var was supposed to be `local maccFood = magicacc * (caster:getMod(MOD_FOOD_MACCP)/100);`

scripts/globals/magic.lua
```diff
@@ -403,6 +403,10 @@ function getMagicHitRate(caster, target, skillType, element, percentBonus, bonus
 
     magicacc = magicacc + bonusAcc;
 
+    -- Add macc% from food
+    local maccFood = magicacc * (caster:getMod(MOD_FOOD_MACCP)/100);
+    magicacc = magicacc + utils.clamp(maccFood, 0, MOD_FOOD_MACC_CAP);
```
santssoft: Will not be?
magicacc = magicacc + utils.clamp(maccFood, 0, caster:getMod(MOD_FOOD_MACC_CAP));

TeoTwawki: Derp, forgot to getMod the 2nd mod before using it.

wrenffxi:
Thanks!  I added the other foods that use this mod to the open Crepes PR.

TeoTwawki:
Removed the crepe script since your other pr already does that (I had forgotten)

PR #3655, opened 2017-01-19
Fixed an issue causing certain 'cannot equip' type armor to not unequip their remove slot when equipped

ThisIsPatrick:
**Re-upload of #3651 with changes suggested by @teschnei** 

Equipping armor with a 'cannot equip footgear' prefix would cause feet to not unequip.
This was due to the item_armor.sql rslot for feet being 256 and the uint8 max being 255.


teschnei:
You just had to push another commit to the branch you were requesting to
merge, but this works too

PR #3675, opened 2017-01-27
Remove wyrm fake 2 hours from mob skill list

Kushdr:
Stops the 2hr ability spam and allows for other mob skills and flying as expected


GMYasiinEra:
this works 100% Era done this about a week ago cuz i was tired of jorm and tiamat standing there doing nothing.. theyre actually far more aggressive and the fight overall becomes about 50% tougher.. we even went ahead and reverted a Magic evasion that was custom added  for for the reason of them actually just sitting there and spamming 2hr

takhlaq:
need you to rebase, looks like a bunch of commits from main repo made it in

TeoTwawki:
You don't need to close a pull request to fix it...

PR #3681, opened 2017-01-30
QC for mods with itemId >= 28200

wrenffxi:


sql/item_mods.sql
```diff
@@ -42266,6 +42347,7 @@ INSERT INTO `item_mods` VALUES (28239,29,2);
 INSERT INTO `item_mods` VALUES (28239,31,47);
 INSERT INTO `item_mods` VALUES (28239,68,33);
 INSERT INTO `item_mods` VALUES (28239,384,41);
+INSERT INTO `item_mods` VALUES (28239,411,10);    -- Laksamana's Bottes: Quick Draw damage +10
```
TeoTwawki: Oddly bg wiki's own source link to a post on their forums seems to indicate the bonus is 5 and not 10
https://www.bluegartr.com/threads/108197-Random-Facts-Thread-Abilities?p=5956830#post5956830

wrenffxi: Hm.  I see that.  Can't find any other source to confirm the +10, so I'm setting it back to 5 in this commit.

PR #3684, opened 2017-01-31
QC for foods (F) and extended usables itemId range

wrenffxi:
QC for foods (F)
extended range of CItemUsables to 0x1FFF since 0x2000 starts puppet items
filled in missing foods in this previously unusable range


TeoTwawki:
@teschnei this reminds me, The range for weapons/armors changed since that block of code was written. The line where it switches over moved from 0x5FFF to 0x5BFF I think but I don't have verification on that so leaving it alone for now.

wrenffxi:
Looks like I missed a line, too.  New commit incoming.

PR #3686, opened 2017-02-01
Fix mods for itemid<18000

Fizztastic:
Adds missing mods and fixes existing mods for items under itemid 18000 including pet mods and latent effects.

### DIFF
--- a/sql/item_latents.sql
+++ b/sql/item_latents.sql
@@ -715,7 +715,6 @@ INSERT INTO `item_latents` VALUES(17719, 366, -2, 15, 2); -- DMG 47~39, decrease
 INSERT INTO `item_latents` VALUES(17719, 366, -2, 15, 3); -- DMG 47~39, decreases by 2 for every person in party (excl. wearer)
 INSERT INTO `item_latents` VALUES(17719, 366, -2, 15, 4); -- DMG 47~39, decreases by 2 for every person in party (excl. wearer)
 INSERT INTO `item_latents` VALUES(17719, 366, -2, 15, 5); -- DMG 47~39, decreases by 2 for every person in party (excl. wearer)
-INSERT INTO `item_latents` VALUES(17719, 366, -2, 15, 6); -- DMG 47~39, decreases by 2 for every person in party (excl. wearer)
 
 -- -------------------------------------------------------
 -- Garuda's Sickle
@@ -1544,11 +1543,9 @@ INSERT INTO `item_latents` VALUES(13288, 374, 10, 2, 75); -- "Cure" potency +10%
 
 -- -------------------------------------------------------
 -- Fencer's Ring
-/*
-    [NOT CODED] Latent effect: Sword enhancement spell damage +5
-*/
+-- Latent effect: Sword enhancement spell damage +5
 -- -------------------------------------------------------
--- INSERT INTO `item_latents` VALUES(13290, ???, 5, 2, 75); -- Sword enhancement spell damage +5 while HP <=75% and TP <=100%
+INSERT INTO `item_latents` VALUES(13290, 432, 5, 2, 75); -- Sword enhancement spell damage +5 while HP <=75% and TP <=100%
 
 -- -------------------------------------------------------
 -- Guardian's Ring
@@ -2706,8 +2703,8 @@ INSERT INTO `item_latents` VALUES(17814, 23, 10, 52, 6); -- Attack +10 in Thunde
 -- -------------------------------------------------------
 -- Desert Boots
 -- -------------------------------------------------------
-INSERT INTO `item_latents` VALUES(14166, 169, 12, 52, 2); -- movement speed +12% in Earth weather
-INSERT INTO `item_latents` VALUES(14167, 169, 12, 52, 2); -- Desert Boots +1
+INSERT INTO `item_latents` VALUES (14166, 169, 12, 52, 2); -- movement speed +12% in Earth weather
+INSERT INTO `item_latents` VALUES (14167, 169, 12, 52, 2); -- Desert Boots +1
 
 -- -------------------------------------------------------
 -- Marabout Sandals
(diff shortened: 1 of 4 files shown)


sql/item_latents.sql
```diff
@@ -711,11 +711,11 @@ INSERT INTO `item_latents` VALUES(17720, 366, 1, 15, 6); -- DMG 36~41, increases
 -- -------------------------------------------------------
 -- Mensur Epee
 -- -------------------------------------------------------
-INSERT INTO `item_latents` VALUES(17662, 366, -2, 15, 2); -- DMG 47~39, decreases by 2 for every person in party (excl. wearer)
-INSERT INTO `item_latents` VALUES(17662, 366, -2, 15, 3); -- DMG 47~39, decreases by 2 for every person in party (excl. wearer)
-INSERT INTO `item_latents` VALUES(17662, 366, -2, 15, 4); -- DMG 47~39, decreases by 2 for every person in party (excl. wearer)
-INSERT INTO `item_latents` VALUES(17662, 366, -2, 15, 5); -- DMG 47~39, decreases by 2 for every person in party (excl. wearer)
-INSERT INTO `item_latents` VALUES(17662, 366, -2, 15, 6); -- DMG 47~39, decreases by 2 for every person in party (excl. wearer)
+INSERT INTO `item_latents` VALUES(17719, 366, -2, 15, 2); -- DMG 47~39, decreases by 2 for every person in party (excl. wearer)
```
TeoTwawki: Would you mind adding a correction here https://github.com/DarkstarProject/darkstar/blob/79e3d0f9d21e39c439bb8b07a73a9514dc907257/sql/item_weapon.sql#L1362 for the same item with a comment on the line letting future editors know why? That has its base damage beginning at 39 instead of 47 and then the latent dropping it to 27

sql/item_latents.sql
```diff
@@ -2777,22 +2776,298 @@ INSERT INTO `item_latents` VALUES(14646, 29, 10, 32, 0); -- Darksday: MDB+10
 -- -------------------------------------------------------
 -- LAVALIER
 -- -------------------------------------------------------
-INSERT INTO `item_latents` VALUES (10961,60,-40,13,14); -- CHARMED:LIGHTRES
-INSERT INTO `item_latents` VALUES (10961,61,-40,13,14); -- CHARMED:DARKRES
+INSERT INTO `item_latents` VALUES(10961,60,-40,13,14); -- CHARMED:LIGHTRES
```
TeoTwawki: Please don't remove the space. these liens were otherwise unchanged, no reason to take that out. Its in the formatting mysql natively dumps to.

Fizztastic: I did a find and replace on that space because the rest of the file doesn't have it but I'll put them back.

sql/item_latents.sql
Fizztastic: Good eye on that one 👍

sql/item_latents.sql
TeoTwawki: I've been actively trying to get our sql files to consistently be the standard output instead of whatever random thing ppl did or poorly configured tool they used (navicat should die in a fire using all those single quotes as kindling!)

sql/item_latents.sql
```diff
@@ -686,18 +682,22 @@ INSERT INTO `item_latents` VALUES(16238, 369, 1, 13, 3);
...
+-- -------------------------------------------------------
 INSERT INTO `item_latents` VALUES(17952, 366, 2, 16, 3);
 INSERT INTO `item_latents` VALUES(17952, 366, 2, 16, 4);
 INSERT INTO `item_latents` VALUES(17952, 366, 2, 16, 5);
+INSERT INTO `item_latents` VALUES(17952, 366, 2, 16, 6);
 
 -- -------------------------------------------------------
 -- Company Sword
 -- -------------------------------------------------------
-INSERT INTO `item_latents` VALUES(17662, 366, 2, 15, 2); -- DMG 46~54, increases by 2 for every person in party (excl. wearer)
```
TeoTwawki: eh? shouldn't this be kicking in with a 2 member party? looking at latent_effect_container.cpp it looks for `GetConditionsValue() <= members` so the comment about excluding wearer would just mean doesn't kick in with party of one

TeoTwawki: looks like it should hit the cap of 54 before party if full as well.

TeoTwawki: nevermind, wikis don't state same thing, ffxiclopedia says "over 2" and bg doesn't - its right as you have it pretty sure..

Fizztastic: that comment is also probably wrong because with a party of 6 the final damage should be 56 (46+5*2).

TeoTwawki: crap. it does cap out at 54 in every source. somethings off. somebody failed at math somewhere and its been repeated since.

http://ffxiclopedia.wikia.com/wiki/Company_Sword
```
Damage on Company Sword increases by 2 for each party member over 2, up to 54 for a party of 6 people.
```
https://www.bg-wiki.com/bg/Company_Sword
```
This weapon's DMG rating increases by 2 for each party member in the player's party excluding themself.
```
http://www.ffxiah.com/item/17662/company-sword
```Damage on Company Sword increases by 2 for each party member over 2, up to 54 (DPS: 12.27) for a party of 6 people.
```

TeoTwawki: The only way this makes sense is if it does start at 3, and ends at 5 instead of 6.

Fizztastic: Looking at every edits made in FFXIclopedia, the consensus seems to be that it starts at 3 characters in the party but it's really all over the place. 

http://ffxiclopedia.wikia.com/wiki/Company_Sword?oldid=195396
Company Sword has a hidden effect, activated when in a party of 3 or more. For every member in the party after the second, add 2 to the Base Damage portion of the damage calculation. 

http://ffxiclopedia.wikia.com/wiki/Company_Sword?oldid=354561
Damage on Company Sword increases by 2 for each party member over 2, up to 51 for a party of 6 people. 

http://ffxiclopedia.wikia.com/wiki/Company_Sword?oldid=389024
Damage on Company Sword increases by 1 for each party member over 2, up to 56 for a party of 6 people. 

http://ffxiclopedia.wikia.com/wiki/Company_Sword?oldid=461658
Damage on Company Sword increases by 2 for each party member over 2, up to 54 for a party of 6 people.

Fizztastic: Assuming BG-Wiki is wrong (it's also the only place that says for every party member excluding wielder) and it does cap at 54 (like multiple sources say it does) then my edit is correct.

TeoTwawki: Sometimes wiki pages are like Chinese whispers 😖

PR #3689, opened 2017-02-01
Foods (G-L)

wrenffxi:


scripts/globals/items/chunk_of_homemade_cheese.lua
```diff
@@ -39,7 +39,6 @@ end;
 
 function onEffectGain(target,effect)
     target:addMod(MOD_HP, 10);
-    target:addMod(MOD_DEF, 40);
```
TeoTwawki: missed nuking the comment above :)

wrenffxi: nuked

PR #3692, opened 2017-02-03
Fix item mods with id >18001 and <23000

Fizztastic:
Adds missing mods and fixes existing mods for itemid between 18001 and 23000 including pet mods and latent effects.
On a side note, (Can I have it?) (Mod) Physical Damage taken II, "Cure" Potency II & Subtle Blow II?

### DIFF
--- a/sql/item_mods.sql
+++ b/sql/item_mods.sql
@@ -21627,6 +21627,7 @@ INSERT INTO `item_mods` VALUES (18013,431,1);
 INSERT INTO `item_mods` VALUES (18018,55,-7);
 INSERT INTO `item_mods` VALUES (18018,56,7);
 INSERT INTO `item_mods` VALUES (18018,431,1);
+INSERT INTO `item_mods` VALUES (18019,421,10);    -- xs_knife Critical Hit Damage+10
 INSERT INTO `item_mods` VALUES (18021,10,4);
 INSERT INTO `item_mods` VALUES (18021,23,8);
 INSERT INTO `item_mods` VALUES (18022,5,11);
@@ -21746,9 +21747,11 @@ INSERT INTO `item_mods` VALUES (18114,73,3);
 INSERT INTO `item_mods` VALUES (18115,25,2);
 INSERT INTO `item_mods` VALUES (18115,73,1);
 INSERT INTO `item_mods` VALUES (18116,25,3);
+INSERT INTO `item_mods` VALUES (18116,73,2);      -- Store TP+2
 INSERT INTO `item_mods` VALUES (18118,431,1);
 INSERT INTO `item_mods` VALUES (18119,431,1);
 INSERT INTO `item_mods` VALUES (18121,288,3);
+INSERT INTO `item_mods` VALUES (18121,73,3);      -- Store TP+3
 INSERT INTO `item_mods` VALUES (18122,8,2);
 INSERT INTO `item_mods` VALUES (18122,12,-1);
 INSERT INTO `item_mods` VALUES (18122,13,-1);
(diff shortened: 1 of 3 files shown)


Hozu:
PDT-% II, assuming it's like MDT-% II, should be easy to copy. https://github.com/DarkstarProject/darkstar/blob/c60730e8ddd0d9877a16bbac0ad1836136523430/src/map/utils/battleutils.cpp#L4207

Ninjistix:
is this still being checked?

TeoTwawki:
@Ninjistix I think you mean "is it still being updated?".

It has sat broken for awhile now. Its waiting on the other pull request I think, which also needs changes before it can be merged.

sql/item_mods.sql
```diff
@@ -21955,56 +21961,56 @@ INSERT INTO `item_mods` VALUES (18281,356,57); -- Stage 4 Ragnarok - Scourge
...
+INSERT INTO `item_mods` VALUES (18295,506,80); -- Extra DMG Chance 8%
 INSERT INTO `item_mods` VALUES (18295,507,200); -- Extra DMG X2
 INSERT INTO `item_mods` VALUES (18299,356,121); -- Stage 4 Gugnir - Geirskogul
 INSERT INTO `item_mods` VALUES (18300,25,20); -- Gungnir LV75 - ACC +20
 INSERT INTO `item_mods` VALUES (18300,355,121); -- Geirskogul
 INSERT INTO `item_mods` VALUES (18300,431,1); -- Additional Effect - scripts\globals\items\gungnir.lua
-INSERT INTO `item_mods` VALUES (18300,506,50); -- Extra DMG Chance %5
-INSERT INTO `item_mods` VALUES (18300,507,200); -- Extra DMG X2
+INSERT INTO `item_mods` VALUES (18300,506,70); -- Extra DMG Chance 7%
+INSERT INTO `item_mods` VALUES (18300,507,250); -- Extra DMG X2.5
```
helixhamin: Is this correct? There is a 7% change that 2.5x extra damage will be done. Here, we have an extra damage proc && an occasionally does x times normal damage. These should not be separate, one depends on the other.

I haven't looked over the other relics, but assume this problem may be in other relics.

From BG: Hidden Effect: Does 2.5x Damage on the first swing of an attack round 7% of the time.

sql/item_mods.sql
```diff
@@ -21955,56 +21961,56 @@ INSERT INTO `item_mods` VALUES (18281,356,57); -- Stage 4 Ragnarok - Scourge
...
 INSERT INTO `item_mods` VALUES (18301,25,25); -- Gungnir LV80 - ACC +25
 INSERT INTO `item_mods` VALUES (18301,355,121); -- Geirskogul
 INSERT INTO `item_mods` VALUES (18301,431,1); -- Additional Effect - scripts\globals\items\gungnir.lua
-INSERT INTO `item_mods` VALUES (18301,506,50); -- Extra DMG Chance %5
-INSERT INTO `item_mods` VALUES (18301,507,200); -- Extra DMG X2
+INSERT INTO `item_mods` VALUES (18301,506,70); -- Extra DMG Chance 7%
+INSERT INTO `item_mods` VALUES (18301,507,250); -- Extra DMG X2.5
```
helixhamin: Same as row 21998.

sql/item_mods.sql
TeoTwawki: We've had a lot of debates about this (search old issues) which is why I'm sticking the others with reviewing this one

helixhamin: Seems like it would be something better under latent, rather than item mods, since you can build in the situational in latents. But that is just my two cents worth.

takhlaq:
need you to reopen once rebased and conflicts are resolved
also has the pr this depends on been merged yet?

TeoTwawki:
@takhlaq github won't allow him to reopen after the branch has changes, so he'll be forced to make a new PR. which was why we left it open so long.

takhlaq:
ah right, guess i'll label future conflicts instead

PR #3694, opened 2017-02-04
Modifiers for PDT II, "Subtle Blow" II and "Cure" potency II

Fizztastic:


src/map/utils/battleutils.cpp
```diff
@@ -2020,7 +2020,7 @@ namespace battleutils
             if (giveTPtoVictim)
             {
                 //account for attacker's subtle blow which reduces the baseTP gain for the defender
-                float sBlowMult = ((100.0f - dsp_cap((float)PAttacker->getMod(Mod::SUBTLE_BLOW), 0.0f, 50.0f)) / 100.0f);
+                float sBlowMult = ((100.0f - (dsp_cap((float)PAttacker->getMod(Mod::SUBTLE_BLOW), 0.0f, 50.0f) + dsp_cap((float)PAttacker->getMod(Mod::SUBTLE_BLOW_II), 0.0f, 25.0f))) / 100.0f);
```
Hozu: Is Subtle Blow II supposed to cap at +25? As in, you can't do something like, 40 Subtle Blow + 35 Subtle Blow II to hit the limit of 75? Same deal later on in this file, line 2141.

Fizztastic: Do you think it's a good assumption to think that it would work the same as PDT/MDT where MDT+MDT II caps at 87.5%?

Fizztastic: Regarding Cure Potency II, the info I found seems to suggest that there's no cap so I'll be removing that check. 
`Equipment that has "Cure potency II" as a stat will increase beyond the 50% cap. `

There is also another known cap that's not currently being enforced but adding that to all the healing spells shouldn't be too much work.
`The casted and received potency caps are separate and multiply with each other for a maximum possible potency multiplier of 1.95 (1.5 x 1.3).`

teschnei:
Cure potency is enforced (of course), it's just the cure received that's
not capped right now (it caps at 30%), but they do not have a combined cap
if that's what you were insinuating

Fizztastic:
I was only insinuating about the Cure Potency received.

src/map/utils/battleutils.cpp
```diff
@@ -4236,6 +4238,8 @@ namespace battleutils
         resist = 1.0f + ( floor( 256.0f * ( PDefender->getMod(Mod::DMGPHYS) / 100.0f ) ) / 256.0f )
                       + ( floor( 256.0f * ( PDefender->getMod(Mod::DMG)     / 100.0f ) ) / 256.0f );
         resist = dsp_cap(resist, 0.5f, 1.5f); //assuming if its floored at .5f its capped at 1.5f but who's stacking +dmgtaken equip anyway???
+        resist = resist + (floor(256.0f * (PDefender->getMod(Mod::DMGPHYS_II) / 100.0f)) / 256.0f);
```
teschnei: physical damage taken II is additive, not multiplicative with physical damage taken

scripts/globals/magic.lua
```diff
@@ -205,6 +205,8 @@ function getCureFinal(caster,spell,basecure,minCure,isBlueMagic)
         potency = 1.5;
     end
 
+    potency = potency + (caster:getMod(MOD_CURE_POTENCY_II) / 100);
```
teschnei: cure potency II is also additive with cure potency

Hozu: Erm, it's done that way. Cure potency is added and capped at 1.5 above this.

teschnei:
ah, oops

PDT still needs to be additive, though (makes me wonder if MDT is)

Hozu:
Oh damn, I guess I did MDT-% II wrong way back when I added it... damn it. The way Subtle Blow II was done is how it should be.

teschnei:
oh actually, MDT II is multiplicative.. lol

PR #3699, opened 2017-02-06
Added warp ring to conquest vendors for all 3 nations

ffxijuggalo:
Added missing code for the warp ring to become purchasable.
Tested and works for all 3 nations.
Also tested the ring itself and was able to warp to set home point after each successful purchase.


scripts/globals/conquest.lua
```diff
@@ -95,7 +95,8 @@ BastInv = {0x80A1,0x000A,0x1055,0x80A0,0x0007,0x1056,0x80A3,0x2328,0x3CB5,
            0x8062,0x7D00,0x34F6,0x8060,0x7D00,0x3E55,0x8061,0x7D00,0x458F,
            0x8072,0x9C40,0x3355,0x8071,0x9C40,0x3638,0x8070,0x9C40,0x36BF,
            0x8080,0xBB80,0x419F,0x8081,0xBB80,0x4431,0x8083,0xBB80,0x44F7,
-           0x8082,0xBB80,0x4714,0x8090,0xDAC0,0x385D,0x80A4,0x1388,0x44B0};
+           0x8082,0xBB80,0x4714,0x8090,0xDAC0,0x385D,0x80A4,0x1388,0x44B0,
+		   0x80A8,0x1388,0x6F7C};
```
Hozu: Hate to be that guy, but you got uneven spacing in here due to a couple tabs.

ffxijuggalo: yeah, i'm working on it. Notepad++ did that when I hit enter after adding the comma on the end of line 98. Trying to refresh my memory on fixing bad commits without having to create another...

TeoTwawki: It called an amend. But now you already have 2 commits so it'll take a squash/rebase operation.

To avoid similar tab mishaps in the future in notepad++ got to **`view`** then **`show symbol`** and then **`show whitespace and tab`** and they will always be visible.

ffxijuggalo: would it be easier to just close this pull request and start another one with the spacing corrected?

helixhamin: Probably better to just keep trying until you get it. Then, in the future, you will know how to resolve it.

ffxijuggalo:
this better? lol
git sure is frustrating sometimes

Issue #3642, opened 2017-01-14
Wanderers have too few HP.


wrenffxi:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30161206_3

**_Server Version_** (type `@revision` in game) **:** f00420430f39fcb6b21224070882738b62454e6c

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

Wanderers in Promy Holla/Dem/Mea/Vahzl have too few HP.  You can easily one-shot VT Wanderers with a single lower tier nuke for crazy XP.

This query reveals the HP/level ratio for all mob groups.  http://pastebin.com/raw/HKstsVWe

It shows Wanderer HP:level is ~5 in Holla/Mea/Dem, and ~8 in Vahzl.  Other XP mobs at their respective levels have ratios of ~25 and ~35.  Wanderers have roughly 20% of the HP of their same-leveled cohort.

I suspect this might be a side effect of reducing HP on Strays, which share a mob family with Wanderers.  I'd recommend setting mob_family_system.hp = 100 where familyid = 255, but this would also give Strays 5x health, and I don't know if that's retail accurate.

Hozu:
Probably should change the existing family to just Stray, and add a new one for Wanderers that is the exact same except for the HP. I can't remember Strays and Wanderers link, and if they do, if they link with each other. If so, there's a place to set different mob families linking with each other.

Hozu:
This issue is fixed now, no?

TeoTwawki:
I forgot to close

PR #3705, opened 2017-02-09
Increased Wanderer hpScale to match other Promyvion mobs.

wrenffxi:
Fix for issue #3642.


TeoTwawki:
Looks logical if anyone objects they can show us the alternative.

PR #3717, opened 2017-02-14
Update some issues with DNC Horos gear

helixhamin:


helixhamin:
I used the 
`    AMNESIARES                = 253, // Enhances "Resist Amnesia" effect`
from modifier.h file, but it mentions at the top of that section:

>     // Resistances to enfeebles - Traits/Job Ability

So not sure if this is correct. If not, then there is no gear resist amnesia enabled, since the 253 mod is the only amnesia mod.

TeoTwawki:
Gear use the same mod as traits there.

helixhamin:
Sooo, all good then?

TeoTwawki:
Good cept for dupe and the potentially confusing to the next person commenting.  Just put what the mod is, like all the pre-existing comments do. And if you click "details" travis says what it broke on https://travis-ci.org/DarkstarProject/darkstar/builds/201417009#L2384

helixhamin:
@TeoTwawki Yep, checked travis, figured it out. Was supposed to integrate the item id by 1...

I also removed the extra comment and left it as just the mod effect note.

TeoTwawki:
"Should also get mp+19" Should? are you not sure or what lol

helixhamin:
o.O
Should as in "The sources confirm they should get it, but it was not originally in the script, so I included and noted".
I will change to just the value and drop the other part.

PR #3719, opened 2017-02-14
made SAM AF3 battlefield demons immune to sleep and silence (again).

wrenffxi:
This was originally corrected #3580, then undone by later commits.


TeoTwawki:
Likely my fault, sorry

PR #3721, opened 2017-02-14
Added missing Aqueduct Spider drops

ffxijuggalo:
This spider didn't have any drops listed, so I added those in.
Also gave it retail variables according to:

http://ffxiclopedia.wikia.com/wiki/Aqueduct_Spider

https://www.bg-wiki.com/bg/Aqueduct_Spider

http://ffxidb.com/zones/27/aqueduct-spider

Lots of research went into making sure this spider is as accurate as possible. But if I happened to overlook something, please let me know so I can get it fixed. The clothcrafters have been antsy for these water spider webs so enjoy!

Updated this PR to include retail variants for Mourning Crawler, King Crawler, and Champion Crawler. All of their drops have been accurately fixed/added, referencing those same websites.


helixhamin:
Hate these spiders. Was merrily running through on my uncapped COP area server, walzing right by everything, and bam! They hit me... Didn't even have time to type !godmode...

PR #3726, opened 2017-02-15
Fe'e attack speed tied to its health.

wrenffxi:
Fe'e (Up in Arms kraken NM) is supposed to attack really fast at the start of the fight, and slow down as it takes damage.  I eyeballed these delay values from watching YouTube vids of the fight.

### DIFF
--- a/scripts/zones/Waughroon_Shrine/mobs/Fe_e.lua
+++ b/scripts/zones/Waughroon_Shrine/mobs/Fe_e.lua
@@ -3,12 +3,37 @@
 --  MOB: fe_e
 -----------------------------------
 
-
 -----------------------------------
 -- onMobSpawn Action
 -----------------------------------
 
 function onMobSpawn(mob)
+end;
+
+-----------------------------------
+-- onMobFight Action
+-----------------------------------
+
+function onMobFight(mob,target)
+
+    local hpp = mob:getHPP();
+
+    if hpp < 14 then
+        mob:setDelay(2700);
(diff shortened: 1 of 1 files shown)


Deadwing888:
This is not even close to correct. 

First of all delay 2700 is 45 seconds. 

Second of all, delay does not actually change as the fight progresses. Delay is constant with him hitting several times per round at the start, then progressively he will lose tentacles (notified through messages) and each time he loses one he hits for less one less attack per round and hits slightly harder. 

At 34% health he switches to an un-named ink breath attack with minor knockback on it.

wrenffxi:
I tested the code, so can say for sure that using the setDelay(2700) isn't 45 seconds between attacks.

Otherwise, completely agree with you.  This isn't retail accurate.  It's just making it swing fast to start, and get slower at six points during the fight.  As it is now the fight is trivial, so wanted to get a fix in quickly to prevent abuse.

Unsure how to adjust number of attacks per round dynamically.  Any suggestions?

Deadwing888:
There's some sort of mobmod for multi attacks. I wouldn't know though because its not on the core I usually work with. Is setDelay now tenths of a delay? It didn't used to be... Not sure why it would ever need to be lol

The big thing to include with this if you're going to fudge it is he needs to hit harder as his attacks slow down.

wrenffxi:
I'm going to close this and implement a quick fix on our server that I'll tweak over time.  If I land on something more closely matching retail, I'll submit it here.

PR #3728, opened 2017-02-17
cleanly removing wyrm 2hrs so they act correctly

Kushdr:
Because rebasing that was way more complicated than doing this


Kushdr:
@takhlaq @TeoTwawki It was way faster to do this than to figure out rebasing on my phone. Same changes but without the mess

TeoTwawki:
@Kushdr all ya had to do was force push this over that same branch though, github automatic update steh PR to match

Kushdr:
Not a git guru so I was unaware. And I was doing it on my phone from work, so I wasn't about to really mess around with it

TeoTwawki:
> Not a git guru 

Me neither. Here have a link. https://git-scm.com/docs/git-push

Theres also clickable things all over for users of tortoisegit who don't wanna command line it.

PR #3734, opened 2017-02-20
Adjust armor mods for mnk specific armor.

helixhamin:
Add dodge effect + to Anchorite's Gaiters +2

Add Magic defense bonus to Anch. Gloves +1

Adjusted Mel. Gaiters & Mel. Gaiters +1  to 5%. ffxicyclopedia states 10% for nq gaiters, but only 5% for +1 gaiters. bg shows 5% for both nq and +1, so going with bg information.

Add magic defense bonus to Hes. Gloves +1

Change the incorrect str and vit + for Tantra Tathlum

Tantra Gaiters +2: Unsure if the base is 256 base or 1024 base. If it is 256 base, tantra gaiters were using the wrong number. If it is 1024 base, then gaiters +1 is using the wrong number. Going with base 256, since this is what is present in bg-wiki.

Bhikku Gaiters/+1: Fix many issues with the item mods applied.

Bhikku Gloves /+1: Someone put tactical guard mod as guard. Since there is no tactical guard mod currently, removed the incorrect mod.

Bhikku Crown /+1: Someone put the perfect counter att bonus as a standard att bonus. It has been properly fixed.


helixhamin:
Sigh, okay, is this pull conflicting?
https://github.com/DarkstarProject/darkstar/pull/3692
Or something else? I rebased, but still not resolved because I am current with master...

sql/item_mods.sql
```diff
@@ -3598,7 +3598,7 @@ INSERT INTO `item_mods` VALUES (11145,1,25);
 INSERT INTO `item_mods` VALUES (11145,8,8);
 INSERT INTO `item_mods` VALUES (11145,25,8);
 INSERT INTO `item_mods` VALUES (11145,386,45);
-INSERT INTO `item_mods` VALUES (11145,429,52);
+INSERT INTO `item_mods` VALUES (11145,429,38);    -- Increases Footwork's attack modifier to +38/256
```
helixhamin: Can anyone tell me if the divisor here should be the 256, or 1024? I am unsure where to look...

sql/item_mods.sql
```diff
@@ -37210,6 +37212,7 @@ INSERT INTO `item_mods` VALUES (27414,31,69);
 INSERT INTO `item_mods` VALUES (27414,68,72);
 INSERT INTO `item_mods` VALUES (27414,384,40);
 INSERT INTO `item_mods` VALUES (27414,386,55);
+INSERT INTO `item_mods` VALUES (27414,429,25);     -- Footwork+10% of 256 = 25.6
```
helixhamin: Same as above, can anyone tell me if the divisor here should be the 256, or 1024? I am unsure where to look...

sql/item_mods.sql
TeoTwawki: modifier.h says this:
`FOOTWORK_ATT_BONUS        = 429, // TODO: Raises the attack bonus of Footwork. (Tantra Gaiters +2 raise 100/1024 to 152/1024)`
status.lua says this:
`MOD_FOOTWORK_ATT_BONUS        = 429 -- Raises the attack bonus of Footwork. (Tantra Gaiters +2 raise 100/1024 to 152/1024)`

Also doesn't appear to actually be implemented right now..
https://github.com/DarkstarProject/darkstar/search?utf8=%E2%9C%93&q=FOOTWORK_ATT_BONUS&type=Code

Seeing mods added to modifier.h that don't actually do anything makes me sad.

helixhamin:
@TeoTwawki I changed the one back, and made the inserted one based on 1024, rather than 256.
I guess it won't work anyways, since the mod doesn't actually do anything, but once it is implemented, we should expect it to work properly.

TeoTwawki:
My **hope** is that once MOD_FOOTWORK_ATT_BONUS is actually doing something we *don't* have to keep reminding everyone its a fraction of whatever, but since it does nothing for now I'll merge and just hope we wind up having to change those lines to something not a freakin fraction later..

PR #3739, opened 2017-02-21
cleaned up some invalid or duplicated recipes

wrenffxi:


TeoTwawki:
Later I'm gonna nuke the ones that don't belong and shift the IDs so that there are no gaps, unless you want to beat me to it.

.. Coz you just *know* ppl will keep +1 ing the highest value at bottom of file instead of filling in those gaps ..

wrenffxi:
@TeoTwawki After we finish going through the remaining crafts -- @R3P0FFXI added a whole bunch recently. We should coordinate so we're not stepping on each other's toes -- we could reorder the table by main craft level + type or something, and get rid of all the gaps in one go.

TeoTwawki:
Sounds good.

R3P0FFXI:
@wrenffxi hit me up on Legions Discord and we can discuss this as we have a ton more stuff to send in still

PR #3741, opened 2017-02-22
Added additional mods for missing items pt1

R3P0FFXI:
Added item mods part 1


TeoTwawki:
> 7,525 additions, 6,755 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

Either you did waaay to much in one go, or you did something very wrong.

R3P0FFXI:
I believe it was the previous that its possible there was to much in the initial batch I thought I had it split up enough in the first go @TeoTwawki I think some of it may also be from the sorting to get the IDs in order as to why it shows so much deleted. It is correct as this si the first batch im pulling out of the file you asked me to cut down.

teschnei:
if you're reordering everything (or a bunch of stuff), then that needs to be in a separate commit (but not necessarily a separate pull request) from actual additions so that we can review the additions properly

TeoTwawki:
Checked, file was already sorted before this. Using textFX on all rows in notepad++ is a bad idea, try again without doing that..,

PR #3747, opened 2017-02-24
Adding missing item mods and item mods pet.

helixhamin:
Also, some spacing editing so the comments align.


TeoTwawki:
need a rebase and then I think its all good

helixhamin:
Finally got the rebase working correctly

PR #3748, opened 2017-02-24
Fix for very large values being set into [GUILD]daily_points

wrenffxi:
rank-up putting player in new guild point pattern with a smaller maxpoints than currPoints


TeoTwawki:
Silence is acceptance /merge

teschnei:
I only thought about getting rid of the extraneous comments, but it wasn't
pressing enough for me to remember to do it

PR #3759, opened 2017-02-28
Fix crashing in Port Windurst. Add comments pointing out airships.

ibm2431:
To quote @takhlaq : "airships suck".


TeoTwawki:
Aren't non standard characters you can't even SEE placed into a text field _fun_ to deal with?

PR #3765, opened 2017-03-02
Npc placeholders for uncaptured NPCs

ibm2431:
As mentioned in #3754 , I wrote a tool which compares our SQL list to a POLUtils dump and fills in placeholder comments for NPCs listed in POLUtils which we're currently lacking in our list.

These comments serve multiple purposes:
1) Their prefix is picked up by my (still in-testing) newer version of NPC batch job and treated as "live" NPCs for calculating/shifting/moving purposes. This improves accuracy _immensely_ such that the tool _stops trying to move NPCs into the wrong slot_ because the slot has the same name as an NPC we currently don't have captured.
2) Make the file easier to read/edit by hand. Knowing what's _supposed_ to be in an ID gap gives insight into what NPCs you're currently examining. For example, take a Moogle in a town zone with an uncaptured slot next to it. How is it different from other Moogles? Now insert a comment saying that there's supposed to be a Symphonic Curator in that empty slot. You now know that Moogle is probably the MH one.
3) Serves as documentation of what still needs capturing! All an enterprising NPC capturer has to do is search for that `-- NC:` prefix!
4) If the tool is rerun after NPC shifts happen, instantly tells us why: "Oh, they added debug Moogles."

My tool starts at the first ID in our SQL list. It doesn't start at the beginning of the list because then it'd just populate every zone with a bunch of mobs. We (usually) have the first non-battle NPC in a zone, so this works out. Once it passes the end of our SQL list, it'll go through the rest of the zone in POLUtils and add whatever we're missing on the tail end. It's also important to note that this only picks up on NPCs which have an entry in POLUtils. It is completely ignorant of NPCs not in POLUtils.

Due to the way SE organized some zones (all battlefields), what we consider NPCs are interspersed with battle mobs. I've gone through and manually removed these mobs while leaving the coffers. However, Throne Room, Throne Room [S], and Empyreal Paradox deserve special mention. We have some Shadow Lords , Promathias, Lillisettes, etc. captured in the list, which honestly, I can't determine are CS NPCs or battle mobs. I _do_ know that both versions of the NPC batch job are constantly trying to move the ID for those NPCs. So I've left the filler mobs in those blocks so those versions we have listed _stop bouncing around_. At some point will need to find out definitively which ones are CS versions, and which are actually battle mobs that don't belong in the file.

@starlightknight I want to do a PR for the darkstar-utils repository. Do I need to keep anything special in mind when compiling the JAR for the included-in-the-repo "precompiled binaries"? When I run Maven it yells about the build being platform dependent. Is this something that needs to be remedied, or is the JAR currently in the repo _already_ platform dependent?

@nyczducky I dedicate this PR to you! This is the list of every NPC we don't have! Get cracking! 😉


starlightknight:
@ibm2431 glad to see someone taking a shot at improving this :-) 

I have seen instances in the file in the past that did not exist in polutils at all that you may need to be weary of with this type of approach. Iirc they were things like certain transports and npcs with special symbols. In this case, in the old job I would shift by the same offset that the previous npcs had been going by, since there was nothing to compare them (if you leave these alone during a shift, you'd end up with duplicate ids so that the file doesn't load, and the zone generally shifts by the same amount once it starts shifting unless there are gaps in the file). 

You should not need to do anything special for the JAR, but the app is kind of platform dependent anyways. There was some incompatibility with POLUtils dumps via Wine and the job that i didn't end up having time to sort out - but that was the primary reason I stopped running the job myself. I no longer use Windows post Windows 10.

ibm2431:
My approach is declaring "named" NPCs in our SQL list (ie: have POLUtils names) as starting and ending points of a "chunk", with the no-names (and the ones not in POLUtils at all) being inside the "chunk".  I then find the updated POLUtils ID for both the starting/ending NPCs and calculate the differences. All the no-names / non-POLUtils-existant NPCs are then shifted by the same amount the starting NPC moved (either forwards or backwards).

I then compare how much the starting NPC shifted vs how much the ending NPC shifted to determine if NPCs were added or removed within the chunk. If our starting NPC's ID stayed the same while our ending NPC's ID increased, then SE added new NPCs within that chunk (and we write a little comment to notify the user). But if our ending NPC ID decreased while the starting ID stayed constant, then obviously NPCs were removed (and we mark the area for manual review).

We then move onto the next chunk, with our previous "ending" NPC being the new starting NPC, and search through our SQL list for the next named NPC to serve as the next ending NPC (or the last NPC if the zone ends before finding a named NPC). Rinse and repeat until the zone is solved.

"What if the named, chunk-ending NPC itself was removed?" We detect that by checking the _next_ NPC in our SQL list and comparing IDs of our prospective ending NPC versus the NPC we list after it. If the prospective ending NPC was removed, but our ID search didn't return -1 (because it found an ID of another instance of that name later on in the POLUtils dump), then the ID returned will be higher than the next-in-line NPC we're checking against. The tool sees this, comments the NPC out as being removed, and carries on through the list looking for a valid "ending" NPC.

I threw a couple different tests at the newer tool and it passed them all, so I actually used it for the last update. It did "okay", but was still a bit oversensitive for my taste (saying things changed which didn't, like coffers and Promathias moving). But placeholder comments for our uncaptured NPCs should resolve any remaining issues for the tools!

takhlaq:
rip custom npcs

nyczducky:
Awesome!

Cloudef:
I've thought few times doing some C command line tool alternatives to POLUtils. This needs to be done anyways sometime if we plan to reimplement open source client.

nyczducky:
WTB @speed255 on Retail.

I'll jump on IRC later to confirm. Looked at all the NC prefixes.

In terms of capturing and uploading based on the tool you gave me, I just wanted to confirm I'm updating the SQL properly before I do any PR.

PR #3768, opened 2017-03-02
Fixed: Corrected aH values to match retail

ffxijuggalo:
This fix fills in auction house categories: Alchemy 1 & 2, Misc. 1, 2, & 3.
Also fills in numerous amounts of other items that were missing in other categories.


TeoTwawki:
Now we need https://wiki.dspt.info/index.php/Items#item_basic updated to reflect the new categories

PR #3771, opened 2017-03-04
rejoin run in progress only if uniqueID > 0

wrenffxi:
removed unused firstDyna variable from hieroglyphics

### DIFF
--- a/scripts/zones/Bastok_Mines/npcs/Trail_Markings.lua
+++ b/scripts/zones/Bastok_Mines/npcs/Trail_Markings.lua
@@ -33,6 +33,7 @@ function onTrigger(player,npc)
         local firstDyna = 0;
         local realDay = os.time();
         local dynaWaitxDay = player:getVar("dynaWaitxDay");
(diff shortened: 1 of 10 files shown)


maxtherabbit:
what is the point of this? what situation could result in the unique id being set to 0?

wrenffxi:
uniqueID will be zero after every completed run.

maxtherabbit:
I see where the server variable is unset whenever the dynamis 
battlefield is cleaned up (leavecode 4) but I don't see where it is ever 
unset for the players. Unless I'm missing something, the players still 
have their DynamisID set to the "unique id" of the last dynamis they 
entered, regardless of whether the run was completed/exited or not.

If I'm correct it would mean that even if the server variable is unset 
(would return 0 on check) then it would never satisfy the conditional 
(player:getVar("DynamisID") == 
GetServerVariable("[DynaWhatever]UniqueID")) unless they had never 
entered a dynamis before(in which case they would not be time-locked 
anyway).

Issue #3394, opened 2016-09-26
Linux Server Crashes (max cpu?) with zlib_compress error/sea al


abriasffxi:
**_I have:_**
- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160831

**_Server Version_** (type `@revision` in game) **:**
6475f7ebaef

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
As discussed in IRC, I was seeing intermittent crashes that started happening when the server (Ubuntu Server 16.04 running on xeon 1270v5) reached moderately high cpu loads.

Found a way to replicate by maxing cpu with stress-ng and then aoeing about 30 mobs in Giddeus.  Consistently crashed 3 times in 4 pulls.  Ran with gdb and here is the backtrace.

```

[26/Sep] [07:09:50][Warning] zlib_compress: ran out of space (3854 : 11 : 3864 : 101)

Thread 1 "dsgame" received signal SIGSEGV, Segmentation fault.
send_parse (buff=<optimized out>, buffsize=buffsize@entry=0x7fffffffe808, from=from@entry=0x7fffffffe820, 
    map_session_data=map_session_data@entry=0x1d47a930) at src/map/map.cpp:682
682         WBUFL(PTempBuff, zlib_compressed_size(PacketSize)) = PacketSize;
(gdb) bt
#0  send_parse (buff=<optimized out>, buffsize=buffsize@entry=0x7fffffffe808, from=from@entry=0x7fffffffe820, 
    map_session_data=map_session_data@entry=0x1d47a930) at src/map/map.cpp:682
#1  0x000000000049551f in do_sockets (rfd=rfd@entry=0x7fffffffe880, next=...) at src/map/map.cpp:382
#2  0x00000000004073f7 in main (argc=1, argv=0x7fffffffea18) at src/common/kernel.cpp:303

```

After consultation with @Cloudef , reverted 47b3a5f, remade, and tried again.  We've done about 15 pulls now with no crash or any other random stuff in gdb.

Cloudef:
[...]

PR #3772, opened 2017-03-04
Attempt at fixing zlib compression errors (#3394)

Cloudef:
Using the data posted by https://github.com/DarkstarProject/darkstar/issues/3394#issuecomment-263086362 @eraffxi I confirmed that zlib_compress always failed even with the old "zlib" routine.

We can get rid of the failure by allowing zlib_compress function to use the whole allocated space of PTempBuff.  

I'm not sure if the routine can fail anymore, but just in case I added another commit that will split the packets in case it does. In anycase, if map.cpp gets refactored, it may also be good idea to reconsider changing the -1 returns in zlib.cpp just to fatal errors, to reduce the error handling overall in codebase, assuming the function doesn't fail anymore.

### DIFF
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -658,33 +658,57 @@ int32 send_parse(int8 *buff, size_t* buffsize, sockaddr_in* from, map_session_da
     uint32 PacketCount = PChar->getPacketCount();
     uint8 packets = 0;
 
-    while (PacketSize > 1300 - FFXI_HEADER_SIZE - 16) //max size for client to accept
-    {
-        *buffsize = FFXI_HEADER_SIZE;
-        PacketList_t packetList = PChar->getPacketList();
-        packets = 0;
+    do {
+        do {
+            *buffsize = FFXI_HEADER_SIZE;
+            PacketList_t packetList = PChar->getPacketList();
+            packets = 0;
 
-        while (!packetList.empty() && *buffsize + packetList.front()->length() < map_config.buffer_size &&
-            packets < PacketCount)
+            while (!packetList.empty() && *buffsize + packetList.front()->length() < map_config.buffer_size &&
+                packets < PacketCount)
+            {
+                PSmallPacket = packetList.front();
+
+                PSmallPacket->sequence(map_session_data->server_packet_id);
+                memcpy(buff + *buffsize, *PSmallPacket, PSmallPacket->length());
+
+                *buffsize += PSmallPacket->length();
+                packetList.pop_front();
+                packets++;
+            }
+
+            PacketCount /= 2;
+
+            //Сжимаем данные без учета заголовка
+            //Возвращаемый размер в 8 раз больше реальных данных
+            PacketSize = zlib_compress(buff + FFXI_HEADER_SIZE, *buffsize - FFXI_HEADER_SIZE, PTempBuff, map_config.buffer_size);
+
+            // handle compression error
+            if (PacketSize == static_cast<uint32>(-1))
+            {
+                continue;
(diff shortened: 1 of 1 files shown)


src/map/map.cpp
```diff
@@ -676,15 +675,25 @@ int32 send_parse(int8 *buff, size_t* buffsize, sockaddr_in* from, map_session_da
...
+
+        PacketCount /= 2;
+
         //Сжимаем данные без учета заголовка
         //Возвращаемый размер в 8 раз больше реальных данных
-        PacketSize = zlib_compress(buff + FFXI_HEADER_SIZE, *buffsize - FFXI_HEADER_SIZE, PTempBuff, *buffsize);
+        PacketSize = zlib_compress(buff + FFXI_HEADER_SIZE, *buffsize - FFXI_HEADER_SIZE, PTempBuff, map_config.buffer_size);
+
+        // handle compression error
+        if (PacketSize == static_cast<int32>(-1))
```
teschnei: comparison of signed with unsigned?

Cloudef:
Oh also, don't merge until tested by people from #3394

src/map/map.cpp
```diff
@@ -676,15 +675,25 @@ int32 send_parse(int8 *buff, size_t* buffsize, sockaddr_in* from, map_session_da
...
+        {
+            packets = 0;
```
teschnei: isn't this already set to 0 at the start of the loop anyways?

src/map/map.cpp
Cloudef: Gah, meant to cast to uint32.

src/map/map.cpp
Cloudef: Yes but, not will not if we exit the loop due to PacketCount <= 0

Cloudef: There will be more problems if that happens though. The code doesn't do any checking whatsoever after that so things are in pretty undefined state. There is that PTempBuf overflowed message, but it doesn't return or anything, it has already corrupted the memory and continues to corrupt memory.

teschnei:
well, == 0 because it's a uint32 ;)

at any rate, I think that character will be stuck if PacketCount gets to 0
and we can't compress the one packet - because it will leave that same
packet in the packet list.  So it seems to me that we must erase at least
one packet

Cloudef:
I think in that case we might want to give up already. Depending how well the first commit fix works, we might not need this second commit at all. And instead just treat the zlib.cpp error paths as fatal errors.

helixhamin:
Well, I have pulled and rebuilt, but it will take me a bit before I can get to dynamis to check.

Cloudef:
To help with testing. The original issue is caused by the amount of packets being compressed and sent to client on one go. So, having either lots of activity / players, or throttling CPU time on server should work. I reproduced this on dynamis windtrust by pulling everything, however it still wasn't very consistent so .. :)

helixhamin:
Well, I cannot say it is fixed, but I was finally able to complete a standard dynamis windurst run. I will check dynamis bastok this weekend to see if I can complete that also.

Cloudef:
@helixhamin If you can complete bastok, I would say that's enough. Unless we get more people here. I was not able to get the zlib warnings anymore in my testing. Did you have those?

Altalus:
Just my 2 cents but we were able to get both Dyna without a crash using this on my server.

Cloudef:
@teschnei I would say it does not at least make the situation worse :)

teschnei:
Nope

teschnei:
Nope as in "i don't object"

PR #3777, opened 2017-03-06
Ix'Aern (DRK) animosity, spawn, and reraise mechanics

wrenffxi:
http://ffxiclopedia.wikia.com/wiki/Ix'aern_(Dark_Knight)

Sheer animosity is sometimes rewarded when killing Aerns in the room containing QM.
Player who gains sheer animosity is sent a message, and has 10 minutes to click QM.
Click on QM indicates when someone else has sheer animosity.
If hated player clicks QM, Ix'Aern DRK and its minions spawn.
Ix'Aern DRK has a high chance to reraise.
When killed, it drops one of vice (25%) or deed (75%).
QM moves every 30 minutes. This does not reset animosity.


maxtherabbit:
I noticed two instances of: player:PrintToPlayer("reset");
In the qm2.lua as well as the Aw_aern.lua. Did you mean to comment these out or replace with something else?

Also why is it you always call os.time() with the parameter 't' when you are using a simple timestamp (not a time table)?

wrenffxi:
Fixed.  Too much copy+paste.  Also made it immune to stun, per wiki page.

maxtherabbit:
thanks!

PR #3779, opened 2017-03-07
COR AF3 sidequests

wrenffxi:
[Against All Odds - Leleroon's Sidequests](http://ffxiclopedia.wikia.com/wiki/Against_All_Odds_-_Leleroon%27s_Sidequests)

This commit adds three COR AF3 sidequests, for the Corsair's Gants, Bottes, and Frac.  This does NOT contain the Cutter battlefield for the Corsair's Tricorne.

### DIFF
--- a/scripts/zones/Nashmau/npcs/Leleroon.lua
+++ b/scripts/zones/Nashmau/npcs/Leleroon.lua
@@ -1,12 +1,11 @@
 -----------------------------------
 -- Area: Nashmau
--- NPC: Leleroon
--- Standard Info NPC -- Corsair SideQuests
+--  NPC: Leleroon
+-- Corsair AF2 and AF3 quests
 -- @pos -14.687 0.000 25.114 53
 -----------------------------------
 package.loaded["scripts/zones/Nashmau/TextIDs"] = nil;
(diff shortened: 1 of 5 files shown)


scripts/zones/Nashmau/npcs/Leleroon.lua
```diff
@@ -1,30 +1,28 @@
...
 -----------------------------------
 package.loaded["scripts/zones/Nashmau/TextIDs"] = nil;
 -----------------------------------
 
-require("scripts/zones/Nashmau/TextIDs");
 require("scripts/globals/settings");
 require("scripts/globals/status");
 require("scripts/globals/quests");
 require("scripts/globals/keyitems");
+require("scripts/zones/Nashmau/TextIDs");
```
TeoTwawki: :slightly_frowning_face:

TeoTwawki: Aside from terrorizing my OCD by re-ordering the requires\*, all looks good.


###### \* *I compulsively order them widest first, in reality we* ***all*** *should be leaving them alone unless they are actually wrong.*

PR #3783, opened 2017-03-08
Fishing segment of quest "A Boy's Dream" (PLD AF2)

wrenffxi:
Players can now fish for odontotyrannus for PLD AF2 quest.


maxtherabbit:
if only it were that easy!

Issue #3782, opened 2017-03-08
Repeating EXP Loss while character left dead overnight


whateversclever1215:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30161227_1

**_Server Version_** (type `@revision` in game) **:**
"Revision is: Unknown" (DSP Server: ERA FFXI)

**_Source Branch_** (master/stable) **:**

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
I was told to report this here, as this is a DSP/core code issue.

I multibox, although that is not relevant to this issue.
I was exping solo with my characters.

It was late and my wife distracted me / pulled me away from the game.
I just decided to let my THF (The only one fighting the mob, a non-aggro lizard at Cape Teriggan outpost) melee the last 40%~ hp of the enemy, and I just went afk. Wouldn't matter if I died one time. My character did die, but I just left it for the morning and went to sleep.

When I woke up, my character was still dead and had not home pointed, first question mark.
Someone had given me a raise while I was asleep.
I didn't accept or deny it because I was afk.
When I came back in the morning I accepted the raise.

Then I learned that I had lost over 2 full levels. I went from a pre-death 40,000/44,000exp at 75, to a post-death 28911/40,000 at Lv73.

I am told that this bug is known somewhat, but I didn't find it when searching your repository.
I think the Raise accept/deny dialog kept me from either homepointing or anything else. I think the timer kept running all night, and I think after every hour it took another death worth of experience.
I don't care about the exp loss, easy to earn back, but you should be aware of this bug.
Hopefully the extra detail that happened to me helps narrow down the cause of the issue.

GMYasiinEra:
I can confirm this as well. We spoke about it and I have tested on the master since then. IF you let the timer run out you lose the exp for a second time upon being homepointed.. Since he didnt get homepointed(pending raise) everytime the hour went by he lost exp.. since he was away for so long, the timer kept rolling back his exp every hour until he finally awoke the next morning. Leaving him nearly 3 levels delvled lol. Now I was pretty harshe in explaining a simple work around but never the less its still a bug.

R3P0FFXI:
[...]

PR #3797, opened 2017-03-12
Players now return to their home point when the death timer runs out.

UnknownX7:
#3782


teschnei:
whoops, forgot to implement that

could you put all that logic into a charutils function and just have both of them call that?  charutils::HomePoint or something - seems a bit more intuitive than packet_system calling onDeathTimer when you select the menu option

UnknownX7:
Done!

PR #3805, opened 2017-03-15
Attempting to address issue with Mythic Aftermaths

ShelbyZ:
- The formulas needed to use math.floor(...) on the equation result
- Need to check if we should apply an aftermath on top of an existing
one (highest wins)
- Need to remove any lower aftermaths when applying a higher
- Need to remove duplicate aftermath because occassionally it does not
self-overwrite (even with sql changes)
- Power level for DA should be 40 not 45
- Duration should be 60, 90, 120

It should be noted that this does lack changes to abilities.sql to mark aftermath 1,2,3 as overwrite because it did not appear (through testing) that this made the overwrite of aftermath level X to aftermath level X (resetting the duration timer)

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -950,17 +950,16 @@ function applyAftermathEffect(player, tp, params)
 
     local apply_power = 0
     if (tp == 3000 and shouldApplyAftermath(player, EFFECT_AFTERMATH_LV3)) then
-        checkAndRemoveLowerAftermath(player, EFFECT_AFTERMATH_LV3);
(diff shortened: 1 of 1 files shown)


teschnei:
You don't need to check an effects existence before trying to delete it - deleting an effect you don't have does nothing.  Actually, checking it first is worse, because it'll have to iterate the effect list twice instead of once.  With that in mind, it does seem a bit much to have a function call that would be replaced by 2 lines (in the AM3 block) and 1 line (in the AM2 block).  You don't need to delete the effect you're adding - that would be handled by the effects sql table.

ShelbyZ:
I see what you mean about the function call.  I hope this is inline with your suggestion(s).

PR #3806, opened 2017-03-15
CORE: add new members' char or party id to MSG_PT_RELOAD

maxtherabbit:


teschnei:
it's not pretty, but I suppose none of the message stuff is in the first place.  If you've already tested, you can go ahead and merge it

maxtherabbit:
no it's not, but I went through about 4 different approaches before landing on this one - I'm pretty confident it's the least expensive way to achieve the needed functionality unless we were to implement some kind of map server party registration list in memory

PR #3807, opened 2017-03-15
Added /blockaid

UnknownX7:
Let me know if there are any other features that /blockaid worked on.

### DIFF
--- a/src/map/ai/states/range_state.cpp
+++ b/src/map/ai/states/range_state.cpp
@@ -98,7 +98,7 @@ bool CRangeState::Update(time_point tick)
             m_errorMsg = std::make_unique<CMessageBasicPacket>(m_PEntity, m_PEntity, 0, 0, MSGBASIC_MOVE_AND_INTERRUPT);
         }
         action_t action;
-        if (m_errorMsg && m_errorMsg->getMessageID() != MSGBASIC_CANNOT_SEE)
+        if (m_errorMsg && static_cast<CMessageBasicPacket*>(m_errorMsg.get())->getMessageID() != MSGBASIC_CANNOT_SEE)
         {
             action.id = m_PEntity->id;
             action.actiontype = ACTION_RANGED_INTERRUPT;
--- a/src/map/ai/states/range_state.cpp
+++ b/src/map/ai/states/range_state.cpp
@@ -98,7 +98,8 @@ bool CRangeState::Update(time_point tick)
             m_errorMsg = std::make_unique<CMessageBasicPacket>(m_PEntity, m_PEntity, 0, 0, MSGBASIC_MOVE_AND_INTERRUPT);
         }
         action_t action;
-        if (m_errorMsg && static_cast<CMessageBasicPacket*>(m_errorMsg.get())->getMessageID() != MSGBASIC_CANNOT_SEE)
+        auto cast_errorMsg = dynamic_cast<CMessageBasicPacket*>(m_errorMsg.get());
+        if (m_errorMsg && (!cast_errorMsg || cast_errorMsg->getMessageID() != MSGBASIC_CANNOT_SEE))
         {
             action.id = m_PEntity->id;
             action.actiontype = ACTION_RANGED_INTERRUPT;
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -37,6 +37,7 @@
 #include "../packets/menu_raisetractor.h"
 #include "../packets/char_health.h"
 #include "../packets/char_appearance.h"
+#include "../packets/message_system.h"
 
 #include "../ai/ai_container.h"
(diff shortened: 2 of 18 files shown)


src/map/entities/charentity.cpp
```diff
@@ -1509,6 +1520,11 @@ void CCharEntity::OnItemFinish(CItemState& state, action_t& action)
 
 CBattleEntity* CCharEntity::IsValidTarget(uint16 targid, uint16 validTargetFlags, std::unique_ptr<CMessageBasicPacket>& errMsg)
 {
+    // check /blockaid
+    CCharEntity* PChar = (CCharEntity*)this->GetEntity(targid, TYPE_PC);
```
teschnei: GetEntity is expensive.  There's no reason to do it since we already get an entity object from CBattleEntity::IsValidTarget - we can check blockaid after we the entity from the base call

src/map/entities/charentity.cpp
```diff
@@ -1509,6 +1520,11 @@ void CCharEntity::OnItemFinish(CItemState& state, action_t& action)
...
+    if (PChar && charutils::IsAidBlocked(this, PChar))
+        return nullptr;
```
teschnei: is there no specialized error message for trying to cast on someone with blockaid?

src/map/entities/charentity.h
```diff
@@ -363,6 +365,7 @@ class CCharEntity : public CBattleEntity
 
     bool			m_isWeaponSkillKill;
     bool            m_isStyleLocked;
+    bool            m_isBlockingAid;
```
teschnei: where is the initialization? Either in the constructor (like everything else) or in here is fine

src/map/packet_system.cpp
```diff
@@ -2866,6 +2896,14 @@ void SmallPacket0x06E(map_session_data_t* session, CCharEntity* PChar, CBasicPac
         return;
     }
 
+    // check /blockaid
+    CCharEntity* PTarget = (CCharEntity*)PChar->GetEntity(targid, TYPE_PC);
```
teschnei: same thing here as earlier - don't do GetEntity if we don't need to.  we already get an entity further down from GetCharFromWorld (and it's worth noting that CCharEntity->GetEntity only gets an entity from this zone).  Additionally, blockaid will need to be checked on the receiving side, because an entity on another server will always fail this check (PTarget would be null, and PInvitee would also be null)

src/map/utils/charutils.cpp
```diff
@@ -4715,6 +4716,29 @@ namespace charutils
         }
     }
 
+    bool IsAidBlocked(CCharEntity* PInitiator, CCharEntity* PTarget) {
+        if (PTarget->getBlockingAid())
+        {
+            bool inAlliance = false;
+            PTarget->ForAlliance([&PInitiator, &PTarget, &inAlliance](CBattleEntity* PEntity) {
```
teschnei: you don't need to capture PTarget (you aren't using it)

src/map/utils/charutils.cpp
```diff
@@ -4715,6 +4716,29 @@ namespace charutils
...
+                if (PEntity->id == PInitiator->id)
+                {
+                    inAlliance = true;
+                }
+            });
+
+            if (!inAlliance)
+            {
+                // Target is blocking assistance
+                PInitiator->pushPacket(new CMessageSystemPacket(0, 0, 225));
```
teschnei: let the caller push the packets - you don't know how they will need to handle it (for example, IsValidTarget using errmsg instead of pushing packets directly)

UnknownX7: There are 2 error messages to different clients, I'm not sure how to send both through IsValidTarget.

UnknownX7:
It should be good now, let me know if you need any more changes.

GMYasiinEra:
i love you lol stay on ERA for ever.... with my custom sandworm people are getting in on the hate list and piggy backing in on doomvoid lol

src/map/ai/states/range_state.cpp
```diff
@@ -98,7 +98,7 @@ bool CRangeState::Update(time_point tick)
             m_errorMsg = std::make_unique<CMessageBasicPacket>(m_PEntity, m_PEntity, 0, 0, MSGBASIC_MOVE_AND_INTERRUPT);
         }
         action_t action;
-        if (m_errorMsg && m_errorMsg->getMessageID() != MSGBASIC_CANNOT_SEE)
+        if (m_errorMsg && static_cast<CMessageBasicPacket*>(m_errorMsg.get())->getMessageID() != MSGBASIC_CANNOT_SEE)
```
teschnei: really ought to dynamic_cast this instead, just in case someone in the future sticks a different packet into there

UnknownX7:
Done.

Issue #3756, opened 2017-02-28
Ranged Weaponskills seem to have been hardcoded with exploitable ranges.


GMYasiinEra:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **30161227_1**

**_Server Version_** (type `@revision` in game) **unknown - master clone from last week**

**_Source Branch_** (master/stable) **master**

**_Additional Information_** (Steps to reproduce/Expected behavior) **Tested on ERA and Master DSP**

So considering SQL calls a value of 16 for every ranged ws. I think someone ninja updated ranger ws's in the core. First noticed this after a January update on ERA. reproduced on my test server. Its super OP alot of Linkshells tend to just Ranger Burn everything that have dangerous AoE moves. Id like to see it made a little more interesting and retail like. Wish I understood the core files more to help.

http://imgur.com/a/ofGXY

teschnei:
Ranger burning things with dangerous tp moves is exactly how retail works

GMYasiinEra:
lol yeah i get that but you still had to run to ws and get back out of harms way tactically after mobs skills like  gates etc.right now you can flat out TP and WS at 20+

Hozu:
Is mob AoE range supposed to take monster size into account or no?

teschnei:
You didn't have to move to use a WS. The only ranged WS that needs to be
close is blast shot

On Feb 28, 2017 9:48 AM, "Hozu" <notifications@github.com> wrote:

Is mob AoE range supposed to take monster size into account or no?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/DarkstarProject/darkstar/issues/3756#issuecomment-283095828>,
or mute the thread

GMYasiinEra:
> You didn't have to move to use a WS. The only ranged WS that needs to be
close is blast shot

DB shows Blast Shot as even further away lol `...` 
Wiki shows 5 yalms `...`
http://ffxiclopedia.wikia.com/wiki/Distance

15.8'	
Maximum range for most ranged Weapon Skills (excluding Blast Arrow, Blast Shot).

17.7'	
Maximum range for Detonator and Empyreal Arrow.

17.2'	
Maximum range for Jishnu's Radiance.

[...]

PR #3810, opened 2017-03-16
Fixed issues with weapon skills and melee distance calculation

UnknownX7:
#3756

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -538,7 +538,6 @@ void CMobController::Move()
     {
         return;
     }
-    float currentDistance = distance(PMob->loc.p, PTarget->loc.p);
     if (PMob->PAI->PathFind->IsFollowingScriptedPath() && PMob->PAI->CanFollowPath())
     {
         PMob->PAI->PathFind->FollowPath();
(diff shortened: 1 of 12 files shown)


teschnei:
"Fixed TP being used if a weapon skill is used out of range."

can you explain that?  Unless I'm misinterpreting the message, that's just wrong

teschnei:
and how would these changes affect spells that have melee range (blue magic, primarily)?

UnknownX7:
> "Fixed TP being used if a weapon skill is used out of range."
>
> can you explain that? Unless I'm misinterpreting the message, that's just wrong

Currently, when the client uses a WS, it will use TP regardless of whether or not it is in range. This isn't reflected on the client until it gains more TP though.

> and how would these changes affect spells that have melee range (blue magic, primarily)?

Magic doesn't currently take model size into account, so it would be unchanged. The only current difference in melee range is for small pets, which had to walk far inside of a large mob to attack when in reality they should have been able to attack by walking within melee range of the mob

teschnei:
so what is it trying to fix? the code and message imply you made character not lose tp if the WS is out of range, but that's wrong

UnknownX7:
Huh, did not remember that retail did that. Sorry, I guess I thought it wasn't supposed to be that way since the client's TP wasn't updating. I've reverted it.

UnknownX7:
I went ahead and reverted the mob movement change as well, mobs will still be able to hit each other at the same moment as opposed to before, where the large mob would always get the first hit since the small mob had to move inside of the larger one in order to get in range.

With the movement change the mob was walking to its actual attack distance, but since mob skills don't factor in sizes, it couldn't use abilities. I believe most job abilities should factor in sizes, but I think it needs a case by case basis for abilities, spells, and weapon skills. I think this is how retail had it, but I think the issue should be addressed in a separate pull request later on if needed rather than here, since its a much larger scale problem and would require a lot more changes.

PR #3811, opened 2017-03-16
pcmd/acmd kick on message server

maxtherabbit:
pcmd/acmd kick across message server
fixes new member of allianced party not getting party flags
updates logic for prioritizing char to update with MSG_PT_RELOAD

### DIFF
--- a/src/map/message.cpp
+++ b/src/map/message.cpp
@@ -297,25 +297,12 @@ namespace message
         {
             if (extra->size() == 8)
             {
-                uint32 partyid = RBUFL(extra->data(), 4);
-                uint32 charid = RBUFL(extra->data(), 0);
-                zoneutils::ForEachZone([partyid, charid](CZone* PZone)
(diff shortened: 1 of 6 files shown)


src/map/message.cpp
```diff
@@ -297,25 +297,12 @@ namespace message
         {
             if (extra->size() == 8)
             {
-                uint32 partyid = RBUFL(extra->data(), 4);
-                uint32 charid = RBUFL(extra->data(), 0);
-                zoneutils::ForEachZone([partyid, charid](CZone* PZone)
+                CCharEntity* PChar = zoneutils::GetCharToUpdate(RBUFL(extra->data(), 4), RBUFL(extra->data(), 0));
```
takhlaq: seems ambiguous in charutils
CParty::GetCharToUpdate would be a lot clearer

maxtherabbit:
The zone list is only available to zoneutils

maxtherabbit:
I could just rename the member function... Like get char for pt reload or
something

PR #3813, opened 2017-03-18
Fixed TA calculations when z is the dependent variable.

UnknownX7:


teschnei:
"No description provided"

fixed? how? what was the problem? why does this fix it?

UnknownX7:
http://pastebin.com/QB1nhzNs

As you got closer to a diagonal line with the mob and you were further away on the X axis vs Z axis, the triangle's angle became 0, creating a straight line.

UnknownX7:
It fixes it because the two points being used to create the slope of the lines will always be equidistant from the player instead of slowly converging. A player at (5,5) vs a monster at (0,0) was causing the code to find (5 + sqrt(2)/4, 5 + sqrt(2)/4) and (5 - sqrt(2)/4, 5 - sqrt(2)/4) as the two points to be used since the min and max of X and Z would converge because the X's min and max was calculated so that it was meant to find the slope of (minZ, maxX) and (maxZ, minX) as compared to the mob's position when the code was actually doing (minZ, minX) and (maxZ, maxX).

Hozu:
So you're saying that under certain situations, the triangle would be more like a rectangle?

UnknownX7:
Somewhat. It will become a very thin triangle like this http://etc.usf.edu/clipart/41700/41728/fc_obtusetri_41728_lg.gif (the monster would be at the top point of the triangle while the TA user is halfway between the 2 points making the base, meaning that the TA user could technically TA off someone slightly behind them as well) until the base has fully rotated to be parallel with the line made by the monster and TA user, which just makes a straight line. However, this point is never reached because the moment that x == z, it swaps back to the other formula (which is correct, the whole point of this change is that the two functions are currently disjoint).

GMYasiinEra:
i haven't gotten around to testing this yet on my own server but coming a Thief the current formula isn't working correctly.

PR #3819, opened 2017-03-24
Update bard af quest with wait duration change

ddumont:


ddumont:
Wait...   the original logic was idiotic...  this needs much more work.

ddumont:
Ok so the basic problem was...   if you buried the ring on Dec 31st, before midnight, you're shit-outta-luck for ever...  you could never complete this quest.   Also, if you did it on Dec 30th but waited too long to complete the quest (IE: let the year laps) you had to wait another year to complete it.

ddumont:
This compromise stores the current hour and makes sure that it's not the same as it was when you buried it.
Granted, if you waited exactly 24 hours it would fail... but I can't put the date and time in playerVars because it's only int(11) big.

ddumont:
I feel like with the updated info, most people would probably wait around for an hour or so... and won't really hit this.   I'm open to other suggestions.

teschnei:
uh.. store os.time() and compare that + (60*60) with os.time() again?

I think os.time is only a 32 bit number which would fit into a player var

ddumont:
nah... 32 bit numbers signed (column is not unsaigned) do not go high enough.  already checked.

ddumont:
2147483647 max val (signed)
can only hold 68.096259734906 years since epoch.  I still wanna play this when I'm 80 :P

teschnei:
Hmm, something I am considering changing then

ddumont:
I also didn't want to deal with upgrade scripts... but if you do that, I'll surely do something like what you mentioned.   It's what I wanted to do to begin with.

ddumont:
ahhh... this should have been minute anyway...

scripts/zones/Xarcabard/npcs/Perennial_Snow.lua
```diff
@@ -25,9 +25,13 @@ function onTrigger(player,npc)
    local circleOfTime = player:getQuestStatus(JEUNO,THE_CIRCLE_OF_TIME);
 
    if (circleOfTime == QUEST_ACCEPTED and player:getVar("circleTime") == 3) then
+       -- http://ffxiclopedia.wikia.com/wiki/The_Circle_of_Time (one earth hour delay)
+       local buried = player:getVar("star_ringburied");
+       local cleansed = buried != tonumber(os.date("%H"));
```
takhlaq: what if they use in same earth hour next day?
storing the buried date as seconds since epoch and subtracting that from current time to see if an hour has elapsed would be better

also, lua != == ~=

teschnei:
Technically you could also store vanadiel time, as it is measured as
seconds since the game was released

ddumont:
You're right I need to fix equality check.   But the wiki says minute, not
hour, I misread.   Do you really think we need to prepare against them
hitting the exact same minute?  Will update syntax when I get home.

scripts/zones/Xarcabard/npcs/Perennial_Snow.lua
ddumont: @takhlaq  see previous conversation, the field is not big enough.

takhlaq:
yep since thats how retail does it

ddumont:
I'll just wait for the schema update then.

teschnei:
If we go that route, Lua has to also use 64 bit numbers, which means
compiling in x64 (which means deprecating Win32).  Not particularily a
problem I suppose, but I need to talk about it with the others first

ddumont:
is that just because of the binding from c to lua?   lua seems to deal with os.time() just fine right now?   Is it going to implode in 20 years when we run out of space?

wrenffxi:
Open PR #3780 already makes this change to Perennial Snow.

teschnei:
Yeah, lua_tointeger converts to int32 on 32bit system and 64 on 64 bit
system
 It's overridable but I don't know what kind of other effects that'd have.
Easier to just say 64 only, probably

ddumont:
@wrenffxi yes but that logic makes it instant... and will cause side effects when the date flips eventually.
ppl in here saying they want it to match retail, which that does not do.

ddumont:
@teschnei what about making a getSince and setNow on player that tell mysql to store "NOW()" and select `col` - NOW() so we can avoid this unpleasentness?

teschnei:
Nah. Sounds too confusing - would like to use as much normal Lua as
possible. Besides, we already have problems with the size in integers for
certain scripts that check the 32nd bit anyways

ddumont:
I dunno I think that would make the code more readable in many areas... but ok.

wrenffxi:
It's not instant. It makes them wait 1 minute, and during that minute they get a "you're too impatient" message if they click. That matches retail. http://ffxiclopedia.wikia.com/wiki/The_Circle_of_Time

Of course, if os.time() isn't future-proof, that's a problem that affects tons of .luas including this one.

ddumont:
ok well then I like your fix better with the retail messages, but yeah the time is not future proof.

TeoTwawki:
> If we go that route, Lua has to also use 64 bit numbers, which means
compiling in x64 (which means deprecating Win32).  Not particularily a
problem I suppose, but I need to talk about it with the others first

@teschnei I wish we had some reliable info on how many ppl downloading the project had 64 bit support in their setups. My only concern is potentially losing future(and current) server owners & contributors. The larger guys like nasomi, era, legion, can handle it for sure but I'd bet more ppl run off their local desktop than anything else esp when it comes it testing contributions.

ddumont:
@TeoTwawki @takhlaq 
I'd like to advocate again for a player:setEventTime(/*String*/ eventName) and player:timeSinceEvent(/*String*/ eventName) that will allow us to put this messiness in mysql and leave it out of lua.

As was mentioned before... this kind of stuff happens all over the codebase and is unsafe. We can migrate existing calls over time.   Could even set up a new table for it.

ddumont:
@teschnei that's going to have an adverse impact on previously saved values in the table.
not to mention that time ticks at a different rate, so all calculations will need to be updated.   I don't think you'll find that task fun at all.

teschnei:
I don't know what you mean - time ticks at a different rate?

The existing values in the table aren't as big of an issue

ddumont:
Oh I thought you meant use vana time instead.  I see you said the epoch...    That would save you for calculations... but as I've **repeatedly** been told in IRC for changes I've wanted to push

> That's not how retail works.

teschnei:
retail doesn't have our db tables, I'm not sure what you mean? how is retail relevant to us changing the way something is internally stored?

ddumont:
So in addition to the code change, and possible other calculations done with os.time() you'll prepare table update scripts to tweak all the values in the player var table?

Even though that table isn't used for only time values?

ddumont:
My comment was that in retail, quests don't stop working because of time overflows.   And it was mostly tongue-in-cheek.  A good spirited ribbing in jest, if you will.

teschnei:
I don't really have a problem with that...  If it's simple enough for a server admin to figure out, I'll let them do it.  For the worse things (key items blob change, etc), I will (and have) written a 20 line program to convert whatever's needed

the "retail works" comment refers mostly to gameplay mechanics though, not how the server is ran

PR #3820, opened 2017-03-25
Core: Handle Latent Refresh/Regain from Augment

loganwest:
Applies Refresh or Regain if PC has those augments on the main-handed weapon. Weapon must be drawn to activate. Regain is divided by map_config.player_tp_multiplier due to being readded by addTP function. This keeps TP gain in-line with what augment value really is. Needs utils.h and item_armor.h for unpackBitsBE and proper class member respectively.

### DIFF
--- a/src/map/status_effect_container.cpp
+++ b/src/map/status_effect_container.cpp
@@ -33,6 +33,7 @@ When a status effect is gained twice on a player. It can do one or more of the f
 
 #include "../common/showmsg.h"
 #include "../common/timer.h"
+#include "../common/utils.h"
 
 #include <string.h>
 #include <array>
@@ -41,6 +42,8 @@ When a status effect is gained twice on a player. It can do one or more of the f
 
 #include "ai/ai_container.h"
 #include "ai/states/inactive_state.h"
+
+#include "items/item_armor.h"
 
 #include "packets/char_health.h"
 #include "packets/char_job_extra.h"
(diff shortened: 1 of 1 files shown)


teschnei:
we already have a latent effect for this case... It makes more sense to allow augments to specify any latent rather than hard coding this specific one, no?

loganwest:
These two augments do not currently work, so I wanted to implement them in some way. I tried the script approach first. The SQL file even has a comment after these "add to scripts." However, because of the logic checks that need to be made every three-second time cycle and the needed isolation from custom config file settings, I figured the best place to put this was in the same function that handles HP/MP/TP each time window. Because of the frequency this check needs to be performed, I needed a fast way to check the logic, decode the augments, and bundle the added effects into the update masks. 

Concerning the need for config isolation, the addTP function multiplies all TP arguments passed into the function. So, without stripping away the config multiplier setting, a disparity arises between what the augmented gear says and what's actually being applied. Not an issue for people running vanilla servers; but people who have messed with the TP settings may scratch their head for a little while wondering why it says +10 regain, but they're getting 15 per tick instead.

ibm2431:
Why only when engaged for Refresh? Is this just for items with "Latent Effect: Refresh", or is it including all Refresh+X Augments? Weapons with Refresh _Augments_ (like [Emissary Path D](https://www.bg-wiki.com/bg/Nolan#Escha_-_Ru.27Aun)) will tic MP even when not engaged.

teschnei:
It's probably for the Moonshade earring augment.

TeoTwawki:
At the time the augment guts got added to the project the only latent augs were moonshade earring, and because of that we didn't add the capability to do latents via augment and were going to fake that by item script since it was just the one item on retail. if there are more items doing latent on retail now then sounds like we should just that latent handling I wanted way back then to the augment table.

ibm2431:
The "Latent Effect: Refresh" on [Moonshade Earring](https://www.bg-wiki.com/bg/Moonshade_Earring) only tics MP when you're _not_ engaged~!

But! This PR is for weapons only, which the Earring is not (well, I guess if you _really_ tried...). I'm drawing a blank trying to recall a weapon with a Latent Refresh _augment_. The only main weapons I can find with even just non-augment normal Latent Refresh are [Mistilteinn](https://www.bg-wiki.com/bg/Mistilteinn) and [Numen Staff](https://www.bg-wiki.com/bg/Numen_Staff), of which Numen is engaged-only while Mistilteinn is always.

TeoTwawki:
Been around 20 days so closing. If there *is* a use-case for this tell us about it and I'll reopen and you can make whatever changes teschnei needs

PR #3822, opened 2017-03-25
Script: Fixed CoP Ring Selection Bug

loganwest:
Event now displays correct order and selection of CoP rings even when a user selection "Choose a different one." Cleaned up script to initialize variables only when needed and re-use common variables already defined. Changed minimum wait day to what's specified on Wiki.  Kept SQL variable naming convention the same to prevent user head->wall damage.

Edit: Cancelled original pull request due to typo in all variables named "ringelection." Corrected to be "ringSelection".

### DIFF
--- a/scripts/zones/Upper_Jeuno/npcs/_6s1.lua
+++ b/scripts/zones/Upper_Jeuno/npcs/_6s1.lua
@@ -1,106 +1,124 @@
------------------------------------
+----------------------------------------------------------------------------------------------------------------------------
 -- Area: Upper Jeuno
 -- NPC:  Marble Bridge Eatery (Door)
 -- @pos -96.6 -0.2 92.3 244
------------------------------------
-package.loaded["scripts/zones/Upper_Jeuno/TextIDs"] = nil;
------------------------------------
-
-require("scripts/globals/settings");
-require("scripts/globals/missions");
-require("scripts/zones/Upper_Jeuno/TextIDs");
-
-local RajasRing = 15543;
-local SattvaRing = 15544;
-local TamasRing = 15545;
-
------------------------------------
+----------------------------------------------------------------------------------------------------------------------------
+package.loaded["scripts/zones/Upper_Jeuno/TextIDs"] = nil
+----------------------------------------------------------------------------------------------------------------------------
+require("scripts/zones/Upper_Jeuno/TextIDs")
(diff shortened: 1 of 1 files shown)


loganwest:
P.S. The CoP ring bug was resolved by re-injecting the ring ID's as parameters into the player:updateEvent() function after OnEventUpdate was called.

takhlaq:
no need to close pull requests
you can amend last commit and force push and github will auto update the pr for us

scripts/zones/Upper_Jeuno/npcs/_6s1.lua
```diff
@@ -1,106 +1,124 @@
------------------------------------
+----------------------------------------------------------------------------------------------------------------------------
```
TeoTwawki: why are you adding this huge dashed line? Stick with the format we already use in literally every file in the project. I am not a fan of unique snowflakes.

scripts/zones/Upper_Jeuno/npcs/_6s1.lua
```diff
@@ -1,106 +1,124 @@
...
-
------------------------------------
+----------------------------------------------------------------------------------------------------------------------------
+package.loaded["scripts/zones/Upper_Jeuno/TextIDs"] = nil
+----------------------------------------------------------------------------------------------------------------------------
+require("scripts/zones/Upper_Jeuno/TextIDs")
+require("scripts/globals/missions")
+require("scripts/globals/settings")
+----------------------------------------------------------------------------------------------------------------------------
+local ring = {
```
TeoTwawki: We're brace on newline ppl here. Its in the unfinished style guide. The one automatically linked to everyone the second the try to contribute. https://github.com/DarkstarProject/darkstar/blob/master/CONTRIBUTING.md

scripts/zones/Upper_Jeuno/npcs/_6s1.lua
loganwest: Stick with your bug then asshole and I won't try to help anymore.

TeoTwawki: Grow up. This isn't 3rd grade. If you can't handle that and see it as some huge insult your skin is way too thin to be coding anything.

PR #3824, opened 2017-03-26
Added Revealers Pants, Revealers Mitts, Telchine Braconi 15 matt mod

R3P0FFXI:
Added Revealers Pants, Revealers Mitts, Telchine Braconi 15 matt mod


TeoTwawki:
ok, this looks legit to me, and looks like the only reason it was closed was you mistakenly adding the same commits to your other pr. so merging.

PR #3831, opened 2017-03-29
That CoP ring bug

TeoTwawki:
I amended someone else's commit. Because I'm an _\<expletive deleted\>_ if I criticize anything ever. Did not test, do not care.


TeoTwawki:
For anyone looking to replace the bad mod that was removed from ardor pendant: it would be the kind that buffs drg's wyvern breath attacks and possibly blu magic breaths, not damage reduction which is what was on it. We have a mod for wyvern breath only in trunk so far, nothing that buffs any other breath damage.

TeoTwawki:
commented wrong pr

PR #3840, opened 2017-04-05
Adding events for Magicked Astrolabe

ShelbyZ:
-- Adding logic to Churano-Shurano to offer the KI if not already
obtained
-- Adding variable to track if the player agreed to purchase
-- Adding KI after second event in series when they accept to pay


ShelbyZ:
Had quite a bit of help figuring out the correct event ids from IRC

PR #3843, opened 2017-04-08
Latents did not check for Storm Spells when checking for Weather - based

ShelbyZ:
- Get the player's status effect and check if under a scholar storm
spell
- Check for Zone Latents on effect add / remove
- Fix Desert Boots to check for earth - storms instead of ice - storms

### DIFF
--- a/src/map/latent_effect_container.cpp
+++ b/src/map/latent_effect_container.cpp
@@ -26,6 +26,7 @@ This file is part of DarkStar-server source code.
 #include "latent_effect.h"
 #include "entities/charentity.h"
 #include "entities/battleentity.h"
+#include "utils/battleutils.h"
 #include "utils/zoneutils.h"
 #include "conquest_system.h"
 #include "modifier.h"
@@ -626,7 +627,9 @@ void CLatentEffectContainer::CheckLatentsEquip(uint8 slot)
                 }
                 break;
             case LATENT_WEATHER_ELEMENT:
-                if (zoneutils::GetWeatherElement(zoneutils::GetZone(m_POwner->getZone())->GetWeather()) == m_LatentEffectList.at(i)->GetConditionsValue())
+            {
+                auto weatherElement = m_LatentEffectList.at(i)->GetConditionsValue();
+                if (weatherElement == zoneutils::GetWeatherElement(zoneutils::GetZone(m_POwner->getZone())->GetWeather()) || weatherElement == zoneutils::GetWeatherElement(battleutils::GetWeather((CBattleEntity*)m_POwner, false)))
                 {
                     m_LatentEffectList.at(i)->Activate();
                 }
@@ -635,6 +638,7 @@ void CLatentEffectContainer::CheckLatentsEquip(uint8 slot)
                     m_LatentEffectList.at(i)->Deactivate();
                 }
                 break;
+            }
             case LATENT_NATION_CONTROL:
                 CheckLatentsZone();
                 break;
@@ -1674,7 +1678,9 @@ void CLatentEffectContainer::CheckLatentsZone()
             }
             break;
         case LATENT_WEATHER_ELEMENT:
-            if (zoneutils::GetWeatherElement(zoneutils::GetZone(m_POwner->getZone())->GetWeather()) == m_LatentEffectList.at(i)->GetConditionsValue())
+        {
+            auto weatherElement = m_LatentEffectList.at(i)->GetConditionsValue();
+            if (weatherElement == zoneutils::GetWeatherElement(zoneutils::GetZone(m_POwner->getZone())->GetWeather()) || weatherElement == zoneutils::GetWeatherElement(battleutils::GetWeather((CBattleEntity*)m_POwner, false)))
             {
                 m_LatentEffectList.at(i)->Activate();
             }
@@ -1683,6 +1689,7 @@ void CLatentEffectContainer::CheckLatentsZone()
                 m_LatentEffectList.at(i)->Deactivate();
             }
             break;
+        }
         case LATENT_NATION_CONTROL:
         {
             //player is logging in/zoning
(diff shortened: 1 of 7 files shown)


teschnei:
battleutils::GetWeather already includes the zone weather (and handles the precedence properly), so you can get rid of the zone weather check

ShelbyZ:
Added some handled for Weather-based changes since they happen in core and only had access to the old weather values causing latents to trigger when losing the weather.

src/map/latent_effect_container.cpp
```diff
@@ -1674,7 +1675,23 @@ void CLatentEffectContainer::CheckLatentsZone()
             }
             break;
         case LATENT_WEATHER_ELEMENT:
-            if (zoneutils::GetWeatherElement(zoneutils::GetZone(m_POwner->getZone())->GetWeather()) == m_LatentEffectList.at(i)->GetConditionsValue())
+            // In the default case we need to pull weather from zone/player
+            if (weather == MAX_WEATHER_ID)
```
TeoTwawki: `if (weather == MAX_WEATHER_ID)` is an impossible condition. If that ever happens, a debug break will be tripped.

ShelbyZ: It is a default condition - to be used where CheckLatentsZone() is called without an element.  The only method that should be passing an element would be CZoneEntities::WeatherChange(...).  Basically we need a way to inform the latent checking of a potentially new weather.  At the time of checking we only have access to the old weather for the zone so it would be an incorrect check.

teschnei:
why does CheckLatentsZone need weather?

ShelbyZ:
When CZoneEntities::WeatherChange(...) calls into the latent_status_container to indicate a weather change has occurred it will fire CheckLatentsZone(), but a call to that will not reflect the current weather change for that zone.  The idea was to supply the weather parameter in this instance to provide the new weather when present or MAX_WEATHER_ID (as a default) for all other cases.

teschnei:
sounds like the weather change is supposed to call CheckLatentsWeather, not Zone

ShelbyZ:
Weather handling (LATENT_WEATHER_ELEMENT) is currently done under CheckLatentsZone(), it could be separated into it's own call that could take a weather parameter.

teschnei:
Sorry, I thought it existed already.  Yes, we need a CheckLatentsWeather then instead.  But you need to keep the weather logic in CheckLatentsZone because that's the one that's checked when you zone in.

src/map/latent_effect_container.cpp
```diff
@@ -1785,3 +1769,33 @@ void CLatentEffectContainer::CheckLatentsZone(uint16 weather)
...
+void CLatentEffectContainer::CheckLatentsWeather(uint16 weather)
+{
+    for (uint16 i = 0; i < m_LatentEffectList.size(); ++i)
+    {
+        if (m_LatentEffectList.at(i)->GetConditionsID() == LATENT_WEATHER_ELEMENT)
+        {
+            // get the scholar weather and convert to a weather element
+            auto sch_weather = zoneutils::GetWeatherElement(battleutils::GetScholarWeather(m_POwner));
+
+            // if either the new weather or the scholar weather satisfy the latent then activate it
```
teschnei: hmm.. this is still wrong, though.  battleutils::GetWeather does it right, so I'd rather have that do it - might have to just add another parameter to it for zone weather override (or separate them and have one call the other using zone->getweather)

ShelbyZ: It just leads me to putting in something like:

WEATHER GetWeather(CBattleEntity* PEntity, bool ignoreScholar, WEATHER zoneWeather = MAX_WEATHER_ID);

So that we can easily determine this call came from either:
- CheckLatentsWeather(...)
- Anywhere else

If that is fine, I will make changes in that direction.

src/map/latent_effect_container.cpp
TeoTwawki: regardless of what you do, if the value for the weather is *ever* the same as the value of MAX_WEATHER_ID your map server is going to crash on a debug break. You should never do either setting it to that or checking for it being that, because it should never be that value.

ShelbyZ: In attempts to avoid using an on-attainable value we could just do:
WEATHER GetWeather(CBattleEntity* PEntity, bool ignoreScholar)
WEATHER GetWeather(CBattleEntity* PEntity, bool ignoreScholar, uint16 zoneWeather)

Where the first method pulls the weather via zoneutils::GetZone(PEntity->getZone())->GetWeather() and calls the second

src/map/status_effect_container.cpp
```diff
@@ -447,6 +448,7 @@ void CStatusEffectContainer::RemoveStatusEffect(uint32 id, bool silent)
         PChar->PLatentEffectContainer->CheckLatentsFoodEffect();
         PChar->PLatentEffectContainer->CheckLatentsStatusEffect();
         PChar->PLatentEffectContainer->CheckLatentsRollSong(HasStatusEffectByFlag(EFFECTFLAG_SONG | EFFECTFLAG_ROLL));
+        PChar->PLatentEffectContainer->CheckLatentsWeather();
```
teschnei: now the last thing: since weather can change via status effect, can you add the weather latent effect check into CheckLatentsStatusEffect instead of calling this?

ziggy52:
you guys do know that sch's spell doesn't really change the weather right? all this trouble for an alternate way to check the but shouldn't you really be checking the legit real weather and then if a match isn't found, check the status effect of sch's spell?

teschnei:
uh.. yes? that's what this does

TeoTwawki:
@ziggy52 problem being dealt with is actually checking weather - it was looking at the zone to trigger before.

ShelbyZ:
@teschnei, @TeoTwawki  is there anything else that needs to be addressed?

TeoTwawki:
@ShelbyZ just so you know we haven't forgotten your PR, just been kinda busy irl 😄

PR #3854, opened 2017-04-19
Update Spirit surge to transfer wyvern TP. Wyvern gets Damage reduction.

soupcans:
I got information about wyvern 40% DMG reduction from these links:
http://forum.square-enix.com/ffxi/threads/11592-Dragoon-Manifesto-thoughts?p=178646#post178646
http://www.ffxivpro.com/forum/topic/24793/dragoon-update-additional-information/

Dragoon pet wyvern was updated with 40% damage reduction from both physical and magical damage in 2011.

Dragoon gets TP from pet wyvern during Spirit Surge
"The wyvern's current TP transfers to the Dragoon."
https://www.bg-wiki.com/bg/Spirit_Surge

### DIFF
--- a/scripts/globals/abilities/call_wyvern.lua
+++ b/scripts/globals/abilities/call_wyvern.lua
@@ -33,6 +33,4 @@ end;
 
 function onUseAbility(player,target,ability)
     player:spawnPet(PET_WYVERN);
-    local pet = player:getPet();
-    pet:addMod(MOD_DMG, -40);
 end;
--- a/scripts/globals/abilities/call_wyvern.lua
+++ b/scripts/globals/abilities/call_wyvern.lua
@@ -33,4 +33,6 @@ end;
 
 function onUseAbility(player,target,ability)
     player:spawnPet(PET_WYVERN);
+    local pet = player:getPet();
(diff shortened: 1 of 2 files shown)


scripts/globals/abilities/call_wyvern.lua
```diff
@@ -33,4 +33,6 @@ end;
 
 function onUseAbility(player,target,ability)
     player:spawnPet(PET_WYVERN);
+    local pet = player:getPet();
+    pet:addMod(MOD_DMG, -40);
```
TeoTwawki: I don't think this persist through zoning and needs to be "baked into" the object before you spawn it using mob_pool_mods.sql or else where the core builds our flying scaled friends stats in the first place. Possibly could put it in the global `scripts\pets\wyvern.lua` but I think thats intended to handle listeners only and dunno if we want to expand that to other things.

The pool ID for a players Wyvern is 5551.

teschnei:
Wyvern.lua's on spawn should be fine

TeoTwawki:
Didn't need to close, all you have to do is push to the same branch you used here, and github auto updates this.

PR #3856, opened 2017-04-19
Update pet wyvern to have 40% damage reduction

soupcans:
I got my information from these links:
http://forum.square-enix.com/ffxi/threads/11592-Dragoon-Manifesto-thoughts?p=178646#post178646
http://www.ffxivpro.com/forum/topic/24793/dragoon-update-additional-information/

Dragoon's wyvern was given 40 percent physical and magical damage reduction in 2011.

### DIFF
--- a/scripts/globals/pets/wyvern.lua
+++ b/scripts/globals/pets/wyvern.lua
@@ -37,6 +37,7 @@ local wyvernTypes = {
 
 function onMobSpawn(mob)
     local master = mob:getMaster()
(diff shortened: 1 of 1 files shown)


teschnei:
Is mod_dmg not 256 base?

TeoTwawki:
no that got changed to base 100 with some stuff in core to make it the correct fraction of 256 automatic. we really need to do that to the other weird fraction ones eventually.

teschnei:
Thanks, just making sure

PR #3857, opened 2017-04-24
Heroes of Abyssea

R3P0FFXI:
Skills and Spellists for Heroes of Abyssea expansion


sql/mob_skill_lists.sql
```diff
@@ -3334,4 +3334,311 @@ INSERT INTO `mob_skill_lists` VALUES ('Durinn',929,2114); -- Hellclap
...
+INSERT INTO `mob_skill_lists` VALUES ('Durinn',929,2118); -- Bilgestorm
+
+INSERT INTO `mob_skill_lists` VALUES ('Amarok',930,1785); -- Lava_Spit
+INSERT INTO `mob_skill_lists` VALUES ('Amarok',930,1786); -- Sulfurous_Breath
+INSERT INTO `mob_skill_lists` VALUES ('Amarok',930,1787); -- Scorching_Lash
+INSERT INTO `mob_skill_lists` VALUES ('Amarok',930,1788); -- Ululation
+INSERT INTO `mob_skill_lists` VALUES ('Amarok',930,1789); -- Magma_Hoplon
+
+INSERT INTO `mob_skill_lists` VALUES ('Ansherekh',931,270); -- Roar
+INSERT INTO `mob_skill_lists` VALUES ('Ansherekh',931,271); -- Razor_Fang
```
TeoTwawki:
```sql
INSERT INTO `mob_skill_lists` VALUES ('Tiger',242,270);
INSERT INTO `mob_skill_lists` VALUES ('Tiger',242,271);
INSERT INTO `mob_skill_lists` VALUES ('Tiger',242,273);
```
so no new list needed for Ansherkh

TeoTwawki: please recheck for other duplicate lists like this one

sql/mob_skill_lists.sql
```diff
@@ -3334,4 +3334,311 @@ INSERT INTO `mob_skill_lists` VALUES ('Durinn',929,2114); -- Hellclap
...
+INSERT INTO `mob_skill_lists` VALUES ('Bennu',933,401); -- Gliding_Spike
+INSERT INTO `mob_skill_lists` VALUES ('Bennu',933,402); -- Feather_Barrier
+INSERT INTO `mob_skill_lists` VALUES ('Bennu',933,922); -- Blind_Vortex
+INSERT INTO `mob_skill_lists` VALUES ('Bennu',933,923); -- Giga_Scream
+-- INSERT INTO `mob_skill_lists` VALUES ('Bennu',933,2703); -- Dread_Wind
+
+INSERT INTO `mob_skill_lists` VALUES ('Bugul_Noz',934,2210); -- Petal_Pirouette
+
+INSERT INTO `mob_skill_lists` VALUES ('Bugul_Noz_post_50',935,2210); -- Petal_Pirouette 
+-- INSERT INTO `mob_skill_lists` VALUES ('Bugul_Noz_post_50',935,2601); -- Petalback_Spin
```
TeoTwawki: This should be a pool check + hp check in `onMobSkillCheck` instead of a second list for just 1 skill.
You are off the hook since the skill does not exist yet. Just use the existing list for Nightshade.
```sql
INSERT INTO `mob_skill_lists` VALUES ('Nightshade',871,2210); -- Petal_Pirouette
-- INSERT INTO `mob_skill_lists` VALUES ('Nightshade',871,2601); -- Petalback_spin
```

sql/mob_skill_lists.sql
```diff
@@ -3334,4 +3334,311 @@ INSERT INTO `mob_skill_lists` VALUES ('Durinn',929,2114); -- Hellclap
...
+INSERT INTO `mob_skill_lists` VALUES ('Ningishzida',991,1832); -- Barofield
+INSERT INTO `mob_skill_lists` VALUES ('Ningishzida',991,1834); -- Trembling
+INSERT INTO `mob_skill_lists` VALUES ('Ningishzida',991,1835); -- Serpentine_Tail
+
+INSERT INTO `mob_skill_lists` VALUES ('Raja',992,2567); -- Stygian_Cyclone
+INSERT INTO `mob_skill_lists` VALUES ('Raja',992,2571); -- Stygian_Sphere
+INSERT INTO `mob_skill_lists` VALUES ('Raja',992,2572); -- Malign_Invocation
+INSERT INTO `mob_skill_lists` VALUES ('Raja',992,2692); -- Royal_Decree
+
+-- Rencounter_Chariot abilities not on wiki reserving 993
```
TeoTwawki: Don't reserve it. we'll just make up a new ID when we actually need to.

sql/mob_skill_lists.sql
teschnei: should it be a pool check?

I think the scripts look cleaner if they change skill lists in their script under 50%

TeoTwawki: its a really small bit in the skillcheck instead of checking if it shoudl change lists every cycle of onmobfight, seems saner to me to just check when we're already checking if it should use the skill if for no other reason than its shorter code. poolcheck only needs to be there if another mob uses it at a different % (or doesn't have a hp check at all). Have an example though:

```lua
function onMobSkillCheck(target, mob, skill)
    if (mob:getHPP() > 50 and mob:getPool() == 3326) then
        return 1;
    end
    return 0;
end;
```
that example is if other NM uses it whenever, but the the specified NM uses it only below 50% hp. 3 lines. most mobs either don't need an onMobFight or have other checks the minions would need to learn to not break.

TeoTwawki: You can always overule me :P but I like less onMobFight breakages

teschnei: yeah.. but you don't think that has the potential to become unruly with a bunch of pool checks?

TeoTwawki: very few times we'll need to use them. reason getPool even exists is for the rare times we need to check which NM this is.

TeoTwawki: let me clarify that a sec

TeoTwawki: what I mean is: its unlikely to have a ton of conditionals in any one skillcheck.

TeoTwawki: if we ran into one that became a rats nest of conditionals, I'd prolly convert it to multiple lists at that point. we almost did that to the ultima/omega skill scripts but I forget what made us do skillcheck exactly. I think it was the ton of hp checks it would have taken in onmobfight.

TeoTwawki: however we do'em in the future, I think in this mobs case it should just use nightshade,s list and we deal with it whenever the other skill gets made

R3P0FFXI:
@TeoTwawki did another quick run through corrected the tiger issue did not see anything else that matched

TeoTwawki:
@R3P0FFXI decided that unless you are already doing it, I'll switch over bugol noz myself after I merge because I'm going to create the other needles scripts and I'll be in there to uncomment those anyway

Issue #2764, opened 2016-02-15
Light/Dark arts not boosting skill if current main job does not have it natively


Hozu:
Base skill (ie. before merits/gear) is the higher of three values: Actual base skill (implemented), B+ - the amount the skill is from its actual cap, but only if the job has it natively (implemented), and some seemingly arbitrary value (not implemented). I say seemingly random because it appears to have no relation to any existing skill ranks, meaning a new one would be needed to be made for it.

As for the data, @teschnei collected it a while ago, and I recorded it. It can be viewed here: http://pastebin.com/S00mVFnf

PR #3858, opened 2017-04-26
Fixed Dark and Light arts not correctly increasing skill levels for…

UnknownX7:
…jobs with less than E skill


Hozu:
What does this commit change the skill levels to? https://github.com/DarkstarProject/darkstar/issues/2764 has data of what the value should be at minimum, gathered from retail.

teschnei:
I think it's just fixing a bug of using MJob instead of MLevel

Hozu:
Ah. I checked the various formula against the data, and it mostly lines up. Largest error is +/-1, and none of the levels divisible by 10 (except 70), 75, and 76-99 are wrong, so it's more or less correct. Accurate for level-capped content, unless there's a 70 cap thing I'm forgetting about.

UnknownX7:
> I think it's just fixing a bug of using MJob instead of MLevel
This is correct.

PR #3863, opened 2017-05-04
Added lines to populate petZoningInfo to allow writing pet info into …

sleepingkirby:
…char_stats table to retain pets on log out and login.

### DIFF
--- a/src/map/utils/petutils.cpp
+++ b/src/map/utils/petutils.cpp
@@ -1096,7 +1096,8 @@ namespace petutils
 
         Pet_t* PPetData = g_PPetList.at(PetID);
 
-        if (PMaster->objtype == TYPE_PC){
+        if (PMaster->objtype == TYPE_PC)
+        {
             ((CCharEntity*)PMaster)->petZoningInfo.petID = PetID;
             ((CCharEntity*)PMaster)->petZoningInfo.respawnPet = true;
 	}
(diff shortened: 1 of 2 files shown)


src/map/utils/petutils.cpp
```diff
@@ -1096,8 +1096,10 @@ namespace petutils
 
         Pet_t* PPetData = g_PPetList.at(PetID);
 
-        if (PMaster->objtype == TYPE_PC)
+        if (PMaster->objtype == TYPE_PC){
```
TeoTwawki: We're braces on newlines folk 'round here pardner 😃

sleepingkirby: Will make that change, but it turns out that it works for a single logout/login then something else goes and resets the pet info to zero. Might as well cancel this request until I figure that out >.<.

TeoTwawki: Its ok to leave open while you work on it and if you figure it out just updating your branch will automatic update the pull request. *Sometimes* we'll close out really old ones we think got abandoned.

TeoTwawki: logout should pretty much be the same as zoning out so surprised this isn't already a thing honestly.

sleepingkirby: Branch updated. ^^/

TeoTwawki: All set and tested now?

sleepingkirby: Yep, Should be all good. My wyvern hasn't disappeared on me while zoning or logging out yet. ^^b

PR #3870, opened 2017-05-11
Add ballista nation flags and their action IDs for quarry, sprint

Scavenge:
nation flags activate when allegiance is set
active nation flags enable the quarry and sprint options on the action menu
using these actions triggers action ID 15 and 16 in the console.

### DIFF
--- a/src/map/packets/char_update.cpp
+++ b/src/map/packets/char_update.cpp
@@ -87,7 +87,7 @@ CCharUpdatePacket::CCharUpdatePacket(CCharEntity* PChar)
     if (PChar->StatusEffectContainer->HasStatusEffect(EFFECT_TERROR))
         flag |= 0x08;
     if (PChar->allegiance > 1)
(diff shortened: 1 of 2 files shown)


src/map/packets/char_update.cpp
```diff
@@ -86,6 +86,15 @@ CCharUpdatePacket::CCharUpdatePacket(CCharEntity* PChar)
     uint8 flag = 0x20;
     if (PChar->StatusEffectContainer->HasStatusEffect(EFFECT_TERROR))
         flag |= 0x08;
+    if (PChar->allegiance > 1)
+    {
+        if (PChar->allegiance == ALLEGIANCE_SAN_DORIA)
+            flag |= 0x40;
```
teschnei: flag |= (allegiance << 5)

Scavenge: oh wow that makes sense.  did not expect it to make sense

PR #3873, opened 2017-05-14
More Abyssea QM/pos fixes and a few mods

Heepster03:
More Abyssea QM/pos fixes and a few mods


TeoTwawki:
er uh..you got multiple things mixed in here, need separate branches/pr's for stuff that isn't renaming QM's

TeoTwawki:
Lettin it slide because I know these are right, but in the future don't mingle

PR #3879, opened 2017-05-18
update allegiance flags for other players and correct typo

Scavenge:


src/map/packets/char_update.cpp
```diff
@@ -87,7 +87,7 @@ CCharUpdatePacket::CCharUpdatePacket(CCharEntity* PChar)
     if (PChar->StatusEffectContainer->HasStatusEffect(EFFECT_TERROR))
         flag |= 0x08;
     if (PChar->allegiance > 1)
-        flag |= (PChar->allegiance << 5);
+        flag = (PChar->allegiance << 5);
```
teschnei: this is wrong - it would always overwrite the 0x20 and terror flags

teschnei: and actually, the default allegiance is 1, which is 0x20 when left shifted 5 times.  So this whole thing can be simplified down to uint8 flag = PChar-> allegiance << 5 and the terror check after

Scavenge: so perhaps my original problem was that it was declaring uint8 flag = 0x20 before any of the other stuff and causing different values

Issue #3872, opened 2017-05-12
Cannot spend Guild Points at any guild rep


Spacemanap:
**_I have:_**

- [X] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [X] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **30170427_0**

**_Server Version_** (type `@revision` in game) **Unknown**

**_Source Branch_** (master/stable) **Master**

**_Additional Information_** (Steps to reproduce/Expected behavior) Players on my server cannot spend GP on guild items. All of the trade reps behave the same; the dialogue works but once a selection of and item is made the item is not added, the GP are not removed, and the dialogue loops back to the "Exchange points for..." option window. Things I've noted:
- Performs this way with HQ crystals (all of them), key items, and regular items.
- When selecting a HQ crystal and after entering the number of crystals desired, the dialogue box "Proceed with the transaction?" displays the correct GP amount possessed by the character. Upon going through the purchase process again without deactivating/reactivating the NPC the GP display will change to the itemID of the previous HQ crystal selected, not the character's actual GP amount. This does not happen with key items or regular items.
- If a character does not have the GP to buy an item, no notification is made and the dialogue loop continues.
- I'm working with the 1/16/2017 version of the crafting.lua file, completely unmodified.
- I last updated my client on 5/11/2017 to see if that would fix the issue (it did not).
- The server log produces no errors and does not display inserted print commands if the command is put into the crafting.lua unionRepresentativeTriggerFinish function, but the same command will display when inserted into the NPC script for the corresponding CSID option.
- Link to screenshots and dump of server log: https://drive.google.com/open?id=0B8DHaluZr_0Ed2FobE1MVXpkOVk
- Link to DSP forum post: https://forums.dspt.info/viewtopic.php?f=15&t=21297&p=35953#p35953

ddumont:
I'm seeing this as well on oldschool

PR #3881, opened 2017-05-20
Scope text ids in Windurst Waters

ddumont:
Possible fix for #3872 
The best way to look at this is with ?ws=1 on the url.


ddumont:
Note:  this should fix the cooking guild merchant

PR #3887, opened 2017-05-25
Adding quest Bastok Too many chefs

dwn134:
2 problems im having with this is the check for fame, and both options for the start of this quest are option 0 , so the quest will start no matter what you pick. Other than that everything works.

### DIFF
--- a/scripts/zones/Bastok_Markets/npcs/Umberto.lua
+++ b/scripts/zones/Bastok_Markets/npcs/Umberto.lua
@@ -2,6 +2,7 @@
 --  Area: Bastok Markets
 --   NPC: Umberto
 --  Type: Quest NPC
+-- Involved in Quest: Too Many Chefs
 -- @zone 235
 -- @pos -56.896 -5 -134.267
 --
@@ -23,7 +24,12 @@ end;
 -----------------------------------
 
 function onTrigger(player,npc)
-    player:startEvent(0x019b);
+
+	if 	(player:getVar("TOO_MANY_CHEFS") == 5) then -- end Quest Too Many Chefs
+		player:startEvent(0x01d9);
+	else	
+		player:startEvent(0x019b);
+	end	
 end;
 
 -----------------------------------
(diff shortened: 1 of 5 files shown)


ibm2431:
> is the check for fame

What's the problem with the fame check? Is the quest available event playing when the person has insufficient fame? Not playing when they supposedly do have the correct fame?

> both options for the start of this quest are option 0 , so the quest will start no matter what you pick

Is this how retail behaves? Do the contents of the event for both options indicate the quest is being accepted?

dwn134:
The fame check problem im have on my test server is i can recieve the quest
no matter what my fame is. I have only tested on my GM account and not a
reg account if that makes a difference.

dwn134:
It could be very possible either option starts the quest. I cant see to find any evience that it wouldnt. In that case disregard that part

TeoTwawki:
> The fame check problem im have on my test server is i can recieve the quest no matter what my fame is. 

show us what your fame check looked like.

> both options for the start of this quest are option 0 

```lua
    -- printf("CSID: %u",csid);
    -- printf("RESULT: %u",option);
```
Wen you uncomment the prints, both menu selections have RESULT of zero on that csid?

dwn134:
Yes both option show as 0 when un commented

TeoTwawki:
Cloned your branch local and took a look. Fame check worked fine for me so guessing you had the fame boost setting used or something..

I also verified retail does start the quest no matter which option you choose, and then cleaned up your [whitespace](https://github.com/DarkstarProject/darkstar/pull/3887/commits/0b8a65148b4a92bce87ecc301461c73c585ff8f4#diff-d11d9c73b0e9aa8160f756b64893bc5dL24). Project uses 4 spaces, not tabs. It looks like you used githubs crappy web editor.

dwn134:
I was using notepad ++  I seem to get some error when i use spaces but tabing to the right location seems to work. Then copy and pasted to my fork. This was my 1st time doing this so im happy it got merged

TeoTwawki:
Whatever you copy and pasted *into* from notepad++ is likely the culprit. We avoid github's app and its web UI for editing files. They encourage bad commit messages, stabs, and other problems.

In notepad++ you can still actually use the tabkey and automatically get 4 spaces if you edit the settings to do so. If you are having trouble with some kind of error, you should see about getting help with getting that fixed. I have a reconfigured portable copy (runs from USB device, no installation) should you need one.

Thanks for contributing. 😄

dwn134:
The issue i was getting was someone sending me a file from pastebin that i
copy and pasted and i had to delete every space in the whole script for it
to work. Its a learning process for sure But thanks you guys for putting
this game together

PR #3882, opened 2017-05-21
Update packet parsing for guild vendor (not honoring quantity)

ddumont:
Fixes #3872

Is there a function for adding multiple items to a player's inventory?

PR #3888, opened 2017-05-26
Version update 2017/04/04 (client: 30170329_0)

ibm2431:
Requesting that this update from March->April gets PR'ed into master, prior to the update up to May @TeoTwawki mentioned in #3882 (which would be March->May), so that commit log has a history of what changed March->April, and then April->May.

Sorry if this would cause you to have to do a rebase, Teo. D:

### DIFF
--- a/scripts/zones/Bastok_Markets/TextIDs.lua
+++ b/scripts/zones/Bastok_Markets/TextIDs.lua
@@ -17,53 +17,53 @@ FULL_INVENTORY_AFTER_TRADE =  6383; --Try trading again after sorting your inven
        NO_MORE_GP_ELIGIBLE =  7099; --You are not eligible to receive guild points at this time.
                GP_OBTAINED =  7104; --Obtained: ?Numeric Parameter 0? guild points.
         NOT_HAVE_ENOUGH_GP =  7105; --You do not have enough guild points.
-    FISHING_MESSAGE_OFFSET =  7200; --You can't fish here.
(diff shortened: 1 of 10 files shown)


TeoTwawki:
That actually isn't necessary, and is gonna confuse my minions and when confused ppl tend make mistakes get frustrated and give up on things.

TeoTwawki:
I'd rather we didn't bother up obsoleted updated just to have extra client revisions in history (unless your server needs that revision)

TeoTwawki:
Just a heads up I am prolly gonna merge this afterall things seem to have already gone sideways with the work in progress so I''l re-sync it to the next client myself,

TeoTwawki:
@ibm2431 so get this, not a single text ID changed since your revision to the current client. NONE. The only text ID's I found changed that weren't in your PR were caused by the default paths in the tool missing their ending slash. Maybe I'll be lucky and the same is true for npc IDs (I can dream right?)

ibm2431:
Oh, just got the chance to check the internet (access is very sporadic), glad it was of help.

You shouldn't have any changes to NPCs other than in Legion~! Those have changed every month. Usually just shifts every NPC in the zone up or down a few IDs as SE adds/removes Ambuscade mobs from month to month.

I haven't seen any other zones (truly) shift in a long time, despite what the tool might like you to believe for a few of them. Feel free to ignore it on things like Apollyon chests, Shadow Lords in Throne Room, etc.

TeoTwawki:
thanks, @Heepster03 is double checking for me before I bump the version check again

PR #3889, opened 2017-05-26
There's no such thing as MOD_PET_ATTP

ddumont:


ddumont:
@TeoTwawki Like this?

ddumont:
I don't think this is all that's needed.  I mean this just calculates a boolean, looks to be related to meriting...    where are these values actually applied?  I am not actually sure we don't need this constant anymore.

ddumont:
I'm concerned just changing this to MOD_ATTP will affect the player and not the pet.   The problem is the buff here is on the player (it does not apply to the pet, it travels with the player).  If you dismiss and resummon the pet, the new pet should get it as long as the master has the buff.

Where do these things get looked at?  Should we bring back a constant for this or can we distinguish?

TeoTwawki:
I *think* what you wanna do is something like:
```lua
local pet = target:getPet();
```
and then instead of doing it to the target, do it to the pet

TeoTwawki:
```lua
    if (pet:addCorsairRoll(caster:getMainJob(), caster:getMerit(MERIT_BUST_DURATION), EFFECT_BEAST_ROLL, effectpower, 0, duration, caster:getID(), total, MOD_ATTP) == false) then
```

ddumont:
hmm.  well the *buff* needs to be tracked on the player... the effect needs to be added to the pet.

I see the beast_roll.lua in effects that does:
````lua
    target:addPetMod(MOD_ATTP, effect:getPower());
end;
````

I'm just not really sure how this gets called... and (off topic) why we even specify the constant here when it should be built in the effect?

ddumont:
if I did what you said... how would the effect be applied to future pets that were changed?

TeoTwawki:
Sorry I'm not real familiar with how cor roll stuff was built and sound like I got it wrong, paging @teschnei

teschnei:
Bust needs to know the modifier type, because its one status for every roll.  So you'll need to figure out a way to fix the pet mods in there.

teschnei:
Basically your current diff will just make beast roll's bust lower YOUR attack instead of your pets (and previously it just did nothing because that mod didn't exist).

ddumont:
Ahhhhhhh   ok I see.    This is just for the bust application.

Where is the code for the normal application?    How are pet rolls tracked across desummon / summon?

ddumont:
@teschnei So basically these pet mods should apply a status of really... nothing to the owner.
And somehow the real status should be modified on the pet in the event that the owner gets the status, loses the status, or summons it while it's active?

ddumont:
ok so, assuming that for pets, the ID here really doesn't matter....   I'm going to change this to nil to reduce confusion, and possibly add a comment.

It looks like AddStatusEffect will properly call the onEffectGain when the effect is granted, which will 
````lua 
target:addPetMod(MOD_ATTP, effect:getPower());
````

Great.  Where might I look to make sure that the effect is added to a pet if it is summoned when the owner already has the effect?

ddumont:
@teschnei How's this? Am I on the right track?

teschnei:
no, no.. it was already correct in the first place (using addPetMod in the effect's lua).  The only thing that was wrong was bust.

PR #3895, opened 2017-05-29
Fixing issues where skill level checks were getting extra bonuses

ShelbyZ:
- getSkillLevel(...) includes merits + level + equip bonus
- getMod(...) inlcudes equip bonus
- many places were calling getSkillLevel + getMod giving extra equip
bonus

### DIFF
--- a/scripts/globals/abilities/elemental_siphon.lua
+++ b/scripts/globals/abilities/elemental_siphon.lua
@@ -43,7 +43,8 @@ function onUseAbility(player,target,ability)
         spiritEle = 3
     end;
     
-    local basePower = player:getSkillLevel(SKILL_SUM) - 50;
+    local pEquipMods = player:getMod(MOD_ENHANCES_ELEMENTAL_SIPHON);
+    local basePower = player:getSkillLevel(SKILL_SUM) + pEquipMods - 50;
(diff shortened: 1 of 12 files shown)


ShelbyZ:
While diving into magic.lua, I noticed that some places were accounting for skill mods (equip bonuses) when doing calculation and did some double checking resulting in what looks like minor double bonuses from equips in a good deal of formulas.

scripts/globals/abilities/elemental_siphon.lua
```diff
@@ -43,8 +43,7 @@ function onUseAbility(player,target,ability)
         spiritEle = 3
     end;
     
-    local pEquipMods = player:getMod(MOD_ENHANCES_ELEMENTAL_SIPHON);
```
Hozu: I'm not sure why you're removing this? This is not a bonus from skill, but a separate modifier on items such as this: http://www.ffxiah.com/item/27439/beck-pigaches

ShelbyZ: Well that throws a wrench into things, I'll need to test that bonus values from equips are reported by getMod and if so remove the skill bonus from the method while keeping in bonus effects.

PR #3896, opened 2017-05-30
Updating NPC IDs for newest Version Update.

Heepster03:
Also updated version.info to 30170427_1


TeoTwawki:
verified myself, just waiting on travis in case of some missing semicolon or something.

ibm2431:
Big jump in city zones. That time of year already, huh?

PR #3898, opened 2017-05-31
More Abyssea Fixes

Heepster03:
Mob Pools - Fixed Sedna as it was unspawnable before.
Mob Groups - Created a group for Rakshas since it was using Quasimodo's ID.
Mob Spawns - Fixed locations for some NMs in Abyssea, , not perfect retail but they are usable now.


ibm2431:
> Fixed locations for some NMs in Abyssea, , not perfect retail but they are usable now.

I recaptured Abyssea some time last year, but never got around to going through and correcting any mobs. Can find my captures at: https://drive.google.com/open?id=0B6J1HCKIjYoTaEFnQzVHemYxWWM

There's some Lua tables in the `/tables/` folder, which should have all the information DSP usually uses processed already (ie: you don't need to parse the packets in `/raw/`).

And if there's any interest in mob levels, those can be found in the `/widescan/` folder (levels are sent in the widescan packet). With the exception of someone cleaving bugards in Grauberg to farm Azdaja (so the skeletons may be inaccurate too), all the zones were dead empty when I was running through them, so mobs should be in their natural state.

Please pardon the occasional Randy and Trusts, I had them out so I wouldn't be entirely ganked by any True Sight NMs I blindly ran into. And I was a little lonely.

Sorry the tables aren't sorted by ID, they were logged in order of receiving the spawn packets. You might be able to sort them with Excel or something (input table into Excel, then just sort by column).

PR #3905, opened 2017-06-07
Adding cut scene to winning "The Shadow Lord" fight when winning for …

sleepingkirby:
…the first time.

Made sure the cutscene happened before being warped out.

### DIFF
--- a/scripts/zones/Throne_Room/bcnms/shadow_lord_battle.lua
+++ b/scripts/zones/Throne_Room/bcnms/shadow_lord_battle.lua
@@ -32,7 +32,7 @@ function onBcnmLeave(player,instance,leavecode)
         if (player:hasCompletedMission(player:getNation(),15)) then
             player:startEvent(0x7d01,1,1,1,instance:getTimeInside(),1,0,1);
         else
-            player:startEvent(0x07,1,1,1,instance:getTimeInside(),1,0,0);
+            player:startEvent(0x7d01,1,1,1,instance:getTimeInside(),1,0,0);
         end
     elseif (leavecode == 4) then
         player:startEvent(0x7d02);
@@ -45,18 +45,18 @@ end;
     
 function onEventFinish(player,csid,option)
     -- print("bc finish csid "..csid.." and option "..option);
-    print("bc finish csid "..csid.." and option "..option);
-    -- if (csid == 0x7d01) then
-    if (csid == 0x07) then
+    if (csid == 0x07d01) then
         if (player:getCurrentMission(player:getNation()) == 15 and player:getVar("MissionStatus") == 3) then
             if ((not player:hasCompletedMission(ZILART, THE_NEW_FRONTIER)) and (player:getCurrentMission(ZILART) ~= THE_NEW_FRONTIER)) then
                 -- Don't add missions we already completed..Players who change nation will hit this.
                 player:addMission(ZILART,THE_NEW_FRONTIER);
             end
-            player:setPos(378.222,-12,-20.299,125,0xA1);
-            player:addKeyItem(SHADOW_FRAGMENT);
-            player:messageSpecial(KEYITEM_OBTAINED,SHADOW_FRAGMENT);
-            player:setVar("MissionStatus",4);
+            player:startEvent(0x07);
         end
+    elseif (csid==7) then
+        player:setPos(378, -12, -20, 125, 0xA1);
+        player:addKeyItem(SHADOW_FRAGMENT);
+        player:messageSpecial(KEYITEM_OBTAINED,SHADOW_FRAGMENT);
+        player:setVar("MissionStatus",4);
     end    
 end;
--- a/scripts/zones/Throne_Room/bcnms/shadow_lord_battle.lua
+++ b/scripts/zones/Throne_Room/bcnms/shadow_lord_battle.lua
@@ -32,7 +32,7 @@ function onBcnmLeave(player,instance,leavecode)
         if (player:hasCompletedMission(player:getNation(),15)) then
             player:startEvent(0x7d01,1,1,1,instance:getTimeInside(),1,0,1);
         else
-            player:startEvent(0x7d01,1,1,1,instance:getTimeInside(),1,0,0);
+            player:startEvent(0x07);
         end
     elseif (leavecode == 4) then
         player:startEvent(0x7d02);
@@ -61,5 +61,7 @@ function onEventFinish(player,csid,option)
             -- You will be transported back to the entrance of Castle Zvahl Baileys
             player:setPos(378.222,-12,-20.299,125,0xA1);
         end
+    elseif (csid == 0x07) then
+        player:startEvent(0x7d01);
     end    
 end;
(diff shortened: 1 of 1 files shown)


teschnei:
i don't think i've ever seen an exit cutscene not be a 32000 numbered one - i think i would have to look into it before merging it

sleepingkirby:
Take your time. That was the only way (short of writing my own lua commands) to get it to run. It 0x07 doesn't actually exit the player from bcnm instance. It just plays the cutscene. But what happens if you let 0x7d01 (32001) run first is that it'll exit and show the player into the Throne Room zone, then play the cutscene that happens inside the BCNM. If you have it start the cs event 7 before 0x7d01 is that it'll try to run 0x7d01 while 7 is still running and fail the exit. it is an option to have it go from event 7 to warpping to the entrance of castle vahzl and setting up the quest there. But for some reason, I remember the retail version warping to the Throne Room zone first before warping them out. I could be totally wrong on that though.

teschnei:
You could check a youtube video if you've no other option.  But I suspect that 0x07 is only for replaying the cutscene at goblin footprints (as it wouldn't want to teleport you around anywhere).

sleepingkirby:
... not sure why I didn't think of that. Probably the 4 hours of sleep >.<. Just checked. A successful fight means warping to the entrance. I'll update.

sleepingkirby:
So, I've updated the script as close as to the retail I can. The original has the player warped by the time the CS ends but we really can't do that with the OnBcnmLeave or OnEventFinish. The closet we can get is the CS finishes, then it warp you from the BCNM to the front of Castle Vahzl. I'll look into making another function called "OnBcnmWin" or something later. I see a win evaluation function that runs on a period check in the c++ code. There should be a way to make a LUA function to run on win conditions being satisfied or something that can then be hooked into.

teschnei:
can you link the video?

sleepingkirby:
Here you go.
https://www.youtube.com/watch?v=qO1hf4FBtpo

teschnei:
What's wrong with OnEventFinish in this case? it looks like that's all this
does

sleepingkirby:
in the video, you beat the fight, the CS happens, you appear at the start of castle vahzl. 
With OnEventFinish, the teleport happens *after* the cs is over. Meaning, you fight, the CS happens, you're brought out of the cs into the bcnm still, the teleport happens. 

I know it may be trivial to some people. But I'd like to get it as close to retail as possible.

teschnei:
I don't think it's possible to do it any other time.  There's nowhere
sooner to do it - you can't zone during the cutscene.

Hozu:
Hm... to me, that video looks like you're put in a black screen on CS end, since that's when the post-battlefield text comes up, then it warps you. So the CS ends, but you don't see yourself in the battlefield zone at the end of it. I'm assuming it's done that way to disguise the fact that you're in the Throne Room zone at the end of the CS, despite the CS showing you at the entrance of the castle.

Hozu:
Huh. I guess it might not be possible due to client limitations. Do note that DSP wants to to be accurate to current retail simply because it can be tested against.

gedads:
here it's not like there's any choice, i went to that area some months ago and i couldn't get the right exit cs to play like it was played before. you can still get some mid fight cs by playing with the parameters but the ending cs you just can't and it's probably because it's done in a new way in retail game.
Most of the new bcnms are like that, it has been like that since high tier mission were added

sleepingkirby:
I think Hozu got it right. I didn't notice that before. Also didn't know you can't zone during a cs. SE did a lot of tricks didn't they? Also learned (by accident) cs 7 automatically warps you out to the front of the keep. I'm gonna ticker with this a little more.

sleepingkirby:
okay... so if you're in the Throne Room and you run cs 7, it warps you out. if you play the cs when you just entered the bcnm, you get warped out. But if get played after you kill shadow lord, you stay in the BCNM... wtf?

sleepingkirby:
Okay, so once you attack and/or the bcnm is locked, it prevents the cs from teleporting you out.

sleepingkirby:
okay, results of my testing:
Playing cs 7 when bcnm is locked --> cs 7 doesn't warp you out.
Playing cs 7 manually when shadow lord dies --> cs 7 doesn't warp you out.
 
I tried playing event 32001 (which just warps you just outside the bcnm into the Throne Room) then when that finishes, play event 7 (the ending cs). But something still prevents you from warping out. Using the manually invoking cs 7 after that warps you out.

So I guess the biggest question is, what about bcnm locking/starting that prevents a cs from warping you out?

gedads:
oneventfinish add a zone player to throne room exit location

gedads:
actually, when you check the latest video, you can see directly the cs appearing outside then player is zoned to bcnm exit (loading screen) then onzone in cs 7...

sleepingkirby:
*checks the latest video* rngsus f**kig christ. This is where I want to roll up a newspaper and start smacking SE on the nose while going "What. did. you. do? What. did. you. break?" If that's how the latest client handles it, that means they broke something and this is a work around. >.< Oh well. At least that explains why I couldn't match the behavior of the older video and that I remember from like 2003.

It looks like to me:
win battle --> event 32001 (warps you outside the BCNM and into the Throne Room) --> a zone warp call to the same location --> event 7 for cutscene --> cut scene drops you off at the entrance to castle vahzl

Sound right to the rest of you guys?

gedads:
Not sure that they broke anything, all recent bcnms are handled like that, i was more thinking about an intention to free space (high tier missions can be really busy when you farm relic items and rem's tale)

Hozu:
So uh, does DSP handle the CS like current retail or not? From what I remember, it does, but I haven't seen it recently.

sleepingkirby:
@gedads it has to be something they broke. Anytime you write an object/class, you need to write in the basics (Create, Read, Update, Delete. Or "CRUD"). The fact that a code can set a setting (prevent cs's from warping you out) and not be able to unset said setting, but instead, requiring a zone to the exact location to unset said setting, means they broke something that they didn't know how or didn't have time to fix. 
Think of it this way. If you lose this fight, you get exited with event 32002. When that happens, it doesn't prevent the next group from entering. If that's the case, then the rezone has nothing to do with clearing the bcnm registrations and/or lock and/or instance and, instead, everything to do with allowing event 7 to run correctly. But we know that in previous incarnations of the game, event 7 ran just fine. In which case, event 7 not being able to run correctly is a new thing that prevented a behavior which was the, in both the aesthetic sense and logistical sense, the better option. 

With that said, I'll make the lua script reflect the new retail behavior and test probably tomorrow. Finally done with d*mn finals, but still got some personal projects to head up. No rest for chaotic evil,  I guess.

sleepingkirby:
@Hozu 
I'm pretty new contribution so I'm not an authority in anyway. From what I've seen from playing around the server and the code is that they try to match the retail. But there's also conflict between retail now and retail, say, back in 2002. A lot has changed (this being an example.), sometimes for the better, sometimes for the worse. And the team has tried to find a way to allow for both behavior via settings or find a happy medium. But it's not always easy.

sleepingkirby:
lua script updated. It doesn't match the retail 100%. I don't know of a way via lua to wait to execute a script after a setpos command finishes and the custom lua wait function I wrote runs out of order. So now it goes:
win fight --> event 32001 warp to outside bcnm --> run event 7 --> setPos() to castle vahzl entrance --> sets KI's and missionstatus to 4.

which, ironically, is cleaner than the SE version...

ibm2431:
Don't forget that you have `onZoneIn` at your disposal. You can `setVar` a flag of some kind, do the `setPos`, check for that flag in the zone you're sending them to, do whatever you need, and then nuke the variable.

PR #3906, opened 2017-06-14
Preliminary Fix for Synth/Bazaar Issues

ShelbyZ:
- I do not think this is the perfect solution, but it should at least
start a conversation or at worst be a bandaid

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -5536,7 +5536,7 @@ void SmallPacket0x10A(map_session_data_t* session, CCharEntity* PChar, CBasicPac
 
     CItem* PItem = PChar->getStorage(LOC_INVENTORY)->GetItem(slotID);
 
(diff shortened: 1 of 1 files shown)


ShelbyZ:
I did some poking around in the packets thinking that maybe the synth process wasn't correctly locking/unlocking the items while a craft was active, but it was preventing me from modifying it in my inventory.  Sadly bazaar pulls the listing and this lockout wasn't present and I went for the 'first idea' that came to mind.  It may be possible to directly modify that item list the bazaar is getting which should be the better solution.

Alice5150:
Tested, works great.

teschnei:
The bazaar just uses your inventory.  Not sure why INV_NOSELECT doesn't
work in the bazaar menu - might want someone to check retail.

This is probably correct though - anything that's locked should not be
bazaarable, regardless of how it was locked.

PR #3909, opened 2017-06-14
Make SMN AF quests available

Brierre:
The Puppet Master will keep repeating, rather than allow people to get the next AF quest.  Added a check for the weapon so it will go on.  Drop the weapon and you will get The Puppet Master again.

### DIFF
--- a/scripts/zones/Windurst_Walls/npcs/_6n2.lua
+++ b/scripts/zones/Windurst_Walls/npcs/_6n2.lua
@@ -41,7 +41,7 @@ function onTrigger(player,npc)
             player:startEvent(0x0181,1125,1125,1125,1125,1125,1125,1125,1125);
         ----------------------------------------------------
         -- The Puppet Master (AF weapon)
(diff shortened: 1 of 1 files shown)


TeoTwawki:
its because its being checked for `~= QUEST_ACCEPTED` instead of `== QUEST_AVAILABLE`. these arent repeatable like this on retail (redoing af quests on retail requires that memory wipe thing, which removes the quests so you can do them again)

```
you can repeat the artifact quests by talking to Vingijard in Lower Jeuno. You do not need to discard your other AF first. You may reset the quests and start them again, even if you still have all of your original AF. But you will need to either discard your af set or store it at an Armor Depository. You must have completed all the quests and opened all the coffers for a particular job before you can reset the quests. 
```
http://ffxiclopedia.wikia.com/wiki/Category:Artifact_Armor

Brierre:
Ok, as is, it just keeps repeating puppet master forever. So with a change to quest available, it would also work to move somebody along without checking for the weapon?

TeoTwawki:
yeah its still triggering on completed because it looks for "different than accepted" and should not

TeoTwawki:
prolly more af quests with similar faulty logic

Brierre:
Perhaps, but this one was changed a little while back to put a couple other quests after this one in the overall order.  We didn't notice before that because it would hit the other quests first, I guess.  I tried reordering them, but then saw that change was done previously to fix something else.  I have to run out to work, so I can't change/test/submit the other thing right now, but later I can get to it.

PR #3911, opened 2017-06-15
Adding lua hooks to modify mob skills on the fly per mob

ShelbyZ:
- Adds onMobWeaponSkillPrecast that can be added to any zone mob lua to
change parameters similar to onSpellPrecast
- Expose methods on lua mob skill object to set distance, aoe, flags,
knockback

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -283,6 +283,9 @@ bool CMobController::MobSkill(int wsList)
         {
             continue;
         }
+
+        luautils::OnMobWeaponSkillPrecast(PMob, PMobSkill);
+
         if (PMobSkill->getValidTargets() == TARGET_ENEMY) //enemy
         {
             PActionTarget = PTarget;
@@ -319,6 +322,8 @@ bool CMobController::TrySpecialSkill()
         ShowError("CAIMobDummy::ActionSpawn Special skill was set but not found! (%d)\n", PMob->getMobMod(MOBMOD_SPECIAL_SKILL));
         return false;
     }
+
+    luautils::OnMobWeaponSkillPrecast(PMob, PSpecialSkill);
 
     if (!IsWeaponSkillEnabled())
     {
--- a/src/map/lua/lua_mobskill.cpp
+++ b/src/map/lua/lua_mobskill.cpp
@@ -167,6 +167,42 @@ inline int32 CLuaMobSkill::getHPP(lua_State* L)
     return 1;
 }
 
+inline int32 CLuaMobSkill::setFlag(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PLuaMobSkill == nullptr);
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, -1) || !lua_isnumber(L, -1));
+
+    m_PLuaMobSkill->setFlag(lua_tointeger(L, -1));
+    return 0;
+}
+
+inline int32 CLuaMobSkill::setAoE(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PLuaMobSkill == nullptr);
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, -1) || !lua_isnumber(L, -1));
+
+    m_PLuaMobSkill->setAoe(lua_tointeger(L, -1));
+    return 0;
+}
+
+inline int32 CLuaMobSkill::setDistance(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PLuaMobSkill == nullptr);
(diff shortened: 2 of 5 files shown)


ShelbyZ:
Example lua addition:

```lua
-- onMobWeaponSkillPrecast

 function onMobWeaponSkillPrecast(mob, skill)
     skill:setAoE(1);
     skill:setFlags(0x008); -- hit all
     skill:setDistance(30);
     skill:setKnockback(3);
 end;
````

Could also use the skill id to filter changes based on specific skills instead of the blanket changes to all skills.

teschnei:
Seems redundant, do you have an example use case?

TeoTwawki:
I've wanted something like this for awhile, but I tried to handle it in the skill scripts themselves rather than the mob script (which turns out even the skillcheck was too late to effect the skilluse). Thumbs up from me.

Bonus goldstar if you manage a working setAnimation binding for them too.

TeoTwawki:
@teschnei multiple mobs with same named skill and diff properties, but lack of a duplicate name in the skill list. Or just avoiding an if else tree in the skill script? I will admit my main reason to want it has to do with *"making things up"* rather than a retail excuse. Which isn't to say one doesn't exist, I just haven't thought of one that would have let me foist it on you (yet).

teschnei:
yeah, I get that, but... are there any actual examples of a mobskill that has different properties that is not a duplicate?

(knockback excluded, that has no business in the mobskill in the first place)

teschnei:
also, if that's what you *just* said, I may have skipped over the second half of your post..

TeoTwawki:
d'aww

pleading eyes

ShelbyZ:
I was thinking from the point of view of an event planner, I want my mob to make use of these skills with non-default properties.  Is something like Spike flail into that category where it should behave differently NQ vs HQ?  If I did modify the skill scripts would the changes be applied before the mob used?

teschnei:
Fafnir and Nidhogg have different Spike Flails, so it doesn't need to be modified in the script (they have entirely different skill IDs).  I think every mobskill that has different properties like that will have its own skill ID (probably because those properties are stored in a similar way on SEs database as ours)

ShelbyZ:
I'll give in, it was more of a convenience factor for mob skill control similar to spell control in the mob's lua.

teschnei:
I think most of it already works with the lua_action (or at least, animation would, in Teo's case, and knockback, as I pointed out earlier).

Distance and flags no, but I'm not sure if those are as important as the other two

teschnei:
actually lua_action has no knockback setter, you could add that there instead

ShelbyZ:
I'm not sure I understand using lua_action, if we are talking about the skill scripts, I would expect to be interacting with a CLuaMobSkill.

teschnei:
eh well.. i don't want yet another binding for something we can already do somewhere else.  we already have so many, it's such a waste to parse the file each time and do nothing with it in 95% of cases

besides, something like knockback and animation... those are part of the action, not the mobskill (same with message but we haven't got around to fixing that atrocity yet)

TeoTwawki:
> besides, something like knockback and animation... those are part of the action, not the mobskill 

aaah ic.

> (same with message but we haven't got around to fixing that atrocity yet)

:rofl:

ShelbyZ:
So the CLuaAction should be calling into an action_t.  If this was to be done the right way, it would require exposing those values on action_t, CLuaAction, and reworking luautils to push out CLuaAction in place of CLuaMobSkill (in the case of mob skills)?

I'm not sure that is something I would address in this commit.  I would rather start from scratch and convert an existing on-something to use the ideal solution.

teschnei:
It just needs a CLuaAction passed in OnMobWeaponSkill, just like the other 3 bindings that use it already.  That's it, just 2-ish lines

teschnei:
I did it in ee106869513866949634026391df5761762d137a, but CLuaAction still has no knockback

ShelbyZ:
Adding the knockback to CLuaAction would solve modifying knockback, but what about control over distance, AoE, and flags?  Would it still be useful to provide those on CLuaMobSkill or is there any route?

teschnei:
Are there any examples of a mobskill that has those properties that are different that don't use a different mob skill ID?

Would they be useful enough to warrant another script call on every mobskill?

ShelbyZ:
I cannot provide any, I'm still arguing for the extensibility of the system.  I would not expect any current mob skills to make use of these changes.  I would lean towards exposing this additional functionality as optional/not required.  The only way it would warrant another call (or any for that matter) would be if DSP moved the mob skills completely out of the database, but I do not suggest that is the direction that should be taken.

I think at this point, the addition of a precast for mob skills does not add value (duplicates existing functionality).  With that in mind, should I close out this merge?

teschnei:
I'll talk with teo about it later maybe first

PR #3932, opened 2017-07-10
Chlevnik additions

CTOON50:
Added spell list that contains Meteor, and added BLM as subjob for Chlevnik to give MP to cast.


TeoTwawki:
You can get away with re-using King Behemoths spell list without creating a new one, sicn ethey are identical. The list ID is 126 and contains only Meteor.

CTOON50:
Completely overlooked that...should have remembered as I just messed around with KB not too long prior.

TeoTwawki:
p.s. I wouldn't mind if you changed that spell lists name field to say something like "Casting_Behemoths" or something else accurately descriptive since its not just KB anymore. Up to you, not required.

(the name field there is unused by anything in the code and is just there for at a glance identification by us fleshy meat puppets)

CTOON50:
There we go, though somehow an extra line was added to the spell list

TeoTwawki:
eh that's ok

PR #3933, opened 2017-07-10
Fixing Quick Draw for Corsair.

Heepster03:
-- It was 1 second, fixed it to 1 minute as it should be.

### DIFF
--- a/sql/abilities.sql
+++ b/sql/abilities.sql
@@ -140,7 +140,7 @@ INSERT INTO `abilities` VALUES (99,'puppet_roll',17,52,1,60,193,420,0,115,2000,0
 INSERT INTO `abilities` VALUES (100,'dancers_roll',17,61,1,60,193,420,0,167,2000,0,6,8.0,1,1,80,0,8,'WOTG');
 INSERT INTO `abilities` VALUES (101,'scholars_roll',17,64,1,60,193,420,0,168,2000,0,6,8.0,1,1,80,0,8,'WOTG');
 INSERT INTO `abilities` VALUES (107,'double-up',17,5,1,8,194,424,0,116,2000,0,6,8.0,1,1,80,0,0,'TOAU');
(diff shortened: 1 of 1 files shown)


teschnei:
What's the problem? It's one charge - the old SQL is right

Heepster03:
The problem is that it's spammable

Heepster03:
Each of the shots are set to 1 second, therefore if quickdraw recharges a "charge" in 1 second, then it's spammable. Setting the QuickDraw to 1min then it isn't spammable and according to bgwiki/ffxiclopedia, it's 1min recast.

teschnei:
I need to verify this before anyone merges.

Heepster03:
https://www.bg-wiki.com/bg/Quick_Draw

http://ffxiclopedia.wikia.com/wiki/Quick_Draw

Take your pick, each clearly state 1min timers.

TeoTwawki:
@heepster03 not what he meant just calm and wait bruh

Heepster03:
Oh, one more thing to add to maybe make things easier.

"Initially you have 2 Quick Draw charges. These recharge at 1 per minute. Each Quick Draw Recast merit will reduce this by 2 seconds to a maximum of -10 seconds with the full 5 merits."

teschnei:
Can't reproduce.  It's working as intended, seems like your server is
missing the abilities_charges table.

teschnei:
Sorry, yes, I know how Quick Draw works (I wrote the entire thing), I just
meant I needed to verify the behaviour you were seeing.

TeoTwawki:
that should leave a log error shouldn't it?

teschnei:
Maybe the table is there but it's empty - but the behaviour he is seeing would be explained by Quick Draw not being joined to abilities_charges properly

TeoTwawki:
investigation results: Old bad data in char_recast table

PR #3937, opened 2017-07-15
Reworking latent_effect_container

ShelbyZ:
- Adding getmod to check the mod value of a player - getmod $modid
- Sorting latent_effect.h and adding placeholders for unused values
- Adding three helper methods to streamline latent processing:
ProcessLatentEffects, ProcessLatentEffect, and ApplyLatentEffect
- Removing GetLatentEffect definition in header - not declared/used
anywhere
- Removing parameters from some latent check functions because the value
is available in the checking code already
- Updating functions that called latent check functions that previously
required parameters
- We currently duplicate "ALOT" of latent checking code and in some
instances we would be mid-loop processing and dive into another loop of
processing for specific latent types (I made this mistake for weather,
but prior cases existed)

### DIFF
--- a/scripts/commands/getmod.lua
+++ b/scripts/commands/getmod.lua
@@ -1,24 +1,31 @@
 ---------------------------------------------------------------------------------------------------
--- func: getmod
--- desc: gets a mod on the target/player
+-- func: getmod <modID>
+-- desc: gets a mod by ID on the player or cursor target
 ---------------------------------------------------------------------------------------------------
 
 cmdprops =
 {
     permission = 3,
-    parameters = "is"
+    parameters = "i"
 };
 
-function onTrigger(player, id, target)
-    local effectTarget = player;
-    if (target ~= nil) then
-        effectTarget = target;
-    end
+function error(player, msg)
+    player:PrintToPlayer(msg);
+    player:PrintToPlayer("@getmod <modID>");
+end;
 
+function onTrigger(player, id)
+    -- validate modID
     if (id == nil) then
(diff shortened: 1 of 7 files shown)


TeoTwawki:
I see you've re-sequenced some latent IDs as well..If you change the ID of a latent you need to update every sql entry that used its old ID, this is why they were to in that order - they were ordered as created instead of logically grouped like that because the ID value is simply +1 to whatever the previous highest ID was.

ShelbyZ:
I can revert those changes,  the values of existing latents _should not_ have changed.  I just liked reading through a list that was 1..2..3 and not 1..2..45..3.

teschnei:
the list was ordered so that you could actually find the latent you were looking for by their logical groups.  You're probably not ever going to skim that list looking for a specific number

TeoTwawki:
I got tripped up because I wasn't looking at the numbers. my bad

scripts/commands/getmod.lua
```diff
@@ -0,0 +1,24 @@
...
+---------------------------------------------------------------------------------------------------
+
+cmdprops =
+{
+    permission = 3,
+    parameters = "is"
+};
+
+function onTrigger(player, id, target)
+    local effectTarget = player;
```
takhlaq: `local effectTarget = GetPlayerByName(target) or player;`

ShelbyZ: I did not make me intention clear when writing this as I had originally tested this on monster and not players.  I'll have an update that uses the cursor target instead of target.

PR #3938, opened 2017-07-15
Bcnm pet cleanup

### LINKED ISSUE #3919: BCNM that involve mob's that have pets
**_I have:_**

30170427_0

**_Server Version_** (type `@revision` in game) **:**
4ca87a3dd8b29dda4f8b2a11443d917eff30ed48

master

This issue is old but has been poorly described by people guessing at what it wrong, and usually they just complain without opening an issue on github. Well now we have one. These STILL aren't being cleared by the battlefield handler at the end, leaving them sitting ~~to attack the next person to do any BCNM in the same zone~~. ~~The fact they can even attack says the code spawning them didn't register them to the BCNM.~~ *players are often unreliable witnesses* 

All their SQL entries are correct: not spawned at start, not required killed to win, not timed repops..So we just need them to be ~~considered part of the BCNM so they get~~ cleaned up.

ShelbyZ:
- Check on cleanup() if any mobs have pets and if they aren't dead, we
should tell them to despawn
- Should fix #3919

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -465,7 +465,7 @@ void CBattlefield::cleanup()
         auto enemy = m_EnemyList.at(i);
         if (enemy->PPet != nullptr && enemy->PPet->isAlive())
         {
-            enemy->PPet->Die();
+            enemy->PPet->PAI->Despawn();
         }
         enemy->PAI->Despawn();
         enemy->status = STATUS_DISAPPEAR;
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -462,9 +462,14 @@ void CBattlefield::cleanup()
     ShowDebug("bcnm cleanup id:%i inst:%i \n", this->getID(), this->getBattlefieldNumber());
     //wipe enmity from all mobs in list if needed
     for (int i = 0; i < m_EnemyList.size(); i++) {
-        m_EnemyList.at(i)->PAI->Despawn();
-        m_EnemyList.at(i)->status = STATUS_DISAPPEAR;
-        m_EnemyList.at(i)->PBCNM = nullptr;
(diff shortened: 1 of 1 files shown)


teschnei:
I'm not merging this unless it's explicitly tested to have fixed the issue.  The pets should be in the enemy list in the first place, so I don't think it makes a difference

ShelbyZ:
The BCNM in question is SMN Maat and I was able to get an avatar from a prior fight into new fight (WHM) by doing the following:

- Start SMN fight and approach Maat and ensure he summons an avatar
- Warp/zone out (do not win)
- Start WHM fight and view that avatar from prior fight is present

BCNM info (first line is Maat and second is his Avatar):

bcnmId | battlefieldNumber | monsterId | conditions
-------- | --------------------- | ----------- | -----------
103 | 1 | 17375263 | 3
103 | 1 | 17375264 | 0

When I test through spawnMonstersForBcnm(...), I see that it returns 2 Rows for the query.  In the while(...) loop I do see that Maat is added to the enemy list, but on the second iteration Sql_NextRow(SqlHandle) returns -1 (SQL_ERROR).  So the Pet never gets added to the container, but this does not prevent Maat from summoning the avatar.

teschnei:
Can you test again with the commit I just pushed?

I'm guessing you've already got it set up to test quickly

teschnei:
Without your fix in this branch, I mean

TeoTwawki:
> but on the second iteration Sql_NextRow(SqlHandle) returns -1 (SQL_ERROR). 

OH HI THERE BUG I HAVE BEEN TRYING TO SPOT

ShelbyZ:
Yeah, this does indeed add the pets to the enemy list, which gets properly removed in cleanup().  Can close out this merge.

TeoTwawki:
thank you both

PR #3948, opened 2017-07-28
Remove some junk from widescan

ymkc:
- hide banners
- hide goblin footprints (and fix typos)
- hide option one/two/three in xarcabard


TeoTwawki:
_- rubbastamp -_

PR #3954, opened 2017-08-02
commands [A]: input validation and error standardization

wrenffxi:
addCurrency: the getCurrency check did not function as intended.  Always returned a non-nil value, even for invalid currency names.

addMission and addQuest logId is restricted to valid Mission logs and Quest logs, respectively.

animateNPC, animateSubNPC can now be targeted via cursor

Fixed some server-crashy inputs.  e.g. `@addkeyitem foo`

lua_baseentity.cpp and chautils.cpp: whitespace fixes. no logic changes.


scripts/commands/addtreasure.lua
```diff
@@ -9,28 +9,43 @@ cmdprops =
     parameters = "isi"
 };
 
+function error(player, msg)
```
TeoTwawki: Overall I like the idea of just standardizing the usage error handling in all the commands like this.

scripts/commands/addkeyitem.lua
```diff
@@ -11,32 +11,41 @@ cmdprops =
     parameters = "ss"
 };
 
+function error(player, msg)
+    player:PrintToPlayer(msg);
+    player:PrintToPlayer("@addkeyitem <ID> <player>");
+end;
+
 function onTrigger(player, keyId, target)
-    if (keyId == nil or keyId == 0) then
```
TeoTwawki: You got a server crash? I didn't get any server crash passing bs strings to the command before when the "or" was added but either when I tried to handle casing or when demo `_G`'d everything we could have broken it I guess.

scripts/commands/addallspells.lua
```diff
@@ -9,69 +9,72 @@ cmdprops =
...
+        660,661,662,663,664,665,666,667,668,669,670,671,672,673,674,675,676,677,678,679,680,681,682,683,684,685,
+        686,687,688,689,690,691,692,693,694,695,696,697,698,699,700,701,702,703,704,705,706,707,708,709,710,711,
+        712,713,714,715,716,717,718,719,720,721,722,723,724,725,726,727,728,736,737,738,739,740,741,742,743,744,
+        745,746,747,748,749,768,769,770,771,772,773,774,775,776,777,778,779,780,781,782,783,784,785,786,787,788,
+        789,790,791,792,793,794,795,796,797,798,799,800,801,802,803,804,805,806,807,808,809,810,811,812,813,814,
+        815,816,817,818,819,820,821,822,823,824,825,826,827,828,829,830,831,832,833,834,835,836,837,838,839,840,
+        841,842,843,844,845,846,847,848,849,850,851,852,853,854,855,856,857,858,859,860,861,862,863,864,865,866,
+        867,868,869,870,871,872,873,874,875,876,877,878,879,880,881,882,883,884,885,886,887,888,889,890,891,892,
+        893,894,895
+        --trusts are causing client crashes when flagged in char spells packets sent outside of zone/log in
```
TeoTwawki: Note to self, go fix that Teo (I know why it happens already)

scripts/commands/addcurrency.lua
```diff
@@ -9,33 +9,125 @@ cmdprops =
...
+end;
 
-    if (currency == nil or amount == nil) then
-        player:PrintToPlayer( "You must enter a valid currency and amount." );
-        player:PrintToPlayer( "@addcurrency <currency type> <amount> <target player>" );
-        return
-    end
+function onTrigger(player,currency,amount,target)
+    local currencies = {
+        sandoria_cp=true,
```
TeoTwawki: What exactly happened when you tried it?

The entire reason these are strings is to avoid having to lua side define them at all (technically bad practice, but super convenient). Last time I tried trying to pass non-existent currency to this error'd out like it was supposed to in the core.

TeoTwawki:
overall I don't see anything that tells me not to merge, I'm just curious about a a few things.

scripts/commands/addcurrency.lua
wrenffxi: If you put in, say, "FOO" as the currency name, then targ:getCurrency("FOO") returns 0, rather than nil.  It's indistinguishable from a valid currency the player's out of, except that it sends a SQL error to the console.  Then the code happily continues, targ:addCurrency("FOO",amount), which sends another error to the console.  Server doesn't crash, but checking for valid currency name lua-side prevents the two console error messages.

scripts/commands/addkeyitem.lua
wrenffxi: With current DSP code, if I do `@addkeyitem foo` the server crashes.

scripts/commands/addcurrency.lua
TeoTwawki: Kind of how its intended though, because if a script uses an invalid currency it should throw the error.\* I'll wait and see if other dev's want the sql error axe'd this way. 

\**I admit our sql error handling needs work though - we're using the generic "it didn't work" which is useless 90% the time unless you are the only person on the test server.*

scripts/commands/addkeyitem.lua
TeoTwawki: was about to ask to see the crash dump, just realized why it would crash - checkin for zero when value could now possibly be a string. depr derp. It would hit debug break in luabaseentity if an invalid string got passed.

TeoTwawki: So caused that the other day when I moved nil checks to fix the *other* bug.

Bad teo, bad.

scripts/commands/addcurrency.lua
teschnei: i don't really like the idea of having to maintain the currency list in two places

the binding should be able to check if it errored and pass nil back

wrenffxi: How's this?  Now binding returns nil if invalid currency type is entered.

teschnei:
the whitespace changes...

I don't like the idea of using -1 (how do you know -1 isn't a valid number for some currency?)

Probably better to return an optional<int>

wrenffxi:
I can't for the life of me figure out how to get my VS2017 to recognize optionals.  Keep getting this message "class template optional is only available with C++17" no matter what tutorials I follow.

teschnei:
Config properties -> c/c++ -> language -> c++ language standard

teschnei:
it'll have some side effects, the linux build might also need a flag change (travis will find out for us)

also I think it drops vs2015

TeoTwawki:
☹️

wrenffxi:
Sorry, have been trying to get it working for 2 hours now and have no more fight left in me.  Tried installing VS2017 Preview 2.  Tried setting C++ language standard to c++latest.

Tried all this stuff.
https://blogs.msdn.microsoft.com/vcblog/2017/05/05/cpp-modules-in-visual-studio-2017/

All roads lead to "class template optional is only available with C++17".

I'm going to back out the addcurrency.lua changes for now and call it a night.

teschnei:
That's strange, it worked fine for me.  I can test adding it for you, if
you like.

teschnei:
Ah, DSP doesn't compile in c++17 anyways because binary_function is
deprecated and removed!

I should look into that first

wrenffxi:
Sure.  I left the code in addcurrency.lua that will catch nil return from binding, so this PR is complete, and it'll work with any code changes down the road that modify lua_baseentity:getCurrency.

TeoTwawki:
@teschnei I'm going to try for a fix that doesn't break anybodies setup if it can be helped. if it cant then rip vs2016

TeoTwawki:
I think its good, merging.

Going to see take care of the sql error message either in the sql_refactor branch or after it has merged. The result might actually increase the console log use though (making error actually give you info about *what caused* said error)

PR #3960, opened 2017-08-03
command input validation and standardization [b-c]

wrenffxi:
Added `@checkquest` command. (Thanks for the code, @kushdr).
`@checkquest`, `@checkmission`, `@checkvar` can now be targeted via cursor.
QUEST_LOGS and MISSION_LOGS globals now allow logName output when player enters numeric logId.
`@changejob` and `@changesjob` now report resulting job/sjob.
`@bring` now sets player rotation and does not zone target if they're in correct zone already.

### DIFF
--- a/scripts/globals/log_ids.lua
+++ b/scripts/globals/log_ids.lua
@@ -219,3 +219,34 @@ ROV =
     ['full_name'] = "Rhapsodies of Vana'diel",
     ['mission_log']= 13
 };
+
+QUEST_LOGS = {
+    [0] = "SANDORIA",
+    [1] = "BASTOK",
+    [2] = "WINDURST",
+    [3] = "JEUNO",
+    [4] = "OTHER_AREAS",
+    [5] = "OUTLANDS",
+    [6] = "TOAU",
+    [7] = "WOTG",
+    [8] = "ABYSSEA",
+    [9] = "SOA",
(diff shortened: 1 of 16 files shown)


whasf:
How about making @bring and @goto work across multiple game servers? They're pretty useless unless you are running all zones on one server.. :(

wrenffxi:
That's an issue outside of my pay scale. :)

Really, I don't know the first thing about multi-server setups.  Will leave that to the experts.

TeoTwawki:
Presuming all tested and good coz its Wren.
❤️

TeoTwawki:
the cross server commands thing is really difficult to solve. In cases of some commands, very not possible right now.

scripts/globals/log_ids.lua
```diff
@@ -219,3 +219,34 @@ ROV =
...
+    [4] = "OTHER_AREAS",
+    [5] = "OUTLANDS",
+    [6] = "TOAU",
+    [7] = "WOTG",
+    [8] = "ABYSSEA",
+    [9] = "SOA",
+    [10] = "COALITION",
+};
+
+MISSION_LOGS = {
```
TeoTwawki: I totally missed this before I hit the biutton, my bad. But I am pretty sure these values was already defined inside the other tables.

PR #3962, opened 2017-08-03
Unify Conquest Guards

ymkc:


src/map/lua/luautils.h
```diff
@@ -167,6 +167,7 @@ namespace luautils
     int32 clearVarFromAll(lua_State *);                                         // Deletes a specific player variable from all players
     int32 terminate(lua_State*);                                                // Logs off all characters and terminates the server
 
+    int32 GetTextIDVariable(lua_State*);                                        // Lua binding to call luautils::GetTextIDVariable
```
teschnei: so it's a lua binding to call a c++ binding to call lua?

I can't think of a single use for this

scripts/globals/conquest.lua
```diff
@@ -211,6 +211,82 @@ function tradeConquestGuard(player,npc,trade,guardnation,guardtype)
...
+                break;
+            end
+        end
+    end
+end;
+
+function finishConquestGuard(player,csid,option,size,inventory,guardnation)
+    local CONQUEST = GetTextIDVariable(player:getZoneID(), "CONQUEST");
+    local ITEM_OBTAINED = GetTextIDVariable(player:getZoneID(), "ITEM_OBTAINED");
+    local ITEM_CANNOT_BE_OBTAINED = GetTextIDVariable(player:getZoneID(), "ITEM_CANNOT_BE_OBTAINED");
```
TeoTwawki: Instead of making up new bindings, just variable-ize the zones textIDs file so its always "whatever zone we are in now".
```lua
        local TextIDs = "scripts/zones/" .. player:getZoneName() .. "/TextIDs";
        package.loaded[TextIDs] = nil;
        require(TextIDs);
```

teschnei:
You don't even have to do that. The correct zone is already loaded! Just
use the textID normally. Same thing the crafting union reps do.

TeoTwawki:
well at least he has the example for next time if he does need to load them

TeoTwawki:
I just wanted to show how to load an "anywhwere" textids,

ymkc:
> You don't even have to do that. The correct zone is already loaded! Just use the textID normally.

I totally didn't see this >.<

ymkc:
is there anything that needs correction/isn't done yet?

TeoTwawki:
@ymkc just needs reviewed and nobody has yet

scripts/globals/conquest.lua
```diff
@@ -211,6 +211,77 @@ function tradeConquestGuard(player,npc,trade,guardnation,guardtype)
...
+                player:updateEvent(2,CPVerify,inventory[Item + 2]);
+                break;
+            end
+        end
+    end
+end;
+
+function finishConquestGuard(player,csid,option,size,inventory,guardnation)
+    if (option == 1) then
+        duration = (player:getRank() + getNationRank(player:getNation()) + 3) * 3600;
```
takhlaq: local

ymkc: turned into local

PR #3963, opened 2017-08-04
corrected Haste value of Rune Copper

dwn134:


TeoTwawki:
Haste is a [fraction of 1024](https://github.com/DarkstarProject/darkstar/blob/master/src/map/modifier.h#L214), not a direct percent (until someone changes the core to do the math for you and then edits allllll the existing haste values. ***not it***!)

so 90 isn't correct. 92 would get you 9% of 1024.

TeoTwawki:
oh actually..rune chopper isn't really 9%..its just under. so this is right after all according to
http://ffxiclopedia.wikia.com/wiki/Haste_%28Status_Effect%29#Precise_Haste_Values
and bg has nothing to contradict that, so merging.

teschnei:
What about Rune Iron?? Or Rune Brass???

TeoTwawki:
Who cares bout those?

GET TO THE CHOPPA

PR #3965, opened 2017-08-04
command input validation and standardization [d-e]

wrenffxi:
`@addcurrency` and `@delcurrency` commands report resulting currency.
`@delcurrency` now handles values larger than player's current currency amount.
`@despawnmob` now reports mob name.


TeoTwawki:
Looks gook to me if I missed something on review either I or wren can fix.

@wrenffxi ty for doing this

PR #3966, opened 2017-08-04
command input validation and standardization [f-g]

wrenffxi:
fixed a typo in changejob from previous commit.
`@getid` now reports target name.
`@getmobaction` can now be targeted with cursor.
`@givegil` now reports resulting amount of gil.
`@goto` now sets player rotation and does not zone player if target is in same zone.


whasf:
Wouldn't you want @goto to work even if you're in the same zone? you may be half way across the map ;p

Hozu:
It looks like it just moves the user to reach the target instead of zoning them. There are advantages and disadvantages to whichever method is used.

Hozu:
Also goto yourself had certain uses, like if you got stuck in an event somehow. But that only worked if you zoned to yourself.

wrenffxi:
I've re-edited `@bring` and `@goto` so that they can target self, and also added a forceZone option.  By default, bring or goto target in same area will not cause a dezone.  But if you set forceZone to 1 it will.

TeoTwawki:
eh, I'd rather **default** behavior be whats its been. coz am lazy and goto myself a lot. Any objections from anybody?

wrenffxi:
That should do it.

PR #3972, opened 2017-08-07
command input validation and standardization [o-r]

wrenffxi:
`@pettp` now sets pet tp to given amount, rather than adding given amount.
`@pos` now reports target position to player, rather than target, when not self-targeted.
`@promote` now warns same- or higher-tiered GMs of attempted targeting.

### DIFF
--- a/scripts/commands/promote.lua
+++ b/scripts/commands/promote.lua
@@ -15,10 +15,12 @@ function error(player, msg)
 end;
 
 function onTrigger(player, target, level)
-    local maxLevel = player:getGMLevel();
-    local targ;
+    -- determine maximum level player can promote to
+    local maxLevel = player:getGMLevel() - 1;
+    if (maxLevel < 1) then maxLevel = 0;
     
     -- validate target
+    local targ;
     if (target == nil) then
         error(player, "You must provide a player name.");
         return;
--- a/scripts/commands/promote.lua
+++ b/scripts/commands/promote.lua
@@ -9,35 +9,43 @@ cmdprops =
     parameters = "si"
 };
 
+function error(player, msg)
+    player:PrintToPlayer(msg);
+    player:PrintToPlayer("@promote <player> <level>");
+end;
+
 function onTrigger(player, target, level)
-    if (level == nil) then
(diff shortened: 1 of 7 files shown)


Hozu:
Kind of strange that the person who originally added the console print to promote didn't add one for exec. Maybe they assumed that it would only be given to the most trusted staff? I feel like it should probably have a console print that include the entire input.

TeoTwawki:
@hozu I have been thinking of adding optional logging to all comands universally - just tale thee tire input for any *valid* command and toss to a string field table with user name and cmnd name. I have something similar already.

TeoTwawki:
@wrenffxi was gonna ask you to fix pos's lack of handling the pricise value with decimal places  when you got to that cmnd but looks like you already did that before I could  <3

scripts/commands/promote.lua
```diff
@@ -9,35 +9,43 @@ cmdprops =
...
+        error(player, "You can not use this command on same or higher tiered GMs.");
         return;
     end
 
-    -- Validate the level..
-    if (level < 0) then 
-        level = 0;
+    -- validate level
+    if (level == nil or level < 0 or level > maxLevel) then
+        error(player, string.format("Invalid level.  Must be 0 to %i.", maxLevel ));
```
TeoTwawki: gms should not be able to create additional gms of identical tier, only lesser gms. original command was designed that way intentionally.

TeoTwawki: am on mobile so if i misread i apologize.

wrenffxi:
Understood. That should do it.

scripts/commands/promote.lua
wrenffxi: Understood.  This should do it.

Hozu:
@TeoTwawki Should make it log commands that require being a GM to use. ie. commands that can be used by everyone don't need to be logged. Or have an option to decide the GM level for command logging.

TeoTwawki:
@hozu mobile kinda butchered my sentence back there but what I intended was configurable and to omit any non existing command to

PR #3974, opened 2017-08-08
command input validation and standardization [send, zone]

### LINKED ISSUE #322: zone GM command error
(at) zone {Behemoth's Dominion}
lands me in Meriphataud Mountains instead of Behemoth's Dominion.

wrenffxi:
`@send` and `@zone` now accept all auto-tralanslate phrases.

You can set x,y,z to non-zero values in zones_list table.  See `@send` for example.  Sometimes 0,0,0 puts the player out of bounds.

Fixes #322 .

### DIFF
--- a/src/map/commandhandler.cpp
+++ b/src/map/commandhandler.cpp
@@ -163,6 +163,11 @@ int32 CCommandHandler::call(CCharEntity* PChar, const int8* commandline)
 
         switch (*parameter)
         {
+        case 'b':
+            lua_pushstring(m_LState, commandline);
+            ++cntparam;
+            break;
+
         case 's':
             if (cmdparameters.size() == 1)
             {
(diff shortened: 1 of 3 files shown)


src/map/commandhandler.cpp
```diff
@@ -163,6 +163,11 @@ int32 CCommandHandler::call(CCharEntity* PChar, const int8* commandline)
 
         switch (*parameter)
         {
+        case 'b':
+            lua_pushstring(m_LState, commandline);
```
TeoTwawki: Nice

TeoTwawki:
>  Sometimes 0,0,0 puts the player out of bounds.

Lots of zones lack a "default" pos. Most will check on entry for if you are at 0/0/0 and move you. For other reasons, I just wanna say so everyone knows that not having a  default pos is its own issue even with your command fixes. :)

PR #3976, opened 2017-08-09
command input validation and standardization [t-z]

wrenffxi:
Added getExp() binding.
`@takexp` no longer sets player xp to high value if more xp is deducted than they have.
`@givexp` and `@takexp` now report resulting xp.

### DIFF
--- a/scripts/commands/givegil.lua
+++ b/scripts/commands/givegil.lua
@@ -36,7 +36,6 @@ function onTrigger(player, amount, target)
 
     -- give gil to target
     targ:addGil(amount);
-    local newAmount = targ:getGil();
-    player:PrintToPlayer(string.format("Gave %i gil to %s.  They now have %i gil.", amount, targ:getName(), newAmount));
+    player:PrintToPlayer(string.format("Gave %i gil to %s.  They now have %i gil.", amount, targ:getName(), targ:getGil()));
     
 end;
--- a/scripts/commands/givexp.lua
+++ b/scripts/commands/givexp.lua
@@ -16,12 +16,6 @@ end;
 
 function onTrigger(player, amount, target)
 
-    -- validate amount
-    if (amount == nil or amount < 1) then
-        error(player, "Invalid amount.");
-        return;
-    end
-
     -- validate target
     local targ;
     if (target == nil) then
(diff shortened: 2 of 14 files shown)


takhlaq:
i feel getexp has no potential retail scenario use so not sure whether it's something we really need a binding for

wrenffxi:
Problem case is this:

Alice has 100 xp into level.
GM does `@takexp 500 Alice`
Now Alice has roughly 65,535 XP.

So either we need a getXP binding that caps deducted amount at current player XP, or we need to fix the integer underflow problem in delExp.

TeoTwawki:
odd that this even happens, when you die in that situation (same function) you'd *delevel* rather than gain a big chunk of exp. I should have tested that better back when I did it instead of assuming it'd work just like dying. Present me is sorry on behalf of past me.

takhlaq:
i'll wait for someone to look into the underflow before merging

TeoTwawki:
cef1353b62f27d92212fe1ac0ba9dc77cde2000b

takhlaq:
we don't need getExp binding anymore right?

TeoTwawki:
Correct. Existing del binding should delevel peeps now instead of turning into to a big exp gain.

wrenffxi:
Had to fix a few more things, but all should be working now:

* XP is now deducted via `@takeexp` even if map_config.exp_loss_level would usually prevent this.
* If amount deducted is larger than two levels, player will now lose one level and be set at 0 XP.  Previously, another underflow would occur.
* If player is level 1, they will no longer drop to level 0.

TeoTwawki:
hmm. That underflow is irritating me. But I don't see an easy way to go about multi deleveling (which wouldn't ever occur on retail.)

I guess when ppl are caught cheating their rear ends off I just reset them to level one instead of trying to reverse the exact exp amount they cheated to get.

TeoTwawki:
@takhlaq stole yer merge!

teschnei:
I think level up has the same quirk, and on retail you would just level
once and be 1 XP away from the next level

TeoTwawki:
yeah that's how modern retail does it. I think there **may** have been a situation at one time in the past where a double level might have been possible as an accidental thing on retail, not sure.

Anyway we should prevent double level up on DSP as well. I think its pretty easy to trigger with book XP+moderately buffed XP gains.

teschnei:
I think we do already, is what I was meaning to say. I recall running into
that years ago.  So we should do the same for level down as well

TeoTwawki:
Been awhile since I actually played enough to notice, guess I missed it

PR #3977, opened 2017-08-09
Hide all moghouse related moogles

ymkc:
These shouldn't be visible in the cities and moghouse is handling them anyway


TeoTwawki:
actually, [mog house ones are on the fly done](https://github.com/DarkstarProject/darkstar/blob/498440b100c7b62a3488d71ebbdffd109f6090a2/src/map/zone_entities.cpp#L581) and just happen to wind up at the same IDs as some other moogles (in most zones). These are event mogs that we don't use yet, so still should hide'em

TeoTwawki:
I think the Bastok markets one has the same ID as the moogle for the mog sack expansion, if you print them in `onTrigger()`, lol.

ymkc:
@teotwawki i see, thanks

PR #3991, opened 2017-08-19
Blighting Brand fixes

Brierre:
is a Weapon, not Gears, and he had the wrong groupid in his mob_spawn_points entry


Whitechaser:
Could Eldritch Edge be broken in a similar way too?

Brierre:
I'll take a look. Right in the middle of something atm!

PR #3992, opened 2017-08-19
Bastok8 1

Brierre:
Bastok Mission 8-1 was just spawning the mobs over and over again.  Now it continues as intended on to the next CS when the three ants are gone.


takhlaq:
please reopen after rebased and conflicts are resolved

PR #3995, opened 2017-08-19
Warm-up will now be used by brown mamool jas.

wrenffxi:


scripts/globals/mobskills/warm-up.lua
```diff
@@ -10,7 +10,13 @@ require("scripts/globals/status");
 ---------------------------------------------
 
 function onMobSkillCheck(target,mob,skill)
-    return 0;
+    -- only brown-skinned mamool should use this move
+    local mobSkin = mob:getModelId();
+    if (mobSkin == 1639 or mobSkin == 1619) then
```
TeoTwawki: Huh, coulda swore the brown and blue had separate skill lists..that must have gotten lost when the skills refactor hit. Since things used to be by family ID they were batch made to match the existing family IDs at the time. I see ID 177 is unused yet there is a big ol block of mamool skills commented out. The blue skin ones get stave toss. Fam ID 177 says "mamool knight" which I am guessing is the mounted ones.

PR #3997, opened 2017-08-19
added Fotia gorget and belt to ws mods

dwn134:


scripts/globals/weaponskills.lua
```diff
@@ -1043,6 +1048,11 @@ function handleWSGorgetBelt(attacker)
                 break;
             end
         end
+
+        if (belt == 28420) then
+                accBonus = accBonus + 10;
+                ftpBonus = ftpBonus + 0.1;
```
takhlaq: 4 space indent

scripts/globals/weaponskills.lua
```diff
@@ -1034,6 +1034,11 @@ function handleWSGorgetBelt(attacker)
             end
         end
 
+        if (neck == 27510) then
+        	    accBonus = accBonus + 10;
```
takhlaq: tabs > 4 spaces

scripts/globals/weaponskills.lua
```diff
@@ -1034,6 +1034,11 @@ function handleWSGorgetBelt(attacker)
...
+                ftpBonus = ftpBonus + 0.1;
```
takhlaq: 4 space indent

scripts/globals/weaponskills.lua
```diff
@@ -1043,6 +1048,11 @@ function handleWSGorgetBelt(attacker)
                 break;
             end
         end
+
+        if (belt == 28420) then
```
takhlaq: not a fan of using ids but guess there isnt really a better way
a comment with which item it is would be nice though, for anyone who needs to look up how this is handled later

wrenffxi:
If this is coded as an equipment slot check rather than an item mod + mod check, then players will receive the bonus when wearing the item even when under level sync.

dwn134:
It coded as all the other gorgets right above though

TeoTwawki:
It's yet another case of person before you got here being naughty.

@takhlaq why you no teach'em how to modifier? 😢

Hozu:
It was like this before I moved the gorget/belt check into its own function so it could be called upon for Atonement. The problem I couldn't figure out at the time was how to avoid doubling up on the bonus on certain WSs that had SC properties that could both count for a gorget, like every single relic WS.

...but thinking about it now it could probably do a for loop that runs for as long as many properties the WS has, comparing each individual gorget mod, and using a break to exit early once it's met. Nested inside another for loop for the 8 gorget elements. And before that a check for the all element gorget mod. Unless someone can figure out a way for it to work with all 8 modifiers.

TeoTwawki:
@dwn134 now you got multiple diff things smushed into one branch. This is why you should make new branches for each thing you work on. 

You'll have to rebase ***do not simply close pr and or delete your entire fork to start over.*** 

Make a new branch to rebase to darkstar master and cherrypick the commits that didn't belong here into that, then rebase or reset this one to eliminate them. If you can't figure it out **ask in discord for help.**

dwn134:
Yea. I hit the new pull request button and I didn't know it would still merge to my last pull

TeoTwawki:
it didn't, git is based on branches. you worked in same branch for all these commits.

dwn134:
fixed

scripts/globals/weaponskills.lua
```diff
@@ -1034,6 +1034,11 @@ function handleWSGorgetBelt(attacker)
             end
         end
 
+        if (neck == 27510) then -- Fotia Gorget
+                accBonus = accBonus + 10;
+                ftpBonus = ftpBonus + 0.1;
```
takhlaq: 4 space indent please

dwn134:
Ah I had that fixed. Whatever I did took it back to the beginning. ><

PR #4010, opened 2017-08-23
Eliminate NPC/Mob IDs from Lua

ymkc:
This is just a suggestion and kind of a proof of concept.

In the long run, it could simplify keeping up with the ever changing IDs and reduce the "ID out of sync" bugs.

### DIFF
--- a/scripts/globals/conquest.lua
+++ b/scripts/globals/conquest.lua
@@ -658,7 +658,7 @@ switch (region): caseof {
   ---------------------------------
     --print("RONFAURE");
 
-    local Doladepaiton = 17187525;
+    local Doladepaiton = GetNPCIDByName(100, "Doladepaiton_RK");
 
     npc  = {
     --
@@ -689,7 +689,7 @@ switch (region): caseof {
   ---------------------------------
     --print("ZULKHEIM");
 
-    local Quanteilleron = 17199709;
+    local Quanteilleron = GetNPCIDByName(103, "Quanteilleron_RK");
 
     npc  = {
     --
@@ -720,7 +720,7 @@ switch (region): caseof {
   ---------------------------------
     --print("NORVALLEN");
 
-    local Chaplion = 17203848;
+    local Chaplion = GetNPCIDByName(104, "Chaplion_RK");
 
     npc  = {
     --
@@ -751,7 +751,7 @@ switch (region): caseof {
   ---------------------------------
     --print("GUSTABERG");
 
-    local Ennigreaud = 17212060;
+    local Ennigreaud = GetNPCIDByName(106, "Ennigreaud_RK");
 
     npc  = {
     --
@@ -782,7 +782,7 @@ switch (region): caseof {
   ---------------------------------
     --print("DERFLAND");
 
-    local Mesachedeau = 17224326;
+    local Mesachedeau = GetNPCIDByName(109, "Mesachedeau_RK");
 
     npc  = {
     --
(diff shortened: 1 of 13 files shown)


TeoTwawki:
Potential snag:
Sometimes we actually have multiples of the same name where it doesn't make sense to rename them to avoid it.

TeoTwawki:
Potential snag:
Sometimes we *might* actually have multiples of the same name where it doesn't make sense to rename them to avoid it. Depends on where we use this.

teschnei:
It only really works if we give everything a unique name. Which just
happens to be exactly what the id is for...

TeoTwawki:
Whats that first value in the function? If its the short version of the index id, we used to use that and moved away from it.

ymkc:
the first id is the zone, as almost all npcs are unique per zone.

it returns the first match so it can be used to get an offset

TeoTwawki:
most zone shave a ton of duplicates the player doesn't realize are there, but I can see a getByName type maneuver possibly working for the conquest entities.

ymkc:
it works for alot of entities, like  special doors, qm's/nms spawn objects, treasure chest/coffers (ignoring spawn point sql atm),  rainbows :)

i know there are a million moogles/ah npcs etc, but those arent the ones hardcoded in lua and running out if sync

ymkc:
btw i am not trying to say this is the final/best solution.

what i am trying to say is, maybe making the db the single source of truth and forgetting about all the rest is easier than having ids all over the scripts and the db

ShelbyZ:
My only suggestion would be to not return the first found entity that matches by name, but instead return a collection (0..n) that match and leave it up to the script author to choose the "right" entity and doesn't prevent a script that needs to find all mobs with similar names and split out their behaviors accordingly.

TeoTwawki:
just my opinion: I think its better to simply not use this in any situation where multiples exist. So thumps up for qm's rainbow and conquest guys at the outpost, thumbs down for those holiday moogles we aren't using yet etc.

ibm2431:
Honestly, I think DSP should have its own unique, unchanging IDs for NPCs, maybe with a defined labeling convention, and use those for Lua scripts.

It would be an additional field in the DB, and the DB would still have to be updated with SE NPC IDs (there really is no way to avoid this without major changes to how DSP handles NPCs). But as @TeoTwawki said, much less updating in Lua scripts (and fewer issues if tools/people overlook certain IDs after shifts).

Could also be used for more clarity in Lua scripts as well, since it wouldn't have to be a number. It's easier to spot a potential error in:
`local Chaplion = getNPCIDByLabel("Chaplion0001")`
than:
`local Chaplion = 17203848`
if someone gets their NPC wires crossed and uses the wrong ID / label.

Could include zone as well if that could be useful for reading/debugging. "024_Prishe_001" specifies the first Prishe in Lufaise, which clearly shouldn't be in a battle script for Empyreal Paradox (Zone 036), or the Prishe you talk to in Empyreal Paradox [A] (Zone 255). Can also use it to return entire collections: "I want all treasure coffers in Quicksand Caves" with a `LIKE '208_TreasureCoffer_%%%'`

TeoTwawki:
@ibm2431 I want to avoid any kind of "nTH copy of this name" stuff. To easy to do things we shouldn't if that happens, and collections will encourage that. Ensuring we grab a single unique entity for a specific zone though, that has some appeal.

ymkc:
accidently deleted branch, pls close if off the table

TeoTwawki:
still interested in the possibility, but I think everyone's attention went elsewhere before making a decision on what if anything to do with it.

is restored branch same content?

ymkc:
think so, didn't touch the branch, just clicked delete on the wrong one

PR #4022, opened 2017-08-27
Changed Ge_Dha_Evileye to have ph

Skarbie:
Made PH lua
Ge_dha_Evileye uses 2hr
Added to MobIDs for zone


TeoTwawki:
There was no need to close the existing pull request. All you had to do was force push over it. I get really annoyed having to say this *over and over and over*.

TeoTwawki:
It also loses all the comments, so there is no history of wtf was said without opening both pages when someone does this. Which in the past has caused a good many mistakes to slip through the cracks and get merged.

Skarbie:
I'm sorry Teo >×< stop new to this, I am trying!! Won't happen again!!!

PR #4028, opened 2017-08-31
Su fix

Skanless:
Wanted to go into the src\maps\packets\char_stats.cpp to ad in 
WBUFB(data,(0x52)) = 2;


TeoTwawki:
Hilarious as the commit log here is, you need to rebase onto darkstar master, and next time start with a clean branch that is 1:1 with darkstar master.

TeoTwawki:
I'm not gonna be around much this week, so I summon @takhlaq to show you the ropes.

maxtherabbit:
also we intentionally didn't hardcode the superiority value into the packet because it's not retail accurate

so this doesn't need to be in trunk anyway

TeoTwawki:
Oh.. I didn't even catch that. just saw the many many many commits and knew somethin went terribly wrong

Issue #4019, opened 2017-08-25
Progression Fix for "Trouble in the Sluice" w/DRG quest flagged


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**30170728_2

**_Server Version_** (type `!revision` in game) **:**unknown

**_Source Branch_** (master/stable) **:**master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
Was not able to complete "Trouble at the Sluice" if you talked to Arminibit who starts you on the path to unlock DRG ("The Holy Crest"). This fixes that so you can progress to finish "Trouble at the Sluice" and then refers you back to the DRG quest if you still want to unlock it.

[Novalmauge.txt](https://github.com/DarkstarProject/darkstar/files/1251041/Novalmauge.txt)

https://github.com/Wiggo32/darkstar/commit/549ec6599988b04edf1b665718d2eb75fae13ea2

I would create a pull request but I can't figure it out.

Wiggo32:
https://github.com/DarkstarProject/darkstar/commit/01c91f6f3b0ad95827e6721b2b13c78482c7083b

PR #4038, opened 2017-09-06
Progression fix for Trouble at the Sluice

Wiggo32:
This tweak to Novalmauge.lua will allow you to complete the quest "Trouble at the Sluice" if you happen to have DRG quest flagged.

https://github.com/DarkstarProject/darkstar/issues/4019

### DIFF
--- a/scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
+++ b/scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
@@ -74,7 +74,7 @@ function onTrigger(player,npc)
     -- The Holy Crest Quest
     if (TheHolyCrest == 1) then
         player:startEvent(0x0006);
-    elseif (TheHolyCrest == 2 and crestCheck == 0)then
+    elseif (TheHolyCrest == 2 and crestCheck == 0) then
         player:startEvent(0x0007);
         player:setVar("theHolyCrestCheck",1);
     -- Trouble at the Sluice Quest


scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
```diff
@@ -67,14 +67,16 @@ function onTrigger(player,npc)
...
     local theRumor = player:getQuestStatus(SANDORIA,THE_RUMOR);
+	local crestCheck = player:getVar("theHolyCrestCheck");
 
     npc:wait(-1);
 
     -- The Holy Crest Quest
     if (TheHolyCrest == 1) then
         player:startEvent(0x0006);
-    elseif (TheHolyCrest == 2) then
+    elseif (TheHolyCrest == 2) and (crestCheck == 0)then
```
TeoTwawki: do:
```lua
if (condition1 == x and condition2 == y) then
```
not:
```
if (condition1 == x) and (condition2 == y) then
```

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
```diff
@@ -88,8 +90,9 @@ function onTrigger(player,npc)
         player:startEvent(0x000b);
     elseif (theRumor == QUEST_COMPLETED) then
         player:startEvent(0x000e); -- Standard dialog after "The Rumor"
+		player:setVar("theHolyCrestCheck",0);
```
TeoTwawki: you have 2 tabs here. this project uses indents made of spaces. configure your editor to use spaces in place of tabs 4 at a  time (nobody ever actually sits there tapping spacebar and counting it, we just make our tabkeys actually create the correct number of spaces)

Since this line is 2 indents deep, you need 8 spaces here.

TeoTwawki: actually I'm not sure this line even belongs here. you've zero'd it out in more than one spot.

Wiggo32: Ok, editor configured... =/ 
and the second zero is there to make sure you are referred back to the reminder to finish the DRG quest if it is flagged.

PR #4048, opened 2017-09-08
Added Puppetmaster

UnknownX7:
This includes Puppetmaster's Automaton AI, most of the attachments, a system for attachments to call skills through, new listeners for certain attachments, several fixes to the attachment system and Automaton entity memory management, ability to skill up Automaton skills, several fixes to pup's skills and items, all of pup's merits, new mods for pup, up to 75 pup gear, a rework for the current pet mod system (pet mods weren't checking pet type properly), a new gm command for adding all attachments to self, all Automaton weaponskills, new functions for Automaton weaponskills since they are a hybrid of mob's and player's, a fix for the quests for unlocking Automaton parts, new enmity/lua functions for Ventriloquy, made Harlequin frame and head the default for Automatons that may be missing it, and other things I probably forgot.

Issues include, almost all attachments, that use skills, have the wrong animation, some attachments are currently not possible with the way some systems are set up (optic fiber for example), some guesswork was required due to information regarding the AI and attachments being obscure or changed, missing post 75 content, skill chain info for the Automaton's weaponskills is located in their lua files, some balancing needed, the Automaton doesn't resummon upon teleporting, and other things I probably forgot.


teschnei:
Why would you wait for it to get this big for this?

scripts/globals/effects/overdrive.lua
```diff
@@ -9,18 +9,50 @@
 -----------------------------------
 
 function onEffectGain(target,effect)
-end;
+    target:addMod(MOD_OVERLOAD_THRESH, 5000)
+    local pet = target:getPet()
+    if pet then
```
Hozu: Should this not check if the pet is an automaton? Otherwise you could have the effect work on pets from /SMN or /BST.

UnknownX7: You're right, I just tested it. Seems Familiar and Reward from BST also allow this on the Automaton if you use pup/bst as well. I'll have to make a new lua callback to get the pet type.

TeoTwawki: > I'll have to make a new lua callback

One already exists for this.

takhlaq:
closing this to prevent duplicate notifications as pull requests are made directly to pup branch now

Issue #4039, opened 2017-09-07
Holy Circle duration should be 3 minutes, AF/AF+1 Bonuses not implemented


Whitechaser:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** **:** 30170707_0

**_Server Version_** **:** Unknown

**_Source Branch_** **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
https://www.bg-wiki.com/bg/Holy_Circle

Dark Knight's Arcane Circle lasts for 3 minutes but Paladin's Holy Circle does not.
https://github.com/DarkstarProject/darkstar/blob/824fbbd7c79eea6d43524c0d96131abffac56881/scripts/globals/abilities/arcane_circle.lua

https://github.com/DarkstarProject/darkstar/blob/824fbbd7c79eea6d43524c0d96131abffac56881/scripts/globals/abilities/holy_circle.lua
(line 25 should be changed to "target:addStatusEffect(EFFECT_HOLY_CIRCLE,1,0,**180**);")

However! Neither of these two abilities take into consideration artifact armor/artifact armor +1's bonus to the duration. Should be simple though, the technique used for Flee for Thief can be copied.

https://github.com/DarkstarProject/darkstar/blob/824fbbd7c79eea6d43524c0d96131abffac56881/scripts/globals/abilities/flee.lua

So just have a variable called "buff" initialized to 180, then check the feet slot for Chaos Sollerets or Chaos Sollerets +1 for DRK (item id 14096, item id 15359) and Gallant Leggings or Gallant Leggings +1 for PLD (item id 14095, item id 15358). Add 90 to buff if they are equipped.

wrenffxi:
You're right that the AF armor extended duration hasn't been coded yet.

You're also looking at old copies of the files.  This is the current flee script.
https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/abilities/flee.lua

In this case, flee duration is extended using modifiers, rather than checking for equipped item IDs.  If you hard code an item id check, players would gain the benefit even while under the effect of level sync that drops them below item level.

Whitechaser:
Ahh, got it! Thank you. Is there a table that shows exactly what certain modifiers return?

[...]

PR #4060, opened 2017-09-11
Added and implemented AF/AF+1 "Enhances Ancient/Arcane/Holy/Warding Circle" mods, corrected Killer effect from Circle abilities

Whitechaser:
I also included the level 99 stuff that extends the duration of those JAs.

Fix for #4039

### DIFF
--- a/scripts/globals/abilities/ancient_circle.lua
+++ b/scripts/globals/abilities/ancient_circle.lua
@@ -22,5 +22,6 @@ end;
 -----------------------------------
 
 function onUseAbility(player,target,ability)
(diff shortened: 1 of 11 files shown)


Whitechaser:
Crud, I have to redo this, the three new mods I used got taken up by Puppetmaster which got merged first.

takhlaq:
git reset soft <commit before the conflicting one>
git stash
git pull
git stash pop
git commit

..probably
might wanna checkout a new branch locally and push that to this pr's remote branch so you can switch back in case anything goes wrong

Whitechaser:
That didn't seem to do anything. Wahaha. Maybe I picked the wrong commit to reset to. Is the PR okay like this though or would you prefer it without the first two?

Issue #3945, opened 2017-07-25
Ancient magic elemental resist debuff missing


xipies:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30170707_0

**_Server Version_** (type `@revision` in game) **:**
8dd70aa

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
Cast ancient magic, tier I or tier II.

Expect elemental resist debuff to be applied. Does not appear to do so. Looking at ancient magic spell lua files, does not appear to have code to apply status effect. Only appears to have a call to apply the direct nuke damage. Compared with ninjutsu elemental nuke spell lua files, which do have a call to apply status effect.

Hozu:
I imagine they weren't included because the resist debuff values aren't known. Or are they?

Hozu:
A quick look at BG wiki states that AM I is -30 resist for 10s, though it doesn't list if it's affected by the spell being resisted or not. It doesn't have a resist amount or duration for AM IIs unfortunately.

xipies:
A few pages below listing -30 for 15s and being identical to ninjutsu. Although, this could just be one big echo chamber (the repeated info).

http://ffxiclopedia.wikia.com/wiki/Resistance
http://ffxi.allakhazam.com/forum.html?forum=263&mid=1142911918147289979&h=50
http://ffxi.allakhazam.com/forum.html?forum=259&mid=11503938669557970
http://www.ffxionline.com/forum/ffxi-game-related/race-job-type-q-a/blue-mage/68200-haste-vs-refueling/page2

Hozu:
I wouldn't be surprised if it was the exact same as for Ninjutsu nuke debufs. Shouldn't be too difficult to verify on retail if needed.

TeoTwawki:
> I wouldn't be surprised if it was the exact same as for Ninjutsu nuke debufs. Shouldn't be too difficult to verify on retail if needed.

My foggy retail day memories say this is the most likely situation

TeoTwawki:
I also tend to trust BG over other sources for hard numbers

zynjec:
Seems this was added with https://github.com/DarkstarProject/darkstar/pull/4061

[...]

PR #4061, opened 2017-09-11
Added -elemental resistance effect to Ancient Magic

Whitechaser:
Which is just the debuff from NIN spells, nerfed to a constant duration of 10 seconds.

Sources: https://www.bg-wiki.com/bg/Category:Ancient_Magic

Band-aid fix for #3945 

Notes: I don't see any reason why AMII should have a different elemental resistance reduction effect compared to AM's -30 magic evasion on the next element. AMII is just more mp-efficient AM, that does more damage on magic bursts. The -elemental resistance in AMII's case is just a hold-over from its ugly days as AM.

P.S. It's been brought up in a number of issues, but while we do not have separate mods for species-specific elemental damage taken and elemental resist, this might end up affecting damage taken for certain mobs as well. Then again, the effect lasts 10 seconds. And that issue is also present for NIN nukes.

### DIFF
--- a/scripts/globals/spells/burst.lua
+++ b/scripts/globals/spells/burst.lua
@@ -27,5 +27,8 @@ function onSpellCast(caster, target, spell)
     spellParams.M100 = 2;
     spellParams.M200 = 2;
 
+    -- no point in making a separate function for this if the only thing they won't have in common is the name
+    handleNinjutsuDebuff(caster,target,spell,30,10,MOD_EARTHRES);
+
     return doElementalNuke(caster, spell, target, spellParams);
 end;
--- a/scripts/globals/spells/burst_ii.lua
+++ b/scripts/globals/spells/burst_ii.lua
@@ -27,5 +27,8 @@ function onSpellCast(caster, target, spell)
     spellParams.M100 = 2;
     spellParams.M200 = 2;
 
+    -- no point in making a separate function for this if the only thing they won't have in common is the name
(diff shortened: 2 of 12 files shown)


Hozu:
Not quite related to this pull, but TIL AM I/IIs were changed to all have fixed enmity of 300 VE/80 CE.

teschnei:
"It's been brought up in a number of issues, but while we do not have separate mods for species-specific elemental damage taken and elemental resist"
huh? we don't?

Hozu:
@Whitechaser you would use family mobmods (or pool ones) for elemental defense using mods like FIREDEF. Bit annoying that elemental resistance is in mob_family_systems table while elemental defense has to be done by mobmods, but both do exist.

Whitechaser:
Oh! Guess the issue I read was far too old.

Hozu:
https://github.com/DarkstarProject/darkstar/blob/master/sql/mob_family_mods.sql#L660

Right now there only appears to be a lone entry, for Sahagins having 50% water element defense.

Whitechaser:
Oh nice! There's something I can do! I'll get to work entering stuff I can verify to be true.

Whitechaser:
Anything else this PR needs? I'd like to return to working on the magic burst parameters after this gets merged.

takhlaq:
don't think @teschnei got round to reviewing it

Whitechaser:
It doesn't seem like this PR affects much in gameplay mechanics, perhaps someone else can review if teschnei is too busy with other aspects of the project?

PR #4064, opened 2017-09-12
assuming this fixes #4032

### LINKED ISSUE #4032: Lamie No.9 Not coded as a NM
**_I have:_**

**_Client Version_** (type `/ver` in game) **:** 30170728_2

**_Server Version_** (type `!revision` in game) **:** Unknown

**_Source Branch_** (master/stable) **:** Master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** Lamie No.9 @ (J-6) on Map Three in Arrapago Reef is coded as a Very Tough normal mob with no loot pool other than standard Lamie pool. Should be a Notorious Monster.

Sauce: http://ffxiclopedia.wikia.com/wiki/Lamie_No.9

Ninjistix:
let me know if i missed anything or did anything wrong, assuming this fixes #4032


TeoTwawki:
Need to set the mob type to notorious which is in mob_pools.sql none of this other stuff relates to the `/check` being wrong. So one more line needed.

Ninjistix:
Will modify it when I get home

Ninjistix:
I also am assuming that is the right position for that value, assuming that I didn't count wrong lol...

TeoTwawki:
btw nm_spawn_points only gets used if the NM's script calls the function to grab a new spawn point. Any NM that can pop at many diff pos would need that script command and the diff pos's added to the table. This one doesn't even have a script yet so that table entry you added for it won't be used until she does.

Ninjistix:
thanks for the heads up

PR #4071, opened 2017-09-14
Fix (Nbu Latteh): Tweaked Nbu, especially for Mom, the Adventurer quest

davismj:
She was giving a bit too much fame, which put you immediately at Bastok fame 2, preventing you from repeating the repeatable quest. Additionally, there was some minor dialog that wasn't included. I also cleaned up the syntax while I was in there.

### DIFF
--- a/scripts/zones/Bastok_Markets/npcs/Nbu_Latteh.lua
+++ b/scripts/zones/Bastok_Markets/npcs/Nbu_Latteh.lua
@@ -25,18 +25,20 @@ end;
 
 function onTrigger(player,npc)
     local pFame = player:getFameLevel(BASTOK);
-    local MomTheAdventurer = player:getQuestStatus(BASTOK,MOM_THE_ADVENTURER)
+    local momTheAdventurer = player:getQuestStatus(BASTOK,MOM_THE_ADVENTURER);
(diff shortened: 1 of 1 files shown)


TeoTwawki:
dammit demo, my finger was on the merge button 😢 

@davismj 
for future reference, I've been de-hexifying things that don't need to be hex values (not a mask, flag, offset, or bitwise, or are simply list). technically cs events are offset of the npc's dat, but realistically in use they are just ducks in a row cs 1 cs 2 etc so I've been decimal'ing those whenever I edit.

Vars I haven't got the team to settle on any single standard yet but I think there is a preference for lowerCamel like you did here and UPPER_SNAKE_CASE whenever that won't do. I'm OCD about making things consistent across the board, so thanks.

I'm not 100% sure what retails fame reward is these days, I know they dropped fame requirements for the tiers quite a bit.

davismj:
I had a player on Kupo complaining that the quest was broken because he
couldn't repeat it, thus the fame drop. Kupo is era so its a bit different.

TeoTwawki:
There's a setting for the fame value. I was gonna merge anyway but demo beat me too it. Just letting you know in case Kupo wants old school fame requirements. I hear good things about you.

PR #4072, opened 2017-09-15
item_basic update for Sept Client Version

Ninjistix:
- some minor flag correction/adjustments
- removal of : from some item names
- minor name corrections
- new items added

also turns out Gil actually has an item number, added it just because... (prob wont hurt anything)

and if someone else could teach me how to update npc id/s when there's a client version update would be great 👍


sql/item_basic.sql
```diff
@@ -20410,6 +20410,456 @@ INSERT INTO `item_basic` VALUES (29332,0,'moogle_storage_slip_21','storage_slip_
 INSERT INTO `item_basic` VALUES (29333,0,'moogle_storage_slip_22','storage_slip_22',1,61504,99,1,0);
 INSERT INTO `item_basic` VALUES (29334,0,'moogle_storage_slip_23','storage_slip_23',1,61504,99,1,0);
 INSERT INTO `item_basic` VALUES (29695,0,'mangled_mess','mangled_mess',1,12352,99,0,0);
+INSERT INTO `item_basic` VALUES (3718,0,'handful_of_adoulinian_tomatoes','ado._tomatoes',1,28752,99,0,0);
+INSERT INTO `item_basic` VALUES (3722,0,'lion_statue','lion_statue',1,61504,99,0,0);
```
TeoTwawki:
```sql
INSERT INTO `item_basic` VALUES (29695,0,'mangled_mess','mangled_mess',1,12352,99,0,0);
INSERT INTO `item_basic` VALUES (3718,0,'handful_of_adoulinian_tomatoes','ado._tomatoes',1,28752,99,0,0);
```
sort by primary key please (item ID)

TeoTwawki: easier to prevent accidental duping and other mistakes

TeoTwawki: Also quick question: your flag changes go according to our enums? 
https://github.com/DarkstarProject/darkstar/blob/master/src/map/items/item.h#L59
We've occasionally had changes that weren't and didn't realize so double checking.

Ninjistix: will reorder it, and also, the flags were done according to flags from https://raw.githubusercontent.com/Windower/Resources/master/xml/items.xml

assumed they were they same, since the reset is the same as what we have (besides the item changes)

TeoTwawki: I don't recall exactly but I believe windower and polutils differed, and ours matched polutils. I may have that wrong, but the way to be ***certain** is to look at our source file and go by what it has.

Ninjistix: i think this one matches polutils, at least its the same process of the last time i updated item_basic

Ninjistix: and just took a look, the majority of old changes are name fixes/correction, there's actually only a few item flag changes, and stack size.. the reset is new items, from previous version changes.

Ninjistix:
@TeoTwawki are you still checking this? and/or #4091

TeoTwawki:
I haven't had a lot of time lately so am foisting it on someone else to check. if nobody does ima threaten with blind merges!

Ninjistix:
i was honestly hoping this would get merged before the oct client version update, but it seems that ship sailed, after this and #4091 get merged ill work on items for oct

PR #4076, opened 2017-09-16
Add message when receiving gil for quest(s) complete

Wiggo32:
There were no notifications you have received any gil when Kopuro-Popuro's quests were completed. Also, fixed some spacing to comply with Teo's standards.


TeoTwawki:
<3 for fixing bad indents

PR #4085, opened 2017-09-20
Make /anon verbose.

zynjec:
Returns appropriate shown/hidden system messages based on FLAG_ANON state.


src/map/packet_system.cpp
```diff
@@ -4325,6 +4325,14 @@ void SmallPacket0x0DC(map_session_data_t* session, CCharEntity* PChar, CBasicPac
         break;
     case 0x0004:
         PChar->nameflags.flags ^= FLAG_ANON;
+        if(PChar->nameflags.flags & FLAG_ANON)
```
takhlaq: space between `if` and parentheses please
probably gonna have to mass replace that soon cause kj forgets it too

PR #4096, opened 2017-09-27
Various Fix to allow Placeholder Respawn

Wiggo32:
This will make sure all PH are able to respawn and fix any issues of 'Lost' mobs. I only changed the ones that followed the simple formula for handling PH's. Any NM's that had more complicated coding or only 1 Placeholder; I did not touch. They can be taken care of on a case by case basis.

### DIFF
--- a/scripts/zones/Palborough_Mines/mobs/Copper_Beetle.lua
+++ b/scripts/zones/Palborough_Mines/mobs/Copper_Beetle.lua
@@ -22,12 +22,14 @@ function onMobDespawn(mob)
         local checkPH = GetServerVariable("[PH]Bu_Ghi_Howlblade");
         if (ToD <= os.time() and GetMobByID(Bu_Ghi_Howlblade):isDead()) then
             -- Set ToD, DisallowRespawn PH repop, set NM respawn time
-            if (math.random(1,10) == 5 and checkPH == 0) then
+            if (math.random(1,10) == 5) then
                 DisallowRespawn(Bu_Ghi_Howlblade, false);
                 UpdateNMSpawnPoint(Bu_Ghi_Howlblade);
                 GetMobByID(Bu_Ghi_Howlblade):setRespawnTime(GetMobRespawnTime(mobID));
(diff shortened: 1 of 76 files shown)


ymkc:
Problem with this might be if the server crashes after setting the PH, but before killing the NM, the db entry will still be there.

so there might be a (multi gameserver aware) cleanup job needed, to remove all the PHs on startup

scripts/zones/Palborough_Mines/mobs/Copper_Beetle.lua
```diff
@@ -22,13 +22,15 @@ function onMobDespawn(mob)
         local checkPH = GetServerVariable("[PH]Bu_Ghi_Howlblade");
         if (ToD <= os.time() and GetMobByID(Bu_Ghi_Howlblade):isDead()) then
             -- Set ToD, DisallowRespawn PH repop, set NM respawn time
-            if (math.random(1,10) == 5 and checkPH == 0) then
```
TeoTwawki: no point in repeating the same commands once the already happened I tested those beetles and the nm last night there is no problem with having the var checked at the same time as the random number, being together doesn't make it "not used"

TeoTwawki:
@ymkc @Wiggo32 
we could possibly switch to using local vars instead of server vars. local vars get auto-wiped on spawn. Would need to discuss on discord.

TeoTwawki:
Wish I'd thpought of that when he showed us the pastebin on discord. 

Another way to clear them would be check the var on spawn and if mobID matched the var's PH ID wipe it. since that'd obv mean ph already popped and its not a deter'd/dissallowed pop due to server reboot. Or do what retail does and startup with NM windows "already open" and a new PH selected so they clear it. (maybe we already do, I'd have to look at every NM's default value sin the SQL+zone scripts..)

TeoTwawki:
Resolution on discord:
localVar and a new listener to replace servervars, death to servervars. Torches, pitchforks, and partially spoiled produce will be handed out in a commit coming soon.

Wiggo32:
Closing... moving on to a better solution.

PR #4102, opened 2017-09-30
path up tor works -- remove autowarp to top of parradamo tor

wrenffxi:


TeoTwawki:
Before anybody says it:
No it doesn't need a setting. Just revert this one commit in your local copy your server runs off. Should be running servers off of a branch in your fork anyway.

PR #4107, opened 2017-10-02
Quest: A Timely Visit

beauxq:


beauxq:
I had a different pull request for this.
But in trying to correct mistakes, I messed up the branch.
So I deleted the branch and made a new one.

scripts/zones/Southern_San_dOria/npcs/Phillone.lua
```diff
@@ -31,7 +32,32 @@ end;
...
+        player:startEvent(0x005b);  -- find knight to save my family
+    elseif (ATimelyVisit == 1 and ATimelyVisitProgress == 9) then
+        player:startEvent(0x0030);  -- go get scepter in forest
+    elseif (ATimelyVisit == 1 and ATimelyVisitProgress == 11) then
+        player:startEvent(0x0015);  -- figuring everything out
+    elseif (ATimelyVisit == 1 and ATimelyVisitProgress == 12) then
+        player:startEvent(0x0031);  -- save my family at ordelles
+    elseif (ATimelyVisit == 1 and ATimelyVisitProgress == 13) then
+        player:startEvent(0x000e);  -- quest completion
+    elseif (ATimelyVisit == 2) then
```
TeoTwawki: QUEST_AVAILABLE = 0;
QUEST_ACCEPTED  = 1;
QUEST_COMPLETED = 2;

scripts/zones/Southern_San_dOria/npcs/Deraquien.lua
```diff
@@ -30,11 +31,48 @@ end;
...
+    --printf("Fame %d, Rank %d, Nation %d, Mission %d, ATV %d, ATVP %d\n", Fame, Rank, Nation, Mission, ATimelyVisit, ATimelyVisitProgress);
+
     local WildcatSandy = player:getVar("WildcatSandy");
 
+    -- Lure of the Wildcat: San d'Oria
     if (player:getQuestStatus(SANDORIA,LURE_OF_THE_WILDCAT_SAN_D_ORIA) == QUEST_ACCEPTED and player:getMaskBit(WildcatSandy,4) == false) then
         player:startEvent(0x032b);
+
+    -- A Timely Visit
+    elseif ((Nation == 0 and Rank >= 2                                      and Fame >= 3 and ATimelyVisit == 0) or
```
TeoTwawki: kill that huge whitespace

scripts/zones/Southern_San_dOria/npcs/Phillone.lua
TeoTwawki: use the quest status variable in other locations where you have them as  raw numbers as well

scripts/zones/Jugner_Forest/npcs/qm1.lua
```diff
@@ -0,0 +1,70 @@
...
+        end;    
+    end;
+end;
+
+-----------------------------------
+-- onEventUpdate
+-----------------------------------
+
+function onEventUpdate(player,csid,option)
+--printf("CSID: %u",csid);
```
TeoTwawki: indent

scripts/zones/Southern_San_dOria/npcs/Deraquien.lua
beauxq: I think it's better human readability for that line to line up with the same checks in the next line.
You don't think so?

TeoTwawki: nope. messes me up reading it because to far apart. doesn't need to line up with whats on line beneath it, lack of a `then` on same line and ending with a diff lua keyword already tells me its below it.

TeoTwawki: I'd just:
```lua
elseif ((Nation == 0 and Rank >= 2 and Fame >= 3 and ATimelyVisit == 0) or (Nation ~= 0 and 
       (Rank >= 3 or (Mission >=5 and Mission < 255)) and Fame >= 3 and ATimelyVisit == 0)) then
```
personally, but doesn't really matter where exactly you linebreak at.

beauxq: It isn't to tell you that there is something below it. It's for understanding the condition that's being checked.
Your example is way more complex to actually understand what the requirements for the quest are.

I'm looking for a way to make the lines shorter.... (I also replace `Nation == 0` with `Nation == NATION_SANDORIA` which makes the lines longer)
but I'm not going to sacrifice ease of understanding what the requirements for the quest are.

beauxq: I considered putting the rank/mission at the end of the line.
I don't know if LUA has short-circuit evaluation, but if it does, that would make it a little bit less efficient.

TeoTwawki: > It isn't to tell you that there is something below it.
I didn't say it was. I am saying I understand the conditions just fine without a big gap between them and the line break can be anywhere that a keyword shows me its still going. 

> Your example is way more complex to actually understand what the requirements for the quest are.

All I did was move your line break. Literately all I did. The code either speaks for itself or doesn't.

> I'm not going to sacrifice ease of understanding what the requirements for the quest are.
That phrasing disappoints me as a reviewer. My job here is to make sure this is retail, that it works, and that it conforms to project style. The project doesn't conform to the pull requester's wishes. If it were in project already we'd change it next time the file got an edit. This isn't some attack on your preferences. (But it is a reminder to finish porting the style guidelines to github..)

@teschnei takeover, I'm not getting into this.

takhlaq: newline like in teo's example please
we try not to use tons of characters per line

teschnei: Yes, Lua has short circuit evaluation, and yes, you should remove that whitespace because it doesn't conform to the project style (this also applies to when you indent a multi-line condition more than one indent).

It's very confusing when looking at it in limited column width (see: git bash, or even the split diffs on this website)

beauxq: What do you think of a local function just to check that condition?

teschnei: Seems like overkill for a relatively simple construct

takhlaq:
```lua
elseif Fame >= 3 and ATimelyVisit == QUEST_AVAILABLE and (Nation == 0 and Rank >= 2) or
        (Nation ~= 0 and (Rank >= 3 or (Mission >= 5 and Mission < 255)) then
```

beauxq: You haven't seen how much longer the lines get when you replace `Nation == 0` with `Nation == NATION_SANDORIA` and replace `ATimelyVisit == 0` with `ATimelyVisit == QUEST_AVAILABLE`

A local function would also remove Fame, Rank, Nation, and Mission from this scope.

teschnei: the complexity isn't going to change by renaming the numbers - it's fine having it the way demolish has it (plus renaming the 0 to NATION_SANDORIA)

beauxq: @takhlaq The Mission <= 255 is required because a player with no mission returns 255 and 255 >= 5

teschnei: hmm.. from the wiki, the requirements are "being able to enter chateau d'oraguille" - but, I think what you have here for the other nations is "currently on mission 5 or later".  That's not right, isn't it? You could have completed a mission but not accepted the next one.
You just want to know if they're currently on or have previously completed 2-3

teschnei: (that is: getCurrentMission(nation) is 5 or hasCompletedMission(nation, 5) is true

beauxq: I didn't know about `hasCompletedMission`
(I got `getCurrentMission` from somewhere on the forums.)

scripts/zones/Southern_San_dOria/npcs/Deraquien.lua
```diff
@@ -30,11 +31,49 @@ end;
 
 function onTrigger(player,npc)
 
+    local Fame = player:getFameLevel(SANDORIA);
+    local Rank = player:getRank();
+    local Nation = player:getNation();
+    -- access to chateau if not sandorian
+    local Access = (player:getCurrentMission(Nation) == 5) or (player:hasCompletedMission(Nation, 5));
```
teschnei: technically, San d'Orians should be checked here as well, the only difference being you check mission 1-3 (where you have access to the Chateau) instead of checking rank

beauxq: What would be nice is a `player:hasChateauAccess()`
I think it could be used more than just in this quest.

teschnei:
I wouldn't be opposed to adding a helper function in quests.lua or
missions.lua for that

TeoTwawki:
> I wouldn't be opposed to adding a helper function in quests.lua or missions.lua for that
>> What would be nice is a player:hasChateauAccess() I think it could be used more than just in this quest.

+1 for adding that helper

Hozu:
Could just add a function for checking Chateau access in utils.lua, no? I don't think there's any reason the core would be checking that?

teschnei:
Yeah, that's what I said, but in quests.lua

beauxq:
I was wondering where the code is currently that checks when the player actually tries to walk into the chateau.

wrenffxi:
@beauxq Hey, I noticed this PR's gone dry a while after your last question.  In case you never found the answer, the current code that checks whether a player should zone into the Chateau via walking occurs here ...

https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Northern_San_dOria/Zone.lua#L23

... which defines a rectanglar cuboid near the Chateau entrance, and here ...

https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Northern_San_dOria/Zone.lua#L89

... which runs whenever a player enters that region from outside of it.  Hope that helps!

PR #4108, opened 2017-10-02
Update Promyvion Mobs to be Level Appropriate

Wiggo32:
Added some groups to make sure the promy mobs spawn with appropriate levels in their respective areas according to ffxiwiki. Also, fixed one if the Thinkers that had the wrong grouping causing it to look like a Weeper. (16842797)

### DIFF
--- a/sql/mob_groups.sql
+++ b/sql/mob_groups.sql
@@ -12429,6 +12429,105 @@ INSERT INTO `mob_groups` VALUES (14549,3314,217,0,0,0,0,0,0,0,0); -- Rakshas
 -- Hovering Hotpot
 INSERT INTO `mob_groups` VALUES (14550,5841,200,3600,0,4584,7000,9999,58,60,0);
 
+-- Wanderers
+INSERT INTO `mob_groups` VALUES (14551,4283,16,1056,0,4057,0,0,22,28,0);  -- Promyvion-Holla Level 1
+INSERT INTO `mob_groups` VALUES (14552,4283,16,1056,0,4057,0,0,28,32,0);  -- Promyvion-Holla Level 2
+INSERT INTO `mob_groups` VALUES (14553,4283,16,1056,0,4057,0,0,32,35,0);  -- Promyvion-Holla Level 3
+INSERT INTO `mob_groups` VALUES (14554,4283,16,1056,0,4057,0,0,35,36,0);  -- Promyvion-Holla Level 4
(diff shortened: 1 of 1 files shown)


TeoTwawki:
ffxiclopedia or bgwiki or one of the other lesser used wikis?
(there are several)

Wiggo32:
I don't know what you guys prefer for reference... not much changed from what was already in the table just basically split into 4 groups. Tell me what to change and I'll change it.

TeoTwawki:
Just curious which source is all. we usually trust bg if they differ, I think yer fine as is but I try to make sure before I hit the merge button.

Wiggo32:
Oh, my bad... misunderstood.  
http://ffxiclopedia.wikia.com/wiki/Promyvion_Guide

TeoTwawki:
thanks

TeoTwawki:
```
    1st: Easy Prey to Tough 
    2nd: Decent Challenge to Very Tough 
    3rd: Tough to Incredibly Tough 
    4th: Tough to Incredibly Tough 
```
That helps a ton. 
For holla/dem/mea an even match would be lv 30 to a 30 capped player back in the day. So to see anything higher than even match the mob would need to be higher than 30.

Using the weepers as a check case:
```
INSERT INTO `mob_groups` VALUES (14567,4317,16,1056,0,4124,0,0,25,30,0);  -- Promyvion-Holla Level 1```
A 1st floor Weeper would never be "tough". That seems to not match your source material. Unless I missed something.

**Comment by Wiggo32** [comment]
Yea cuz the thinkers were "tough" on the first floor. 
```
INSERT INTO `mob_groups` VALUES (14608,3897,16,1056,0,3461,0,0,29,31,0);  -- Promyvion-Holla Level 1
```
Since there isn't any reference on the specific levels of the mobs per specific floors I went with how I recall seeing them when I used to run through there with my NPC fellow. If you want the weepers to be tough, we can change that. 

(I actually never found a "decent challenge" thinker back in the day on the first floor either but they have it listed there on ffxiwiki as the low level so thats what I made it)

TeoTwawki:
Fair enough. I'll see if anyone has any objections 1st so I'm not "that guy". Tough to gauge this stuff without getting on retail, But I think its more accurate than just randomly creating each one 25-37.

Wiggo32:
Understood... I'm also ready to make changes if someone has a better idea of how it should be. Just let me know.

PR #4109, opened 2017-10-03
Store per-account expansion and feature bitmasks in account table.

zynjec:
Default bitmasks are the same as the values that were in login_darkstar.conf.

image

### DIFF
--- a/conf/login_darkstar.conf
+++ b/conf/login_darkstar.conf
@@ -55,54 +55,6 @@ mysql_database:  dspdb
 #Search Server Port
 search_server_port: 54002
 
-#Expansion Icons - 2 Bytes
-#
-#Byte 1 - Zilart to A Shantotto Ascension
-#00000000 Bit0 - Not Used - Original FFXI bit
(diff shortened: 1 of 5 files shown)


maxtherabbit:
but y tho

TeoTwawki:
@zynjec you know the expansions lit up is purely cosmetic right?

zynjec:
I do understand it's cosmetic.

But more retail-like, also planning to add an expansion check to zone line requests (0x05E), which would make it a bit more than cosmetic.

image

TeoTwawki:
just checkin ^^

TeoTwawki:
was on mobile earlier. was just saying we ought to get that error message in eventually because right now game doesn't care and just zones you anyway.

TeoTwawki:
was on mobile earlier and didn't see the screenshot.*

takhlaq:
why was this closed?

zynjec:
Bad at git!

takhlaq:
might also want to enable them all in char create query

tagban:
I love this idea. Thanks!

PR #4117, opened 2017-10-08
Warnings

ShelbyZ:
- Targeting warnings found during gcc7 build of system
- Most changes fall into narrowing values/implicit conversion, unused return types, non const char* string constants, and empty format printing

### DIFF
--- a/src/common/socket.h
+++ b/src/common/socket.h
@@ -9,7 +9,9 @@
 
 
 #ifdef WIN32
-    #define FD_SETSIZE 1000
+    #ifndef FD_SETSIZE
+        #define FD_SETSIZE 1000
+    #endif
 	#include <winsock2.h>
 	typedef long in_addr_t;
 #else
--- a/src/common/socket.h
+++ b/src/common/socket.h
@@ -9,9 +9,7 @@
 
 
 #ifdef WIN32
-    #ifndef FD_SETSIZE
-        #define FD_SETSIZE 1000
-    #endif
+    #define FD_SETSIZE 1000
 	#include <winsock2.h>
 	typedef long in_addr_t;
 #else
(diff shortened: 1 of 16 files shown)


takhlaq:
hmmm should the `puts`/`printf` stuff just use `ShowMessage`  instead @teschnei

src/common/zlib.cpp
```diff
@@ -79,7 +79,11 @@ static bool read_to_vector(const std::string &file, std::vector<uint32> &vec)
     fseek(f, 0, SEEK_SET);
 
     vec.resize(size / sizeof(uint32));
-    fread(vec.data(), sizeof(uint32), vec.size(), f);
+    if (fread(vec.data(), sizeof(uint32), vec.size(), f) == -1)
```
Cloudef: Probably better check is to do `!= vec.size()`

src/search/search.cpp
```diff
@@ -126,14 +126,14 @@ void PrintPacket(char* data, int size)
         if (((y + 1) % 16) == 0)
         {
             message[48] = '\n';
-            printf(message);
+            puts(message);
```
Cloudef: Note, puts appends \n at the end  of string. Either use printf("%s", message); or fputs()

src/common/socket.h
```diff
@@ -9,7 +9,9 @@
 
 
 #ifdef WIN32
-    #define FD_SETSIZE 1000
+    #ifndef FD_SETSIZE
```
teschnei: this is just hiding the real problem, unfortunately (the real problem being that include order is not right - this should always be set before including winsock2, but something is including winsock2 before this that I haven't found yet)

this would just not override the fd size at all

PR #4129, opened 2017-10-16
Battlefield Overhaul

wrenffxi:
Rewrote bcnm.lua.  Battlefield registration and entry now controlled by a few LUA tables.  It is now possible to have multiple battlefields appear in the menu for any given trade.

Limbus arena instances now match client.  GMs can now !goto limbus battlefields.  Players who disconnect from limbus will automatically be readded to their run if it's still in progress.  Their position will be set to the entrance of their instance.

Removed as many of the differences between BCNMs, Dynamis, and Limbus as possible -- in many cases we had separate functions for doing the same things for different battlefield types. I condensed those.

BCNM chests now disappear after being opened.

Removed all IsMobDead() calls.

Added a where clause to the SQL query that writes battlefield record back into DB.

### DIFF
--- a/scripts/globals/bcnm.lua
+++ b/scripts/globals/bcnm.lua
@@ -1,761 +1,956 @@
-require("scripts/globals/status")
+--[[
+
+--]]
+
 require("scripts/globals/keyitems")
 require("scripts/globals/missions")
 require("scripts/globals/quests")
+require("scripts/globals/status")
 
--- NEW SYSTEM BCNM NOTES
--- The "core" functions TradeBCNM EventUpdateBCNM EventTriggerBCNM EventFinishBCNM all return TRUE if the action performed is covered by the function.
--- This means all the old code will still be executed if the new functions don't support it. This means that there is effectively 'backwards compatibility' with the old system.
+-----------------------------------------------
+-- battlefields by zone
+-- captured from client 2017-10-02
+-----------------------------------------------
 
--- array to map (for each zone) the item id of the valid trade item with the bcnmid in the database
--- e.g. zone, {itemid, bcnmid, itemid, bcnmid, itemid, bcnmid}
--- DO NOT INCLUDE MAAT FIGHTS
-itemid_bcnmid_map =
-{
-    6,   {0, 0}, -- Bearclaw_Pinnacle
-    8,   {0, 0}, -- Boneyard_Gully
-    10,  {0, 0}, -- The_Shrouded_Maw
-    13,  {0, 0}, -- Mine_Shaft_2716
-    17,  {0, 0}, -- Spire of Holla
-    19,  {0, 0}, -- Spire of Dem
-    21,  {0, 0}, -- Spire of Mea
-    23,  {0, 0}, -- Spire of Vahzl
-    29,  {0, 0}, -- Riverne Site #B01
-    31,  {0, 0}, -- Monarch Linn
-    32,  {0, 0}, -- Sealion's Den
-    35,  {0, 0}, -- The Garden of RuHmet
-    36,  {0, 0}, -- Empyreal Paradox
-    57,  {0, 0}, -- Talacca Cove
-    64,  {0, 0}, -- Navukgo Execution Chamber
-    67,  {0, 0}, -- Jade Sepulcher
-    139, {1177, 4, 1552, 10, 1553, 11, 1131, 12, 1175, 15, 1180, 17}, -- Horlais Peak
-    140, {1551, 34, 1552, 35, 1552, 36}, -- Ghelsba Outpost
-    144, {1166, 68, 1178, 81, 1553, 76, 1180, 82, 1130, 79, 1552, 73}, -- Waughroon Shrine
-    146, {1553, 107, 1551, 105, 1177, 100}, -- Balgas Dias
-    163, {1130, 129, 1130, 130}, -- Sacrificial Chamber
-    168, {0, 0}, -- Chamber of Oracles
-    170, {0, 0}, -- Full Moon Fountain
-    180, {1550, 293}, -- LaLoff Amphitheater
-    181, {0, 0}, -- The Celestial Nexus
-    201, {1546, 418, 1174, 417}, -- Cloister of Gales
-    202, {1548, 450, 1172, 449}, -- Cloister of Storms
-    203, {1545, 482, 1171, 481}, -- Cloister of Frost
-    206, {0, 0}, -- Qu'Bia Arena
-    207, {1544, 545}, -- Cloister of Flames
-    209, {1547, 578, 1169, 577}, -- Cloister of Tremors
-    211, {1549, 609} -- Cloister of Tides
-}
+--[[
+    [zoneId] = {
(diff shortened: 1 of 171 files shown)


takhlaq:
you'll want to control when to update the client's position too
see how bcnmshit does it (though the luautils::onEventUpdate return flag for setting position update shit doesn't work
also grab the loot tables i moved over to lua (fuck sql)

i'll give the rest a proper look over in a bit

scripts/globals/bcnm.lua
```diff
@@ -1,761 +1,956 @@
...
+-- get battlefield id for a given zone and bit
+-----------------------------------------------
 
-    print("UPDATE csid "..csid.." option "..option)
-    -- seen: option 2, 3, 0 in that order
-    if (csid == 0x7d03 and option == 2) then -- leaving a BCNM the player is currently in.
-        player:updateEvent(3)
-        return true
-    elseif (csid == 0x7d03 and option == 3) then -- leaving a BCNM the player is currently in.
-        player:u
```
takhlaq: you'll wanna handle party members registration too with return codes, and depending on that in 0x05C handler update their position
https://github.com/DarkstarProject/darkstar/blob/bcnmshit/scripts/globals/bcnm.lua#L503

src/map/packet_system.cpp
```diff
@@ -2631,11 +2631,12 @@ void SmallPacket0x05C(map_session_data_t* session, CCharEntity* PChar, CBasicPac
         PChar->loc.p.x = RBUFF(data, (0x04));
         PChar->loc.p.y = RBUFF(data, (0x08));
         PChar->loc.p.z = RBUFF(data, (0x0C));
+        uint16 extras = RBUFW(data, (0x14));
         PChar->loc.p.rotation = RBUFB(data, (0x1F));
 
         if (RBUFB(data, (0x1E)) != 0)
         {
-            luautils::OnEventUpdate(PChar, EventID, 0);
+            luautils::OnEventUpdate(PChar, EventID, 0, extras);
```
takhlaq: https://github.com/DarkstarProject/darkstar/blob/bcnmshit/src/map/packet_system.cpp#L2628

src/map/battlefield_handler.cpp
```diff
@@ -36,183 +36,122 @@
...
-{
+CBattlefieldHandler::CBattlefieldHandler(uint16 zoneid) {
     m_ZoneId = zoneid;
 
-    //Dynamis zone (need to add COP dyna zone)
-    //added ghelsba outpost here, 1 battlefield only
-    if (m_ZoneId > 184 && m_ZoneId < 189 || m_ZoneId > 133 && m_ZoneId < 136 || m_ZoneId == 140 || m_ZoneId == 35 || m_ZoneId > 38 && m_ZoneId < 43)
-    {
+    // The Garden of RuHmet, Dynamis zones, and Ghelsba Outpost
+    if (m_ZoneId == 35 || m_ZoneId > 38 && m_ZoneId < 43 || m_ZoneId > 133 && m_ZoneId < 136 || m_ZoneId == 140 || m_ZoneId > 184 && m_ZoneId < 189 ) {
```
takhlaq: not sure we should keep these ids hardcoded (i hate them)
maybe something like https://github.com/DarkstarProject/darkstar/blob/bcnmshit/src/map/battlefield_handler.cpp#L57

src/map/battlefield_handler.cpp
```diff
@@ -36,183 +36,122 @@
...
+
 void CBattlefieldHandler::handleBattlefields(time_point tick) {
     for (int i = 0; i < m_MaxBattlefields; i++) {
         if (m_Battlefields[i] != nullptr) { //handle it!
             CBattlefield* PBattlefield = m_Battlefields[i];
             int instzone = PBattlefield->getZoneId();
 
-            //handle time remaining prompts (since its useful!) Prompts every minute
+            // update timers
             auto time_elapsed = std::chrono::duration_cast<std::chrono::seconds>(tick - PBattlefield->getStartTime());
```
takhlaq: maybe move this to lua like instances do or on my branch
https://github.com/DarkstarProject/darkstar/blob/bcnmshit/scripts/globals/battlefield.lua#L57

also i think kj was saying the way instances do it is how retail handles it (can grab https://github.com/DarkstarProject/darkstar/blob/bcnmshit/scripts/globals/battlefield.lua#L110 if you feel like handling it in lua instead)

src/map/battlefield_handler.cpp
```diff
@@ -602,104 +405,43 @@ uint8 CBattlefieldHandler::findBattlefieldIDFor(CCharEntity* PChar) {
...
-int CBattlefieldHandler::getUniqueDynaID(uint16 id) {
 
-    CBattlefield* PBattlefield = m_Battlefields[0];
-    return PBattlefield->getDynaUniqueID();
-}
+//========================DYNAMIS FUNCTIONS=============================================//
 
-int CBattlefieldHandler::registerDynamis(uint16 id, CCharEntity* PChar) {
-    if (!hasFreeBattlefield()) {
+int CBattlefieldHandler::registerDynamis(uint16 bnum, uint16 id, CCharEntity* PChar) {
```
takhlaq: ideally dynamis functions would be removed entirely, since there's no need for them besides spawning next wave of mobs (so probably wanna handle tick in lua for special snowflakes like dyna and temenos)

wrenffxi: removed registerDynamis in further commit

src/map/packet_system.cpp
wrenffxi: Removed onEventUpdate 4-parameter overload in further commit.

wrenffxi:
Grr.  One of the additional commits I added has introduced the old "Another event was called while an event was still running" error when you open the BCNM treasure chest. :(

TeoTwawki:
>  (they aren't bcnms in the first place)

@teschnei old timey dyna *was* but no longer is. Some of limbus's events suggested they used the same system back then. Dunno about modern retail limbus at all haven't done it since 2010 or so.

TeoTwawki:
p.s.
My take is lua-it till we remove-a-it

wrenffxi:
Pushed latest changes to this PR.  I can't figure out why opening BCNM chest now causes "Another event was called ..." message and blackscreen. :(  This wasn't happening when I initially submitted the PR, so it's got to be caused by one of the additional commits.  Anyone have an idea?

takhlaq:
```cpp
-    if (m_PlayerList.empty())
 -        cleanup();
```
maybe that or something went wrong when changing the bcnm type

takhlaq:
i'll be redoing my changes on your fork, please dont push anything to `bf` branch (unless i ask you to of course)

takhlaq:
https://github.com/DarkstarProject/darkstar/tree/bcnmcrap

wrenffxi:
So I've got BCNMs working in my branch again, but haven't pushed anything else to :bf.  Let me know if you want these additional commits.

takhlaq:
make any additional prs to bcnmcrap branch

PR #4130, opened 2017-10-17
Msvc warnings

ShelbyZ:
- Enable project warnnings /w3
- _CRT_SECURE_NO_WARNINGS to avoid MS secure functions that would break
cross-platform
- Add _CRT_NONSTDC_NO_DEPRECATE/_WINSOCK_DEPRECATED_NO_WARNINGS to
silence POSIX/deprecated winsock methods
- Remove /ZI from release builds as target was /Zi which causes a warning from the additional command
- most issues are type conversions
- some involved type changes
- some involved loop conversion to ranged-based
- std::any/std::any_cast to handle void* for CTaskMgr


src/map/entities/npcentity.cpp
```diff
@@ -29,6 +29,52 @@
...
+/************************************************************************
+*																		*
+*  Таймер для закрывания дверей.										*
+*																		*
+*  Вообще можно подумать о реализации универсального таймера для		*
+*  изменения анимации NPC, а не только закрытия дверей.					*
+*																		*
+************************************************************************/
+
+int32 close_door(time_point tick, CTaskMgr::CTask* PTask)
```
teschnei: i delete these, they should stay deleted

src/map/packets/server_message.cpp
```diff
@@ -35,24 +35,24 @@ CServerMessagePacket::CServerMessagePacket(const string_t message, int8 language
     WBUFB(data, (0x05) ) = 1;
     WBUFB(data, (0x06) ) = 1;
     WBUFB(data, (0x07) ) = language;
-    WBUFL(data, (0x08) ) = timestamp == 0 ? time(0) : timestamp;
+    WBUFL(data, (0x08) ) = timestamp == 0 ? UINT32_MIN : timestamp;
```
teschnei: these aren't equivalent, i don't think? time(0) vs UINT32_MIN?

takhlaq: iirc time(0) returns current timestamp

PR #4131, opened 2017-10-17
Updated Charybdis's lottery spawn logic

Wiggo32:
Might need to make an adjustment.

### DIFF
--- a/scripts/zones/Sea_Serpent_Grotto/mobs/Charybdis.lua
+++ b/scripts/zones/Sea_Serpent_Grotto/mobs/Charybdis.lua
@@ -25,16 +25,4 @@ end;
 -----------------------------------
 
 function onMobDespawn(mob)
-
-    -- Set Charybdis Window Open Time
-    local wait = math.random(28800,43200) -- 8-12 hours
-    SetServerVariable("[POP]Charybdis", os.time() + wait );
-    DisallowRespawn(mob:getID(), true);
-
-    -- Set PH back to normal, then set to respawn spawn
-    local PH = GetServerVariable("[PH]Charybdis");
-    SetServerVariable("[PH]Charybdis", 0);
-    DisallowRespawn(PH, false);
-    GetMobByID(PH):setRespawnTime(GetMobRespawnTime(PH));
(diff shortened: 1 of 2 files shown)


scripts/zones/Sea_Serpent_Grotto/MobIDs.lua
```diff
@@ -5,13 +5,10 @@
...
 
 -- Charybdis
-Charybdis = 17498522;
 Charybdis_PH =
 {
-    [1] = 17498520,
-    [2] = 17498518,
-    [17498520] = '1', -- -138.181, 48.389, -338.001
-    [17498518] = '1', -- -212.407, 38.538, -342.544
+    [17498518] = 17498522;  -- -138.181, 48.389, -338.001
```
TeoTwawki: You need a comma after rows that aren't the end of the table

Wiggo32: I've been trying to figure out where the comma goes on this one for a while now. I apologize but it's just eluding me. I think you're referring to where the old code has commas but the new code has semicolons? Forgive me but, I thought it could be either or. I just followed Lotus' example:
https://github.com/DarkstarProject/darkstar/blob/9efe30a3f86f4b0787e3fb2146f0e1d8efd820c1/scripts/zones/South_Gustaberg/MobIDs.lua#L8

teschnei:
oh hey I did do that!

the lua docs say you can use semicolons, but its better to use those as
"section" delimiters using commas in between (i should have used commas
there)

TeoTwawki:
Oh. I though lua always ignored semicolon completely and didnt even noticed you used them on lizzy lol

scripts/zones/Sea_Serpent_Grotto/MobIDs.lua
TeoTwawki: [17498518] = 17498522,  -- -138.181, 48.389, -338.001

is what you want

the final line won't need it

Wiggo32:
no problem! I'll fix it!

PR #4135, opened 2017-10-18
Updated Ellyllon's lottery spawn logic

Wiggo32:


sql/nm_spawn_points.sql
```diff
@@ -6544,6 +6544,10 @@ INSERT INTO `nm_spawn_points` VALUES (17195318,10,-331.000, -15.000, -112.000);
 INSERT INTO `nm_spawn_points` VALUES (17195318,11,-320.000, -15.000, -14.000);
 INSERT INTO `nm_spawn_points` VALUES (17195318,12,-343.000, -7.000, 50.000);
 
+-- Ellyllon
+INSERT INTO `nm_spawn_points` VALUES (17403939,0,216.000, 8.129, -157.100);
```
TeoTwawki: This table only gets used to enable giving an NM multiple positions. No point using it if he's only ever spawning in one precise spot. Can either have more pos's for him or else use the mob_spawn_points entry to set it and remove `UpdateNMSpawnPoint` 

I don't think anybody will care which you choose since he's only in that one room.

With most other NM it is preferable to have more positions if the area if pops at is well known from retail to get that "gotta hunt it" feel.  The project hasn't required the positions to be precise (so far) since getting exact points is tricky even by packet capture. **Existing NMs you convert should already have nm_spawn_points entries if they need them.** :)


Rest looks good.

Wiggo32: You're absolutely right! I wasn't sure about just adding arbitrary points though. But, since the NM is only in that room I put three more around the room to give that " retail hunt" feel you described. There ya go sir!

PR #4143, opened 2017-10-20
Update Spawn Logic for Charybdis - Take 2

Wiggo32:


TeoTwawki:
push to same branch of same fork next time so you won't have to make a 2nd pr that doesnt share its review comments  etc..

Issue #4041, opened 2017-09-07
!delquest / !addquest don't like 4 as a logID


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**30170728_2

**_Server Version_** (type `!revision` in game) **:**unknown

**_Source Branch_** (master/stable) **:**master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
!delquest and !addquest both give an "invalid logID" error when attempting to use '4' (the ID for OTHER_AREAS) as a logID.

DFPercush:
See also [PR #4150](https://github.com/DarkstarProject/darkstar/pull/4150) - completequest was overlooked here

metalfiiish:
Issue can be closed as it was resolved in PR #4150

PR #4144, opened 2017-10-20
Fixed quest commands for Other Areas log (global name conflict between log_ids.lua and zone.lua)

DFPercush:
GM Commands addquest, delquest, and checkquest were broken for log id 4, other areas, due to _G["OTHER_AREAS"] evaluating to 4 upon being set by zone.lua, rather than the table set up in log_ids.lua. Because the integer value in zone.lua is a dependency for other modules, I left it alone and appended _LOG to the log_id table var, because that variable name is read dynamically from a config further down in the same file. Commands tested ok now. 
/ver 30170905_5
Revision is: Unknown (guessing branch bcnmcrap is the latest, building from 0df5b8)

### DIFF
--- a/scripts/commands/addquest.lua
+++ b/scripts/commands/addquest.lua
@@ -28,7 +28,12 @@ function onTrigger(player, logId, questId, target)
         logId = QUEST_LOGS[logId];
     end
     if (logId ~= nil) then
-        logId = _G[string.upper(logId)];
+		local logTableTest = _G[string.upper(logId)];
+		if (type(logTableTest) == "table") then
+			logId = logTableTest;
+		else
+			logId = _G[string.upper(logId .. "_LOG")];
+		end
     end
     if ((type(logId) == "table") and logId.quest_log ~= nil) then
         logName = logId.full_name;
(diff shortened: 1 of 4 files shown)


takhlaq:
thanks for the pr!
could you switch to master and redo your changes there 
should be able to just checkout master and cherry pick this commit, and then force push to this pull request's branch

Wiggo32:
https://github.com/DarkstarProject/darkstar/issues/4041
resolves this issue

DFPercush:
By cherry pick, do you mean just edit the changes again? Because I'm deathly afraid of merging a whole bunch of crap that shouldn't be a part of master. I wonder if there's a way to just apply the diff from this commit to master without including all the previous commits on the bcnm branch. My git usage has not included this scenario, so forgive me while I figure this out.

teschnei:
You just described exactly what git cherry-pick does ;)

DFPercush:
How does it look now? The automatic check is complaining but I think it's confused about switching the base.

teschnei:
looks right

btw - I'm guessing the commands won't accept OTHER_AREAS anymore.  it would probably be a good idea to add a check in add/del/checkquest if the log was typed as OTHER_AREAS and just translate it to OTHER_AREAS_LOG

DFPercush:
Figured I would just make the _LOG suffix ubiquitous in case others need to be renamed in the future. Try without, then with, if neither is a table with ['quest_log'] then fail validity test as normal.

takhlaq:
looks good, thanks!

PR #4146, opened 2017-10-20
Fix -Flame spouts now despawn every 3 hrs

Wiggo32:
Adjusted the FlameSpout value indicators to correctly align with the FlameSpout NPC that allows players to pass.


TeoTwawki:
which client ver was this tested on? just so I know ahead when I check the base ID at after we update version.info

Wiggo32:
30170905_5

TeoTwawki:
crap. didnt occur to me to ask/check/look just after I hit the button.. is 17617204 not a flame spout? coz its prolly just that  umber off by 5 cozing the rest and I shouldn't be juggling my phone and hitting merge haha guys dont kill me >.>

PR #4151, opened 2017-10-23
avoid stl exception erasing from an empty vector

DFPercush:
Added a more explicit loop to find and delete ids from CBattlefield::m_registeredPlayers. This vector was not being populated in my tests (solo), so at size 0, remove_if was initializing to past the end() of the container, which throws an out of range exception from erase().

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -272,7 +272,20 @@ bool CBattlefield::InsertEntity(CBaseEntity* PEntity, bool enter, BATTLEFIELDMOB
             {
                 ApplyLevelCap(static_cast<CCharEntity*>( PEntity ));
                 m_EnteredPlayers.push_back(PEntity->id);
-                m_RegisteredPlayers.erase(std::remove_if(m_RegisteredPlayers.begin(), m_RegisteredPlayers.end(), [PEntity](auto id) { return PEntity->id == id; }));
+                //m_RegisteredPlayers.erase(std::remove_if(m_RegisteredPlayers.begin(), m_RegisteredPlayers.end(), [PEntity](auto id) { return PEntity->id == id; }));
+                bool wasRegistered = false;
+                do
+                {
+                    for (auto itRegPc = m_RegisteredPlayers.begin(); itRegPc != m_RegisteredPlayers.end(); itRegPc++)
+                    {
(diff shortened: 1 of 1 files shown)


takhlaq:
would 
```cpp
if (m_RegisteredPlayers.size())
  
  m_RegisteredPlayers.erase(std::remove_if(m_RegisteredPlayers.begin(), m_RegisteredPlayers.end(), [PEntity](auto id) { return PEntity->id == id; }));
```
not work?

DFPercush:
No. Try this example:
```C++
#include <vector>
#include <algorithm>
struct TestEntity{int id;};
int main(int argc, char**argv)
{
    std::vector<int> m_RegisteredPlayers;
    m_RegisteredPlayers.push_back({ 1 });
    m_RegisteredPlayers.push_back({ 2 });
    m_RegisteredPlayers.push_back({ 101 });
    m_RegisteredPlayers.push_back({ 102 });
    TestEntity needle{ 99 };
    TestEntity* PEntity = &needle;
    if (m_RegisteredPlayers.size())
        m_RegisteredPlayers.erase(std::remove_if(m_RegisteredPlayers.begin(), m_RegisteredPlayers.end(), [PEntity](auto id) { return PEntity->id == id; }));
    return 0;
}
```
If the needle is not in the haystack, remove_if will initialize to end() and be out of range. This is why STL really.... bugs me, hhhhah. As far as why my code needs that many nested loops, iterators tend to be invalidated when you modify the container. erase() does return an iterator past the element which is being deleted, and this theoretically could be assigned to the loop iterator, but the 'for' loop will advance that iterator again before doing another comparison, so it would skip over an element. TBH, I don't see anywhere m_RegisteredPlayers is being added to, I think that's an unfinished thought. The only way it could possibly happen is through a lua binding. So in practice, on this revision, that simple if (size) test might work only because its true path will never be executed. But it sows the seeds for future headaches.

teschnei:
The only reason your test code fails is because you are misusing erase/remove

```
    m_RegisteredPlayers.erase(std::remove_if(m_RegisteredPlayers.begin(), m_RegisteredPlayers.end(), [PEntity](auto id) { return PEntity->id == id; }), m_RegisteredPlayers.end());
```

PR #4160, opened 2017-10-26
Implement Sprinter's Shoes

Wiggo32:
Who wants to run a little faster?

### DIFF
--- a/scripts/globals/items/sprinters_shoes.lua
+++ b/scripts/globals/items/sprinters_shoes.lua
@@ -3,8 +3,8 @@
 -- Item: Sprinter's Shoes
 -- Item Effect: Quickening for 60 minutes
 -----------------------------------------
-
 require("scripts/globals/settings");
+require("scripts/globals/msg");
 
 -----------------------------------------
 -- OnItemCheck
@@ -20,7 +20,6 @@ end;
 -----------------------------------------
 
 function onItemUse(target)
-    target:delStatusEffect(EFFECT_QUICKENING);
     target:addStatusEffect(EFFECT_QUICKENING, 10, 0, 3600);
-    target:messageBasic(266,0, EFFECT_QUICKENING);
+    target:messageBasic(msgBasic.GAINS_EFFECT,0, EFFECT_QUICKENING);
 end;
--- a/scripts/globals/items/sprinters_shoes.lua
+++ b/scripts/globals/items/sprinters_shoes.lua
@@ -1,0 +1,26 @@
+-----------------------------------------
+-- ID: 15754
+-- Item: Sprinter's Shoes
+-- Item Effect: Quickening for 60 minutes
+-----------------------------------------
+
+require("scripts/globals/settings");
+
+-----------------------------------------
+-- OnItemCheck
+-----------------------------------------
+
+function onItemCheck(target)
+    return 0;
+end;
+
+
+-----------------------------------------
+-- OnItemUse
+-----------------------------------------
+
+function onItemUse(target)
+    target:delStatusEffect(EFFECT_QUICKENING);
+    target:addStatusEffect(EFFECT_QUICKENING, 10, 0, 3600);
+    target:messageBasic(266,0, EFFECT_QUICKENING);
+end;


scripts/globals/items/sprinters_shoes.lua
```diff
@@ -0,0 +1,26 @@
...
+    return 0;
+end;
+
+
+-----------------------------------------
+-- OnItemUse
+-----------------------------------------
+
+function onItemUse(target)
+    target:delStatusEffect(EFFECT_QUICKENING);
```
TeoTwawki: no need to delStatusEffect, overwriting is baked into addStatusEffect and controlled by https://github.com/DarkstarProject/darkstar/blob/master/sql/status_effects.sql#L31

scripts/globals/items/sprinters_shoes.lua
```diff
@@ -0,0 +1,26 @@
...
+    target:addStatusEffect(EFFECT_QUICKENING, 10, 0, 3600);
+    target:messageBasic(266,0, EFFECT_QUICKENING);
```
TeoTwawki: adding the effect already should trigger this message. Even if it didn't,  we should
```lua
    GAINS_EFFECT           = 266, -- <target> gains the effect of <status>.
```
to the msg global instead off using "266" directly.

scripts/globals/items/sprinters_shoes.lua
TeoTwawki: https://github.com/DarkstarProject/darkstar/blob/master/src/map/status_effect.h#L34
```cpp
enum EFFECTOVERWRITE
{
    EFFECTOVERWRITE_EQUAL_HIGHER = 0, // only overwrite if equal or higher
    EFFECTOVERWRITE_HIGHER = 1, // only overwrite if higher
    EFFECTOVERWRITE_NEVER = 2, // never overwrite
    EFFECTOVERWRITE_ALWAYS = 3, // always overwrite no matter
    EFFECTOVERWRITE_IGNORE = 4 // ignore dupes
};
```

Wiggo32: Ahhh, yes deleting that line still has the effect be overwritten. Nice!

scripts/globals/items/sprinters_shoes.lua
Wiggo32: However, I cannot get any of these items to trigger the message without the messageBasic line
https://github.com/DarkstarProject/darkstar/blob/2ca782d2c7cbdd3582827ea995c294d99595f176/scripts/globals/items/hermes_quencher.lua#L31
Maybe I'm doing something wrong? It triggers a message when the effect wears off though.

scripts/globals/items/sprinters_shoes.lua
Wiggo32: But, how do I deal with the conflict with things that have flee? 
http://ffxiclopedia.wikia.com/wiki/Powder_Boots
```
Notes: Does not overwrite or stack with effects of Sprinter's Shoes.
```

scripts/globals/items/sprinters_shoes.lua
TeoTwawki: I dunno seems to be missing globally, was sure it worked at one point..We even have a flag to suppress the msg in place..

Pull/rebase onto latest and then try this.

[sprinters_shoes.zip](https://github.com/DarkstarProject/darkstar/files/1420084/sprinters_shoes.zip)

scripts/globals/items/sprinters_shoes.lua
TeoTwawki: well those are 2 different effects. you just have one effect script check for the other (the effect, not the item) and behave accordingly. Or if the effect straight up won't apply that's even easier: we have an sql column for that.

scripts/globals/items/sprinters_shoes.lua
Wiggo32: Yes, that works now as I see you've added the GAINS_EFFECT msg.

TeoTwawki:
Free weeks coming up on retail soon, merging as is and will cleanup anything off about it during free week

PR #4161, opened 2017-10-26
Inclusion of Text IDs in Kazham Zone.lua

Wiggo32:
Conquest Updates were saying strange things... all better now.


TeoTwawki:
huh..howd that get missed/lost...

PR #4172, opened 2017-10-29
Handle pets in instances

ymkc:
Allow to call pets in instanced battles

### DIFF
--- a/src/map/instance_loader.cpp
+++ b/src/map/instance_loader.cpp
@@ -296,7 +296,5 @@ CInstance* CInstanceLoader::LoadInstance(CInstance* instance)
         instance = nullptr;
     }
 
-    //TODO: pets
-
     return instance;
 }
--- a/src/map/utils/petutils.cpp
+++ b/src/map/utils/petutils.cpp
@@ -43,6 +43,7 @@ This file is part of DarkStar-server source code.
 #include "../mob_spell_list.h"
 #include "../enmity_container.h"
 #include "../items/item_weapon.h"
+#include "../zone_instance.h"
 
 #include "../ai/ai_container.h"
 #include "../ai/controllers/mob_controller.h"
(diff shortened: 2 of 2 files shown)


src/map/instance_loader.cpp
```diff
@@ -44,6 +45,8 @@ CInstanceLoader::CInstanceLoader(uint8 instanceid, CZone* PZone, CCharEntity* PR
     zone = PZone;
     CInstance* instance = ((CZoneInstance*)PZone)->CreateInstance(instanceid);
 
+    petutils::LoadPetList();
```
teschnei: I think this line isn't needed

if it is, we have to do it a different way because it will corrupt any query running on the main thread at the same time (probably by passing the sql handle in or something - but the pet list is global anyways so it should be loaded already)

src/map/utils/petutils.cpp
```diff
@@ -778,10 +779,18 @@ namespace petutils
...
             PMaster->StatusEffectContainer->CopyConfrontationEffect(PPet);
 
-            PMaster->PPet = PPet;
             PPet->PMaster = PMaster;
 
-            PMaster->loc.zone->InsertPET(PPet);
+            if (PMaster->PInstance)
+            {
+                PPet->PInstance = PMaster->PInstance;
+                PMaster->PInstance->InsertPET(PPet);
```
teschnei: it looks like CZoneInstance already takes care of calling InsertPET on the right instance, so it should be possible to only set PPet->PInstance, and then do PMaster->loc.zone->InsertPET(PPet) for either way (eliminating the else, basically)

ymkc: ok I will test this in a bit

ymkc: yep, works

src/map/instance.cpp
```diff
@@ -239,6 +239,9 @@ void CInstance::ClearEntities()
         if (static_cast<CBattleEntity*>(entity.second)->isAlive())
         {
             entity.second->PAI->ClearStateStack();
+
+            if (entity.second->objtype == TYPE_MOB)
```
teschnei: what scenario did you encounter where this was needed?  I think the purpose of it was simply to make things stop attacking you when you complete the instance (and get put into a cutscene immediately)

ymkc: i believe it was a cheap fix for the cs not triggering when char was in the middle of a ws or casting, canceling player action

ymkc: also, but not 100% sure anymore, i think it sometimes also happend that mobs kept on attacking or reaggroed very quickly

teschnei: hmm.. let's remove it for now, and then put up a separate issue for it when you've confirmed it

PR #4173, opened 2017-10-29
Fix for wipeTime in instances

ymkc:
Not sure this is a linux only issue, haven't see reports on this


teschnei:
it might be a 32-bit vs 64-bit issue, can you describe what the issue actually was? and whether you're on 64 or 32? the default for linux is 64 IIRC

teschnei:
```
    lua_pushinteger(L, (lua_Integer)std::chrono::duration_cast<std::chrono::milliseconds>(get_server_start_time() - m_PLuaInstance->GetWipeTime()).count());
```
nevermind, I think that getWipeTime is just pushing the wrong number (the subtraction is backwards)

PR #4174, opened 2017-10-30
Cmake files for DSP

ShelbyZ:
- unified build for unix/windows
- solutions generated into build/{target}
- compiler/linker flags **should** be fairly close
- has step to generate GIT_VER
- fixes weird issue with const static defined in navmesh.h (only unix emitted link error)


CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
+cmake_minimum_required(VERSION 3.7)
+project(darkstar)
+
+set(CMAKE_VERBOSE_MAKEFILE ON)
+
+include(cmake/compiler.cmake)
+
+if(UNIX)
+    set(lib_path /usr/lib /usr/local/lib /opt/local/lib)
```
Cloudef: This isn't really good. Should use pkg-config, hardcoding paths like this is asking for trouble. The cmake Find* modules aren't very good either.

cmake/compiler.cmake
```diff
@@ -0,0 +1,41 @@
+set_property(GLOBAL PROPERTY CXX_STANDARD 17)
+set_property(GLOBAL PROPERTY CXX_STANDARD_REQUIRED ON)
+set_property(GLOBAL PROPERTY CXX_EXTENSIONS ON)
+set_property(GLOBAL PROPERTY LINKER_LANGUAGE CXX)
+
+add_definitions(-DFMT_HEADER_ONLY)
+
+if(UNIX)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -Wall -Werror -WFatal-errors -fsigned-char")
```
Cloudef: Consider add_compiler_options instead, -std=c++17 is redundant due to CXX_STANDARD above

Cloudef: Also -WFatal-errors should be -Wfatal-errors

src/login/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+    ../common/sql.cpp
+    ../common/taskmgr.cpp
+    ../common/timer.cpp
+    ../common/utils.cpp
+    ../common/zlib.cpp
+)
+
+if(UNIX)
+    target_include_directories(DSConnect-server PRIVATE
+        /usr/include/mysql
```
Cloudef: Can't assume this, see the hint about pkg-config earlier

CMakeLists.txt
Cloudef: https://cmake.org/cmake/help/v3.0/module/FindPkgConfig.html

src/map/CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+    attack.cpp
+    attackround.cpp
+    zone_entities.cpp
+    zone_instance.cpp
+)
+
+if(UNIX)
+    target_include_directories(DSGame-server PRIVATE
+        /usr/include/mysql
+        /usr/include/luajit-2.0
```
Cloudef: Here too

src/search/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+    ../common/showmsg.cpp
+    ../common/sql.cpp
+    ../common/taskmgr.cpp
+    ../common/timer.cpp
+    ../common/utils.cpp
+)
+
+if(UNIX)
+    target_include_directories(DSConnect-server PRIVATE
+        /usr/include/mysql
```
Cloudef: And here

cmake/compiler.cmake
ShelbyZ: I need to fix up some of the higher up variables to get the std fixed.

CMakeLists.txt
ShelbyZ: I had not seen this as a way for checking for libs/include directories.  I like it.

Cloudef:
:+1: on the unix dependency changes

tagban:
Can we make sure to still support systems not requiring cmake?  Just wanted to confirm, been in other projects that use cmake and its driven me nuts.

TeoTwawki:
GIT_VER got nuked from project

CMakeSettings.json
```diff
@@ -0,0 +1,55 @@
+{
+  // See https://go.microsoft.com//fwlink//?linkid=834763 for more information about this file.
```
TeoTwawki: https://stackoverflow.com/questions/244777/can-comments-be-used-in-json#244858
tl;dr //comments are a violation of spec, suggested solutions are a "_comments" field or parsing through JSMin before use

TeoTwawki:
@tagban 
> been in other projects that use cmake and its driven me nuts.

possibly because ppl were dumb in its use? I eman, I've seen a lot of crappy cmakes too, but I have also seen a few nice ones.

Cloudef:
tbh, cmake's inferior to platform specific build files. But I guess it's ok if maintaining them is too much of hassle. Cmake does lots of very stupid things can't even create freestanding build files.

tagban:
As long as we don't switch to cmake from having the code easily buildable on windows I'm fine with it. Cmake being required would end my ability to utilize DSP.

takhlaq:
cmake is fine on windows as long as you use commandline cmake and not vs2017 built in cmake support

ShelbyZ:
Closing/Putting a hold on this for now.  Need to re-evaluate the expected value gain out of moving from solution/makefiles to a unified build.

PR #4185, opened 2017-11-02
fix gcc warnings in search project

ShelbyZ:
- align types in blowfish used as iterator
- remove unused params (msg, val) from md52
- order parameter initialization for CTask
- remove unused leftover from EncodeStringSignature(...)

### DIFF
--- a/src/search/data_loader.cpp
+++ b/src/search/data_loader.cpp
@@ -253,7 +253,7 @@ std::list<SearchEntity*> CDataLoader::GetPlayersList(search_req sr, int* count)
                 continue;
 
             // filter by nation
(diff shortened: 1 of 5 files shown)


src/search/data_loader.cpp
```diff
@@ -253,7 +253,7 @@ std::list<SearchEntity*> CDataLoader::GetPlayersList(search_req sr, int* count)
                 continue;
 
             // filter by nation
-            if (sr.nation != 255 && !sr.nation == PPlayer->nation)
+            if (sr.nation != 255 && sr.nation != PPlayer->nation)
```
takhlaq: past me says hello

PR #4191, opened 2017-11-03
Fixing BCNM 100% Group Drop Rate

EulerPath:
Cleaning up code in map/utils/battlefieldutils.cpp

The bug in the code looks to be that each group's total roll will always be the sum of all of the rolls in each group.  This logic does not make sense because each group should always have a roll between 0 and 1000.  If a group only has a 5% drop rate the group should only drop an item 5% of the time, not each time.  I tested this code on the BCNM Up in Arms.  Before the code change Walkure Mask would always drop, now It does not drop all the time.

### DIFF
--- a/src/map/utils/battlefieldutils.cpp
+++ b/src/map/utils/battlefieldutils.cpp
@@ -292,41 +292,6 @@ namespace battlefieldutils {
         }
     }
 
-
-    uint8 getMaxLootGroups(CBattlefield* battlefield) {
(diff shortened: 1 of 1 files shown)


EulerPath:
I found this bug when I was trying to code creeping doom.  I noticed multiple hacks in the sql file due to this bug.  For example, I noticed in Up in Arms the Kraken Club was moved to a different group that had multiple other items dropping so that the club would not drop so much.

I am planning on going over the sql file and take away these hacks.

PR #4193, opened 2017-11-04
NM pop logic for ZM5 and Bastok 8-1

wrenffxi:
Headstone Pilgrimage stones that pop NMs now use local variables
Bastok 8-1 Antican NMs use same system.
IDs moved into zone ID files.
Carthi and Tipha immune to sleep http://ffxiclopedia.wikia.com/wiki/Headstone_Pilgrimage#Fire_Fragment
Axesarion immune to stun http://ffxiclopedia.wikia.com/wiki/Headstone_Pilgrimage#Wind_Fragment
Clicking Headstone/QM6 during battles will now display proper message.


TeoTwawki:
Do we need to nil the MobIDs.lua ?

wrenffxi:
Hm, my current understanding is that we nil files not in the globals to account for changing values.  Maybe I misunderstand what the package.loaded["foo"] = nil does?

TeoTwawki:
it wipes them out but they should be unchanging after being loaded at startup in that file (unless I seriously missed a major out of the ordinary change you did in this pr). We avoid nil'ing anything we don't have to because it'll reload on every use then. If theres something I missed that needs it slap me an I'll reread in the morning. :)

At one point in an issue conversation I thought our mobIDs might have a similar situation to our text IDs, I was incorrect on that though.

wrenffxi:
(from discord chat with ya): I nil-ed the mobIDs here because those files share CERMET_HEADSTONE constant

PR #4198, opened 2017-11-05
[Abyssea-Altepa] Popped NMs

wrenffxi:
QM and NM ids are now tabled in MobIDs.lua
Added missing QM scripts
Set popped NM groups to scripted SPAWNTYPE
Other Abyssea zones can use the common event functions in globals/abyssea.lua


wrenffxi:
**NOTE: put this PR on hold a bit.  Going to redo the way this works.**

PR #4199, opened 2017-11-05
Entire quest progression for Koblakiq including NM fights

Wiggo32:
This implements the 3 quests (Missionary Moblin, For the Birds, and Better the Demon You Know) for the Oldton Movalpolos NPC: Koblakiq.
NM trigger conditions / abilities taken from: http://ffxiclopedia.wikia.com/wiki/Koblakiq
The NM fight with Marquis Andrealphus has him a little overpowered according to 'wiki' testimonies but I don't have retail specs so I left him a little on the 'stronger' side. Easily changed if need be. He does use his 2hr as well as "escape" the party member with hate at 66% and 33%. Unless solo where no "escape" happens.
As always let me know how to make it better and hope this is useful!


Wiggo32:
For this, we don't really have a mob spell or mob skill or mob ability for "casting escape on a player". So, I've decided that one of the next things I was going to do was to create one.  This is most likely going to change from what I originally have so, let's close this until that skill/spell/ability (whatever as I haven't captured from retail yet) is done properly.

PR #4205, opened 2017-11-06
Allow catching "multiples" of fish with Rigs

Wiggo32:
Implements usage of Sabiki Rig, Robber Rig, and Rogue Rig in the catching of multiple fish.
Code is very 'simple' and amateur but works by sending a "caught fish" message multiple times since there doesn't seem to be a way to use plurals with the messageoffset.
List of 'allowed' fish can be modified easily as I only included the fish I was able to confirm using:
http://www.great-blue.jp/?lng=en

### DIFF
--- a/src/map/utils/fishingutils.cpp
+++ b/src/map/utils/fishingutils.cpp
@@ -380,8 +380,52 @@ void FishingAction(CCharEntity* PChar, FISHACTION action, uint16 stamina, uint32
 
                 // TODO: анализируем RodFlag
 
-				charutils::AddItem(PChar, LOC_INVENTORY, PFish->getID(), 1);
+                // Implement possibility of catching up to 3 fish with Rogue Rig, Robber Rig, or Sabiki Rig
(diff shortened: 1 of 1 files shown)


Hozu:
Wouldn't it be a better idea to add a column to the fishing_lure table for the lures that can catch multiple fish and the eligible fish for those lures? That way it can be updated more easily as opposed to being hardcoded in the core.

Wiggo32:
Yes, it would! I don't really know how to do that though.

teschnei:
item ID checks in the core are something I'd like to avoid (that's usually better suited for scripts)

Like Hozu said, could be a db column in fishing_lure.  Also probably a flag in fishing_zone (or item_basic or something) to flag fish that can be caught in multiples instead of checking all their IDs

Wiggo32:
problem with that is, the message is pushed from core so to accomplish what you're saying (which would be a good idea) I think fishing needs to be completely reworked. So, lets close this because I'm not ready to tackle an entire rework at this point and I don't plan to change my code from what I have already done.

PR #4207, opened 2017-11-07
added poison to blumagic resists and correct damage calc of poison br…

dwn134:
based on breath spells


scripts/globals/bluemagic.lua
```diff
@@ -411,6 +411,8 @@ function getBlueEffectDuration(caster,resist,effect)
         duration = math.random(60,120) + resist * 15; -- 60- 120 -- Needs confirmation but capped max duration based on White Magic Spell Slow
     elseif (effect == EFFECT_SILENCE) then
         duration = math.random(60,180) + resist * 15; -- 60- 180 -- Needs confirmation but capped max duration based on White Magic Spell Silence
+    elseif (effect == EFFECT_POISON) then
```
TeoTwawki: Have always disliked this if/elseif stack, but I'll revisit that another day

PR #4214, opened 2017-11-08
Blu magic breath changes

dwn134:


Hozu:
I would have done this ages ago, but I was hesitant because when BLU mobs use those spells, they'll just kill players in one shot most of the time. Clearly the mob versions of the breaths are slightly different to account for their greatly inflated HP.

Also, it doesn't look like you're doing a resist check? It would probably just be the skill and magic accuracy though, no attribute difference check like dINT.

dwn134:
I'm not sure how to do that. I'm a beginner still but maybe a new section in blumagic.lua for breaths. I see it has a section for physical and magical then make a new section for breath. If this gets merged then maybe someone can build off of it but this is the correct damage calculations and is headway to the finished result

TeoTwawki:
Don't forget to make 1st instance of a variable "local". A lot of the corrections you are doing have a damage variable globally because lua does global by default and the previous editor(s) made the same mistake so you need the first instance of it to be `local damage =`

TeoTwawki:
> Clearly the mob versions of the breaths are slightly different to account for their greatly inflated HP.

Yeah I spoze that'll be a prob for blue magic casting mamools. mobskill version safe at least, though a few NMs already have a similar prob with skill script since NMs have hugely more HP than normal mobs (a certain eruca comes to mind..)

TeoTwawki:
Occurs to me that instead of doing the same math in each script, the same resist checks in each script, and the same checks for breath damage taken mod sin each script, we should just be calling a global for doing breath damage in all these scripts.

dwn134:
But each spell has a different formula. No 2 are the same. I can see making a global for resists and gear check. There are 3 pieces of gear that enhance the damage +10% magus head/+1 And saurian helm that could be added to the global

TeoTwawki:
formula can be handled as function input values

dwn134:
then you can plug in values om M and N in the spell script

TeoTwawki:
I'm really sorry we didn't get back to you for such a long time. Wasn't an intent to reject here, we just didn't get to it.

dwn134:
Np. I'm not sure how I was going to do a resist function. In all honesty I was hoping to get the ball rolling on what needs to be changed and so thought to it

PR #4217, opened 2017-11-09
Added Koga Shuriken Pouch as a usable item

pandorraffxi:
Added:
    scripts/globals/items/koga_shuriken_pouch.lua
    Koga Sh. Pouch in  sql/item_usable.sql

Signed-off-by: pandorraffxi <rahulabon@hotmail.com>

### DIFF
--- a/scripts/globals/items/koga_shuriken_pouch.lua
+++ b/scripts/globals/items/koga_shuriken_pouch.lua
@@ -1,0 +1,25 @@
+-----------------------------------------
+-- ID: 6300
+-- Item: Koga Shr. Pouch
+-- When used, you will obtain one stack of Koga Shurikens
+-----------------------------------------
+
+-----------------------------------------
+-- OnItemCheck
+-----------------------------------------
+
+function onItemCheck(target)
+    local result = 0;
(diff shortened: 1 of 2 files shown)


scripts/globals/items/koga_shuriken_pouch.lua
```diff
@@ -0,0 +1,25 @@
...
+-----------------------------------------
+-- OnItemCheck
+-----------------------------------------
+
+function onItemCheck(target)
+    local result = 0;
+    if (target:getFreeSlotsCount() == 0) then
+        result = 308;
+    end
+return result;
```
TeoTwawki: missing indent here

pandorraffxi: @TeoTwawki - used a shell of another item that didn't have the indent. Will be more mindful in the future as I work more :)

TeoTwawki: Been fixing old files indention and other messes as I locate/see them so that people don't copy :) happens a lot so don't feel bad about it

takhlaq:
thanks for the pr!

PR #4218, opened 2017-11-10
Update to Simurgh.lua to allow NM to Spawn

pandorraffxi:
Updated /scripts/zones/Rolanberry_Fields/mobs/Simurgh.lua

Changed line 38 from:
UpdateNMSpawnPoint(mob:getID());
To:
UpdateNMSpawnPoint(Simurgh));

Signed-off-by: pandorraffxi <rahulabon@hotmail.com>

### DIFF
--- a/scripts/zones/Rolanberry_Fields/mobs/Simurgh.lua
+++ b/scripts/zones/Rolanberry_Fields/mobs/Simurgh.lua
@@ -35,7 +35,7 @@ end;
 function onMobDespawn(mob)
 
     -- Set Simurgh's spawnpoint and respawn time (21-24 hours)
-    UpdateNMSpawnPoint(mob:getID());
+    UpdateNMSpawnPoint(Simurgh);
     mob:setRespawnTime(math.random(75600,86400));
 
 end;


takhlaq:
this doesnt do anything besides error out preventing setting the 21-24h respawn time

Wiggo32:
You could make him 'rage' too! =)

pandorraffxi:
Ahh I see it now, should be added to MobIDs.lua instead of editing the simurgh.lua...sorry!

TeoTwawki:
no point in being in mobids even - its his own id doesnt need to pass through there

TeoTwawki:
also not seeing any spawn problem tested by setting respawn time very low.

wrenffxi:
Quick explanation on this one: Pandorra's on my server, and we have some custom code that preserves certain NM pop cooldowns through server resets.  It's not retail accurate, so I assume ya'll don't want it.  If you do, let me know.  Anyway, we broke, and later fixed, our code recently.  Pandorra didn't realize it was our issue rather than DSP's.  AFAIK your Simurgh works as expected.

TeoTwawki:
Ah ok. I did something similar for another servers ground kings.

PR #4220, opened 2017-11-11
more msgBasic cleanup

wrenffxi:
This should be all the remaining hard-coded numeric message ids.  Except for a couple in bcnm.lua, which I'm not touching at the moment.


TeoTwawki:
❤️

PR #4222, opened 2017-11-11
All SetRespawnTime() calls changed to :setRespawnTime() binding calls

wrenffxi:
Can now remove the redundant SetRespawnTime() function in zone.lua


teschnei:
If SetRespawnTime global binding is totally unused now, can you also remove
it?

I'd like to remove all of the global calls that just work on an ID with
Getmob this (if needed) and an entity function binding eventually

scripts/globals/zone.lua
```diff
@@ -89,19 +89,3 @@ function SetExplorerMoogles(moogle)
         end
     end
 end;
-
------------------------------------
--- SetRespawnTime
-----------------------------------
-
-function SetRespawnTime(id, minTime, maxTime)
-    -- This function is redundant should place the mob:setRespawnTime() and UpdateNMSpawnPoint back in the individual zones.
```
TeoTwawki: ❤️

wrenffxi:
The removal of global binding is included in the PR.  Also, resolved the merge conflicts.

teschnei:
Oh, it's just a global function, not a binding (was looking for a cpp
change). That's all good then

PR #4230, opened 2017-11-14
Version update November2017

dragonarmor:
I am new to this so don't flame me if I did the pull wrong. Thanks
Npc shifts
Lua script updates
TextID's
Version Info


sql/npc_list.sql
```diff
@@ -1346,49 +1346,49 @@ INSERT INTO `npc_list` VALUES (16859197,'2','     ',0,-80.000,-1.000,-40.000,7,4
...
-INSERT INTO `npc_list` VALUES (16859470,'qm1','???',0,280.001,-2.328,280.000,32774,40,40,0,0,0,0,3,0x0000340000000000000000000000000000000000,0,'COP',0);
+INSERT INTO `npc_list` VALUES (16859465,'_0kc','kiokudama_a',0,300.000,-2.500,-220.000,1,40,40,9,0,0,0,3,0x0200000000000000000000000000000000000000,0,'COP',0);
+INSERT INTO `npc_list` VALUES (16859466,'_0kd','kiokudama_b',0,-260.000,-2.500,-180.000,1,40,40,9,0,0,0,3,0x0200000000000000000000000000000000000000,0,'COP',0);
+INSERT INTO `npc_list` VALUES (16859467,'kiokudama_a','kiokudama_c',0,299.920,-0.044,179.982,32769,40,40,0,0,0,0,2051,0x0000340000000000000000000000000000000000,0,'COP',1);
+INSERT INTO `npc_list` VALUES (16859468,'kiokudama_b','???',0,340.261,-0.053,-220.029,32769,40,40,0,0,0,0,2051,0x0000340000000000000000000000000000000000,0,'COP',1);
+INSERT INTO `npc_list` VALUES (16859469,'kiokudama_c','',0,-299.906,-0.045,-180.055,32769,40,40,0,0,0,0,2051,0x0000340000000000000000000000000000000000,0,'COP',1);
+INSERT INTO `npc_list` VALUES (16859470,'qm1','Moogle',0,280.001,-2.328,280.000,32774,40,40,0,0,0,0,3,0x0000340000000000000000000000000000000000,0,'COP',0);
```
TeoTwawki: this is not sync'd up properly

some manual review is required even with the auto tools

sql/npc_list.sql
```diff
@@ -2356,10 +2356,10 @@ INSERT INTO `npc_list` VALUES (16904392,'Goblin_Footprint','Goblin Footprint',21
 -- Sealion's Den (Zone 32)
 -- ------------------------------------------------------------
 
-INSERT INTO `npc_list` VALUES (16908370,'Tenzen','Tenzen',0,0.000,0.000,0.000,0,40,40,0,0,0,6,27,0x0000A20500000000000000000000000000000000,32,'COP',1);
-INSERT INTO `npc_list` VALUES (16908371,'Makki-Chebukki','Makki-Chebukki',0,0.000,0.000,0.000,0,40,40,0,0,0,6,27,0x0000A50500000000000000000000000000000000,32,'COP',1);
-INSERT INTO `npc_list` VALUES (16908372,'Kukki-Chebukki','Kukki-Chebukki',0,0.000,0.000,0.000,0,40,40,0,0,0,6,27,0x0000A40500000000000000000000000000000000,32,'COP',1);
-INSERT INTO `npc_list` VALUES (16908373,'Cherukiki','Cherukiki',0,0.000,0.000,0.000,0,40,40,0,0,0,6,27,0x0000A30500000000000000000000000000000000,32,'COP',1);
+INSERT INTO `npc_list` VALUES (16908358,'Tenzen','Tenzen',0,0.000,0.000,0.000,0,40,40,0,0,0,6,27,0x0000A20500000000000000000000000000000000,32,'COP',1);
```
ibm2431: Make sure when you run the tool that it's not replacing IDs with "earlier copies" of the mob with the same name in the zone. There are both "battle" and cutscene clones for these characters, and if you don't watch the tools like a hawk they'll try to replace what's supposed to be the cutscene version's ID with the ID for the battle version because the battle version is listed first in POLUtils.

You can check the current NPC list with POLUtils to see if the "old" IDs (16908370-16908373) would still be valid for those characters (names and order match up). As a general rule in such cases, unless you have reason to believe SE shifted the IDs in that zone (removing 12 NPCs in this case), the ID should remain the same despite the tool trying to tell you they "shifted".

Same thing goes for the armory crates in Remnants zones. The crate captured isn't the first in the POLUtil list, and the tool just loves to bounce it around.

dragonarmor: Ah yes thanks for that information and you are indeed correct. I will cross check polutils and correct these.

TeoTwawki:
Thanks, all looks good now so merged. if I missed anything we'll fix from here. :)

PR #4232, opened 2017-11-15
Lottery NM overhaul

wrenffxi:
moved lottery spawn mechanic moved to globals/mobs instead of repeated in every placeholder lua
set lottery NMs to spawntype lottery in mob_groups
added Nocuous weapon NM
removed silica from jugner forest land pugils drop pool


scripts/zones/Inner_Horutoto_Ruins/mobs/Nocuous_Weapon.lua
```diff
@@ -0,0 +1,42 @@
...
+        return SUBEFFECT_POISON, msgBasic.ADD_EFFECT_STATUS, EFFECT_POISON;
+    end
+end;
+
+function onMobDeath(mob, player, isKiller)
+end;
+
+function onMobDespawn(mob)
+    SetServerVariable("[POP]Nocuous_Weapon", os.time() + 3600); -- 1 hour
+    DisallowRespawn(mob:getID(), true);
```
TeoTwawki: Sorry we didn't send you the memo: we switched to using a listener in the placeholders now, but haven't updated all the old scripts yet https://github.com/DarkstarProject/darkstar/commit/5c06b263b421310989184d497f8773a75c6282f8#diff-e1863da932c3fbbf882f221b398ae513R42

wrenffxi: Understood.  Hold this PR open and I'll make the switch tomorrow.

teschnei:
one thing to consider: there's not really any reason to keep each NM in its own table.  I didn't do it for Leaping Lizzy, but you could combine them all into one table.  You'd have to update the mobs.lua functions a bit to accommodate this, though.  Overall.. bit cleaner.  But you have to make special care that you don't call them the same name though.  Or else you get the textIDs problem

scripts/globals/mobs.lua
```diff
@@ -37,4 +37,51 @@ function onMobDeathEx(mob, player, isKiller, isWeaponSkillKill)
...
+        end
+    end
+    return false;
+end
+
+-- potential lottery placeholder was killed
+function phOnDespawn(ph,phList,chance,cooldown)
+    local phId = ph:getID();
+    for k,v in pairs(phList) do
+        if (k == phId) then
```
teschnei: if you're looking for a specific key, you should just do phList[phId] instead of looping

wrenffxi: fixed

PR #4236, opened 2017-11-16
Fixed Index in /Globals/BCNM for Promy Dem and Mea

pandorraffxi:
The current DSP format was not allowing players to receive the option to pick the battlefield due to the formatting of the Index of 19 and 21.

Most likely there are other issues within that index array and I will hash those out when I have time available to see if they do give an issue.


takhlaq:
there was one other somewhere in that table, a couple zones before/after

pandorraffxi:
Checked the rest of them and doesn't look like there would have been anymore with issues.

PR #4249, opened 2017-11-18
Fix clamp problems in UpdateHealth

ShelbyZ:
- std::clamp protects against lo > hi, but we need to adjust what we clamp

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -154,10 +154,6 @@ void CBattleEntity::UpdateHealth()
     {
         health.hp = std::clamp(health.hp, 0, health.modhp);
     }
-    if (health.hp == 0)
-    {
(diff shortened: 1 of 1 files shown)


src/map/entities/battleentity.cpp
```diff
@@ -149,8 +146,27 @@ void CBattleEntity::UpdateHealth()
...
-    health.mp = std::clamp(health.mp, 0, health.modmp);
+    if (0 > health.modhp)
+    {
+        health.hp = std::clamp(health.hp, health.modhp, 0);
+    }
+    else
+    {
+        health.hp = std::clamp(health.hp, 0, health.modhp);
+    }
+    if (health.hp == 0)
```
ShelbyZ: This indeed is a bug, makes you immortal which is pretty cool

PR #4253, opened 2017-11-18
NPCID off by one

ymkc:


TeoTwawki:
About to merge the client update pr, I'll recheck this after

TeoTwawki:
was correct after update, so merged

PR #4256, opened 2017-11-19
attohwa antlion, miasma, gasponia fixes

wrenffxi:
Removed :hideModel() from antlion mixins and pit ambush.  It doesn't seem to do anything.

For some reason, mob:hideName() in the SPAWN event trigger in mixin wasn't hiding the name unless called with a timer.  It's ugly, but it works.  Maybe one of you can find a more elegant solution.

Added script for Trench Antlion, and made Ambusher Antlion a proper lottery NM off of it.

Removed an unnecessary loop in the Gasponia code.

Removed and commented the code in zone's onGameHour.  The hard-coded NPC id in there was several version updates wrong, and generally the miasmas need retail testing and recoding.

Fixed some comments and requires in various mob files.


teschnei:
"Removed :hideModel() from antlion mixins and pit ambush. It doesn't seem to do anything."

if you walk into spawn range (40 or 50 yalms) of the antlion when it's burrowed, is it still invisible? we have an issue with worms sometimes and I wasn't sure if it was because animationsub doesn't animate when you're out of range or something else

wrenffxi:
crashing for the night. will test this out tomorrow and let ya know.

wrenffxi:
Update: If I walk 55 yalms away from the burrowed antlion, then walk back toward it, it remains invisible the entire time.

TeoTwawki:
I don't think it was one of those things that could happen every time (the worms) but its easy enough to modify it back in if we need to,

Still bothers me hidename on spawn is refusing to work without that `mob:timer()`

teschnei:
Oh shoot, I didn't notice that.  That's totally on the core side to fix -
can I get that fixed before we merge this?

teschnei:
Okay, if you merge master back into your branch again, you should be able to remove those timers

wrenffxi:
Much better. Thanks! Updated PR.

Issue #4257, opened 2017-11-19
promy strays/receptacles recently 'fixed' incorrectly


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30170905_5

**_Server Version_** (type `!revision` in game) **:** unknown

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Recent PR: https://github.com/DarkstarProject/darkstar/pull/4194/files 'fixes' strays so that they respawn every 30 sec.

Just went to promy dem on retail killed all the strays and then started timing after i killed the last stray
stray 1 popped in about 6 min
stray 2 popped 6 min later
stray 3 popped 2 min later
killed them again and then engaged the receptacle
stray 1 popped in about 20 sec
stray 2 popped 18 sec later
stray 3 popped 18 sec later
went to 2nd floor - 5 strays - killed them all and started timing
stray 1 - about 6 min
stray 2 - 3 min later
stray 3 - 2 min later
stray 4 - 1 min later
stray 5 - 1 min later
killed them again and engaged
stray 1 - about 20 sec
stray 2 - 18 sec
stray 3 - 16 sec
stray 4 - 17 sec
stray 5 - 17 sec
that's about all i had patience for. theres a huge difference in stray pop time whether or not the receptacle is engaged. dont ask me what the exact formula is... i just know its not a flat 30 sec.

also, with the addition of this PR there is a knockback effect with every empty seed whether it connects or not. in retail, there is no knockback with empty seed when the receptacle misses or is absorbed by shadows. i did have knockback with even the slightest bit of damage though. not sure why this was removed with this PR because it used to work correctly.

TeoTwawki:
let us know if still happens after both client and server are on version 30171103_1

TeoTwawki:
misread disregard

TeoTwawki:
Doesn't look like empty seed was touched in that PR.

Wiggo32:
well... empty seed mysteriously started adding knockback upon missing sometime after this PR. included issue... it shouldnt

Delkathus:
[...]

Issue #4201, opened 2017-11-06
Promyvion mobs "strays" are dropping gil


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30170905_5

**_Server Version_** (type `!revision` in game) **:** unknown

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Promy strays drop gil by default and without enabling the "all mobs drop gil" setting. Having a hard time figuring out where to change this...

wrenffxi:
Believe this was fixed in ab5dedd9d99dc850a105c7c218f108e808e0e3f9.

Issue #433, opened 2014-08-23
Memory Receptacles spawning too many Strays


Fiocitrine:
The Memory Receptacles in Promyvion-Dem are spawning two Strays at a time. I believe in retail it only did one at a time. Plus, on the first floor there can be as many as 6 Strays out which there are only suppose to be 3-4. We were getting raped by these fuckers. Might want to take a look at the other Prommies as well. 

Cheers,
Fio

PR #4263, opened 2017-11-20
Hopefully the last MR/Stray related PR of the year.

wrenffxi:
Strays no longer drop gil
Strays spawn every 5.5 minutes if MR is not engaged
Strays spawn in 20 second intervals if MR is engaged
Strays spawn at MR location and move outward

Addresses #433, #4201, and #4257 .

edit: and before anyone says "why does the MR spawn a stray immediately when engaged, instead of waiting 20 seconds", I considered this, and decided that resetting the 20 second timer onMobEngaged is problematic because on DSP mobs deaggro quickly when they're not smacking you.  MRs have no autoattack, so if you aren't meleeing them, they go white quickly, which would then reset the timer when someone nukes it, and a stray would never spawn during the fight.

### DIFF
--- a/scripts/zones/Promyvion-Dem/mobs/Memory_Receptacle.lua
+++ b/scripts/zones/Promyvion-Dem/mobs/Memory_Receptacle.lua
@@ -2,10 +2,8 @@
 -- Area: Promyvion-Dem
 --  MOB: Memory Receptacle
 -----------------------------------
-package.loaded["scripts/zones/Promyvion-Dem/TextIDs"] = nil;
 package.loaded["scripts/zones/Promyvion-Dem/MobIDs"] = nil;
 -----------------------------------
-require("scripts/zones/Promyvion-Dem/TextIDs");
 require("scripts/zones/Promyvion-Dem/MobIDs");
 require("scripts/globals/status");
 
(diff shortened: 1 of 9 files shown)


takhlaq:
summoning @Wiggo32 for review (i dont care to watch the vid)

Wiggo32:
Yea, that's better than what it was for 'retail' accuracy.

TeoTwawki:
Merging, since 'approval'

Wiggo32:
If you put something like this along with the other one it will behave more like retail. Then you can keep the strays scripted (128) so they aren't able to pop all at the same time. Retail doesn't have strays popping at the same time even if they are all mass killed at the same time. Also, with this I'm not sure you need that Stray.lua except for the gil mod.

```
function checkStrayIdle(mob)
    local mobId = mob:getID();
    local numStrays = MEMORY_RECEPTACLES[mobId][2];
    
    if (os.time() > mob:getLocalVar("nextStray")) then
        for i = mobId + 1, mobId + numStrays do
            local stray = GetMobByID(i);
            if (not stray:isSpawned()) then
                mob:setLocalVar("nextStray", os.time() + 330);
                SpawnMob(stray:getID());
                break;
            end
        end
    else
        mob:AnimationSub(2);
    end
end;

function onMobRoam(mob)
    checkStrayIdle(mob);
end;
```

Wiggo32:
no, don't think you can do that exactly, the local vars would conflict... but something along the line?

Issue #4243, opened 2017-11-18
Rottata wrong dialogue


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30170905_5

**_Server Version_** (type `!revision` in game) **:** unknown

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Rottata tells me that regions which are currently under my home country's control are NOT currently under my home country's control when asking for teleports to said regions. I don't know if this is a global problem with all outpost teleporters but Rottata for sure.

TeoTwawki:
let us know if still happens after both client and server are on version 30171103_1 (just updated master branch)

wrenffxi:
It's still a problem.  The prices are calculated correctly, so it's only a cosmetic issue.  The problem lies here

https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Port_Windurst/npcs/Rottata.lua#L46

The fifth argument is, correctly, the player's CP.  The third argument however, to which we're passing CP, is instead supposed to control Rottata's next spoken dialog.  This is confirmed in the dialog extract:
```
That region is ≺Multiple Choice (Parameter 2)≻[no longer under your home country's control./currently under your home country's control.] You will require ≺Numeric Parameter 1≻ gil or ≺Numeric Parameter 3≻ conquest ≺Singular/Plural Choice (Parameter 3)≻[point/points] to teleport.
```

Problem is, I cannot figure out what that third parameter needs to be set to in order to change Rottata's dialogue.  I tried all the obvious things.  Numbers around 0.  Powers-of-two.  Powers-of-two minus one.  The 1073741823-ish menu offset plus and minus obvious variables.

So we know why it's happening, just not yet how to fix it.

TeoTwawki:
thanks for the update

Wiggo32:
Parameter 3 controls the control- starting with 
32 for Ronfaure
64 -Zulkheim
128 - Norvallen
256- Gustaberg
512- Derfland
1024- Saratabaruta
2048- Kolshushu
4096- Aragoneau
8192- Faurengandi
16384- Valduana
32768- Qufim
65536- Lil Teilor
131072- Kuzotz
[...]

PR #4264, opened 2017-11-21
OP Warp NPC dialog fix

wrenffxi:
OP warp NPC dialog now reflects whether chosen region is under nation control.

Addresses #4243


takhlaq:
@wrenffxi here's how it looks using bit operations
https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/conquest.lua#L537-L619

PR #4271, opened 2017-11-23
Less cbasetypes

ShelbyZ:
- removed cbasetypes defines for types and min/max
- removed dead/unused defines/macros
- fixed related casting issues for int8_t to char
- general slimming of cbasetypes


Cloudef:
Yay

Cloudef:
There was also issue with the codebase relying on -fsigned-char, should pinpoint why's that sometime.

ShelbyZ:
Probably need to drop in -Werror and remove -fsigned-char to see what breaks.

Cloudef:
You may need to use `-funsigned-char` actually, unless you are on ARM.

teschnei:
why not `using uint32 = std::uint32_t` instead of changing 300 files?

takhlaq:
it's worth keeping in mind we have a fair amount of code which doesn't use fixed width integers, though i didn't bother checking if this pr those too

ShelbyZ:
I made the ~~mistake~~ choice to do ctrl+h to replace all.  It could be walked back.  It's fine either way, just let me know.

teschnei:
I'd prefer it with a using, both because it's way less merge conflicts and
it's easier to just type what we've always been typing (which is also
shorter than the std types)

Cloudef:
I think avoiding lots of file changes is fair reason.

Cloudef:
@takhlaq Would probably eliminate code that doesn't use stdint types in another PR. Also, it's not neccessary to mass replace those unless they have portability problems. Mainly for ARM there's only issue with unsigned char, so somewhere we should use int8_t instead of char (code relies on the fact it's signed).

takhlaq:
gcc 32 bit and 64 bit use different int widths

teschnei:
I'll merge this once conflicts are resolved (sorry for taking so long getting to it)

PR #4273, opened 2017-11-24
Chocobo digging fixes for crystals/elemental ores and zone-in delay

xipies:
Fixes digging crystals/elemental ores from giving fire crystal/fire ore, if additional required conditions are not met.

Fixes the comparison logic for the zone-in delay.

Implements setting the zone-in time so the zone-in delay check will be effective. Adds an event in player.lua that fires after zone-in [esp. after CZone::CharZoneIn(CCharEntity* PChar), which clears local vars].

I did look at other lua bindings, but did not see a suitable one. OnGameIn, which fires for player.lua, fires before this, so any local vars set there get cleared. AfterZoneIn only fires for Zone.lua files.

Renames the ZoneInTime var (removes "[DIG]" prefix) since it could be useful for other scripts besides digging.


teschnei:
There's no reason to clear local vars at all for chars on zoning, would it help you if I removed that?

(I don't want to add another binding when it's not needed)

TeoTwawki:
after it clears only at zone out and not at zone in, that should be able to work at line 28

xipies:
If it only clears local vars on zone out, and not zone in, I think that would work fine. Then it would not need the new binding.

teschnei:
Just pushed it, can you try that?

xipies:
Yeah, I will work on the change. Thanks for the update.

xipies:
Worked like a charm. Pushed the changes. I think I did the revert for the binding OK? It is only showing two files changed now in this PR.

TeoTwawki:
if the intent is to set both at login and at zonein, it looks alright now.

xipies:
Oops, let me check/change. Probably should just be at zonein.

xipies:
I tested with only being set during zone in. In that test, it did not set the time (for delay comparison), if I logged in, say, already on chocobo. I could dig immediately.

I think I would keep on login and zonein, in that case, unless someone has retail evidence to implement otherwise (e.g., without delay on login). I think it should be good to merge now.

TeoTwawki:
Yeah i wasn't sayin it was wrong, I was saying it was right if thats what you intended

xipies:
Sure, np. I just want to do an actual test to see what it did, since I had not even thought to test it that way before. :)

PR #4285, opened 2017-11-25
Adjust Bastok Mission 6-2 fight

ymkc:
adjust Salamander NM level, make sneak pop possible and always drop 6 frag rocks


scripts/zones/Ifrits_Cauldron/mobs/Magma.lua
```diff
@@ -3,24 +3,17 @@
 --  MOB: Magma
 -----------------------------------
 
------------------------------------
--- onMobSpawn Action
------------------------------------
-
 function onMobSpawn(mob)
+    DespawnMob(mob:getID(), 180);
```
TeoTwawki: whats this intended for?

scripts/zones/Ifrits_Cauldron/npcs/qm4.lua
```diff
@@ -10,44 +10,27 @@ package.loaded["scripts/zones/Ifrits_Cauldron/TextIDs"] = nil;
...
 
------------------------------------
--- onTrade Action
------------------------------------
-
 function onTrade(player,npc,trade)
-
     if (player:getCurrentMission(BASTOK) == THE_PIRATE_S_COVE and player:getVar("MissionStatus") == 2) then
-        if (GetMobAction(17616897) == 0 and GetMobAction(17616898) == 0 and trade:hasItemQty(646,1) and trade:getItemCount() == 1) then
+        if (GetMobByID(17616897):isDead() and GetMobByID(17616898):isDead() and trade:hasItemQty(646,1) and trade:getItemCount() == 1) then
```
TeoTwawki: might want to check isSpawned instead using the not operator, its clearer

scripts/zones/Ifrits_Cauldron/mobs/Magma.lua
TeoTwawki: sidenote: kinda bugs me that this one has the onMobInit after instead of before like the others, even though I know it doesn't effect anything really.

ymkc: c&p mistake :)

TeoTwawki: so, this DespawnMob still exists in the most resent version
was that intended and you just meant the function order was copy paste error?

ymkc: it is itended for salamander, as he doesnt need to be defeated. makes no sense on magma as he spawns claimed by the trading player

TeoTwawki: ah ok,

Issue #4082, opened 2017-09-18
Synthing Flint Stones Buggy


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**30170728_2

**_Server Version_** (type `!revision` in game) **:**unknown

**_Source Branch_** (master/stable) **:**master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

Alchemy lvl 50 status 'Journeyman' 
While synthing 8 flint stones from a stack of 12 to make glass fiber which caps at 61,
if the synth breaks, you will get notification that you have lost flint stones, however the stack will remain greyed out and indicate all 12 remaining. upon zoning, you have full use of the 12 flint stones to attempt synth again.

If the stones are broken up into individual stones (no stack) and placed each in its own slot for the synth, upon synth failure message and notification of losing items the stones that were reported lost are actually lost.

TeoTwawki:
#463
#2373

metalfiiish:
This can be closed. Fixed in patch #4286.

Issue #2373, opened 2015-11-05
"Darkened" synth materials cannot be used (see screenshot)


Fiocitrine:
Kind of hard to explain: Trying to synth hemp gorget with grass thread. I can only use the stack of three and not the full stacks. 

Here: http://prntscr.com/8zct77

Hozu:
Did this happen on a success or a break? I know that if you break a Blessed Briault synth, this happens with Galateia every time.

teschnei:
Probably good idea to let the developers know if you found a way to
reproduce it every time

Hozu:
I forgot this issue existed until now, but yeah, break a Blessed Briault synth to see this bug in action - I'm not the only one on the server I play on who ran into this.

Fiocitrine:
I haven't found a way/tested. But it was most likely due to a break as I was skilling up Clothcraft.

nesstea:
It happens when you break and the stack exists after the synth. I am not certain if it happens when you crystal break and/or lose items break.

TeoTwawki:
#463

Issue #463, opened 2014-09-05
[Clothcraft] Galateia not selectable after a failed synth (See screen)


lwrand:
![screen](https://cloud.githubusercontent.com/assets/8465871/4168384/bbd42ba4-351b-11e4-9581-6778267bf728.png)

If you relog, Galateias are selectable again.

/lwrand,

JoeCT:
I just tried to reproduce this glitch and have been unable to do so.  After failed attempts at synthesizing items with galateia as an ingredient I am still able to select the item from my inventory without logging out.

TeoTwawki:
From comments in #2373 : 

> I forgot this issue existed until now, but yeah, break a Blessed Briault synth to see this bug in action - I'm not the only one on the server I play on who ran into this.

> It happens when you break and the stack exists after the synth. I am not certain if it happens when you crystal break and/or lose items break.

Someone confirm the trigger so we can get this fixed so we don't have year+ old bugs sitting on the tracker.

Deadwing888:
Can confirm this is still a bug. 

Additionally: if this synth is done with 2 galateia stacked and the synth critical breaks to lose one galateia two will still be grayed out and you will have two usable galateia after zoning.

metalfiiish:
This can be closed. Fixed in patch #4286.

PR #4286, opened 2017-11-26
Synthesis Fail Patch

metalfiiish:
Fixes issue where failed synth would cause items to remain locked and would not remove quantity. Players would end up zoning or login/logout to resolve issue.


metalfiiish:
Relevant open bugs:
https://github.com/DarkstarProject/darkstar/issues/463
https://github.com/DarkstarProject/darkstar/issues/4082
https://github.com/DarkstarProject/darkstar/issues/2373

PR #4289, opened 2017-11-26
Adjust Bastok Mission 7-1 fight

ymkc:


sql/mob_pools.sql
```diff
@@ -2774,7 +2774,7 @@ INSERT INTO `mob_pools` VALUES (2713,'Mokka','Mokka',165,0x0000BF060000000000000
 INSERT INTO `mob_pools` VALUES (2714,'Mokke','Mokke',165,0x0000BF0600000000000000000000000000000000,4,4,7,240,100,0,1,1,1,16,0,0,0,3,0,0,2,1,0,165);
 INSERT INTO `mob_pools` VALUES (2715,'Mokku','Mokku',165,0x0000BF0600000000000000000000000000000000,4,4,7,240,100,0,1,1,1,16,0,0,0,3,0,0,2,1,0,165);
 INSERT INTO `mob_pools` VALUES (2716,'Mokkuralfi','Mokkuralfi',112,0x0000080700000000000000000000000000000000,4,4,12,240,100,0,1,1,1,16,0,0,0,3,0,0,2,1,0,112);
-INSERT INTO `mob_pools` VALUES (2717,'Mokkurkalfi','Mokkurkalfi',135,0x0000B00100000000000000000000000000000000,1,1,11,320,100,0,1,0,0,2,0,0,0,3,0,0,0,1,0,135);
+INSERT INTO `mob_pools` VALUES (2717,'Mokkurkalfi','Mokkurkalfi',500,0x0000B00100000000000000000000000000000000,1,1,11,320,100,0,1,0,0,2,0,0,0,3,0,0,0,1,0,135);
```
TeoTwawki: what necessitated the new family ID? both families have identical data looks like

TeoTwawki: oh nvm I see it! man that was hard to spot. 1 lil value for detection.

TeoTwawki: we really should have some kinda override for that instead of having to create a while new family for one mob

PR #4293, opened 2017-11-27
Added missing NMs

wrenffxi:
Huwasi
Lii_Jixa_the_Somnolist
Mucoid_Mass
Patripatan
Saa_Doyi_the_Fervid
Teporingo
Thoon


Hozu:
What about their unique spell list and/or mobskill list that some of these have? ie. uses only Crystal Rain, or Katon: Ichi. Shouldn't those be implemented as well?

wrenffxi:
Yes. This PR deals only with the coordinates/pop conditions, model/family data, and drops.  No special fight behavior -- other than resistances -- are yet coded.

Issue #4128, opened 2017-10-15
Goblintrap working?


Akhawi:
**_I have:Question about Goblintrap NM is wroking whit the dropp it is supose to dropp_**

- [] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version 30170707_0_** (type `/ver` in game) **:** 

**_Server Version 0c16d47cd63dbba1a3db80c7c12df57144769e81_** (type `!revision` in game) **:** 

**_Source Branch Master_** (master/stable) **:** 

**_Additional Information have tried to pop Goblintrap, but it dose not pop, altho it is lottery pop, but this nm might not work. _** (Steps to reproduce/Expected behavior) **:**

metalfiiish:
Currently not implemented. Looks like the spawn points are in the database but the  East Ronfaure (S)/mobs/Battrap.lua is not built to spawn him onMobDeath.

metalfiiish:
Can be closed as this has now been implemented.

PR #4296, opened 2017-11-28
Goblintrap implement

metalfiiish:
Implemented Goblintrap and it's drop.

Corresponding issue#:
https://github.com/DarkstarProject/darkstar/issues/4128


scripts/zones/East_Ronfaure_[S]/mobs/Battrap.lua
```diff
@@ -0,0 +1,42 @@
...
+
+function onMobDeath(mob, player, isKiller)
+
+    local GOBLIN_TRAP_PH = GOBLIN_TRAP - 1;
+
+    if (mob:getID() == GOBLIN_TRAP_PH) then
+
+        local ToD = GetServerVariable("[POP]Goblintrap");
+
+        if (ToD <= os.time() and GetMobAction(GOBLIN_TRAP) == 0) then
```
Comment (TeoTwawki) [inline scripts/zones/East_Ronfaure_[S]/mobs/Battrap.lua]: we stopped using get mob action.

also we started using an entirely different method for lotto pops where we don't even need a script for the NM anymore.
https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/mobs.lua#L54

metalfiiish:
Ok, so the UpdateNMSpawn fails because the NM is not in the NM spawn points. Do I just move him there and delete him from the regular spawn table?

TeoTwawki:
mob spawn table entry is still required. nm spawn table just gives nms the addional repop positions to try and make them nit totally predictable - its a hacky way of avoiding the missing retail feature where nms can use any spawn position a mob in the widescan grouping their ph resides could have

TeoTwawki:
I think I'll get the retail way of spawn positions in as an xmas present to the project :wink:

metalfiiish:
Well in this case the PH is only a specific mob in the list of ten. The NM spawns just next to him every time If I read it correct, though I never played in this zone before, was after my time.

TeoTwawki:
if he does have multiple pos in retail then won't need to use the update funtion

TeoTwawki:
*if he does not*

metalfiiish:
How would I run that function without the UpdateNMSpawn inside it?

metalfiiish:
So I updated it but now my PR won't work because the appveryor PR is pending like 40+ additions to the same sql table. Mine doesn't conflict with ID's. What do I do?

sql/mob_groups.sql
```diff
@@ -10889,7 +10889,7 @@ INSERT INTO `mob_groups` VALUES (10988,4653,82,10000,0,0,15000,9999,83,85,0);
 INSERT INTO `mob_groups` VALUES (10989,4654,222,240,0,0,100000,9999,99,99,0);
 INSERT INTO `mob_groups` VALUES (11001,4657,81,330,128,2797,8000,0,72,76,0);
 INSERT INTO `mob_groups` VALUES (11002,4658,81,330,128,560,4444,0,50,51,0);
-INSERT INTO `mob_groups` VALUES (11003,4655,81,330,128,0,4800,0,60,60,0);
+INSERT INTO `mob_groups` VALUES (11003,4655,81,330,128,4595,4800,0,60,60,0);
```
wrenffxi: `spawntype` column should be set to 32 (SPAWNTYPE_LOTTERY) rather than 128 (SPAWNTYPE_SCRIPTED)

scripts/zones/East_Ronfaure_[S]/mobs/Battrap.lua
```diff
@@ -0,0 +1,25 @@
+-----------------------------------
+-- Area: East Ronfaure [S]
+--  MOB: Battrap
+-- Note:  Place Holder for Goblintrap NM
+-- !pos 156 0 -438 81
+-----------------------------------
+
+require("scripts/zones/East_Ronfaure_[S]/MobIDs");
+
```
Comment (wrenffxi) [inline scripts/zones/East_Ronfaure_[S]/mobs/Battrap.lua]: should also require("scripts/globals/mobs");
this defines the phOnDespawn() function.

scripts/zones/East_Ronfaure_[S]/mobs/Battrap.lua
```diff
@@ -0,0 +1,25 @@
...
+function onMobSpawn(mob)
+end;
+
+-----------------------------------
+-- onMobDeath
+-----------------------------------
+
+function onMobDeath(mob, player, isKiller)
+
+    phOnDespawn(mob,GOBLINTRAP_PH,5,310);
```
Comment (wrenffxi) [inline scripts/zones/East_Ronfaure_[S]/mobs/Battrap.lua]: onMobDeath should be an empty function (it must exist, though, or the core complains)
```
end;
```

phOnDespawn() should be called in onMobDespawn().  According to wiki the cooldown is 60 minutes, or 3600 seconds.

```
function onMobDespawn(mob)
    phOnDespawn(mob,GOBLINTRAP_PH,5,3600); -- 1 hour
end;
```

metalfiiish:
Updated with wrenffxi's comments.

wrenffxi:
Tested and working.

Issue #4066, opened 2017-09-13
Gallijaux and Joulet missing carp item IDs in quest dialogue


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**30170728_2

**_Server Version_** (type `!revision` in game) **:**unknown

**_Source Branch_** (master/stable) **:**master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
Need to put 4401 and 4289 after the cs that starts the quests for both npcs. This should be in the function onTrigger block so it would be something like:

player:startEvent(304,4401,4289);

and the same thing for the other one.

metalfiiish:
This was resolved in the #4297 PR. Along with other missing features from those NPC's.

PR #4297, opened 2017-11-28
Lu shang quest fixes

metalfiiish:
Fixed the issues with initial quest cutscene no passing carp item ID's as per issue:
https://github.com/DarkstarProject/darkstar/issues/4066

Updated both quest givers to give standard dialog after quest starts.

Updated both to utilize Demolish's new npc_utils:fishingAnimation.

NPC's have a minor issue where sometimes they don't fix their rotation (should turn back around after talking with a player). Talking with them again will fix it.


metalfiiish:
The only lingering thing is now they turn around (face away from player) while talking to players during the initial cutscene to start quest. Other than that they are perfect.

teschnei:
are they actually supposed to stop fishing on trigger?

for the event yes, since the event bugs out otherwise, but do you know for sure that the showText should have them putting their rod away?

metalfiiish:
well showtext forces them to look at you with their fishing rod in your face, unless I am missing a param that tells them not to turn.

Trading them, they should not turn at all for sure.

takhlaq:
looks like retail uses an event, and they're not supposed to turn around https://www.youtube.com/watch?v=Xz0yOPfjSQc

metalfiiish:
Yep, then this PR so far looks to match retail, since the changes kj advised resolved thhat.

PR #4298, opened 2017-11-28
Promy: prevent parties being split when moving from floor 2 to 3

wrenffxi:


Wiggo32:
we going to fix the rest of the promy issues too?

wrenffxi:
Like the people who maintain DSP, I do this stuff in my free time because I love playing FFXI on a private server and want to give back to the community.  We all only have so much time and energy to put into it, and tend to work on whatever pieces of the code base catch our interest.  I greatly appreciate the effort you put into identifying, tracking, and testing issues.  It's discouraging to see comments on our work that say "okay, but can you do this now?"

I get it.  You want to help make things retail accurate, and that's a good goal.  But the list of things to work on is endless, and grows at least as quickly as we develop.  We need more people who will dig into code, learn how to use git, submit pull requests, take constructive feedback from the project admins.

I'm not saying that needs to be you.  Like I said, creating and maintaining issues is important too.  Contributions to public projects tend to be thankless, so I want to thank you for yours.  But I encourage you not to follow active developers around and request more work be done, or ask them to focus on one issue or another.  We all keep tabs on the open issues list and work on things that interest us at the moment.  If getting Promyies as retail-accurate as possible truly drives you, I do encourage you to dive in and make some pull requests of your own.  That's how this project stays alive.

Thanks for listening. ~ Wren

Wiggo32:
I completely understand Wren. Based on my experiences though, they're not likely to merge this being that it only provides a "let's fix this part for now" solution to the entire issue.  The process has been explained to me in great detail on more than one occasion, I'm only giving you a heads up. I totally get the value of time and how precious it is (we only have so much). And while I don't expect anyone to just drop everything and fix all the issues that are on the tracker, my comment was merely to bring to attention that there were some other things pending that were related to this... just in case it was missed. No disrespect intended. And we all do appreciate all the hard work you do to make the project better.

TeoTwawki:
Theres a big diff between being asked to use a better method provided by devs or tack on one tiny thing to your pr and being hounded to fix every mildly related issue.

PR #4308, opened 2017-11-30
Msg incorrectly used same object for PSender/PTarget

ShelbyZ:
- check the CMagicState for the target


teschnei:
You probably need to check every msg that can show up there and make sure that the state's target is actually the correct parameter for that

teschnei:
MSGBASIC_IS_INTERRUPTED, MSGBASIC_IS_PARALYZED and MSGBASIC_IS_INTIMIDATED

teschnei:
while the second commit is useful, can you move it to a separate pull request?

PR #4310, opened 2017-11-30
Core: Reorganized lua_baseentity.cpp and lua_baseentity.h

Cryptexx:
No code modifications were performed. This was strictly a code reorganization for logical flow and grouping of functionality/use. Files compile without error.

Changes made:

1. Reorganized all functions into logical categories.
2. Organized header paths
3. Provided real examples of each function call.
4. Identified functions not being used.
5. Identified a function or two that does the same calculation as a core function elsewhere already.
6. Provided relevant notes, including variables that should be passed and what return should be expected.
7. Translated all Russian to English in .cpp (To do: translate in header)
8. Cleaned up code, removed extra lines, tabs, and standardized punctuation, etc.
9. Added comments above each function to provide the following details:

/*
Name:
Purpose:
Example:
Notes:
*/

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -21,20 +21,63 @@
 ===========================================================================
 */
 
-#include "../../common/showmsg.h"
-#include "../../common/timer.h"
-#include "../../common/utils.h"
-#include "../../common/kernel.h"
 #include <math.h>
 
+#include "lua_battlefield.h"
 #include "lua_baseentity.h"
 #include "lua_instance.h"
 #include "lua_item.h"
 #include "lua_statuseffect.h"
 #include "lua_trade_container.h"
-#include "lua_battlefield.h"
-#include "lua_zone.h"
 #include "luautils.h"
+#include "lua_zone.h"
+
+#include "../../common/kernel.h"
+#include "../../common/showmsg.h"
+#include "../../common/timer.h"
+#include "../../common/utils.h"
+
+#include "../ability.h"
+#include "../alliance.h"
+#include "../enmity_container.h"
+#include "../guild.h"
+#include "../instance.h"
+#include "../item_container.h"
+#include "../latent_effect_container.h"
+#include "../map.h"
+#include "../message.h"
+#include "../mob_modifier.h"
+#include "../mobskill.h"
+#include "../mob_spell_container.h"
+#include "../recast_container.h"
+#include "../spell.h"
+#include "../status_effect_container.h"
+#include "../trade_container.h"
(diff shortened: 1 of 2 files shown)


takhlaq:
oh man we needed this so bad
from a quick glance it looks great but i'll let @teschnei merge

Wiggo32:
some misspelled words in the explanations but probably not a big deal as long as the core function code hasn't changed

ReachServer:
Where?

Wiggo32:
https://github.com/Cryptexx/darkstar/blob/14074f2f403876e4ef7f32f48faacdd13da39776/src/map/lua/lua_baseentity.cpp#L13493
but I supposed you *could* have meant for it to say "steamItem"... doesn't make sense for a descriptive.
and we have russian here which goes against "7. Translated all Russian to English in .cpp"
https://github.com/Cryptexx/darkstar/blob/14074f2f403876e4ef7f32f48faacdd13da39776/src/map/lua/lua_baseentity.cpp#L947
And these aren't the only spots, just what I found from 'skimming' it's a very large document to proof read and should be tasked to the contributor. However, I *did* say that it's not a big deal because it's only comments, just as long as the core function code doesn't have any accidental mistakes.

Wiggo32:
https://github.com/Cryptexx/darkstar/blob/14074f2f403876e4ef7f32f48faacdd13da39776/src/map/lua/lua_baseentity.cpp#L287
"standad"

ReachServer:
Thanks for identifying. Was expecting a few typos in a 14000 page document. I'll fix them at a later time. If I had no grammar/spelling mistakes in that behemoth, I'd be a Lvl 1200 copy/paste/edit grand ninja wizard.

takhlaq:
so he doesn't have to deal with fuck ton of conflicts, i'd like to get this in as soon as possible (seemed fine from a glance) though i'll wait for kj or teo to look over it too

TeoTwawki:
@Cryptexx 
```cpp
*  Notes   : Available hex codes: 0x1C, 0xD, 0xF, 0x1F, ___, ___
*          : Can modify the name shown through explicit declaration
```
That's on PrintToPlayer, I'd like to just direct people to the header that defines the message types instead of listing them here.

https://github.com/DarkstarProject/darkstar/blob/master/src/map/packets/chat_message.h#L31
Something like "see packets\filename.h for message types" in these situations

Other than that suggestion if we are going to actually use these headers for something, I am good with it and am more than good to see the thing organized instead of tacking on whatever at bottom of file from here out 👍

teschnei:
I don't need to review this, there are no code changes

PR #4313, opened 2017-11-30
Added Maat's 2hr dialog to the list

Wiggo32:


metalfiiish:
Wouldn't we only want to add that once he is implemented to use it?

TeoTwawki:
We can have it there ahead so the person doing it can easy find. or alternativly use offsets from maats first message for all of them since they are all together.

(person who impliments can figure out how they want to use it)

TeoTwawki:
should prolly do the other maats too u less those already exist (I cant look atm, rl being a pain)

Hozu:
If all the dialogue is consecutive text IDs, wouldn't it be better to use just one variable for the first one and do it incrementally? ie. MAAT_FIGHT_TEXT = 7628, and when you want to refer to the 2hr text you would do MAAT_FIGHT_TEXT +5

TeoTwawki:
thats what i meant by offsets

teschnei:
Reading/writing scripts and having "MAAT_FIGHT_TEXT + 5" isn't very helpful - you wouldn't have a clue if it was the right textid or not without looking it up

If your concern is the auto textid updater having problem... that's an issue for the textid updater

Hozu:
I figured that it would be consistent with what the animated weapons have. Meh.

teschnei:
Yeah.. we have it both ways already.  I usually do it this way because I prefer clarity in the scripts

PR #4314, opened 2017-11-30
Update lightreaver.lua

Ninjistix:
adjusted proc rate

### DIFF
--- a/scripts/globals/items/lightreaver.lua
+++ b/scripts/globals/items/lightreaver.lua
@@ -18,7 +18,7 @@ function onAdditionalEffect(player,target,damage)
 
     -- Now continue with normal resistance and random number checks..
     local procRate = 3;
-    if (procRate >= math.random(1,100) or applyResistanceAddEffect(player,target,ELE_DARK,0) <= 0.5) then
+    if (procRate <= math.random(1,100) or applyResistanceAddEffect(player,target,ELE_DARK,0) <= 0.5) then
         -- If random number beats procRate OR if effect is resisted, return all zero
         return 0,0,0;
     else


scripts/globals/items/lightreaver.lua
```diff
@@ -18,7 +18,7 @@ function onAdditionalEffect(player,target,damage)
 
     -- Now continue with normal resistance and random number checks..
     local procRate = 3;
-    if (procRate >= math.random(1,100) or applyResistanceAddEffect(player,target,ELE_DARK,0) <= 0.5) then
+    if (procRate <= math.random(1,100) or applyResistanceAddEffect(player,target,ELE_DARK,0) <= 0.5) then
```
TeoTwawki: if 3 is less than or equal to 1to100? it's going to be less than very often.

teschnei:
it has a 97% proc chance?

R3P0FFXI:
@TeoTwawki @teschnei There is actually no data on BG wiki or FFXIwiki regarding the proc rate of this. The additional Effect is Death so this is most likely a "guess"

teschnei:
you are going to guess that it procs 97% of the time?

Ninjistix:
this was reported on the server "Lightreaver Great axe seems to have 95%-99% proc rate with death on any mob which isn't Impossible to gauge."

scripts/globals/items/lightreaver.lua
TeoTwawki: @teschnei actually, look at the rest githubs not showing us
I derped

TeoTwawki:
reversed my checks withotu reversing my signage
github didn't display the lined beneath sicne he didn't change them
he's right

TeoTwawki:
``lua
    -- Now continue with normal resistance and random number checks..
    local procRate = 3;
    if (procRate <= math.random(1,100) or applyResistanceAddEffect(player,target,ELE_DARK,0) <= 0.5) then
        -- If random number beats procRate OR if effect is resisted, return all zero
        return 0,0,0;
    else
        -- It's clobberin time!
        target:setHP(0)
        return SUBEFFECT_DEATH, msgBasic.ADD_EFFECT_STATUS, EFFECT_KO;
    end
end;
```

TeoTwawki:
@R3P0FFXI 
thats not the check for it going off
its the check for it NOT going off

teschnei:
takeaway: teo writes really confusing code sometimes

PR #4318, opened 2017-12-01
Re-Engage Delay Option Added to map_darkstar.conf

ReachServer:
Added ability to set re-engage delay to on or off through the map_config. 1 = on (by default); 0 = off (no delay).

### DIFF
--- a/conf/map_darkstar.conf
+++ b/conf/map_darkstar.conf
@@ -92,6 +92,9 @@ disable_gear_scaling: 0
 #Enable/disable jobs other than BST and RNG having widescan
 all_jobs_widescan: 1
 
+#Enable/disable re-engage delay when acquiring new targets (1 = enabled; 0 = disabled)
+reengage_delay: 1
+
 #Modifier to apply to player speed. 0 means default speed of 40, where 20 would mean speed 60 or -10 would mean speed 30.
 speed_mod: 0
 
--- a/src/map/ai/controllers/player_controller.cpp
+++ b/src/map/ai/controllers/player_controller.cpp
@@ -69,7 +69,9 @@ bool CPlayerController::Engage(uint16 targid)
     {
         if (distance(PChar->loc.p, PTarget->loc.p) < 30)
         {
-            if (m_LastAttackTime + std::chrono::milliseconds(PChar->GetWeaponDelay(false)) < server_clock::now())
+            auto nextAttack = map_config.reengage_delay ? m_LastAttackTime + std::chrono::milliseconds(PChar->GetWeaponDelay(false)) : m_LastAttackTime;
+
+            if (nextAttack < server_clock::now())
             {
                 if (CController::Engage(targid))
                 {
--- a/src/map/map.h
+++ b/src/map/map.h
@@ -91,6 +91,7 @@ struct map_config_t
     bool   level_sync_enable;         // Enable/disable Level Sync
     bool   disable_gear_scaling;      // Disables ability to equip higher level gear when level cap/sync effect is on player.
     bool   all_jobs_widescan;         // Enable/disable jobs other than BST and RNG having widescan.
+    bool   reengage_delay;            // Enable/disable re-engage delay when acquiring new targets (1=enabled; 0=disabled)
     int8   speed_mod;                 // Modifier to add to player speed
     int8   mob_speed_mod;             // Modifier to add to monster speed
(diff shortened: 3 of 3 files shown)


TeoTwawki:
I tried to get this in previuosly as a % reduction from full to none so you could do things like half delay as well as no delay. got rejected. 😟

TeoTwawki:
#749 @teschnei

TeoTwawki:
stinking mobile..i wasnt trying to assign

ReachServer:
This just provides an on/off switch so even if they put in some wild value, it's either going to evaluate to true or not, so won't break anything or cause conflicts.

Not retail by any means, but this seems to be pretty popular with private servers so figured I'd put the option out there for centralized control. Didn't know it was requested before though.

teschnei:
Problem: engaging resets your swing timer to 2seconds, so this is easily abusable by anyone using a long delay weapon.  Fortunately, the last attack time is saved outside of attack_state (in the player_controller, which you've probably noticed since the line you changed involves it), so it should be possible to set attack_state's swing timer to remaining attack time (which is 0 without the setting) + 2 seconds in its constructor

TeoTwawki:
@ReachServer I understood what you have going on. was saying i previously tried to do a setting that did this and more, and linked you to the previous reasoning it didnt get into the project.

... and then tagged the big chief to remind him before accidey assigning him coz tiny touchscreens  suck...

ReachServer:
Ah, I see in the old PR now, that's quite the exploit. Hmm..

TeoTwawki:
its not so bad unless you go really low or no delay. but setting got rejected on the premise a one liner wasn't a good use of a conf option

TeoTwawki:
btw if this gets closed i refuse to be the guy that closes it 😉 coz its like the 2nd most thing i get forum messaged for

teschnei:
If it gets updated with my suggestion, it will work well and I have no problem merging it

ReachServer:
I'll update it tomorrow teschnei, thanks

Wiggo32:
@TeoTwawki ```p.s.
i like to run with half sized engage delays myself```
testing in retail since free play I've become pretty familiar with the weapon delay and I, myself believe that dsp is just a little too long anyway. No official testing with timers and such, but I just cut mine in half in the core and it feels and plays a lot more 'retail realistic'. Just a mention...

Wiggo32:
sidenote: I wouldn't use this feature if it were to get merged. I don't want my delay gone, I just want it to play closer to retail. So, I would have to keep my own custom core modification anyhow.

teschnei:
It could probably be a delay (in seconds) to lower the check by.  If you want it gone, set it to like 30 or something.

Also, since engaging sets the swing timer to 2s, we could lower the engage delay by up to 2 seconds for the default.

TeoTwawki:
I'm happy with anything that satisfies the majority of folk who come asking for this,

ReachServer:
Which file is the 2s determined? I have two ideas regarding this - make the opening attack not a fixed 2s, but instead scale in some way with the delay of the weapon (not retail behavior, but would help mitigate any dirty tricks). Is the 'macro attack' your referring to utilizing a change in main weapons over and over? I think that would exploit this line in charutils.cpp:

```
if (PChar->PAI->IsEngaged())
                    {
                        auto state = dynamic_cast<CAttackState*>(PChar->PAI->GetCurrentState());
                        if (state) state->**ResetAttackTimer**();
                    }
```

teschnei:
src/map/ai/states/attack_state has a member variable (m_attackTime) which is the time remaining until your next auto attack.  It's initialized with 2s currently.  The proposed solution is: make the config option a number instead of a boolean.  In player_controller, use that time as an offset to compare current server tick vs m_LastAttackTime.  Then, in attack_state, instead of initializing to 2s, it should initialize to max(0, current tick - (m_LastAttackTime + weapon delay)) + 2s.  That should mean it's always 2s if you don't change the setting, but increases to (time remaining since your last auto attack + 2s) if you re-engage earlier.

It could be a bit tricky since anything other than players don't have m_LastAttackTime.  If you want, I can do that part and you can do the config option part.

TeoTwawki:
This going to be continued at some point?

takhlaq:
closing for now so i don't keep looking at it
reopen when requested changes have been made or hop in discord if you'd like some help

PR #4319, opened 2017-12-01
Reorganized /w minor corrections guild_item_points.sql

Wiggo32:
I don't know if you guys want this but it's something I did because I got tired of seeing the Goldsmith guild ask for Beetle Earring +1 https://github.com/DarkstarProject/darkstar/blob/b069092f78bbfb878d50fc0c53970b76a67f3ad3/sql/guild_item_points.sql#L471
or the Alchemy guild ask for an argute belt https://github.com/DarkstarProject/darkstar/blob/b069092f78bbfb878d50fc0c53970b76a67f3ad3/sql/guild_item_points.sql#L1048
I didn't change much unless it was blatantly wrong or missing (like some of the HQ stuff), just mostly reorganized. I figure it could help when someone is ready to make that major update and change ALL of the values because retail gives way more guild points than they used to. Up to you guys, doesn't matter either way. It will eventually need to be completely updated anyhow.

### DIFF
--- a/sql/guild_item_points.sql
+++ b/sql/guild_item_points.sql
@@ -25,1278 +25,1515 @@ CREATE TABLE `guild_item_points` (
 -- ----------------------------
 -- Records 
 -- ----------------------------
-INSERT INTO `guild_item_points` VALUES ('0', '2850', '9', '2850', '6240', '3');
-INSERT INTO `guild_item_points` VALUES ('0', '4288', '6', '1200', '5040', '2');
-INSERT INTO `guild_item_points` VALUES ('0', '4288', '6', '1200', '5040', '3');
-INSERT INTO `guild_item_points` VALUES ('0', '4288', '6', '1200', '5040', '5');
-INSERT INTO `guild_item_points` VALUES ('0', '4289', '1', '45', '1600', '1');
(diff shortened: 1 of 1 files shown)


teschnei:
Need to have the reorganizing and the minor fixes into separate commits so we can actually see what was changed

Wiggo32:
Yea, that makes sense. Didn't even think of that when I started doing it weeks ago. Originally just planned on doing it for my own benefit cause I wasn't sure something like this would be useful/wanted. Oh well, I'll remember for next time I guess.

Wiggo32:
or if you think it's useful and you know of an easy way I could redo this where it won't take me a week, I could do it. or... if you think it's just not worth the effort because like I said earlier, it's all going to change anyway eventually, I'd rather spend my time doing something else.

teschnei:
The process would probably be: undo the minor fixes, commit --amend to change the previous commit to only the reorganizations, then redo the minor fixes, and commit those into a new commit, and then force push to this branch - not too hard on it's own.  The hard part is remembering all the minor fixes to undo.  If it were me doing it, I'd load both the old and new data sets into a db table and then compare which ones have changed.  If there's not many, then it would probably only take a few minutes

takhlaq:
branch was deleted

PR #4325, opened 2017-12-02
Moved some status effect deletion logic around

### LINKED ISSUE #4324: Server crashes when attempting to rest while sublimation charging
**_I have:_**

**_Client Version_** (type `/ver` in game) **:** 30171103_1

**_Server Version_** (type `!revision` in game) **:** a72563e

1. Changed Job to Scholar lvl 50
2. Activated Sublimation
3. Rested while HP draining
That got me a crash 3 times in a row. It said "DSGame-server.exe has stopped working" so I think that's actually a crash. Not a debug break or anything.

teschnei:
RemovedStatusEffect only flags for deletion
New function DeleteStatusEffects actually does the deletion

Can no longer accidentally run into undefined behaviour when deleting an effect
inside OnEffectTick

Fixes #4324 

Additional information:

This technically hasn't always been an issue - must have come about when I moved some effect stuff around.  However, I think that prior to that, it was still undefined behaviour (or wrong behaviour) rather than being a crash.  If I had to guess, some effects ticks would randomly get skipped instead.  So when I changed the effect container, it just made the bug more obvious.

Also, this has a side effect of only sending the effect updating packets once per tick instead of a whole set for each expiring effect (if multiple effects expire on the same tick)

### DIFF
--- a/src/map/status_effect_container.cpp
+++ b/src/map/status_effect_container.cpp
@@ -259,7 +259,8 @@ bool CStatusEffectContainer::CanGainStatusEffect(CStatusEffect* PStatusEffect)
                 // slow i remote
                 return true;
             }
-            else if (statusEffect == EFFECT_DIA || statusEffect == EFFECT_BIO)
+
+            if (PStatusEffect->GetTier() != 0 && negativeEffect->GetTier() != 0)
             {
                 return PStatusEffect->GetTier() == negativeEffect->GetTier() ? statusEffect > negativeId : PStatusEffect->GetTier() > negativeEffect->GetTier();
             }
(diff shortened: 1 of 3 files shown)


src/map/status_effect_container.cpp
```diff
@@ -460,14 +474,31 @@ void CStatusEffectContainer::RemoveStatusEffect(uint32 id, bool silent)
...
+{
+    CStatusEffect* PStatusEffect = m_StatusEffectList.at(id);
+    PStatusEffect->deleted = true;
+    if (m_POwner->objtype == TYPE_PC)
+    {
+        CCharEntity* PChar = (CCharEntity*)m_POwner;
+
+        if (PStatusEffect->GetIcon() != 0)
+        {
+            if (silent == false)
```
takhlaq: `!silent` cause fuck reading

PR #4326, opened 2017-12-02
Less buffer macros

ShelbyZ:
- Adding ref(buf, index) to socket.h
- removing RBUF* macros for ref<T>
- removing WBUF* macros for ref<T>
- fixing changes resulting from


src/map/lua/luautils.cpp
```diff
@@ -1771,27 +1771,40 @@ namespace luautils
 
     int32 OnEffectTick(CBattleEntity* PEntity, CStatusEffect* PStatusEffect)
     {
-        lua_prepscript("scripts/%s.lua", PStatusEffect->GetName());
-
-        if (prepFile(File, "onEffectTick"))
+        try
```
teschnei: unintentional testing code?

ShelbyZ: yeah, leftovers frame last night

src/common/socket.h
```diff
@@ -303,5 +290,16 @@ extern int32 naddr_;   // # of ip addresses
 	int32 sendudp(int32 fd,void *buff,size_t nbytes,int32 flags,const struct sockaddr *from,socklen_t addrlen);
 #endif 
 
+template<typename T, typename U>
+T& ref(U* buf, std::size_t index)
+{
+    return *reinterpret_cast<T*>(buf + index);
+}
+
+template<typename T, typename U = void*>
```
teschnei: what is this?

ShelbyZ:
```cpp
void* thing;
thing + 1; // error, void* unknown size
```

If I don't provide the overload, it means adding more (uint8*) into message, login_auth, etc.  Where we are using zmq and get data() (void*).

teschnei:
but why have a template parameter U if you don't use it anywhere?

teschnei:
Also... we should cast there because it's a void pointer, in my opinion

ShelbyZ:
Is the suggestion to move to:

```cpp
template<typename T>
T& ref(void* buf, std::size_t index)
{
    return *reinterpret_cast<T*>((uint8*)buf + index);
}
```

teschnei:
I'd rather just get rid of it entirely and cast void pointers to uint8
pointers.  We could theoretically want a buffer that's not uint8...

There's not that many void pointers anyways I don't think.  Should be just
the zmq buffers which we can't control the type of

PR #4336, opened 2017-12-04
Implements rusk food effects.

xipies:
Used info from BG forum for HQ rate percentage.

### DIFF
--- a/src/map/utils/synthutils.cpp
+++ b/src/map/utils/synthutils.cpp
@@ -400,7 +400,7 @@ uint8 calcSynthResult(CCharEntity* PChar)
 
                     // Using x/512 calculation for HQ success rate modifier
                     // see: https://www.bluegartr.com/threads/130586-CraftyMath-v2-Post-September-2017-Update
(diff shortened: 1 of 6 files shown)


src/map/utils/synthutils.cpp
```diff
@@ -396,6 +396,12 @@ uint8 calcSynthResult(CCharEntity* PChar)
                         default: chance = 0.000; break;
                     }
 
+                    int16 modSynthHqRate = PChar->getMod(Mod::SYNTH_HQ_RATE);
+
+                    // Using x/512 calculation for HQ success rate modifier
+                    // see: https://www.bluegartr.com/threads/130586-CraftyMath-v2-Post-September-2017-Update
+                    chance += (double)modSynthHqRate * 0.001953125;
```
teschnei: If it's /512, can you just write it that way in the code?  The compiler should optimize it to the same thing either way, and it's clearer that way to see what it's doing

chance += (double)(modSynthHqRate / 512.); should probably work

PR #4345, opened 2017-12-05
Adjust nin quest fight

ymkc:
try to bring it closer to retail/wikis (= making it solo friendlier for low levels)


TeoTwawki:
so..I can sneak pull and hold one leech till the other 2 depop now?

ymkc:
yes, that was the intention

PR #4349, opened 2017-12-06
Lottery NMs Tottering Toby and Blubbery Bulge

wrenffxi:


scripts/zones/Davoi/mobs/Davoi_Mush.lua
```diff
@@ -1,31 +1,14 @@
...
-            local BlubberyBulge = 17387920;
-            DisallowRespawn(BlubberyBulge, false);
-            GetMobByID(BlubberyBulge):setRespawnTime(GetMobRespawnTime(mobID));
-            SetServerVariable("[PH]Blubbery_Bulge", mobID);
-            DisallowRespawn(mobID, true);
-        end
-    end
-end;
\ No newline at end of file
+    phOnDespawn(mob,BLUBBERY_BULGE_PH,20,3600); -- 1 hour
```
TeoTwawki: you found cooldown info for him?

scripts/zones/Batallia_Downs/mobs/Stalking_Sapling.lua
```diff
@@ -1,27 +1,17 @@
...
-        if (ToD <= os.time() and not GetMobByID(TOTTERING_TOBY):isSpawned()) then
-            if (math.random(1,20) == 5) then
-                UpdateNMSpawnPoint(TOTTERING_TOBY);
-                GetMobByID(TOTTERING_TOBY):setRespawnTime(GetMobRespawnTime(mobID));
-                SetServerVariable("[PH]Tottering_Toby", mobID);
-                DisallowRespawn(mobID, true);
-            end
-        end
-    end
+    phOnDespawn(mob,TOTTERING_TOBY_PH,20,3600); -- 1 hour
```
TeoTwawki: Toby can take way longer than an hr, so either the spawn rate is much lower or the cooldown can be much longer.

scripts/zones/Davoi/mobs/Davoi_Mush.lua
TeoTwawki: Meh actually I'm fine with the assumed 1hr instead, coz who camps this guy seriously anyway..carry on!

scripts/zones/Batallia_Downs/mobs/Stalking_Sapling.lua
wrenffxi: I was using the information from [here](http://ffxiclopedia.wikia.com/wiki/Talk:Tottering_Toby) primarily.  It seems that when people camp the correct placeholder and kill it asap, toby's cooldown is assumed to be about an hour.

TeoTwawki: which would indicate he's lower rate. I had to do him for several diff magian weapons on retail when those were new. we were murdering every sapling zone wide, wasn't a single one left alive in map. unfortunately theres now ay to get handle one what the rate really should be (I suspect SE has tiers for it they group NMs into) so I guess the real question is..Do you really want to go with 1 in 20 chance there? I guess Ima merge it for the same reason I changed my mind about caring if blubbery is 100% accurate, and anyone objecting can PR the changes. No real reason to matter when magian trials aren't done yet

Issue #3196, opened 2016-05-23
Kreutzet spams Stormwind


DanteMccloud:
**_I have:_**
- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160427_0

**_Server Version_** (type `@revision` in game) **:**
value returned was "Revision is: Unknown"

**_Source Branch_** (master/stable) **:**
Master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
NM Kreutzet in Cape Teriggan spams Stormwind. Fights start out fine until he uses stormwind for the first time and then he just keeps spamming it and nothing else. I can't tell if he is stuck at a loop or just not loosing TP after he uses the move. I think the problem is within the Stormwind.lua itself but I'm unsure.

dazusu:
Uh, he's meant to spam it.... and he needs to land 3 before it will stop.

What we do need to fix is AOE abilities that don't require a target stopping when the player runs out of range. I believe @teschnei  said the fix is to change the ability to target the mob instead of a player.

DanteMccloud:
He is suppose to use it consecutively 3 times and stop regardless of if his attack hits or not. Currently you can test him with Godmode and after he starts his first stormwind he just loops it with no auto attacks inbetween not 3 times and stop but infinite looping. Unless retail updated how this mechanic works since I last fought him which is 100% possible I suppose.

"Stormwind will always be used 3 times in succession with no other actions inbetween"

dazusu:
On retail, if the player runs out of range - Stormwind still goes off/animates.
On DSP, if the player runs out of range - Stormwind does not complete.

I'll look at the spam issue, though.

DanteMccloud:
Just re-tested it in fear of maybe it "missing" causing it to not count 1 2 3 and be over. I attacked him until he got to about 90% hp and used his first Stormwind. I turned off godmode and put my weapon away (no TP being fed) he used stormwind about 37 times in a row after that each of them landing and dealing a nice lump of damage. (had to toggle godmode on and off a few times to not die while testing it, but it's looping for sure not just 3 in a row. If there is anything else I can do to help please let me know.

nyczducky:
[...]

PR #4363, opened 2017-12-09
Fix loops on Panzerfaust and Stormwind

ymkc:
unfortunately not 100% matching retail, as they do not continue the chain when they have to catch up to PC. but still way better than an infinite loop


teschnei:
what was wrong with the old way?

ymkc:
i never saw it work, mobs always got stuck in a loop when using them
and just moved despots panzerfaust part into the mob's lua to make it better visible

teschnei:
Yeah, moving it to the mobs lua is good.  Is mob:actionQueueAbility() always returning false?  That should be fixed if it is rather than working around it.

ymkc:
i think it is

teschnei:
I think I'll just get rid of it, actually.

Can you fix horrid roar too? horrid_roar_3.lua (it's the only thing other than panzerfaust/stormwind using actionQueueAbility)

ymkc:
sure, will do

and is there a way to check whether an mobskill was successfully executed?
maybe add a listener on WEAPONSKILL_STATE_EXIT?

teschnei:
this is for repeating the move if it wasn't executed successfully?

TeoTwawki:
whether it lands or not, according to issue #3196

TeoTwawki:
err, sorry, think I misunderstood

ymkc:
yes, to mimic retail

PR #4366, opened 2017-12-09
Change armored goblins' element

ymkc:
armored gobs should drop lightning crystals. not sure this is the best/retail correct way


TeoTwawki:
It of the 3 existing goblin families 2 of them have identical values. ID's 327 and 373 are identical and this new 500 is the same except for the crystal. ID  133 has stat differences. We've a lot of extra families from back when skill lists were tied to it (and from when NMs required their own ID because of that). 

What we could do here is merge these 2 duplicates and then use the freed up ID for this (which is really just changing one value in said duplicate).

Thoughts?

TeoTwawki:
I guess thought was "yeah ok I'll just do that" :)

PR #4377, opened 2017-12-12
Don't show player CS event about killing Quadav if they missed getting the CS at Iron Eater in Bastok that directs them here..

TeoTwawki:
1. get cs Iron Eater from metalworks
2. go kill quadavs
3. on zone out get cs telling if you killed enough or not - if not, you et sent back to kill more, this was happening despite never doing step one leading people to think the mission was broken when in reality they simply never started..

### DIFF
--- a/scripts/zones/Pashhow_Marshlands/Zone.lua
+++ b/scripts/zones/Pashhow_Marshlands/Zone.lua
@@ -4,45 +4,43 @@
 --
 -----------------------------------
 package.loaded[ "scripts/zones/Pashhow_Marshlands/TextIDs"] = nil;
-package.loaded[ "scripts/globals/missions"] = nil;
-package.loaded["scripts/globals/chocobo_digging"] = nil;
 -----------------------------------
-
-require("scripts/globals/zone");
-require("scripts/globals/settings");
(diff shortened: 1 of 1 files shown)


scripts/zones/Pashhow_Marshlands/Zone.lua
```diff
@@ -74,9 +64,9 @@ function onZoneIn( player, prevZone)
 
     if (prevZone == 147 and player:getCurrentMission(BASTOK) == THE_FOUR_MUSKETEERS) then
         missionStatus = player:getVar("MissionStatus");
-        if (missionStatus < 22) then
+        if (missionStatus > 0 and missionStatus < 22) then
             cs = 10;
-        elseif (missionStatus == 22) then
+        elseif (missionStatus > 0 and missionStatus == 22) then
```
takhlaq: wat

TeoTwawki: woops

TeoTwawki: fix'd the derp that fixed derp!

PR #4382, opened 2017-12-14
added lua binding to check if entity is inside zone region

takhlaq:
as per
```
[1:19 PM] Poke: Yeah i didnt know if there was an "inRegion (you confirmed no) or somethign that just allows it to see if in an area
[1:20 PM] Poke: seems like it would be here. but I guess retail ticks in the region
[1:20 PM] Poke: read something like that how the attohwa chasm plants tick in region on retail to keep poison on if it wears
```

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -2425,6 +2425,30 @@ inline int32 CLuaBaseEntity::isInMogHouse(lua_State* L)
 }
 
 /************************************************************************
+*  Function: isInZoneRegion(id)
+*  Purpose : Returns true if a entity is within region (defined by zone:registerRegion)
+*  Example : if player:isInZoneRegion() then -- poison and die
+*  Notes   :
+************************************************************************/
+
+inline int32 CLuaBaseEntity::isInZoneRegion(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+
+    bool ret = false;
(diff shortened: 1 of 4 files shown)


teschnei:
Can we just do it in the regular tick? And have a function like.. get
current region?

takhlaq:
i was wanting to change it to use a vector of regions per player in case theyre withing multiple regions at once

teschnei:
Then player:isInRegion instead of get current, but the rest still stands
(not using a new binding)

takhlaq:
> Can we just do it in the regular tick?

do you mean get rid of the new luautils binding and use status effect tick instead?
im not sure if there's any other cases in retail where we'd need to check if player is within a certain area for a certain duration so i stuck it in the zone's lua but using poison effect tick instead seems dirty

i was also thinking to just push (region, players) instead of per player cause it's dumb

teschnei:
Really, we don't even need anything new for this. You can just add a
listener for effect tick on region enter and remove it on region exit. I'd
rather do that than add yet another Lua caller

takhlaq:
you can ya, but i added to zone region in case any other content works the same way (certain actions within an area which isnt limited purely to battlefield/instance)
if you cant think of any then i'll just close this since it was more to just cover all bases instead of just for that one case with poison

teschnei:
the status effect tick listener doesn't have anything to do with actual effects, it's just an event that fires when effects tick (every 3 seconds) regardless of how many effects you have.  So using a listener (either tick or status effect tick) added on region enter and removed on region exit does exactly what you did in this PR

Rigginit:
Will this change allow scripts to work inside of a region where no 'status effects" are involved?

teschnei:
Closed because it's already possible and doesn't need any extra work

PR #4383, opened 2017-12-14
cap action packets at 30 yalms during combat

takhlaq:


teschnei:
Hmm, is it actually retail, though?

If it's for the "alliance lag", I don't think capping action packets at 30 really makes much of a difference.  How often do you have tons of people farther than 30 yalms but closer than 50?

R3P0FFXI:
@teschnei @takhlaq  the issue isnt being reported by players across the zone its being reported by players in a small BCNM area or in a party or in a big alliance fight. In retail i think it is 50 yalms. It seems to be something related directly to players in the party/alliance.

takhlaq:
>Hmm, is it actually retail, though?

ya, check your backlog in windower

src/map/zone_entities.cpp
```diff
@@ -819,12 +819,16 @@ void CZoneEntities::PushPacket(CBaseEntity* PEntity, GLOBAL_MESSAGE_TYPE message
             }
             case CHAR_INRANGE:
             {
+                // todo: rewrite packet handlers and use enums instead of rawdog packet ids
+                // 30 yalms if action packet and in combat, 50 otherwise
+                const int checkDistanceSq = packet->id() == 0x0028 && PEntity->PAI->IsEngaged() ? 900 : 2500;
```
teschnei: why the engaged check?  i don't think that's part of anything that was said in windower

takhlaq: i made poopy

Issue #4357, opened 2017-12-07
Zanshin Attacks triggering when not engaged


R3P0FFXI:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 
30171103_1

**_Server Version_** (type `!revision` in game) **:** 
https://www.dropbox.com/s/vaymv7t93uliya5/crashdump.7z?dl=0

**_Source Branch_** (master/stable) **:** 
Master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Become a Samurai
KIll shit without even taking out your weapon
I only noticed it with Seigan/Third Eye up while running pulling mobs. Agro'd mobs would get claimed and saw a few GK skillups. Not sure if this is Seigan specific, not heavily tested.

Due to the nature of zanshin, it's possible it affects things like counter and parry too, although I haven't noticed it.

dwn134:
It effects war retaliation also

metalfiiish:
I think it's just a matter of adding PTarget->PAI->isEngaged()  or PTarget->PAI->isAutoAttackEnabled() in various spots of battleentity.cpp for OnAttack function. I will try to reproduce and test if this could fix it when I get some time.

teschnei:
Retaliations check is in HandleSpikesEffect in battleutils and Seigans
counter is in CheckCounter in attack.cpp, you'd want to check engaged in
those places

metalfiiish:
We should also look at IsParried from attackutils no? Shouldn't parry if weapon isn't drawn.

teschnei:
Only if it is currently a problem

teschnei:
GetParryRate checks weapon drawn, so it should be fine already

metalfiiish:
ok cool. I'll test it as well but it sounds like it should be ok.

metalfiiish:
This can be closed. Was able to test the fix resolved this issue.

PR #4385, opened 2017-12-15
{CORE} Fix various counter attacks to check if entity is engaged/weapon drawn.

metalfiiish:
Fixes for the issues with Zanshin, counter attacks and retaliation effect:
https://github.com/DarkstarProject/darkstar/issues/4357


src/map/attack.cpp
```diff
@@ -346,6 +351,11 @@ bool CAttack::IsCountered()
 
 bool CAttack::CheckCounter()
 {
+    if (!m_victim->PAI->IsEngaged())
+    {
+        return m_isCountered;
```
teschnei: i know that m_isCountered is initialized to false.. but can we set it to false again before returning here, just in case someone doesn't change how it works later?

metalfiiish:
You got it boss. Updated it to set false before returning the value, just in case someone changes the default value of m_isCountered.

PR #4390, opened 2017-12-16
Updated Expected Client Version for December 2017.

zynjec:
Expected Client Version wasn't updated with TextID commit.


takhlaq:
can you pr this to stable too

PR #4392, opened 2017-12-17
Adding validation method for handleQuestVar

ShelbyZ:


scripts/globals/quests.lua
```diff
@@ -31,43 +31,63 @@ local check_enum =
...
-            var = quest.vars.main
-            vartype = dsp.quests.enums.var_types.char_var
-        elseif var = quest.vars.additional[varname] then
-            vartype = var.type
-        end
+    local var, vartype;
+    if quest.vars.main == varname then
+        var = quest.vars.main
+        vartype = dsp.quests.enums.var_types.char_var
+    elseif var == quest.vars.additional[varname] then
```
takhlaq: the single = was deliberate (assignment) since we check `var` later

scripts/globals/quests.lua
```diff
@@ -31,43 +31,63 @@ local check_enum =
...
+        print(prefix .. "entity cannot be nil")
+        success = false
+    elseif quest == nil then
+        print(prefix .. "quest cannot be nil")
+        success = false
+    elseif varname == nil or varname == '' then
+        print(prefix .. "varname cannot be nil or empty")
+        success = false
+    end
+    -- todo what is caller?
```
takhlaq: caller was supposed to just be a debug string for any errors, just i forgot to rename it

takhlaq: though actually it can probably fuck off

PR #4393, opened 2017-12-17
Updated onItemCheck and Nation Signet Staves

metalfiiish:
Updated onItemCheck to take into account the caster/item owner for checks such as the signet staves wherein a target can be any party member aligned with the correct nation.


src/map/ai/states/item_state.cpp
```diff
@@ -97,7 +97,7 @@ CItemState::CItemState(CCharEntity* PEntity, uint16 targid, uint8 loc, uint8 slo
         }
         else
         {
-            throw CStateInitException(std::make_unique<CMessageBasicPacket>(m_PEntity, m_PEntity, param, 0, error == -1 ? 0 : error));
+            throw CStateInitException(std::make_unique<CMessageBasicPacket>(m_PEntity, PTarget, param, 0, error == -1 ? 0 : error));
```
teschnei: we did something similar in magic_state, but it would cause a crash because sometimes PTarget was null - can you do PTarget ? PTarget : m_PEntity instead of just PTarget?

src/map/lua/luautils.cpp
```diff
@@ -1933,12 +1933,14 @@ namespace luautils
             return 56;
         }
 
-        CLuaBaseEntity LuaBaseEntity(PTarget);
-        Lunar<CLuaBaseEntity>::push(LuaHandle, &LuaBaseEntity);
+        CLuaBaseEntity LuaBaseEntityTarget(PTarget);
+        CLuaBaseEntity LuaBaseEntityCaster(PCaster);
+        Lunar<CLuaBaseEntity>::push(LuaHandle, &LuaBaseEntityTarget);
+        Lunar<CLuaBaseEntity>::push(LuaHandle, &LuaBaseEntityCaster);
```
teschnei: problem: the lua order used to be (caster, param) but now it's (caster, target, param) - any script that used param will break unless you updated them all (even then, we'd be breaking custom scripts too)

So we should probably just do (caster, param, target) to keep all old scripts working

teschnei: On that note, if you wanted to keep it in the same order as the C++ signature, you could put PCaster after param and have a default value of nullptr

teschnei:
I didn't review the Lua, Teo can take a quick look at it though

TeoTwawki:
If we're actually going to add the user to the function the checks look ok for that, I guess. Don't really care for that but I don't have a better suggestion either so it is what it is.

scripts/globals/items/federation_signet_staff.lua
```diff
@@ -4,9 +4,29 @@
...
 
-function onItemCheck(target)
+function onItemCheck(target, caster)
+   
+    if (target:getNation() ~= NATION_WINDURST) then
+        return msgBasic.ITEM_CANNOT_USE_ON
+    end
+    
+    -- If target's current region is not a conquest region or not a nation city involved with conquest
+    if (target:getCurrentRegion() > REGION_JEUNO) then
```
TeoTwawki: I am not sure, but retail might actually let you waste the charge and give yourself signet in an area it doesn't do anything with (nothings tops you from traipsing around ToAH with signet on, you just don't get CP/xtals/etc there)

src/map/ai/states/item_state.cpp
```diff
@@ -85,7 +85,7 @@ CItemState::CItemState(CCharEntity* PEntity, uint16 targid, uint8 loc, uint8 slo
         throw CStateInitException(std::move(m_errorMsg));
```
takhlaq: teschnei 3 hours ago Owner
@teschnei 
>we did something similar in magic_state, but it would cause a crash because sometimes PTarget was null - can you do PTarget ? PTarget : m_PEntity instead of just PTarget?

see line above, github wont let me mark it for comment

takhlaq: wew guess that was shit also, what i meant to point to was 
https://github.com/metalfiiish/darkstar/blob/8dd93a4a4686265f36202ebae5c1e1befb03fd8e/src/map/ai/states/item_state.cpp#L83

src/map/lua/luautils.cpp
metalfiiish: Originally it was only taking Target, param. In my head I had added caster as the third option already, so now I am super confused.

scripts/globals/items/federation_signet_staff.lua
metalfiiish: I was just going off wiki that says you can only use it in areas with conquest/regions.
http://ffxiclopedia.wikia.com/wiki/Kingdom_Signet_Staff

If we get retail proof that ffxiclopedia is wrong I'm down to remove that.

src/map/ai/states/item_state.cpp
metalfiiish: So shouldn't I just do what Lotus said and make it do an or 'PTarget ? PTarget : m_PEntity instead of just PTarget'?

src/map/lua/luautils.cpp
teschnei: i mixed up caster and target...

it used to be (target,param) but now it's (target,caster,param).  we can't do that because an existing script will try to use the 2nd argument as "param" but it got moved to the third argument, so they will fail if they try to use it
so we should make it (target,param,caster)

metalfiiish:
Updated to use the recommended changes.

src/map/lua/luautils.cpp
metalfiiish: Also, since param is an optional field if I leave it where it was and add my new caster to the end, it requires manually calling param or it gets confused and says it tried to convert basentity to integer and freaks out.

scripts/globals/items/federation_signet_staff.lua
TeoTwawki: "I got  source for it teo <link>" 

is sufficient

metalfiiish: What about <3 you teo whilst I show you some leg :p

TeoTwawki:
@teschnei @metalfiiish
if you both want I could do a mass replace on the luas to make it caster,target,params
the 1st 2 will always exist even if caster will go 99% unused. let me know.

src/map/lua/luautils.cpp
metalfiiish: Can you find a single script that actually uses the param though? I was worried of that too but found none with a git grep and a regular grep against the scripts folder. Like I said, I only found one thing using that params and it was inside a c++.

metalfiiish:
Can you find one example of a script using the params? I couldn't lol...

TeoTwawki:
lol ill laugh so hard if there were none cant look easily from mobile

src/map/lua/luautils.cpp
teschnei: the Lua param order, not the c++ param order

src/map/ai/states/item_state.cpp
takhlaq: no cause PTarget would never be null where he told you to change it since the state throws an exception if it is in the line i linked

metalfiiish: So I am a bit lost now on who's changes to do to get this PR slated. A cliff notes for what you two want for me to do exactly?

takhlaq: you can ignore the stuff about PTarget, but would be best to move
`Lunar<CLuaBaseEntity>::push(LuaHandle, &LuaBaseEntityCaster);`
to after `lua_pushinteger(L, param);` so we dont need to run any mass replaces

metalfiiish:
Updated the PR to put the original ordering of params and allow caster to be null.

PR #4394, opened 2017-12-17
BCNM content, step 2: fill in missing bcnm mob coordinates

wrenffxi:
Mob coordinates for first instance were provided by Brierre.  I infilled the coordinates for 2nd and 3rd instances, by ~~using average coordinate offsets from known battlefields in same zone~~ adjusting 1st instance coords by the entrance deltas.


wrenffxi:
PLEASE HOLD REVIEW.  I realize that coordinates in known battlefields have some weird outliers that are throwing off the averages.  I'll find the correct offsets from client.

TeoTwawki:
position should be calculable. each "copy" of the battlefield is a known precise distance form each other, so if you have an exact pos of a mob in one of them you have the other 2 as well, by offsetting it.

teschnei:
What I did for CoP8-4 was check the offset from my position when entering
the bcnm, and then put the other mobs at the offset of the position from
entering the other bcnms (just using @where or whatever after fiddling with
the entry CS to send me to the other bcnms)

wrenffxi:
Right.  I figured I could bypass all the calculate-from-client CS stuff by using already-coded battlefields, on the assumption that previous mob offsets were calculated the same way.  But I'm seeing a lot of offset variability in mob_spawn_points for already-implemented battlefields, which means a lot of current records in mob_spawn_points need fixing.

teschnei:
Yep, when I did CoP8-4 they were already in the db but the positions were
slightly off or they were facing the wrong way.. so I redid them

TeoTwawki:
it's likely because our `!where` output is imprecise (drops off decimal places, **and** the client will adjust your vertical placement) and ppl didn't double check their positional. When I verify pos I turn off wall hack and `!pos` to the output to see how far off it was. But if you have 1 "known good" pos for battlefield instacne1, then no. 2 and 3 are able to be mathed from that without it. if that 1st post is slightly off though..mobs in floor and crap..

takhlaq:
>(just using `@where` or whatever after fiddling with the entry CS to send me to the other bcnms)
> `@where`
> `@`

TeoTwawki:
image

lol

wrenffxi:
OK, that should do it.  I found all the coordinate offsets via battlefield warp event, and used them to offset all the 2nd and 3rd instance mob coordinates, both old and new.  Ready for review.

wrenffxi:
If anyone wants the data,

[The raw entrance coords](https://pastebin.com/raw/ZuZb1EQD)

[The 2nd and 3rd instance offsets from 1st instance coordinates](https://pastebin.com/raw/TpvwBE9j)

PR #4396, opened 2017-12-17
Small 2hr/NM fixes

ymkc:
- Mobs with Manafont active should cast non-stop while effect is up
- Added a check to phOnDespawn to prevent PHs from disappering when having luck on the rng on multi-ph nms


scripts/globals/mobs.lua
```diff
@@ -59,14 +59,16 @@ function phOnDespawn(ph,phList,chance,cooldown)
         local nm = GetMobByID(nmId);
         if (nm ~= nil) then
             local pop = nm:getLocalVar("pop");
-            if (os.time() > pop and not lotteryPrimed(phList) and math.random(100) <= chance) then
+            local flag = nm:getLocalVar("flag");
```
TeoTwawki: could we call that something more descriptive?

TeoTwawki:
Small not really important note: since retail changed the old 2hrs to being 1 hour..I've been trying to call them "job special" instead. hence "jobSpec" table. Been naming the local vars like "hpTrigger" or "benedictionTrigger" and such to avoid time references on non time based things. I'll still merge without that it is not a problem, just hoping to get other opinions and get us all on some kind of consensus on what to call stuff in the future and then we can get all the script on the same style later.

scripts/globals/mobs.lua
ymkc: trying hard ... :)

ymkc:
@teotwawki hate to think of var names anyway, so will use whatever will be decided on

scripts/globals/mobs.lua
```diff
@@ -59,14 +59,16 @@ function phOnDespawn(ph,phList,chance,cooldown)
         local nm = GetMobByID(nmId);
         if (nm ~= nil) then
             local pop = nm:getLocalVar("pop");
-            if (os.time() > pop and not lotteryPrimed(phList) and math.random(100) <= chance) then
+            local swapped = nm:getLocalVar("swapped");
```
wrenffxi: what's the point of this swapped localVar? the lotteryPrimed() function already tells us if a lottery NM is already set to spawn in place of its PH.

ymkc: oh i see, it's done via nm:getRespawnTime
@wrenffxi totally didn't see it, thanx

PR #4402, opened 2017-12-19
changed PW skin to be correct model and Deverger tp moves

dwn134:


scripts/zones/Aydeewa_Subterrane/mobs/Pandemonium_Warden.lua
```diff
@@ -64,7 +64,7 @@ function onMobFight(mob,target)
 
     ------------------------ FORM CHANGES ------------------------
     if (mobHPP <= 15 and change == 13) then -- Final Form, pets take Dvger form as well
-        mob:setModelId(1839);
+        mob:setModelId(1840);
```
TeoTwawki: @dwn134 I just severely F'd up reviewing this

```lua
    if (mobHPP <= 15 and change == 13) then -- Final Form, pets take Dvger form as well
        mob:setModelId(1840);
        mob:setLocalVar("change", 14);
        for i = 1, 8 do
            if petStatus[i] == 0 then
                SpawnMob(petIDs[i]):updateEnmity(target);
            end
            GetMobByID(petIDs[i]):setModelId(1840);
end
```
By merging your pr without noticing first, we just made warned and lamps have same model and same tp list restrictions. Would you mind making a second pr adjusting the pets?

TeoTwawki: I think what happened was before you change dit, they had their model ID's flipped, so pets prolly needed the value you changed warden *from*...

dwn134:
I have a complete rework of pw made. If you want i can push the full pw script and then I still have phase change abilites to recode. I have a full working pw and pets made and was just going to submit a phase at a time

TeoTwawki:
we can do that later, I'd like to break as little existing stuff as possible along the way. I'll just see if I can go find and swap the borked pet value here.

TeoTwawki:
#4403

PR #4403, opened 2017-12-19
Fix Pandemonium Lamp skin IDs..TODO: use skillList changes, instead of model ID checks, and NUKE duplicate skill scripts!

TeoTwawki:
```
TODO: use skillList changes, in stead of model ID checks, and NUKE duplicate skill scripts!
```

@dwn134 so, the existing cruddy PW implementation was written back when family ID was the skillList ID, and when we were allowed to make multiple differently named skills with the same ID. The current system requires the skill ID's to be unique (which I hate, should that table joined on name instead of ID) but anyway thought I'd tag you here to point out the todo so you are aware of it in that version you want to PR

### DIFF
--- a/scripts/globals/mobskills/hellclap.lua
+++ b/scripts/globals/mobskills/hellclap.lua
@@ -11,7 +11,7 @@ function onMobSkillCheck(target,mob,skill)
   if(mob:getFamily() == 91) then
     local mobSkin = mob:getModelId();
 
-    if (mobSkin == 1840) then
+    if (mobSkin == 1839) then
         return 0;
     else
         return 1;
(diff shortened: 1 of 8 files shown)


dwn134:
I'm not exactly sure how you mean to use skillList changes. On my current version I have skills set to use as a var that looks at phase changes. And removed all the abilites that are listed as PW_Skill but was changing them to modelid since that's how you have it currently. Able to give an example of what you mean please?

TeoTwawki:
I mean we can actually set a mobs skill list on the fly now.
https://github.com/DarkstarProject/darkstar/blob/5078b8b1b71c30af4a5fbcb9075787e5a9a3cd3f/scripts/zones/Abyssea-Konschtat/mobs/Balaur.lua#L36

TeoTwawki:
-nudge-
somebody wanna merge so my mistake can be undone?

PR #4404, opened 2017-12-20
Pandemonium Warden Changes

dwn134:
will be cleaning up all the mobskills code next as we wont be needing the familyid checks anymore for this mob


scripts/globals/mobskills/brainjack.lua
```diff
@@ -8,9 +8,10 @@ require("scripts/globals/monstertpmoves");
 ---------------------------------------------------
 
 function onMobSkillCheck(target,mob,skill)
-    local mobSkin = mob:getModelId();
+    -- See PW's script
+    local phase = mob:getLocalVar("phase");
 
-    if (mobSkin == 1825) then
+    if (phase == 2) then
```
TeoTwawki: Just a head sup about a few things:

I know the other mobs that uses these skills aren't in the project yet, but this is one of the reasons it needs to be handled as skill list swaps instead of a localvar or the previous modelId checks. 

its how the skill wound up with that dumb prefix_ because it used to be 2 scripts and when that changed one of them got deleted (since PW was "implemented" at the time its version is the one that stayed). This Pr is likely going to get conflicts with 2 others that are open. One of which is waiting on the BCNM branch I think. This one may wind up waiting awhile and then need some conflict resolution love to fix.

dwn134:
I only have the phase check on the chariots because he uses that form 4 times and can only use 1 to move per phase. I figured atm he's the only mob using these moves so ok. All the other moves in the game I can remove the ID checks

dwn134:
I can set the 4 chariot skills as separate mobskillsets in the database so it only references the right chariot move per phase if you want to go that route. Have any request for the name of the skillset and id number?

dwn134:
Ok I've made the chariot phases have different skillList ids and removed the modelid checks from the mobskills. Also changed the modelid to go with which chariot style you fight

TeoTwawki:
> Have any request for the name of the skillset and id number?

anything telling us what it is at a glance will do fine <3

ID can use any gap ppl may have created by deleting past ones, or else just +1 the highest list ID we have.

PR #4407, opened 2017-12-20
smithing recipes

wrenffxi:


TeoTwawki:
Has references to sources lookin very legit, my only concern is the gaps created by some removals. Contributors have a bad tendency to not notice these to fill them back in with new entries later, just +1 'ing whatever is at the very bottom. I was gonna ask you to add a comment so they'd know this, but it also looks you you added new things. The synth ID is kinda arbitrary, so maybe you'd like to just fill them in now?

And I have no idea wtf was up with those enfeebling kits.
```sql
--- INSERT INTO `synth_recipes` VALUES (3472,2,0,0,0,0,0,0,0,0,0,4099,4241,2773,2773,2774,2774,2775,4162,0,0,2782,2782,2782,2782,1,1,1,1); -- enfeeblement_kit_of_silence (dupe of 3533 with type=2?)```

I'm tempted to put a completely unused string column in just to preserve some human readable "wtf is this" that survives redumps.

wrenffxi:
I've got one more set of recipes to go (woodworking is the last batch).  when I do the woodworking PR I plan to then do a commit that sorts the table and removes the gaps.

TeoTwawki:
Alright, I'll hold you to that!

PR #4408, opened 2017-12-21
BCNM content, step 3: add lua scripts for bcnm and mobs

wrenffxi:
job_special mixin for basic mob 2hr handling
COMBAT_TICK and ROAM_TICK triggers (thanks shelbyz)
LUAs for all missing bcnms
LUAs for some missing bcnm mobs

**note that the new bcnms are not yet complete** -- they still need loot, and the mobs need more SQL inserts (groups, immunities, skill and spell lists, etc.)  Those will be in steps 4 and 5, to follow. :)

### DIFF
--- a/documentation/AI_Events.txt
+++ b/documentation/AI_Events.txt
@@ -31,6 +31,7 @@ WEATHER_CHANGE - Entity, weather, weather element
 EXPERIENCE_POINTS - Entity, exp
 
 TICK - Entity
-EFFECTS_TICK - Entity
+COMBAT_TICK - Mob
+ROAM_TICK - Mob
 EFFECT_GAIN - Entity, Effect
 EFFECT_LOSE - Entity, Effect
--- a/documentation/AI_Events.txt
+++ b/documentation/AI_Events.txt
@@ -31,7 +31,9 @@ WEATHER_CHANGE - Entity, weather, weather element
 EXPERIENCE_POINTS - Entity, exp
 
 TICK - Entity
-COMBAT_TICK - Mob
-ROAM_TICK - Mob
+EFFECTS_TICK - Entity
 EFFECT_GAIN - Entity, Effect
 EFFECT_LOSE - Entity, Effect
+
+COMBAT_TICK - Mob
+ROAM_TICK - Mob
--- a/scripts/mixins/job_special.lua
+++ b/scripts/mixins/job_special.lua
@@ -1,0 +1,58 @@
+-- for mobs with job-specific special ability
+
+require("scripts/globals/mixins");
+require("scripts/globals/status");
+
+g_mixins = g_mixins or {};
+
+g_mixins.job_special = function(mob)
+
+    mob:addListener("SPAWN", "JOB_SPECIAL_SPAWN", function(mob)
+        mob:setLocalVar("useMainSpecAtHPP", math.random(40,60));
+    end);
+    
+    mob:addListener("COMBAT_TICK", "JOB_SPECIAL_CTICK", function(mob)
+        local defaultAbility = {
+            [JOBS.WAR] = jobSpec.MIGHTY_STRIKES,
+            [JOBS.MNK] = jobSpec.HUNDRED_FISTS,
+            [JOBS.WHM] = jobSpec.BENEDICTION,
+            [JOBS.BLM] = jobSpec.MANAFONT,
+            [JOBS.RDM] = jobSpec.CHAINSPELL,
+            [JOBS.THF] = jobSpec.PERFECT_DODGE,
+            [JOBS.PLD] = jobSpec.INVINCIBLE,
+            [JOBS.DRK] = jobSpec.BLOOD_WEAPON,
+            [JOBS.BST] = jobSpec.FAMILIAR,
+            [JOBS.BRD] = jobSpec.SOUL_VOICE,
+            [JOBS.SAM] = jobSpec.MEIKYO_SHISUI,
+            [JOBS.NIN] = jobSpec.MIJIN_GAKURE,
+            [JOBS.DRG] = jobSpec.CALL_WYVERN,
+            [JOBS.SMN] = jobSpec.ASTRAL_FLOW,
+        };
+
+        if (mob:getLocalVar("usedMainSpec") == 0 and mob:getHPP() < mob:getLocalVar("useMainSpecAtHPP")) then
+            local default = defaultAbility[mob:getMainJob()];
+            local spec = mob:getLocalVar("mainSpec");
+            if (spec > 0) then
+                mob:useMobAbility(spec);
+            elseif (default ~= nil) then
+                mob:useMobAbility(default);
+            end
+            mob:setLocalVar("usedMainSpec", 1);
+        end
+
+        if (mob:getLocalVar("usedSubSpec") == 0 and mob:getHPP() < mob:getLocalVar("useSubSpecAtHPP")) then
+            local default = defaultAbility[mob:getSubJob()];
+            local spec = mob:getLocalVar("subSpec");
+            if (spec > 0) then
+                mob:useMobAbility(spec);
+            elseif (default ~= nil) then
+                mob:useMobAbility(default);
+            end
+            mob:setLocalVar("usedSubSpec", 1);
+        end
+
+    end);
+    
+end
+
+return g_mixins.job_special;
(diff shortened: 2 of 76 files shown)


scripts/mixins/job_special.lua
```diff
@@ -0,0 +1,52 @@
...
+
+        mob:setLocalVar("usedJobSpec", 0);
+        mob:setLocalVar("useJobSpecAtPct", math.random(40,60));
+        local ability = defaultAbility[mob:getMainJob()];
+        if (mob:getLocalVar("jobSpec") == 0 and ability ~= nil) then
+            mob:setLocalVar("jobSpec", ability);
+        end
+    end);
+    
+    mob:addListener("WEAPONSKILL_STATE_ENTER", "JOB_SPECIAL_WS", function(mob, skillID)
```
teschnei: status_effect_tick is probably a better event to listen on (ticks every 3 sec), unless you know that they only 2hour right after using a mob ability

scripts/mixins/job_special.lua
```diff
@@ -0,0 +1,52 @@
...
+            local spec = mob:getLocalVar("jobSpec");
+            if (spec > 0) then
+                mob:useMobAbility(spec);
+                mob:setLocalVar("usedJobSpec", 1);
+            end
+        end
+    end);
+    
+    mob:addListener("DEATH", "JOB_SPECIAL_DEATH", function(mob)
+        mob:setLocalVar("usedJobSpec", 0);
```
teschnei: i'm... 85% sure that local vars get all reset on spawn, so this is probably unnecessary

scripts/mixins/job_special.lua
```diff
@@ -0,0 +1,52 @@
...
+            [JOBS.NIN] = jobSpec.MIJIN_GAKURE,
+            [JOBS.DRG] = jobSpec.CALL_WYVERN,
+            [JOBS.SMN] = jobSpec.ASTRAL_FLOW,
+        };
+
+        mob:setLocalVar("usedJobSpec", 0);
+        mob:setLocalVar("useJobSpecAtPct", math.random(40,60));
+        local ability = defaultAbility[mob:getMainJob()];
+        if (mob:getLocalVar("jobSpec") == 0 and ability ~= nil) then
+            mob:setLocalVar("jobSpec", ability);
```
teschnei: instead of setting this here... I think you should just do nothing in spawn except set the percent, and then in JOB_SPECIAL_WS, check the localvar.  if it's zero, then use the ability in the defaults table, otherwise use the ability in the var.  that way, you can override the ability with whatever you want in the script's OnMobSpawn (note: i am PRETTY SURE that OnMobSpawn runs before the listener, but if it doesn't, that can probably be easily fixed)

scripts/mixins/job_special.lua
wrenffxi: thanks, was trying to figure that out.

teschnei: they should all (hopefully) be listed in documentation/AIEvents.txt

scripts/mixins/job_special.lua
TeoTwawki: they do get wiped.

scripts/mixins/job_special.lua
```diff
@@ -0,0 +1,52 @@
...
+            [JOBS.BRD] = jobSpec.SOUL_VOICE,
+            [JOBS.SAM] = jobSpec.MEIKYO_SHISUI,
+            [JOBS.NIN] = jobSpec.MIJIN_GAKURE,
+            [JOBS.DRG] = jobSpec.CALL_WYVERN,
+            [JOBS.SMN] = jobSpec.ASTRAL_FLOW,
+        };
+
+        mob:setLocalVar("usedJobSpec", 0);
+        mob:setLocalVar("useJobSpecAtPct", math.random(40,60));
+        local ability = defaultAbility[mob:getMainJob()];
```
TeoTwawki: We really do not want to assume a mobs mainjob will decide what 2hr is uses. There are ,many examples of that not being the case.

TeoTwawki: I'd rather we just specify which 2hr we are trying to call as an argument to a function.

teschnei: I think it's fine to have an overrideable default be their main job (as I was saying in my other comment 2 lines down)

TeoTwawki: only saw you mention subjob sorta skimmed over most of your comment to lines down because the bulk of it isn't talking about this. something set the percent something localvar something override something about when the listener runs.

I missed one out of 4, spoke about the only bit that worries me. I disagree about bothering with a default instead of being explicit about what we want to happen but you da boss.

scripts/zones/Balgas_Dais/mobs/Bitoso.lua
```diff
@@ -0,0 +1,22 @@
...
+function onMobSpawn(mob)
+    mob:setMobMod(MOBMOD_MAGIC_COOL, 4);
+end;
+
+function onMobFight (mob,target)
+end
+
+function onMagicHit(caster, target, spell)
+    if (spell:getID() == 356) then
+        spell:setFlag(SPELLFLAG_WIPE_SHADOWS);
```
TeoTwawki: :D 

(I'm just happy to see someone find a use for that outside of virelai)

scripts/mixins/job_special.lua
wrenffxi: Shelbyz showed how to add combat_tick and roam_tick, so I've implemented that here.

scripts/mixins/job_special.lua
wrenffxi: agreed. i made this change below.

scripts/mixins/job_special.lua
wrenffxi: huh. i knew that, and added that code anyway. don't know why.  removed death listener.

scripts/mixins/job_special.lua
wrenffxi: it'll use the default main job ability if one isn't specific in onMobSpawn().  figure that lets us write less code for the most common case, but gives flexibility for specific mobs that use weird 1hrs

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -790,6 +791,7 @@ void CMobController::DoRoamTick(time_point tick)
                 else if (PMob->m_roamFlags & ROAMFLAG_EVENT)
                 {
                     // allow custom event action
+                    PMob->PAI->EventHandler.triggerListener("ROAM_TICK", PMob);
```
teschnei: I'm not 100% sure on this, but I think we only needed the one ROAM_TICK down where OnMobRoam is - I think OnMobRoamAction is in addition to that one.  But.. I'm not 100% sure.  Could you try adding a listener to OnMobRoam to a mob with ROAMFLAG_EVENT, and see if it prints more than once per 3sec?

teschnei: And if it does, remove this one, and make sure that the same listener still prints once every 3 sec

documentation/AI_Events.txt
```diff
@@ -31,6 +31,7 @@ WEATHER_CHANGE - Entity, weather, weather element
 EXPERIENCE_POINTS - Entity, exp
 
 TICK - Entity
-EFFECTS_TICK - Entity
+COMBAT_TICK - Mob
```
teschnei: ahh don't delete EFFECTS_TICK, as I put it back in yesterday

wrenffxi: done

src/map/ai/controllers/mob_controller.cpp
wrenffxi: that went a little over my head. paging @ShelbyZ for backup

ShelbyZ: Remove this line (794), so we avoid firing ROAM_TICK twice for mobs who have ROAMFLAG_EVENT (bcnm, etc) and it should be fine.

wrenffxi: done

wrenffxi:
Done with new commits to this PR.  All yours for review.

TeoTwawki:
teschnei will prolly get to it after the 1st or 2nd 👍

wrenffxi:
rebased

TeoTwawki:
@takhlaq can this be merged without causing you probs in the other branch? I'd like to get it done so we can start using the mixin included here.

PR #4417, opened 2017-12-22
Abyssea-La_Theine/Mob Data Retail

dragonarmor:
Tell me what you guys think and I will push more abyssea zones.
Mob HP was pulled from retail using Scoreboard.
Respawn times are pulled from retail.
Pop item mobs seem to be about 3 mins respawn.
Timed spawned Nm's 15 min respawn.
All other normal mobs 5 mins.
Min and Max levels were pulled from ffxiclopedia wiki. NM's levels are a guess. up for suggestions on those.
MP not real sure what to do with that one.


TeoTwawki:
Don't set HP this way. the field for hp and mp in that table to overrides for situations like NMs that have more hp/mp than a mob of their level normally would. Mob HP/MP is normally calculated form it's family and job(s)

In the case of abyssea mobs, there is an added factor of them spawning in as higher level if their previous spawn was killed more quickly, and down-leveling back to their base when idle for too long. We don't have data for the time increments or level increments involved. (dear folks with retail access, experiment with this please, and I can demonstrate how to script it so yall can gitrdone as ya please)

If you get us an accurate base level for these that's merge-able.

dragonarmor:
ok misunderstood the table structure. Seems like there will be a lot of guess work. I cant see someone spending the time to gather that kind of data.

TeoTwawki:
btw, pop timing you listed in your first comment is *probably* spot on.

ibm2431:
@dragonarmor : I did a recapture for all Abyssea zones last year. During, I captured widescan packets which send out mob levels. You can find my recaptures here:

https://drive.google.com/open?id=0B6J1HCKIjYoTaEFnQzVHemYxWWM

Widescan data is in the widescan folder, obviously.

This capture was done on Leviathan, at times it was _more_ dead than usual. The only soul I ever saw was someone cleaving the Skeletons and Bugards in Grauberg to pop Azdaja. And maybe someone working on pops for Apademak (Snolls / Marids). Therefore, other than those, I'd consider it safe to assume these levels are the lowest that the mobs will delevel/despawn to.

Levels for NMs are sent out in the packet as well. Those are included as well.

Apologies for the data not being sorted; the files were written to as the packets came in. You can probably sort them first to make your life easier.

dragonarmor:
@ibm2431 : Awesome information thanks for that. there was literally no one in the Abyssea-La_Theine when i captured the HP data. I always took the first kill data for each mob so i think it's safe to say we can assume your level data is the lowest and should match up with my HP data. So would this suffice for an accurate base level?? and if this does not go into the mob_groups table, where do we apply the HP data?

TeoTwawki:
HP is auto-calculated by mobs job+family like all its other stats. It's per-level,

You know how you stats change based on your job but your race sets up basic stats like taru get more int than galka? Think of the mob_family_table as the monsters race data.

dragonarmor:
@TeoTwawki : I see how that works, but what I'm missing is I can one shot all those mobs. Why is their hp so low? I must be missing the obvious.

TeoTwawki:
you can one shot them because their level is not there yet to calculate HP from. 

```sql
INSERT INTO `mob_family_system` VALUES (5,'Amoeban',21,'Voragean',0,40,100,120,3,3,3,3,3,3,3,3,3,3,3,1,1,1,1,1,1,1,1,1,1,1,1,0,32,0);
```

lv0 x 100hp = 0 maxhp

dragonarmor:
Thanks for the script! let me have look with that.

PR #4420, opened 2017-12-24
Core: Cross-Server Warping to Entities

Cryptexx:
Added functionality for warping to NPC's and Mobs across game servers using different ports based on the gotoPlayer functionality. If entity is not spawned (status = STATUS_DISAPPEAR), then you will be warped to the default spawn position as indicated in mob_spawn_points.sql. Tested with over 20 Mobs and NPC's, both spawned and not spawned.

Usage: player:GoToEntity(16785735)


Cryptexx:
P.S. in src/login/message_server.cpp it looks like I changed a lot of stuff, but I really didn't. I just tabbed over one space to properly indent the switch statement. I also added a couple line breaks for visual ease and redid the spacing on two or three queries...plus remove the \ character in favor of double quotes on each line as a method of statement continuation. That seems to be the common theme among all our queries, so standardized it according to the norm.

src/map/lua/lua_baseentity.cpp
```diff
@@ -2732,6 +2732,55 @@ inline int32 CLuaBaseEntity::resetPlayer(lua_State *L)
...
+        ref<bool>  (&buf,  0) = true; // Toggle for message routing; goes to entity server first
+        ref<bool>  (&buf,  1) = spawnedOnly; // Specification for Spawned Only or Any 
+        ref<uint16>(&buf,  2) = targetZone;
+        ref<uint16>(&buf,  4) = playerZone;
+        ref<uint32>(&buf,  6) = targetID;
+        ref<uint16>(&buf, 10) = playerID;
+
+        message::send(MSG_SEND_TO_ENTITY, &buf[0], sizeof(buf), nullptr);
+
+        if (PChar->status != STATUS_DISAPPEAR)
```
teschnei: this is technically a race condition: message::send does not block, so you have no way of knowing whether the message went through yet.  It's probably better to just return nothing since it can't realistically be able to determine if you are warping or not

teschnei:
other than that return value, it looks good and i'll merge it after that's fixed

PR #4423, opened 2017-12-26
item_basic updates for dec client

Ninjistix:
some minor adjustments and new items for dec client version


TeoTwawki:
thanks ninji

PR #4426, opened 2017-12-27
Garrison Framework

metalfiiish:
What is not done:
-NPC's for battle are not implemented. Would require 18 groupids per garrison zone or proper refactoring of how CORE handles trusts/dynamic npcs used with players.
-Only West Ronfaure Garrison Output Guard is setup. Simple to copy paste to all other Garrison zone Outpost Guards once the Battle NPC's are built out.
-Set mob Positions per garrison zone. Will get to this shortly.


scripts/globals/msg.lua
```diff
@@ -216,6 +216,7 @@ msgBasic =
     ADD_EFFECT_HP_HEAL     = 167, -- Additional effect: The <player> recovers <number> HP.
     ADD_EFFECT_DISPEL      = 168, -- Additional effect: <target>'s <Status Effect> effect disappears!
     ADD_EFFECT_WARP        = 169, -- Additional effect: Warp! (used by Halloween staves)
+    LEVEL_RESTRICTED       = 314; -- <target>'s level is currently restricted to 0. Equipment affected by the level restriction will be adjusted accordingly.
```
ShelbyZ: Shouldn't this comment indicate that it is restricted to <number> so that it is apparent that passing a value with alter the message.

scripts/zones/West_Ronfaure/npcs/Doladepaiton_RK.lua
```diff
@@ -9,27 +9,80 @@ package.loaded["scripts/zones/West_Ronfaure/TextIDs"] = nil;
...
+-----------------------------------
+function onSpawn(npc)
+    npc:setLocalVar("GarrisonState", IDLE_GARRISON)
+end
+
 -----------------------------------
 -- onTrade Action
 -----------------------------------
 
 function onTrade(player,npc,trade)
```
ShelbyZ: You might want to comment out the print statements to avoid filling the console with output or alternatively have method/variable in garrison.lua that defines a logging method and the server operator can toggle that value to debug garrison.

```lua
--garrison.lua

LOG_GARRISON = false

function garrisonPrint(msg)
  if LOG_GARRISON then
    print(msg)
  end
end
```

scripts/globals/msg.lua
metalfiiish: Sure can. I had copy pasted it straight from message.log but it makes sense to modify it like you said.

scripts/zones/West_Ronfaure/npcs/Doladepaiton_RK.lua
metalfiiish: The print statement were only temporary, mainly pushed this up for scrutiny on the framework, but I will make sure they are all gone by the time we get it prepared for merging.

scripts/globals/garrison.lua
```diff
@@ -0,0 +1,530 @@
...
+--* Ensure SQL flag for confrontation is removed on zone, can update similar to this (allows confrontation to wipe on zoning):
+--        update status_effects set flags = flags | integer_value_for_EFFECTFLAG_ON_ZONE where statusid = integer_value_for_EFFECT_BATTLEFIELD;
+--* Once NPC allies are in change updateClaim to attack NPC's instead of player
+--* Update ALL mobs pos in sql
+--* Update instantiateMob to return newMob
+--* Build out other conquest npc scripts per region OP npc
+
+
+-->>>>UPDATE Mob scripts
+--setDropID(4596)
```
Hozu: Do Garrison mobs drop seals and/or crystals? If not, you're going to have to add a mobmod for that (can't use the no drops one because you have mannequin parts) here:

https://github.com/DarkstarProject/darkstar/blob/9b5036e877bd6a28c291b2b0e895997af2ea113c/src/map/entities/mobentity.cpp#L859

and a couple of lines above that for crystals.

metalfiiish: Not sure about crystals but I was able to get a seal in retail the other day during testing.

takhlaq:
closing for now, tag one of us when it's ready to be reopened and reviewed

PR #4428, opened 2017-12-29
tension spring attachments adding rattp on unequip/maneuverlose

wrenffxi:


wrenffxi:
Yup. This is unrelated to the broader issue of onManeuverLose() not properly accounting for buffs previously given in onManeuverGain(), but it's a separate issue that exacerbates the problem with this attachment in particular.  Thanks Aizen for the heads up.

Issue #3804, opened 2017-03-14
NPC / Quest Trading issue


whateversclever1215:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30161227_1

**_Server Version_** (type `@revision` in game) **:**
Revision is: Unknown (ERA)

**_Source Branch_** (master/stable) **:**

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

I ran into this issue while raising fame on an alt.
I was doing the quest:
http://ffxiclopedia.wikia.com/wiki/Only_the_Best
involving the NPC Melyon in Selbina but I don't believe the quest or npc are important details.

I have capped fame on many many many alts on retail, so I am aware of how trading items to NPCs works.

On retail, when you trade a stack of 12(or really anything beyond the 3 millioncorn Melyon asks for), what happens is it will take 3 out of the stack, finish the quest, and the rest will just still be in your bag.

On era, when you trade a stack of 12, or any number except for the exact 3, the trade simply fails to go through.

This is another issue that is extremely unlikely to affect people too badly, but it is not working like retail, and this may give clues to any other issues people have so I am reporting it.

teschnei:
It's an issue with the way the trades are handled in scripts. We do have
the functionality to do it correctly (used in like.. two places), but every
single script that uses trades would have to be rewritten to use it

whateversclever1215:
So there's trading behavior different from retail and a bunch of people using custom trading scripts to take advantage of the difference? ok then, I didn't know scripting trades was possible in the first place, maybe this change made that possible?

teschnei:
Sorry, the problem is in our NPC scripts with how they handle the trade.  When you trade an NPC, the NPC only checks exactly what you have placed into the trade container.  We do have some Lua bindings to support the proper way of doing it (I wrote them), but most NPCs don't use those bindings.  So fixing this would mean rewriting the Lua script for every NPC that accepts trades (ie. would take a long time).

whateversclever1215:
Oh, so just not worth it. Got you.
I was only ticketing this in the event it gave insight into an unsolved issue.
I will close, this isn't an issue really.

teschnei:
[...]

PR #4431, opened 2017-12-29
Fixes related to npc_utils, trade:confirmItem, and player:confirmTrade

wrenffxi:
Fixed an issue where npcUtil.completeQuest() would crash server when given an area without a corresponding fame area.

Converted several bastok markets quest NPCS to use npc_util.lua

### DIFF
--- a/scripts/globals/npc_util.lua
+++ b/scripts/globals/npc_util.lua
@@ -212,7 +212,8 @@ end
         { {640,2}, 641 }        -- copper ore x2, tin ore x1
         { 640, {"gil", 200} }   -- copper ore x1, gil x200
 ******************************************************************************* --]]
-function npcUtil.tradeHas(trade, items)
+function npcUtil.tradeHas(trade, items, exact)
+    if (type(exact) ~= "boolean") then exact = false; end
 
     -- create table of traded items, with key/val of itemId/itemQty
     local tradedItems = {};
@@ -253,14 +254,47 @@ function npcUtil.tradeHas(trade, items)
         return false;
     end
 
-    -- determine whether all needed items have been traded, and confirm those needed. return true or false.
+    -- determine whether all needed items have been traded. return false if not.
     for k, v in pairs(neededItems) do
         local tradedQty = (tradedItems[k] == nil) and 0 or tradedItems[k];
-        if (v > tradedQty or not trade:confirmItem(k,v)) then
+        if (v > tradedQty) then
             return false;
+        else
+            tradedItems[k] = tradedQty - v;
         end
     end
+    
+    -- if an exact trade was requested, check if any excess items were traded. if so, return false.
+    if (exact) then
+        for k, v in pairs(tradedItems) do
+            if (v > 0) then
+                return false;
+            end
+        end
+    end
+    
+    -- confirm items
+    for k, v in pairs(neededItems) do
+        trade:confirmItem(k,v)
+    end
     return true;
+end
+
+--[[ *******************************************************************************
+    check whether trade has exactly required items
+        if yes, confirm all the items and return true
+        if no, return false
+        
+    valid examples of items:
+        640                     -- copper ore x1
+        { 640, 641 }            -- copper ore x1, tin ore x1
+        { 640, 640 }            -- copper ore x2
+        { {640,2} }             -- copper ore x2
+        { {640,2}, 641 }        -- copper ore x2, tin ore x1
+        { 640, {"gil", 200} }   -- copper ore x1, gil x200
+******************************************************************************* --]]
+function npcUtil.tradeHasExactly(trade, items)
+    return npcUtil.tradeHas(trade, items, true);
 end
 
 function npcUtil.genTmask(player,title)
--- a/scripts/globals/npc_util.lua
+++ b/scripts/globals/npc_util.lua
@@ -1,7 +1,18 @@
+--[[
+    Helper functions for common NPC tasks.
+    
+    npcUtil.pickNewPosition(npc, positionTable, allowCurrentPosition)
+    npcUtil.giveItem(player, items)
+    npcUtil.giveKeyItem(player, keyitems)
+    npcUtil.completeQuest(player, area, quest, options)
+    npcUtil.tradeHas(trade, items)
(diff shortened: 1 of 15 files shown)


TeoTwawki:
I hate the f out of half of npc_utils.lua

pretty much everything in it related to trades, quests, and missions. All needs to die in a fire. Please, don't convert things over to it. Its a crap wrapper on top of a crap way of handling these things. teschnei put in a better method for checking trades and we haven't really started making use of it yet.

TeoTwawki:
anyone that wants to ask me "whats wrong with x teo" hit me up on discord..

TeoTwawki:
re: latest revision.
thats a good deal better actually, thank you.

wrenffxi:
Following a discord convo with Teo last night, I added a new commit to this PR.  Most importantly, it fixes a couple issues for confirming trades.  Then I redid the npcUtil functions to use these new bindings, and updated existing scripts that use npcUtil.

TeoTwawki:
when teschnei has time to look, if he says its good its good 👍

ibm2431:
What's the behavior for when excess items the NPC didn't ask for are included in the window? Does the entire trade get rejected, or does it go through and the NPC only takes the items asked?

Retail will sometimes accept the trade, but other times reject it. I gave [a few examples](https://github.com/DarkstarProject/darkstar/issues/3804#issuecomment-286730720) when trading was brought up in #3804

Retail does rejections when players have "options". For example, if the trade-with-excess-items is accepted by DSP (even if it only "takes" items listed in the tradeHas check)...

When upgrading some piece of AF, our code might look something like:
```
if tradeHas(BLM_feet, feet_upgrade_item)
  giveItem(BLM_feet_plus_1)
elseif tradeHas(WHM_feet, feet_upgrade_item)
  giveItem(WHM_feet_plus_1)
```

But if the player traded their BLM feet, the upgrade item, and accidentally includes their WHM feet in the same window, what should happen? Retail will deny the trade outright, so the player doesn't accidentally get the wrong reward. From what I can tell with these changes (and apologies if I'm wrong, it's late) it looks like DSP would accept the trade + take items, and just ignore "excess" items. This could lead to undesirable behavior / outcomes if done by default.

I don't know how best to handle accepting / taking. Sorry for raising the issue without suggesting a solution.

wrenffxi:
if you use npcUtil.tradeHas(), it flags the required gil and items (only) as "to be traded".  If you then player:confirmTrade(), it will only remove those items from the player, and return any excess ones.

If you want ability to reject a trade that doesn't have exact items, that would be easy to add.  Could either add an optional "mustBeExact" parameter to npcUtil.tradeHas(), or could write a new npcUtil.tradeHasExactly() or something.  Lemme know.

TeoTwawki:
there is something slap-myself level important I forgot about. remember how I was saying on discord that I forget the details of why originally don't like a thing, then wind up looking again and just seeing other things that are also bad? I remembered enough to worry about how new ppl might mistakenly use stuff, and not *why*. Now I remembered why. We normally begin trades in onTrade then finish them in onEventFinish (75 to 90% will involve an event). Those ones that involve an event we don't want the trade finishing before the event, it leads to exploitable scenarios and or lost items. Eeep.

*dons his duncecap*

ibm2431:
Personally, I'd lean towards tradeHasExactly() over a passed parameter. When "written as English" it's easier to say/understand versus an extra param:

`tradeHas({BLM Feet, Upgrade Item}, mustBeExact)` == "Trade has BLM Feet and Upgrade Item, but the trade must be exact"

`tradeHasExactly(BLM Feet, Upgrade Item)` == "Trade has exactly BLM Feet and Upgrade Item"

Also, having it as a "separate" function (even if implementation ends up using 'base' tradeHas() under the hood) places both behaviors (accept-ignoring-excess and reject-because-excess) as being "equal", with no "default" behavior that someone needs to always remember.

tradeHas() and tradeHasExactly() would be used in a given situation as you'd think about it in English, versus the optional param implementation resulting in, "Okay, I need to check this trade for items, so I need to use tradeHas(). Which behavior does it do by default, and what was that optional parameter to switch it?"

scripts/globals/npc_util.lua
```diff
@@ -38,175 +49,254 @@ function npcUtil.pickNewPosition(npc, positionTable, allowCurrentPosition)
...
+            gil = 200,
+            xp = 1000,
+            title = ENTRANCE_DENIED,
+        });
+******************************************************************************* --]]
+function npcUtil.completeQuest(player, area, quest, params)
+    params = params or {};
+
+    -- load text ids
+    local TextIDs = "scripts/zones/" .. player:getZoneName() .. "/TextIDs";
```
teschnei: shouldn't the text IDs have already been loaded by the script that called this function?

scripts/globals/npc_util.lua
```diff
@@ -38,175 +49,254 @@ function npcUtil.pickNewPosition(npc, positionTable, allowCurrentPosition)
...
+            elseif (type(v) == "table" and #v == 2 and type(v[1]) == "number" and type(v[2]) == "number") then
+                itemId = v[1];
+                itemQty = v[2];
+            elseif (type(v) == "table" and #v == 2 and type(v[1]) == "string" and type(v[2]) == "number" and string.lower(v[1]) == "gil") then
+                itemId = 65535;
+                itemQty = v[2];
             else
-                itemCount[v] = itemCount[v] + 1
+                print("ERROR: invalid value contained within items parameter given to npcUtil.tradeHas.");
+                goto continue;
```
teschnei: instead of a goto.. can you just wrap the neededItems[itemId] line with itemId ~= nil and itemQty ~= nil?

I really don't like gotos

wrenffxi: Done.

scripts/globals/npc_util.lua
wrenffxi: Trying to make it new contributor-proof.  How much faith do you have in your script makers? 🎲

teschnei:
Actually, I had been considering putting all text IDs into a global table
indexed by zone, so this would make that easier to transition to whenever
we get to that point, so it's good

wrenffxi:
rebased

PR #4433, opened 2017-12-30
Atma/Synth Recipe Errors

Raelious:
Fixed a couple atma errors, and one in the divine sap hq rate giving the nq result


TeoTwawki:
thanks!

PR #4434, opened 2017-12-30
Mission fights

ymkc:


scripts/zones/QuBia_Arena/mobs/Archlich_Taber_quoan.lua
```diff
@@ -6,78 +6,42 @@
 
 require("scripts/globals/status");
 
------------------------------------
--- onMobSpawn Action
------------------------------------
+function onMobInitialize(mob)
+    mob:setMobMod(MOBMOD_SCRIPTED_2HOUR, 1);
```
TeoTwawki: same as https://github.com/DarkstarProject/darkstar/pull/4435#discussion-diff-159130038R9

TeoTwawki:
~~Pretty sure at least one of the 3 dragons *doesn't* do draw in.~~ They all have slight differences and I recall being able to drag one away from the spawn point. I think dark dragon doesn't and the other 2 do.
wiki doesn't mention it on the dark dragon page, does on the other 2 dragons. Fits what I remember but I could still be wrong...Hmm. Meh, ima leave it like you have it for now, kiting is not really a standard strat here anyway.

PR #4435, opened 2017-12-30
Small fixes

ymkc:


scripts/zones/Gustav_Tunnel/mobs/Nanoplasm.lua
```diff
@@ -4,23 +4,16 @@
 -----------------------------------
 package.loaded["scripts/zones/Gustav_Tunnel/TextIDs"] = nil;
 -----------------------------------
+
```
TeoTwawki: uneeded newline. don't need to add or remove them as you touch files, I've a pr that mass eliminates the excess whitespace along with all the block comments but its waiting on bcnm branch merge coz @takhlaq is concerned about conflicts. Which is why I'll have to rebase and handle a bunch of conflicts myself.

TeoTwawki: Just letting you know so you don't modify a bunch more of them. More conflicts :x

scripts/zones/Quicksand_Caves/mobs/Centurio_X-I.lua
```diff
@@ -11,7 +11,7 @@ function onMobInitialize(mob)
 end;
 
 function onMobSpawn(mob)
-    mob:addMod(MOD_SILENTRES, 35);
+    mob:addMod(MOD_SILENCERES, 35);
```
TeoTwawki: good catch, I wonder how that didn't get reported as a crash before..

Also occurs to me we have these mods to make it resist sleeps when its race/job should already have given is sleep and dark resistances to handle that. Someone will have to look into that someday /shrug (not anything to do with your pr, I just noticed it coz we're here, you can ignore it)

scripts/zones/Upper_Delkfutts_Tower/mobs/Mimas.lua
```diff
@@ -0,0 +1,28 @@
+-----------------------------------
+-- Area: Upper Delkfutt's Tower
+--  MOB: Mimas
+-----------------------------------
+
+require("scripts/globals/status");
+
+function onMobInitialize(mob)
+    mob:setMobMod(MOBMOD_SCRIPTED_2HOUR, 1);
```
TeoTwawki: Unneeded, and that mobmod will be deleted soon™

when we call an ability with `useMobAbility()` the skillcheck is ignored and that ability just goes.

sql/mob_pools.sql
```diff
@@ -2720,7 +2720,7 @@ INSERT INTO `mob_pools` VALUES (2659,'Mikiruru','Mikiruru',196,0x000014070000000
 INSERT INTO `mob_pools` VALUES (2660,'Mildaunegeux','Mildaunegeux',359,0x0500D90100000000000000000000000000000000,1,1,1,180,100,0,1,1,1,18,513,32,231,157,0,0,0,1,0,359);
 INSERT INTO `mob_pools` VALUES (2661,'Millenary_Mossback','Millenary_Mossback',2,0x0000900100000000000000000000000000000000,1,1,7,380,100,0,1,1,1,0,0,0,7,133,0,0,0,1,0,2);
 INSERT INTO `mob_pools` VALUES (2662,'Million_Eyes','Million_Eyes',139,0x0000800100000000000000000000000000000000,4,4,12,280,100,0,1,0,0,0,0,0,151,133,0,0,10,1,0,139);
-INSERT INTO `mob_pools` VALUES (2663,'Mimas','Mimas',126,0x0000D00200000000000000000000000000000000,1,2,6,240,100,0,1,0,1,2,0,32,178,159,0,0,0,1,0,126);
+INSERT INTO `mob_pools` VALUES (2663,'Mimas','Mimas',126,0x0000D00200000000000000000000000000000000,2,2,6,240,100,0,1,0,1,2,0,32,178,159,0,0,0,1,0,126);
```
TeoTwawki: Technically this might not be wrong. He single-swings and animates as one attack per round even though his 2hr is hundred fists. Wiki often mistakes 2hr for equaling job but a lot of the time that isn't the case.

The other possibility is that his main job really is monk and something else should force him to single-swing. We _could_ theoretically calculate his stats from retail damage samples to find out for sure.

TeoTwawki: ..I'll leave this for someone else to make the call on, coz I dunno for sure the right answer.

ymkc: i see, i think you are right, makes more sense him being war with mnk 2hr, than being a mnk with single attack

just remove it?

scripts/zones/Upper_Delkfutts_Tower/mobs/Mimas.lua
ymkc: i added it in case core would ever pick up 2hr support, but looking at wren's pr it probably won't

sql/mob_pools.sql
TeoTwawki: can, we can figure out for sure which it should be later if you'd rather leave it be for now

ymkc: did so

sql/mob_pools.sql
```diff
@@ -3245,7 +3245,7 @@ INSERT INTO `mob_pools` VALUES (3184,'Poroggo_Prince','Poroggo_Prince',196,0x000
 INSERT INTO `mob_pools` VALUES (3185,'Poroggo_Seducteur','Poroggo_Seducteur',196,0x0000130700000000000000000000000000000000,4,4,12,240,100,0,1,1,1,0,0,0,0,0,0,0,2,1,0,196);
 INSERT INTO `mob_pools` VALUES (3186,'Poroggo_Servant','Poroggo_Servant',196,0x0000BF0800000000000000000000000000000000,1,1,0,240,100,0,0,0,0,0,0,0,0,3,0,0,0,1,0,196);
 INSERT INTO `mob_pools` VALUES (3187,'Poroggo_s_Toady','Poroggo_s_Toady',196,0x0000BF0800000000000000000000000000000000,1,1,0,240,100,0,0,0,0,0,0,0,0,0,0,0,0,1,0,196);
-INSERT INTO `mob_pools` VALUES (3188,'Porphyrion','Porphyrion',126,0x0000830200000000000000000000000000000000,11,1,1,240,100,0,1,0,1,2,0,32,456,159,0,0,0,1,0,126);
+INSERT INTO `mob_pools` VALUES (3188,'Porphyrion','Porphyrion',126,0x0000830200000000000000000000000000000000,11,11,5,240,100,0,1,0,1,2,0,32,456,159,0,0,0,1,0,126);
```
TeoTwawki: For future reference:
The cmbSkill column does wrong headed things and should even exist really, mobs don't really have the attack type their model might lead one to believe (mimas isn't really hitting you with an axes damage type) but for the moment it can cause strange behavior like dragoons multi hitting, so I recommend leaving it be in the future. The subjob, I don't know what this mob should have there. I suspect that many retail mobs default to a war sub. Unlike players they don't get "non sub" from job changing to the same main and sub at a moogle. In this case I'm merging as the rng/rng you set here.

PR #4437, opened 2017-12-30
Manawell template

areckx:
What else needs to be done here? Is there SQL? Stuff in abilities.{h,cpp}?

### DIFF
--- a/scripts/globals/abilities/manawell.lua
+++ b/scripts/globals/abilities/manawell.lua
@@ -1,8 +1,9 @@
 -----------------------------------
--- Ability: Manafont
--- Eliminates the cost of magic spells.
--- Obtained: Black Mage Level 1
--- Recast Time: 1:00:00
+-- Ability: Manawell
+-- Eliminates the cost of next magic spell for self or others in party, not alliance or others.
+-- Does not stack with the effects of Spontaneity
+-- Obtained: Black Mage Level 95
+-- Recast Time: 0:10:00
 -- Duration: 0:01:00
 -----------------------------------
 
@@ -22,5 +23,5 @@ end;
 -----------------------------------
 
 function onUseAbility(player, target, ability)
-    player:addStatusEffect(EFFECT_MANAFONT,1,0,60);
+    player:addStatusEffect(EFFECT_MANAWELL,1,0,10);
 end;


R3P0FFXI:
EFFECT_MANAWELL does not even exist you would need to build it

areckx:
Right, that's what I was asking. I am looking throught the cpp files I see stuff there for the two hours but not the regular abilities.

TeoTwawki:
effect scripts: `\scripts\globals\effects\`
ability scripts: `\scripts\globals\abilities\` (you found that one)
ability sql table: `\sql\abilities.sql`

since bg says "The next spell the target casts costs 0 MP and does not require MP to cast, like Manafont." I assume (the ability didn't exist back when i played retail) you'll need to go into where ever we handle manafont and place a similar check for manawell  that *probably* means either src/map/ai/states/magic_state.cpp or `\scripts\globals\magic.lua` but I haven't looked. 

You can likely search trunk for "manafont" to find it using github's search tool up above on this page or grep locally or a 3rd party search tool (I like the one at nirsoft, windows native search sucks).

areckx:
Thanks. I've been using Kate and it has a similar search tool to Sublime that searches all files in the project folder. I'm sure this ability could be implemented, along with the other Black Mage abilities. I certainly am biased, as I play BLM. :)

areckx:
So looking at line 307 of src/map/ability.h ::
`ABILITY_MANAWELL           = 257,`

the ability is already indexed. 

Line 226 of src/map/ai/states/magic_state.cpp ::
` // check has mp available
    else if (!m_PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_MANAFONT) &&
        !(m_flags & MAGICFLAGS_IGNORE_MP) && battleutils::CalculateSpellCost(m_PEntity, GetSpell()) > m_PEntity->health.mp)
    {
        if (m_PEntity->objtype == TYPE_MOB && m_PEntity->health.maxmp == 0)
        {
            ShowWarning("CMagicState::ValidCast Mob (%u) tried to cast magic with no mp!\n", m_PEntity->id);
        }
        m_errorMsg = std::make_unique<CMessageBasicPacket>(m_PEntity, m_PEntity, static_cast<uint16>(m_PSpell->getID()), 0, MSGBASIC_NOT_ENOUGH_MP);
        return false;
    }
    return true;
}`

So that might be the place I can put it. I finished compiling a test server that matches this current commit and after I get it configured and confirm it works I will try out a copy-pasta version of manafont and match to any SQL/abilities index to see if I can get it to do _something_ in the log or debug flags.

TeoTwawki:
> So looking at line 307 of src/map/ability.h ::
ABILITY_MANAWELL = 257,
>
> the ability is already indexed.

Most enumeration for everything under the sun is handled. not 100% of everything, but like, 99%
If you click line numbers here on github you can link directly to places you wanna show ppl, like this:
https://github.com/DarkstarProject/darkstar/blob/master/src/map/ability.h#L307

😉

areckx:
Found this in sql/abilities.sql  ::
--
-- Table structure for table `abilities`
--

DROP TABLE IF EXISTS `abilities`;
CREATE TABLE `abilities` (
  `abilityId` smallint(5) unsigned NOT NULL,
  `name` tinytext,
  `job` tinyint(2) unsigned NOT NULL DEFAULT '0',
  `level` tinyint(2) unsigned NOT NULL DEFAULT '0',
  `validTarget` smallint(3) unsigned NOT NULL DEFAULT '0',
  `recastTime` smallint(5) unsigned NOT NULL DEFAULT '0',
  `recastId` smallint(5) unsigned NOT NULL DEFAULT '0',
  `message1` smallint(5) unsigned NOT NULL DEFAULT '0',
  `message2` smallint(5) unsigned NOT NULL DEFAULT '0',
  `animation` smallint(5) unsigned NOT NULL DEFAULT '0',
  `animationTime` smallint(4) unsigned NOT NULL DEFAULT '0',
  `castTime` smallint(4) unsigned NOT NULL DEFAULT '0',
  `actionType` tinyint(2) unsigned NOT NULL DEFAULT '6',
  `range` float(3,1) unsigned NOT NULL DEFAULT '0.0',
  `isAOE` tinyint(3) unsigned NOT NULL DEFAULT '0',
  `CE` smallint(5) NOT NULL DEFAULT '0',
  `VE` smallint(5) NOT NULL DEFAULT '0',
  `meritModID` smallint(4) NOT NULL DEFAULT '0',
  `addType` smallint(2) NOT NULL DEFAULT '0',
  `content_tag` varchar(7) DEFAULT NULL,
  PRIMARY KEY (`abilityId`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8 AVG_ROW_LENGTH=56;

TeoTwawki:
please use our discord room for this. every comment here sends out email and notifications to every project member. we can reopen the pull request when you are ready.

PR #4439, opened 2017-12-31
5-3T1: Spiral fixes

wrenffxi:
removed unnecessary loop in Gargoyle lua
players must kill the gargoyle to open the door. this prevents players from tucking into a corner where the gargoyle can't path and waiting for it to despawn.  or letting it kill them, then reraising after it despawns.


TeoTwawki:
You probably want to wrap it in a check for isKiller then so that the code doesn't re-run per party member.

wrenffxi:
Whoops.  Right you are.  Done.

PR #4448, opened 2018-01-01
Remove unneeded nil lines from headers.

TeoTwawki:


TeoTwawki:
*REEEEEEEE*based

wrenffxi:
a lot of those mobids nils are needed.

batallia downs mobIds contains CERMET_HEADSTONE which conflicts with headstones in other zones.
cape terrigan, buburimu contains SIGNPOST_OFFSET which does the same
etc

TeoTwawki:
we're going to have to change the var names it'll get lost like this because its unexpected.

TeoTwawki:
Todo (likely I'll get to it as soon as BCNM branch gets merged to master, as I have pr waiting that needs gone first): rename mobIDs.lua to EntityIDs.lua sicn eit now contains NPC's as well, and set both all of those and TextIDs.lua's to be returned tables so nothing needs nil'd ever again. See also: https://github.com/DarkstarProject/darkstar-utils/issues/15

TeoTwawki:
coming back to this later

PR #4451, opened 2018-01-01
Manawell

areckx:
Here it is. :D 

Can someone confirm the behavior of retail version when trying to use on someone who already has Spontaneity and vice-versa? I implemented it so that one will just overwrite the other, but I'm not sure if that's the expected result.


TeoTwawki:
> modified spontaneity so no stack

Both abilities are lv 95 on bg wiki..How would you even stack them unless server is custom (and thus not our problem)? non issue.

areckx:
I mean to say that a rdm can use spontaneity on a player or self, and the
effect won't stack with the effect of manawell

TeoTwawki:
tl;dr teo thought these were self target, ignore

areckx:
Ah yes that just needs a change to the status_effects.h

areckx:
It still needs to have the overwrite or inability to be overwritten(can someone with retail confirm?)  because the abilities last for 1 minute or until next spell. Otherwise, the player could have a BLM cast manawell and then a rdm cast spontaneity and gain both effects.

areckx:
Can't believe I forgot about this one.

Wiggo32:
Gaining both effects is possible:
https://youtu.be/7vYuXiI63M0

Issue #3673, opened 2017-01-27
"aftermath" effects, and how joe random contributor can help with them


TeoTwawki:
Blatantly discarding template - it is irrelevant for this. Laying it out so interested parties can tackle.

- ~~We have 4 effect scripts for this. ***four***. I don't think we need _four_ to get this right.~~ nvm, its a nightmare to condense and we don't wanna addeffectEx for the extra icons mythics apparently use

- Some scripts do `addStatusEffect(EFFECT_AFTERMATH)`

- Some other scripts do a static item ID check and a job check before calling a wrapper that applies 1 of: EFFECT_AFTERMATH_LV2, EFFECT_AFTERMATH_LV2, or EFFECT_AFTERMATH_LV3. 

  - A ton of weapons that should have aftermath effects don't because of the static itemID check, this needs to be a modifier. grab the item pointer so it can be the object you reference instead of the player, then getMod on the *item* and *not the player* to solve this.

  - The job check likely doesn't need to exist? I am aware of no situation where the aftermath is available to only one job on a weapon multiple jobs can equip who both have access to the weaponskill.

  - Effects applied by this wrapper have an unused "params" argument and defaultly fill in everything with the same power/subpower every mythic weapons aftermath is identical. these 3 effect scripts only do attack or acc or magic attack or magic accuracy.

  - Pretty sure some scripts call the version the multiple effect scripts that doesn't have the bonus's that item would grant and instead wind up with a generic default buff. Have not personally checked.

  - Empyrean WS don't have theirs set up proper yet either. So all 3 lines of uber-weapon are gimped.

TeoTwawki:
Dear _joe random contributor_, pick a bullet point son.

ramenkat:
- The job check likely doesn't need to exist? I am aware of no situation where the aftermath is available to only one job on a weapon multiple jobs can equip who both have access to the weapon skill. 

Take anything I say with a grain of salt, since I don't code (like at all) but ... I think you are right on this point. Any job that can equip the relic weapon should be able to gain the aftermath effect if they have access to the ws. It especially doesn't matter if the weapon is one job only (Gungnir, for example). 

/end $0.02 for joe random contributor >:3

Issue #3665, opened 2017-01-24
Levels of aftermath "appear" to stack, no overwrites


ffxijuggalo:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30161227_1

**_Server Version_** (type `@revision` in game) **:** 7b4b683f8c429c1595f28e3d95566b7bce3b2b0f

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** After using mandalic stab to check and see if the aftermath effect would activate, I noticed that the levels don't overwrite each other as the level increases. Screenshot included showing all 3 levels of aftermath active:
![shaggy_2017 01 24_083529](https://cloud.githubusercontent.com/assets/22580624/22251659/396b25a0-e211-11e6-99da-9048661c7b4b.png)

Another screenshot from the chatlog showing them wearing off individually:
![shaggy_2017 01 24_084458](https://cloud.githubusercontent.com/assets/22580624/22251721/6e20e7ee-e211-11e6-96e4-aeefa4568f2f.png)

zynjec:
This was fixed, closing.

PR #4453, opened 2018-01-01
Aftermath Rewrite (Relic and Mythic)

nfaletra:
- New Mod::Aftermath. Added to item_mods.sql to appropriate Relic and Mythic item IDs. This mod dictates whether aftermath should be applied when using the corresponding weaponskill and at what tier based on reqLvl/iLvl
- Added Mods for double/triple damage. These need to stack with other increased damage mods while under aftermath effects
- Logic in weaponskills.lua and individual weaponskill lua scripts has changed drastically
- Aftermaths that should apply to pets does now
- Added some functions to lua_baseentity.h/cpp to assist with new lua code

Addresses Issues:
#3673 
#3665


R3P0FFXI:
@nfaletra /cry

Hozu:
I'm trying to figure out how you're handling the "occasionally attacks twice/thrice" thing here. You're not doing it with double/triple attack, are you?

nfaletra:
I didn't find anything in existing code for "Occasionally Attacks twice/thrice" and used double/triple attack. Is the math underneath different for retail? If it is, I can implement that. What was there before was attempting to use double/triple attack.

Hozu:
Yeah, it wasn't done properly before. You'll have to add at least one mod to handle it. Thankfully, it looks like you've already figured out how to get mods for individual weapons so it shouldn't be too difficult.

I believe you'll need to add the check in this function:

https://github.com/DarkstarProject/darkstar/blob/master/src/map/utils/attackutils.cpp#L39

nfaletra:
That function actually isn't called anywhere. However, the one in battleutils.cpp, which does the exact same thing, is. Not sure how I want to mod it yet as, according to bg-wiki, the proc rates on the Mythic Aftermath is a little different than normal occ att twice/thrice.

Hozu:
https://github.com/DarkstarProject/darkstar/blob/master/src/map/attackround.cpp#L208

Isn't it called there and a couple of lines below? attackround.h is included, and that includes utils/attackutils.h.

nfaletra:
That calls the version in battleutils.cpp

nfaletra:
Implemented and tested. Works puuuurdy good.

Hozu:
Hm.. I have a couple of concerns. Using the existing function for OAX as-is doesn't account for the different OAX rates across the Mythic aftermath levels (https://www.bg-wiki.com/bg/Mythic_Aftermath). I was thinking each weapon should have a mod for OATwice and a second for OAThrice to account for this (in the aftermath script probably). The getHitCount function would have to have another distribution check specifically for the Mythic aftermath based on whether or not the OATwice mod is greater than 0. ie. compare against the mod amount held by OATwice and OAThrice.

Another concern is making sure the OAX only applies to the mythic weapon when dual-wielding. I remember in my initial implementation of the Jailer weapons, it ended up applying the effect to both weapons so I had to add checks for which hand the weapon is in (excluding h2h). So uh... test for that. :P Should be easier for you though since Mythic aftermath requires the Mythic to be in the main hand, right? I honestly don't know, never having had the chance to use them on retail.

Third concern is that shouldn't the Mythic OAX mods apply to weaponskills? Unless I'm blind, I don't see where you added that?

nfaletra:
I'll look into it. Thanks!

nfaletra:
I tweaked the mods to be Mythic specific since they are the exception to the OAT rule. They CAN apply to weapon skills and job abilities. I found where to add the checks so that the extra hits apply to Jump, but I can't figure out where multi-hit is applied to weapons skills.

Also, I'm applying the mods to the weapon itself rather than the player entity. This should prevent the effects from proccing on the offhand weapon. I tried searching for where Double Attack was being applied since it should be applying to weaponskills, as well, but I can't seem to find it. Is it in lua?

maxtherabbit:
why are you using the power of the effect to determine the twice/thrice? the way you have aftermath_lv3.lua written currently it will apply BOTH twice and thrice mods to the same weapon...

maxtherabbit:
wouldn't it be better to just change the effect subtype to OCCASIONALLY_ATTACKS_THRICE

nfaletra:
The description on the weapons says "Occasionally attacks twice or thrice" when the weapon is level 95 or higher. The getHitCount distribution values are slightly off from what bg-wiki has (30% for 3, 50% for 2 rathar than 20% and 40% respectively).

With that in mind, I see if thrice procs and skip twice, otherwise see if twice procs. Does that make sense? Did I understand the question?

maxtherabbit:
oh I see, my fault for not knowing how the mythics work I guess

teschnei:
Here's how I'm thinking is the best way to do this:
The WS scripts have nothing in them for aftermath at all.
The weapons that have aftermath get the SCRIPT_EQUIP flag toggled in their item table, so that we can utilize onItemCheck in the weapons scripts.

In onItemCheck, on equip, we will add a listener for WEAPONSKILL_USE, which fires after a WS is used.
```lua
player:addListener("WEAPONSKILL_USE", "AFTERMATH", function(user, target, wsid, tp, action)
  -- figure out if it should apply based on tp/action/etc
end)
```
(this should probably also only happen in the main hand slot, and the ranged weapon would need a different identifier)

And on unequip, we remove the same listener
```lua
player:removeListener("AFTERMATH")
```

Inside the inner function, you can also apply it to pets, etc.  This'll keep all the mods in their own scripts, so we won't need any sort of if/elses to figure out which mod to apply, etc.  We could even name all the different tiers of weapons so they each get their own script (kogarasumaru_85.lua, etc) to keep it even more distinct, if we wanted.  Doing it this way also means we don't have to worry about aftermath on non-RMEA weapons, since they won't have the aftermath listeners.

That's... pretty much it.  Almost everything else related to aftermath can go away at that point.  This doesn't address the mythic OAX or empyrean ODD mods, though.  I think we should handle that in a separate pull request, as a separate issue, after we discuss it a bit more

nfaletra:
I like it.

In equip/unequip, I think we should check if the aftermath effect is already active and apply/remove the mods from the player appropriately. That will keep people from gaining aftermath, then switching to a different weapon but still getting the effects. And then switching back will re-enable those effects. This would probably require some kind of tracking as to what weapon applied the aftermath. Perhaps a cached off itemId when aftermath is applied?

I was even thinking about defining aftermath and what mods they apply in its own DB table. And have an item_aftermaths table that links itemIds to aftermathIds (similar to item_mods). That way we can keep the scripts stupid when it comes to applying mods. Not entirely sold on this. Was just thinking about it on my way to work. It would clean up the logic in the weapon scripts and keep all of the data involved in one location. Rather than having to crack open 10 different weapon scripts to make changes. Also we could keep the weapon scripts to a single file instead of one for each level of weapon (75-119/III). After editing god knows how many weaponskill scripts, I'm a big fan of making this as generic as possible.

What do you think?

teschnei:
I think that's making it more complex than it needs to be. For
equip/unequip, can probably just shove the itemid in the effect somewhere
if there's space

nfaletra:
Which part is more complex than it needs to be?

I just don't want to see mandau_75.lua to mandau_199III.lua. I'd like to make that unnecessary. It's like 9-10 scripts per weapon. 10-15 weapons per category (Relic, Mythic, Empyrean). That's a lot of copy+pasting. And a lot more testing. I'd rather test each tier of weapon (2 tiers for Relic/Empyrean, 3 tiers for Mythic) instead of testing EVERY weapon.

The way the aftermath effect is currently working (not in my PR, what is there now), it adds mods to the player. How do we check how that mod was applied? Removing the mod but keeping the effect active (so you still have the icon, etc.) seems like the only way to keep people from exploiting it. Is there space in StatusEffect to keep extra data like the itemId? I'm almost tempted to keep track of a current aftermath class with relevant data in it on the player. That might be taking it too far, but it will be very clear what is trying to be accomplished.

teschnei:
Oh, we can keep 1 script per weapon then, and just table the itemid to the
aftermath effect like you said, except do it as a lua table inside the
weapon's script instead of the database.

For your other question, it doesn't really matter because all aftermath
should be removed if you change any weapons, so the effect will remove the
modifiers (that should be done in the core instead of scripts, where
changing gear would reset TP, if it isn't already)

nfaletra:
Ah. Okay. That will work. I don't believe the effect is removed when unequipping the weapon right now. But I can make sure that happens.

nfaletra:
Quick note for myself: In retail, anything that resets TP (swapping/unequipping main/offhand/ranged, zoning) will cause the Aftermath effect to wear. EZ LYFE

teschnei:
Search tp = 0 in charutils, that should be the 4 places you need to do it

nfaletra:
Yup. And adding EFFECT_FLAG_ON_ZONE to the sql will make it wear on zoning. :D

PR #4460, opened 2018-01-02
Bridges in RoMaeve now rise during full moon.

zynjec:
- Bridges now go up and down based on moon cycle.
image
- Cleaned up some formatting.


scripts/zones/RoMaeve/Zone.lua
```diff
@@ -62,64 +50,38 @@ function onZoneIn(player,prevZone)
...
-
-    -- Full moon + "clear" weather stuff (actually "sunshine" weather, widespread misconception since Ro'Maeve does not have "clear" weather ever)
-    local Moongate_Offset = 17277195; -- _3e0 in npc_list
-    local hour = VanadielHour();
-
-    if (IsMoonFull() == true and GetNPCByID(Moongate_Offset):getWeather() == WEATHER_SUNSHINE) then
-        GetNPCByID(Moongate_Offset):openDoor(432); -- 3 game hours worth of seconds
-        GetNPCByID(Moongate_Offset+1):openDoor(432);
-        GetNPCByID(Moongate_Offset+7):openDoor(432); -- visual part of Qu'Hau Spring
+    local MOONGATE_OFFSET = 17277195; -- _3e0
```
teschnei: tfw wren is removing all these

zynjec: I can move it to a NPCIDs if you prefer, it was just a quick fix to what was there.

TeoTwawki: wrens been putting the NPC values in the mobIDs.lua, which I intend to rename to enitityIDs.lua for semantics sake when I take care of assaults "ids.lua"

zynjec: I actually initially added it to MobIDs, then backtracked as I didn't think having conflicting entity types was a great idea.

I can throw it in MobIDs for now if that is what your renaming plan is.

Hozu:
I remember when I attempted to do this, there was an oddity when you went out of range of the bridge when it was up, then went back to the bridge. When I did that, the bridge was entirely missing, but it could still be walked on, I think.

Is that still the case?

zynjec:
@Hozu That is indeed still the case, but it also happens on retail (happened to me when I was checking this out last free play), so while its a visual error it's at least retail accurate ;)

Hozu:
Seriously? It's a retail bug? Damn it.

PR #4465, opened 2018-01-03
Fixed CCharSyncPacket and latents being checked if no effects changed

teschnei:


src/map/status_effect_container.cpp
```diff
@@ -459,23 +459,23 @@ void CStatusEffectContainer::DeleteStatusEffects()
...
+                UpdateStatusIcons();
+            }
+
+            //check for latents
+            PChar->PLatentEffectContainer->CheckLatentsFoodEffect();
+            PChar->PLatentEffectContainer->CheckLatentsStatusEffect();
+            PChar->PLatentEffectContainer->CheckLatentsRollSong();
+            PChar->pushPacket(new CCharSyncPacket(PChar));
+        }
         m_POwner->UpdateHealth();
```
ShelbyZ: Should this just call:

```cpp
m_POwner->updatemask |= UPDATE_HP;
```

teschnei:
No, I was wrong when we talked about that earlier (in the latent_effect
PR): UpdateHealth is the one that toggles UPDATE_HP, and UPDATE_HP just
sends the packets

PR #4467, opened 2018-01-03
Dec2017-TextID Fixes/All Zones Cross Checked-Stable Branch

dragonarmor:


TeoTwawki:
are you sure that you are sure that you got'em all right this time?

dragonarmor:
Well unfortunately the batch files report no errors. I have learned that this is going to require a lot of manual cross checking polutils. We can leave this open for the time being if you like or have someone cross check my work.

dragonarmor:
here is an example Lufaise_Meadows:
6400 There is nothing out of the ordinary here.<Prompt>
7649 There is nothing out of the ordinary here.<Prompt>
7736 There is nothing out of the ordinary here.<Prompt>

So is it ok for the id batch to use any of the above? or the lowest one? i.e. 6400
the batch job will sometimes use a higher entry

TeoTwawki:
I've run into that before. See on retail every NPC has its own, but on darkstar we just reuse the same one. Any of the3 will work identically form the players perspective.

The reason they have multiples is due to how SE stores these strings they are actually a part of an entity.

With cs events you can see this as an NPC with a physical model will turn to face you. With messageSpecial strings that doesn't happen, and these strings here are usually used for objects with no visible model anyway.

dragonarmor:
Sounds good. rechecked and this should wrap it up.

PR #4470, opened 2018-01-03
Enhance 0x0DC handling

MalRD:
This is an early PR to get some feedback on the implementation. Listed are the completed items and items I believed may need to be done.

- [x] Handle /displayhead [subcommand] in 0x0DC (#486)
- [x] Handle /mentor [subcommand] in 0x0DC
- [x] Handle Help Desk=>Config=>Cancel New Adventurer Status in 0x0DC
- [x] Modify appearance packet to respect /displayhead
- [x] Add migration script for chars.mentor and chars.isnewplayer
- [x] Rename charutils::mentorMode to follow save pattern
- [x] Improve 0x0DC switch handling
- [x] Correct /anon [subcommand] in 0x0DC
- [x] Correct /inv [subcommand] in 0x0DC
- [x] Correct /autogroup [subcommand] in 0x0DC
- [x] Correct /recruit [subcommand] in 0x0DC
- [x] Correct /autotarget [subcommand] in 0x0DC
- [x] Correct /away in 0x0DC
- [x] Correct /online in 0x0DC
- [x] Move unnamed flag masks from FLAGTYPE to new enum?
- [ ] Figure out better name for unnamedFlags

**Improve 0x0DC switch handling**
Currently, the three 0x0DC packet cases that I handled are returning from the switch statement to ensure that the packet response matches what retail captures showed. My testing shows that /anon, /inv, /away, /online, /autogroup, /recruit, and /autotarget also follow this packet flow and not the one that DSP currently uses. I'm unsure what the case for the current implementation is so I haven't modified it yet, but my plan is to unify the response for 0x0DC to match the three I added.

From the packets I've captured, the flow for 0x0DC in the general case: 

1. Outgoing packet sent.
2. Incoming response packet(s) received (1-3 packets, depending on functionality)
3. One second later, CharUpdate (0x37) is received.
4. One second later, CharSync (0x67) is received.

I was unable to get the  CharUpdate and CharSync to be received on the same second sending only one request. 

If a second 0x0DC packet is sent before the CharUpdate packet has been received, however, the Update and Sync packet will be received at the same time. Update always arrives before Sync when testing in low-packet situations (Mog House, empty zone).

The way that Update and Sync are currently handled doesn't look like it will work for retail-accurate packet pushing. With the current implementation, Update and Sync are received by the client immediately, while they are delayed on retail. We may want to create a new issue dedicated to Update/Sync work to resolve that, but I'm not familiar enough with DSP or FFXI's packet system to know if this is intended or not.


src/common/mmo.h
```diff
@@ -57,10 +57,19 @@ enum FLAGTYPE : uint32
     FLAG_GM_LEAD        = 0x06000000,
     FLAG_GM_PRODUCER    = 0x07000000,
     FLAG_BAZAAR         = 0x80000000,
+};
 
-//  FLAG_AUTOGROUP      = 0x00000000,
-//  FLAG_NEWPLAYER      = 0x00000000,
-//  FLAG_MENTOR         = 0x00000000,
+enum NFLAGTYPE : uint32
```
TeoTwawki: >  Figure out better name for unnamedFlags

maybe "flags"/"nameflags"(same thing) and "unnamedFlags" could be "entityFlags" and "playerFlags" respectively. The former have near identical analogs in the flags used on mobs and npc but the later are player specific at least in usage (I wouldn't be surprised if you could force them into non players, I can already make GM goblins and bazaar icon's leaping lizzies.)

whasf:
@teschnei does this look good to merge?

teschnei:
The code is good

I still wanted a better name for the flags, but not badly enough to keep the PR unmerged.  We can fix it later if someone ever thinks of a better name

PR #4476, opened 2018-01-05
purulent_ooze.lua Damage Fix

dragonarmor:
Should be baseDamage


TeoTwawki:
I might make a tiny unrelated adjustment, just because now that I've seen the pre-existing code I can't un-see it.

PR #4478, opened 2018-01-05
Add script for gravitic horn. Animation and other sql still needed.

TeoTwawki:


scripts/globals/mobskills/gravitic_horn.lua
```diff
@@ -0,0 +1,41 @@
...
+    local damage = 0;
+
+    -- estimation based on "Throat Stab-like damage"
+    if (currentHP / target:getMaxHP() > 0.2) then
+        baseDamage = currentHP * .95;
+    else
+        baseDamage = currentHP;
+    end
+
+    -- Because shell matters, but we don't want to calculate damage normally via MobMagicalMove since this is a % attack
```
TeoTwawki: Wasn't sure how to handle this, need eyes other than my own on it to be sure this was how I should do it.

Its a magical wind move that is % based instead of the usual calculated damage. But I don't know the correct percent, so I'm using what throat stab had as all either wiki says is percent based of "throat stab like". I'm assuming SE has a standard "critical damage" and throat stabs math is it.

Some NM antlions use this, including Orcus who I'd like to implement in the near future.

TeoTwawki: *nudge*
If its good lets merge it, if not guide me bruh.

PR #4480, opened 2018-01-06
!F - Relic and Mythic Aftermaths

nfaletra:
* Added SCRIPT_EQUIP flag to all Relic weapons
* Handling onItemCheck in Relic weapon scripts, listening for weaponskill use in order to apply aftermath
* Added weaponskillids.lua to define Weaponskill IDs for human readability in LUA
* Removed all aftermath logic from specific weaponskill scripts
* Added listener to EFFECT_LOSE to remove mods when Aftermath Effect wears off
* Added EFFECTFLAG_ON_ZONE to aftermath effects so it wears on zoning
* Aftermath now wears whenever TP is reset (swapping any weapon slot)
* This is a first pass to clear the method of applying aftermath. Mythic/Empyrean to follow.


Hozu:
Stage 4 relic weapons should also get the aftermath effect, for when they grant access to the relic WS inside Dynamis.

src/map/lua/lua_baseentity.cpp
```diff
@@ -11229,6 +11229,39 @@ inline int32 CLuaBaseEntity::getWeaponDmgRank(lua_State *L)
...
 }
 
+/************************************************************************
+*  Function: getWeaponID()
+*  Purpose : Returns the item ID for the weapon in the passed in SLOT
+*  Example : player:getWeaponID()
+*  Notes   : Primarily used in REM weapon scripts to determine which version of the weapon is being used
+************************************************************************/
+
+inline int32 CLuaBaseEntity::getWeaponID(lua_State* L)
```
teschnei: what does this do that getEquipID doesn't?

src/map/utils/charutils.cpp
```diff
@@ -96,6 +96,12 @@ This file is part of DarkStar-server source code.
 #include "petutils.h"
 #include "zoneutils.h"
 
+enum ITEMCHECK
```
teschnei: probably put this in item_armor.h since it's directly related to SCRIPTTYPE

teschnei: also, it should be an enum class

scripts/globals/items/amanomurakumo.lua
```diff
@@ -1,13 +1,85 @@
...
+    end
+end
+
+function onItemCheck(player, param, caster)
+    if (param == ITEMCHECK_EQUIP) then
+        player:addListener("WEAPONSKILL_USE", "AFTERMATH", onWeaponskill);
+    elseif (param == ITEMCHECK_UNEQUIP) then
+        -- Make sure we clean up the effect and mods
+        if (player:hasStatusEffect(EFFECT_AFTERMATH)) then
+            aftermathLost(player, player:getStatusEffect(EFFECT_AFTERMATH))
```
teschnei: i believe this will delete the modifiers twice - once here, and then once again immediately after when the effect is deleted as a result of the TP reset.  Better off just doing nothing here except removing the listener and letting the effect expire on its own

scripts/globals/items/amanomurakumo.lua
```diff
@@ -1,13 +1,85 @@
...
+            removeAftermathEffect(target, aftermathTable[itemId]);
+            -- Remove the effect listener
+            target:removeListener("AFTERMATH_LOST");
+        end
+    end
+end
+
+function onItemCheck(player, param, caster)
+    if (param == ITEMCHECK_EQUIP) then
+        player:addListener("WEAPONSKILL_USE", "AFTERMATH", onWeaponskill);
```
teschnei: my preference is, unless you're using onWeaponskill more than once, just define it inline as the parameter (and the same thing for aftermathLost)

scripts/globals/weaponskills.lua
```diff
@@ -1023,78 +1023,21 @@ function takeWeaponskillDamage(defender, attacker, params, primary, finaldmg, sl
...
-function shouldApplyAftermath(player, effect)
-    local result = true;
-    if (effect == EFFECT_AFTERMATH_LV1 and (player:hasStatusEffect(EFFECT_AFTERMATH_LV2) or player:hasStatusEffect(EFFECT_AFTERMATH_LV3))) then
-        result = false;
-    elseif (effect == EFFECT_AFTERMATH_LV2 and player:hasStatusEffect(EFFECT_AFTERMATH_LV3)) then
-        result = false;
-    end;
+function addAftermathEffect(player, tp, params)
+    player:addStatusEffect(EFFECT_AFTERMATH, params.power, 0, params.duration(tp));
+    for i,mod in ipairs(params.mods) do
```
teschnei: convention is, if you're not using the index, just call it _

scripts/globals/items/amanomurakumo.lua
nfaletra: How exactly are modifiers deleted on effect wearing? effects/aftermath.lua is going away.

src/map/lua/lua_baseentity.cpp
nfaletra: Nothing. I didn't find getEquipID right away. Will removed and swap over.

scripts/globals/items/amanomurakumo.lua
nfaletra: Also, I tested it and delModifier was only getting called once when aftermath wore off. Tested both natural wearing and weapon swapping.

scripts/globals/items/amanomurakumo.lua
nfaletra: Binding a listener in a listener inline was getting really ugly. What variables to reference where was extremely confusing. This was more readable. Also, aftermathLost gets called when the item is unequipped to make sure the mods are removed from the player.

scripts/globals/weaponskills.lua
nfaletra: Can do.

src/map/utils/charutils.cpp
nfaletra: word.

nfaletra:
@Hozu I can't find any reliable info that states you get Aftermath with Stage 4 Relics in Dynamis. Only thing I found that mentions it in passing is FFXICLOPEDIA. BG-WIKI doesn't have any info about it. Do we have anyone on retail with a Stage 4 Relic that can confirm this?

scripts/globals/items/amanomurakumo.lua
teschnei: ah I was a bit confused on the way the listeners were ordered.  I think it would be a tiny bit easier to just delStatusEffect here instead, actually, even though it does get deleted later.  Doing it here should fire AFTERMATH_LOSE while the equip slot is still the weapon, so you wouldn't have to clean up manually here, and then rely on AFTERMATH_LOSE doing nothing when the effect is deleted by the core after

teschnei:
One last thing... I think the aftermath listener names must be namespaced.  They definitely need to have different names between the melee and ranged ones, but it's probably safer to just name them all related to their weapon (AFTERMATH_AMANO and AFTERMATH_AMANO_LOSE, for example) in case a job can (or will be able to) dual wield two of them

scripts/globals/items/amanomurakumo.lua
nfaletra: Ah okay. That makes sense. I’ll give that a try.

nfaletra:
Samurai is actually a good example of that. Good call.

nfaletra:
@Hozu Found it! It was on the weaponskill page.

https://www.bg-wiki.com/bg/Final_Heaven

I’ll add to the stage 4 versions.

nfaletra:
Sorry for the delay guys. I should be able to sneak in a bit work this evening.

scripts/globals/items/amanomurakumo.lua
nfaletra: So I could have sworn I tested item swapping to make aftermath wear. It's not deleting the mods because the item isn't equipped anymore when it tries to grab the itemId.

src/map/utils/charutils.cpp
```diff
@@ -1540,26 +1540,28 @@ namespace charutils
...
+                luautils::OnItemCheck(PChar, PItem, ITEMCHECK::UNEQUIP, nullptr);
+            }
+
             //todo: issues as item 0 reference is being handled as a real equipment piece
             //      thought to be source of nin bug
             PChar->equip[equipSlotID] = 0;
             PChar->equipLoc[equipSlotID] = 0;
 
-            if (((CItemArmor*)PItem)->getScriptType() & SCRIPT_EQUIP)
+            PChar->m_EquipFlag = 0;
```
teschnei: we only need to redo the equipflag if the item had script_equip

scripts/globals/items/amanomurakumo.lua
```diff
@@ -1,13 +1,90 @@
...
 -- Additional Effect: 10% Attack Down
 -----------------------------------------
 require("scripts/globals/status");
 require("scripts/globals/magic");
 require("scripts/globals/msg");
+require("scripts/globals/weaponskills");
+require("scripts/globals/weaponskillids");
 -----------------------------------
 
+NAME_WEAPONSKILL = "AFTERMATH_AMANO";
```
teschnei: these should be local vars rather than global vars

teschnei:
I don't remember if we discussed it, but moving the lua call to before the item IDs are zeroed is what I would have suggested.  Other than the small comment on localing the vars, relics should be good.  Should be good to do the rest of the weapons

scripts/globals/items/amanomurakumo.lua
nfaletra: referenced in multiple places. Only one place to edit if needs changing?

nfaletra: oh you mean local. derp thought I fixed that.

teschnei: I just mean making it local NAME_WEAPONSKILL instead of NAME_WEAPONSKILL so that the variable doesn't end up in the global namespace (like how aftermathTable is local)

nfaletra: I did it in mandau.lua and that was it. -_-

nfaletra: Aaaaand fixed.

nfaletra:
To clarify, did you want me to add Mythic and Empyrean to this PR before it gets merged?

teschnei:
Yep. Even though the relic is better, don't want to have a version where
empyrean and mythic were removed.  If you want, you can close this one, and
then reopen it once you want me to review again or you're finished

On Jan 9, 2018 7:47 AM, "Autkast" <notifications@github.com> wrote:

To clarify, did you want me to add Mythic and Empyrean to this PR before it
gets merged?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/DarkstarProject/darkstar/pull/4480#issuecomment-356304382>,
or mute the thread

nfaletra:
Mythic aftermaths have been added. With these changes, the previous aftermath implementation (that didn't actually function) is covered. I'd like to get this merged in and then start on Empyrean weapons. Empyrean weapons were never hooked up, so nothing has been removed at this point, just fixed.

Modifiers were added for Occ does double/triple damage and Occ attacks twice/thrice. Mythics/Empyreans use the former, only Mythics use the latter. If there are issues with how the code is implemented to use these mods, please let me know.

src/map/utils/attackutils.cpp
```diff
@@ -151,15 +151,32 @@ namespace attackutils
...
             return damage;
         }
+
         uint32 originalDamage = damage;
-        auto occ_extra_dmg = battleutils::GetScaledItemModifier(PChar, PWeapon, Mod::OCC_DO_EXTRA_DMG);
-        auto occ_extra_dmg_chance = battleutils::GetScaledItemModifier(PChar, PWeapon, Mod::EXTRA_DMG_CHANCE);
-        if (occ_extra_dmg > 0 && occ_extra_dmg_chance > 0)
+        int16 occ_do_triple_dmg = PChar->getMod(Mod::REM_OCC_DO_TRIPLE_DMG) / 10;
+        int16 occ_do_double_dmg = PChar->getMod(Mod::REM_OCC_DO_DOUBLE_DMG) / 10;
+        float occ_extra_dmg = battleutils::GetScaledItemModifier(PChar, PWeapon, Mod::OCC_DO_EXTRA_DMG) / 100.f;
```
teschnei: there needs to be a ranged version of occ do extra damage, otherwise getting empyrean aftermath on a melee weapon would give the double/triple damage to ranged attacks

also, should it be mainhand only as well?

teschnei:
I think weaponskills are missing the occasionally attacks 2-3 times from mythic weapons as well

nfaletra:
I think that covers it. Ended up passing the weapon slot along to CheckForDamageMultiplier. Let me know if there was an existing way to check for that. Didn't seem like it since it was in a utils class.

TeoTwawki:
104 files changed..
So much for 
image

scripts/globals/weaponskills.lua
```diff
@@ -900,6 +902,10 @@ function getMultiAttacks(attacker, numHits)
             bonusHits = bonusHits + 2;
         elseif(chance < doubleRate + tripleRate + quadRate) then
             bonusHits = bonusHits + 1;
+        elseif(chance < oaThriceRate) then
```
teschnei: this isn't right, but after getting thorny to test for me, the whole loop isn't right either. for your case, it needs to be after assassins/warriors, and it should only be when i == 1 (the first hit)

i'll update the formula to be right, then you'll have to merge master into your branch again

teschnei: oh sorry, the position relative to assassins/warriors charge doesn't matter, thats just removing the effect but it still should be mainhand only

PR #4484, opened 2018-01-07
Some small stuff

ymkc:


sql/mob_family_system.sql
```diff
@@ -468,8 +468,7 @@ INSERT INTO `mob_family_system` VALUES (400,'Tegmine',2,'Aquan',0,40,108,120,4,3
 INSERT INTO `mob_family_system` VALUES (401,'Martinet',3,'Arcana',1,40,108,90,1,3,4,3,6,6,5,3,3,3,3,1,1,1,1,1,1,1,1,1.25,1,1,1,2,288,0);
 INSERT INTO `mob_family_system` VALUES (402,'Aqrabuamelu',20,'Vermin',3,40,90,120,3,3,4,3,3,6,5,3,3,3,3,1,1,1,1,1,1.25,1,1,1,0.875,1.25,0.875,4,257,0);
 INSERT INTO `mob_family_system` VALUES (403,'Autarch',3,'Arcana',1,40,108,90,1,3,4,3,6,6,5,3,3,3,3,1,1,1,1,1,1,1,1,1.25,1,1,1,2,288,0);
-INSERT INTO `mob_family_system` VALUES (404,'CureRabbit',6,'Beast',0,40,96,120,3,1,4,2,4,4,6,3,3,3,3,1,1,1,1,1,1,1,1,1.25,1.25,1,1.25,4,257,0);
-INSERT INTO `mob_family_system` VALUES (405,'SnowRabbit',6,'Beast',0,40,96,120,3,1,4,2,4,4,6,3,3,3,3,1,1,1,1,1,1,1,1,1.25,1.25,1,1.25,4,257,1);
+INSERT INTO `mob_family_system` VALUES (404,'CureRabbit',6,'Beast',0,40,96,120,3,1,4,2,4,4,6,3,3,3,3,1,1,1,1,1,1,1,1,1.25,1.25,1,1.25,4,257,1);
```
TeoTwawki: comment the freed up ID
```sql
-- 405 available for use
```
or something

Hozu:
Those two rabbit families shouldn't be merged - all snow rabbits (should) have a family mobmod for +10% EXP gain. I suppose you could apply the mods to the individual mob pools, but it's probably less confusing to just leave it as-is.

TeoTwawki:
I think that'd be better as a pool mod instead of having 2 identical families just to keep a family mod entry. And we should rename a lot of these (the text is arbitrary for human readability). Though I'd have ditched the other ID of the 2 just so some noob doesn't later get confused and make me explain that family ID and skill list ID do not have to match.

But this PR isn't wrong. 

I'd like to see us use subfamilies the way http://ffxiclopedia.wikia.com/wiki/Category:Rabbits does at some point. Any family mod that is specific to a family row that has an identical data set on another row should be moved out of family mods. Heck, I wouldn't mind seeing the entire family mod table itself wipe dout in favor of scripting these via an onSpawn mixin.

ymkc:
@Hozu but those are already pool mods, atleast for the variable and polar hares

sql/mob_family_system.sql
TeoTwawki: *nudge*

PR #4499, opened 2018-01-16
Updated Conquest offsets for December 2017 patch.

### LINKED ISSUE #4498: Xarcabard telepoint is missing
**_I have:_**

30171205_0

**_Server Version_** (type `!revision` in game) **:** 
459685fa63386640ffaa8759f61b76dae3973d29

master

Go to Xarcabard and position where the telepoint is supposed to be, it's not there. NPC looks ok to me in database

zynjec:
Fixes #4498

### DIFF
--- a/scripts/globals/conquest.lua
+++ b/scripts/globals/conquest.lua
@@ -658,7 +658,7 @@ switch (region): caseof {
   ---------------------------------
     --print("RONFAURE");
 
-    local Doladepaiton = 17187525;
+    local Doladepaiton = 17187523;
 
     npc  = {
     --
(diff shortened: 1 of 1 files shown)


scripts/globals/conquest.lua
```diff
@@ -1386,4 +1386,4 @@ function conquestUpdate(zone, player, updateType, messageBase)
             end
         end
     end
-end;
+end;
```
ShelbyZ: No newline, what world we live in.

TeoTwawki: I'd say something about how we should leave those whatever they are as we find'em but my auto format tools keep adding it without my noticing so I'd be a hypocrite ^^ some day I'll look into isolating that feature from my dsp local copy >.>

PR #4500, opened 2018-01-16
Riverne A01 portal IDs

DFPercush:
Some spatial displacements in Riverne A01 send you to the wrong destination, after a certain client update, making it difficult to reach monarch linn for CoP mission. Mapped new IDs, changed npc script and tested all paths both ways. (Edit: A particular unstable displacement appears to not be coded, but seems like a useless dead end anyway)

### DIFF
--- a/scripts/zones/Riverne-Site_A01/npcs/Spatial_Displacement.lua
+++ b/scripts/zones/Riverne-Site_A01/npcs/Spatial_Displacement.lua
@@ -25,67 +25,9 @@ function onTrigger(player,npc)
         player:startEvent(3);
     elseif (id == base+2) then 
         player:startEvent(4);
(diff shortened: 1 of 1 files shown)


DFPercush:
auto build needs to fix its internet, not my fault. But closing this one and submitting a new PR which includes Riverne B01 as well.

TeoTwawki:
You can send any change to the same branch of the same fork and github will automatically update the open pull request. If you do something that changes the history you can use a force push. Even if you "start over" for some reason, you can force push over the branch and github will auto-magically put the same thing into the pull request that you have on the copy you were working on.



tl;dr So in the future, avoid closing them and if you run into problems fixing it ask us for help if need be.

Issue #4281, opened 2017-11-25
Tame shouldn't remove sneak or invisible


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30171103_1

**_Server Version_** (type `!revision` in game) **:** 2fe3c08

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
In retail, Tame does not remove sneak or invisible. Confirmed this many many times during my bout of Taming the same mob for 8 hours recently.
In dsp, both sneak and invisible popped off when I attempted to tame a mob.

Wiggo32:
This has been corrected: https://github.com/DarkstarProject/darkstar/pull/4518

PR #4514, opened 2018-01-25
Added ABILITY_TAME to the If include keeping sneak/invis on

Wiggo32:
Takes care of https://github.com/DarkstarProject/darkstar/issues/4281

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -927,7 +927,7 @@ void CCharEntity::OnAbility(CAbilityState& state, action_t& action)
         }
 
         // remove invisible if aggresive
-        if (PAbility->getID() != ABILITY_FIGHT)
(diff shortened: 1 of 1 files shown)


src/map/entities/charentity.cpp
```diff
@@ -927,7 +927,7 @@ void CCharEntity::OnAbility(CAbilityState& state, action_t& action)
         }
 
         // remove invisible if aggresive
-        if (PAbility->getID() != ABILITY_FIGHT)
+        if (PAbility->getID() != ABILITY_TAME || PAbility->getID() != ABILITY_FIGHT)
```
bendangelo: This should be `&&` right? If the ability is not either then remove invisible.

Wiggo32: This is the exact coding that kj told me to write and PR.  Ask him if && would be better...

teschnei:
Yes, it should be and

Wiggo32:
I dont know what I did... im making a new one...

PR #4519, opened 2018-01-27
sky portals / pincerstone, typo npcId -> portalId

DFPercush:
client /ver 30171205_0
GIT_VER 761961e4f484c6648da0d07f0edbc280ffce13ef (master)

Noticed that at least some of the pincerstones in sky (Ru'Aun Gardens) were not activating the teleport pads. Script was attempting to index a nil value on line 21 because tried to use (npcId - 1) instead of (RUAUN_PINCERSTONES[npcId] - 1). Simple one line fix. Tested going around the circle.

### DIFF
--- a/scripts/zones/RuAun_Gardens/npcs/Pincerstone.lua
+++ b/scripts/zones/RuAun_Gardens/npcs/Pincerstone.lua
@@ -18,7 +18,7 @@ function onTrigger(player,npc)
     if (portalId ~= nil) then
         local portal = GetNPCByID(portalId);
         if (portal:getAnimation() == ANIMATION_CLOSE_DOOR) then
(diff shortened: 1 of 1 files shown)


wrenffxi:
If this code isn't working it's not because npcId needs to be portalId.  ncpId is correct.

npcId is the id of the thing you press.
npcId - 1 is the id of the color change / whoosh for the pincerstone.
portalId is the id of the portal you open.

When you push the npc, both npcId-1 and portalId NPCs should openDoor.

So if this code doesn't work, issue is either missing/inccorect NPC ID in npclist.sql, or out-of-sync NPC IDs between npclist.sql and scripts/zones/RuAun_Gardens/MobIDs.lua.

PR #4532, opened 2018-02-01
Disable AH and mailbox menu's while in jail.

TeoTwawki:
It is possible (easy even) to get the client to bring up these menus while in jail even if the players AI is disabled.

We should in the future look into handling the fact you can pop up any menu any place at any time in general..

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -3811,7 +3811,7 @@ void SmallPacket0x0AD(map_session_data_t* session, CCharEntity* PChar, CBasicPac
 
 void SmallPacket0x0B5(map_session_data_t* session, CCharEntity* PChar, CBasicPacket data)
 {
-    if (data.ref<uint8>(0x06) == '!' && CmdHandler.call(PChar, (const int8*)data[7]) == 0)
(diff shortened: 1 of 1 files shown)


src/map/packet_system.cpp
```diff
@@ -3801,7 +3811,7 @@ void SmallPacket0x0AD(map_session_data_t* session, CCharEntity* PChar, CBasicPac
 
 void SmallPacket0x0B5(map_session_data_t* session, CCharEntity* PChar, CBasicPacket data)
 {
-    if (data.ref<uint8>(0x06) == '!' && CmdHandler.call(PChar, (const int8*)data[7]) == 0)
+    if (data.ref<uint8>(0x06) == '!' && CmdHandler.call(PChar, (const int8*)data[7]) == 0 && !jailutils::InPrison(PChar))
```
takhlaq: may as well stick the jail check before the command handler call

teschnei:
It doesn't, but we can save some processing by not bothering to load and
compile the Lua if we know it can't call it anyways

TeoTwawki:
> save some processing by not bothering to load and compile the Lua if we know it can't call it anyways

that sounds like it matters :)

PR #4535, opened 2018-02-03
THF Job Abilities/Traits/Merits up to 75

nfaletra:
Went through all THF job abilities, traits, and merits up to level 75.

* Fixed Group 2 merits that were not implemented or not functioning correctly
* Added mods to enhance specific THF job ablities. Some were just adding mods to gear in item_mods table or needed some C++ implementation

* Also fixed the Job Traits list from displaying group 2 merits that had not been merited yet
** This required adding a column to the traits table to handle the meritID it is linked to

* Replaced distance() with distanceSquared() where applicable. sqrts are expensive, boys.

Note: Feint is functioning as it did in retail prior to TH 2.0 (TH Procs). If/when TH 2.0 is implemented, modifying Feint to affect the proc rate should be trivial

Will work on 75+ after this. Probably a PR per ability, trait, etc. to make things easier on reviewers.


nfaletra:
You know what? I forgot to fix the resist check in steal.lua. Need to figure out a way to check dispel resistance in a job ability.

Hozu:
Couple of concerns.

1. Any crit rate modification you did also needs to be added to the weaponskills.lua global script. Otherwise they won't affect weaponskills. Same with the SNEAK_ATK_DEX thing, I believe.

2. The stealStatusEffect function was originally used for Thinker mobs, which can steal food, unlike the Aura Steal effect. Not sure of the best way to address that.

3. Perhaps the Hide duration mod should be stored as a value out of 100? In the event that any future modifiers affect by a duration other than +100%.

maxtherabbit:
2) could be resolved easily enough by adding an optional argument that
defaults to true for "able to steal food"

nfaletra:
1) I'll make sure that gets added
2) Sounds like we need a new EFFECTFLAG for Aura Steal (stealStatusEffect takes in an EFFECTFLAG and defaults to EFFECTFLAG_DISPELABLE. I'll make sure the Thinker mob abilities still function, too. StealStatusEffect was pretty crashy before my changes.
3) Easy enough to change to keep it consistent with other percentage modifiers

Thanks for the feedback.

takhlaq:
@teschnei watdo

nfaletra:
Sorry guys, I've been busy. I need to finish setting up a new Effect Flag for Aura Steal in effects sql and this should be ready.

takhlaq:
sorry wasnt meaning to rush you, thought it was waiting for review

i'll close for now so we dont keep pinging you about it and you can drop a comment when its ready so we can reopen it and review

PR #4551, opened 2018-02-11
Removed revision related stuff

teschnei:
They are pretty useless in git, assuming people are using it properly


TeoTwawki:
Need to update the template that tells them to give this to us on bug reports then.

Lets use DSP_VER and have it match whasf's releases maybe?

teschnei:
Yeah, I was going to update that if this got merged.  We can talk about a
replacement in discord before merging this, I just needed to get it out of
my local copy for now

PR #4558, opened 2018-02-14
Animations (Unstable Displacements, Diabolos Tiles, Dangruf Geysers)

wrenffxi:


src/map/lua/luautils.cpp
```diff
@@ -285,11 +285,7 @@ namespace luautils
 
     int32 SendEntityVisualPacket(lua_State* L)
     {
-        if ((!lua_isnil(L, 1) && lua_isnumber(L, 1)) &&
-            (!lua_isnil(L, 2) && lua_isnumber(L, 2)) &&
-            (!lua_isnil(L, 3) && lua_isnumber(L, 3)) &&
-            (!lua_isnil(L, 4) && lua_isnumber(L, 4)) &&
-            (!lua_isnil(L, 5) && lua_isnumber(L, 5)))
```
TeoTwawki: not sure why we had args 5 being checked for not nil and is number if we only even use 2 args.

takhlaq:
hmmm not sure if it's worth moving the door trade/trigger stuff to globals like in your psoxja pr...
@teschnei @TeoTwawki watdo

takhlaq:
actually, yeah might as well

TeoTwawki:
My opinion is if we need a lua function or wrapper for a bunch of things (for exmaple) to call in the onTrigger, instead of just using onTrigger, then its worth a global. I think this PR is fine like it is.

TeoTwawki:
Meh, I thought it was short enough code to not need it but Wren did it anyway. I'm cool with it either way.

PR #4568, opened 2018-02-20
pso'xja entity IDs

wrenffxi:
remove IDs from entity luas
indentation


takhlaq:
from a quick look through, there's a lot of duplicate code which could instead be put into a global function
@teschnei maybe we should make a `scripts/zones/ZONE_NAME/[globals|helpers]/` directory for anything not quest specific and have npcs/mobs require those?

takhlaq:
i say not quest specific since when quests are rewritten their helper functions will also be in the quest script itself

teschnei:
i honestly don't have an issue with that suggestion. scripts/zones/<zone>/globals.lua.  Honestly then we could just require textids and mobids from globals.lua and have all the script inside the zone just require the zones global.lua

but just creating a global.lua in the zone's subdirectory for now with common functions is good enough

TeoTwawki:
if we do that lets nix the requirement of having to nil things at the same time.

TeoTwawki:
moving the text IDs into it I mean

teschnei:
We can't do that without a large scale replacement. Let's not discuss it
here

TeoTwawki:
> We can't do that without a large scale replacement. Let's not discuss it here

the textId moving into globals.lua would need a mass replace anyway because of all the existing textId usages

teschnei:
Yes, when we get there. For now I am okay with just this one zone

takhlaq:
dat lines added/removed ratio 👌👌👌👌

PR #4595, opened 2018-02-25
Ruminator/Brooder mob_pools adjustments

neuromancerxi:
Brooder and Ruminator in Aby-La Theine were both missing models, family, aggro, true_detection, and skill lists. Mobs are both Thinker Family, applied Thinker family, modelid, aggro, true_detection and skill list to Brooder and Ruminator.

Verified Model appearance against Stable Branch and client ver 30171205_0.


TeoTwawki:
Probably never got added due to missing abyssite pop system for it. I might do that soonish.

PR #4599, opened 2018-02-27
Outer Horutoto Ruins NMs, entity IDs

wrenffxi:
Outer Horutoto IDs removed from entity luas.
Bomb King/Doppelgangers no more server vars.
Added job special to mission NMs.
Added sleep + silence immunity to Queen of Coins.


sql/mob_spawn_points.sql
```diff
@@ -55678,12 +55678,12 @@ INSERT INTO `mob_spawn_points` VALUES (17572138,'Three_of_Batons','Three of Bato
 INSERT INTO `mob_spawn_points` VALUES (17572139,'Three_of_Swords','Three of Swords',9716,0,0,0,127);
 INSERT INTO `mob_spawn_points` VALUES (17572140,'Three_of_Coins','Three of Coins',9714,0,0,0,127);
 INSERT INTO `mob_spawn_points` VALUES (17572141,'Balloon','Balloon',9658,509,-1,709,18);
-INSERT INTO `mob_spawn_points` VALUES (17572142,'Bomb_King','Bomb King',9662,0,0,0,98);
-INSERT INTO `mob_spawn_points` VALUES (17572143,'Doppelganger_Gog','Doppelganger Gog',9668,0,0,0,32);
-INSERT INTO `mob_spawn_points` VALUES (17572144,'Doppelganger_Dio','Doppelganger Dio',9667,310,-1,709,2);
+INSERT INTO `mob_spawn_points` VALUES (17572142,'Bomb_King','Bomb King',9662,509,-1,709,98);
```
TeoTwawki: Never even realized there were multiple copies :thinking: I wonder if the reason I was able to nail these NM so much on retail was there were actually 2 separate cooldowns the entire time..

TeoTwawki: Looks just like the leaping lizzy scenario everyone on wiki argued about

PR #4601, opened 2018-02-27
Escape the user and password strings before stuffing into queries

TeoTwawki:
I'm sick and tired of people claiming they send this project tons of fixes that we just sit on. The only person I have ever seen *even mention* this was atom0s. You who doubt, search the issues and pull requests yourself.

Note:

This can probably be written better, I did it as a quick thing on demiurge recently and hadn't asked for DSP's C++ ppl to look it over yet. I'm not that great in C++ land

### DIFF
--- a/src/login/login_auth.cpp
+++ b/src/login/login_auth.cpp
@@ -184,7 +184,7 @@ int32 login_parse(int32 fd)
                 }
 
                 if (numCons > 1) {
-                    ShowInfo("login_parse:" CL_WHITE"<%s>" CL_RESET" has logged in %i times! Removing older logins.\n", name.c_str(), numCons);
+                    ShowInfo("login_parse:" CL_WHITE"<%s>" CL_RESET" has logged in %i times! Removing older logins.\n", escaped_name, numCons);
                     for (int j = 0; j < (numCons - 1); j++) {
                         for (login_sd_list_t::iterator i = login_sd_list.begin(); i != login_sd_list.end(); ++i) {
                             if ((*i)->accid == sd->accid) {
@@ -197,13 +197,13 @@ int32 login_parse(int32 fd)
                 }
                 //////
 
-                ShowInfo("login_parse:" CL_WHITE"<%s>" CL_RESET" was connected\n", name.c_str(), status);
+                ShowInfo("login_parse:" CL_WHITE"<%s>" CL_RESET" was connected\n", escaped_name, status);
                 return 0;
             }
             else {
                 session[fd]->wdata.resize(1);
                 ref<uint8>(session[fd]->wdata.data(), 0) = LOGIN_ERROR;
-                ShowWarning("login_parse: unexisting user" CL_WHITE"<%s>" CL_RESET" tried to connect\n", name.c_str());
+                ShowWarning("login_parse: unexisting user" CL_WHITE"<%s>" CL_RESET" tried to connect\n", escaped_name);
                 do_close_login(sd, fd);
             }
         }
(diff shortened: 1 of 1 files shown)


RunAwayDSP:
I'll admit i have never sent this issue even while knowing about it for years, my apologies about not informing of a security risk in lobby. The most I have done is discuss it with other server admins. I am happy to see the issue being dealt with.

src/login/login_auth.cpp
```diff
@@ -107,7 +107,9 @@ int32 login_parse(int32 fd)
             const char* fmtQuery = "SELECT accounts.id,accounts.status \
 									FROM accounts \
 									WHERE accounts.login = '%s' AND accounts.password = PASSWORD('%s')";
-            int32 ret = Sql_Query(SqlHandle, fmtQuery, name.c_str(), password.c_str());
+            Sql_EscapeString(SqlHandle, chrName, chrName);
```
teschnei: Sql_EscapeString's comments: 
`/// The output buffer must be at least strlen(from)*2+1 in size.`
Because of this, it is probably not safe to use the same buffer as input and output.  We probably need to allocate a new buffer with the size of name.length()*2 +1 (and same for password)
we would probably want something like
`char[16*2+1] escaped_name` and so on

I don't have a compiler in github obviously, so just message me on discord if you have trouble getting that working

TeoTwawki: :+1:

TeoTwawki:
Updated with safer buffer

TeoTwawki:
tested still working

src/login/login_auth.cpp
```diff
@@ -248,7 +253,7 @@ int32 login_parse(int32 fd)
                 fmtQuery = "INSERT INTO accounts(id,login,password,timecreate,timelastmodify,status,priv)\
 									   VALUES(%d,'%s',PASSWORD('%s'),'%s',NULL,%d,%d);";
 
-                if (Sql_Query(SqlHandle, fmtQuery, accid, name.c_str(), password.c_str(),
+                if (Sql_Query(SqlHandle, fmtQuery, accid, escaped_name, escaped_pass,
```
teschnei: isn't escaped_pass empty here? it was only escaped inside LOGIN_CREATE

TeoTwawki: fixed. dunno how I lost the line. was pushing from the build I was using.

TeoTwawki:
Wait..Are we sure password max length even is enforced? I don't think it is.

teschnei:
what do you mean? it only reads 32 characters in memory

TeoTwawki:
> what do you mean? it only reads 32 characters in memory

we should prolly limit xiloaders field then, is what I mean

Cloudef:
tbh, should use prepared sql statements instead of escaping.

src/login/login_auth.cpp
```diff
@@ -107,7 +109,9 @@ int32 login_parse(int32 fd)
             const char* fmtQuery = "SELECT accounts.id,accounts.status \
 									FROM accounts \
 									WHERE accounts.login = '%s' AND accounts.password = PASSWORD('%s')";
-            int32 ret = Sql_Query(SqlHandle, fmtQuery, name.c_str(), password.c_str());
+            Sql_EscapeString(SqlHandle, escaped_name, name.c_str());
```
Cloudef: Does this function assume output buffer is of same length as input buffer? Sounds bit unsafe.
I would at least stick assert before to make sure there's no buffer overflow.

teschnei: it assumes the output buffer is at least 2 times plus 1 bigger than the input (which is what I was talking about earlier)

teschnei:
prepared statements would be nice, but I don't think anyone has time to move things to that

TeoTwawki:
> prepared statements would be nice, but I don't think anyone has time to move things to that'

Yeah I'd like us to move to that in the future but this was quick

PR #4613, opened 2018-03-01
Add capability to log GM command use to database.

TeoTwawki:
Note: if this is merged then from here out we require a minimum of **mysql 5.6,5** or **MariaDB 10.0.1** as anything older doesn't allow a default value of CURRENT_TIME on datestamp columns.


TeoTwawki:
woops I forgot to fix that buffer size..will reopen after I do that

PR #4614, opened 2018-03-02
COR: Roll Clean-up, add unimplemented Rolls 65+

neuromancerxi:
Adds the Following Rolls:
Naturalist's Roll*
Runeist's Roll
Bolter's Roll
Caster's Roll
Courser's Roll***
Blitzer's Roll
Tactician's Roll
Allie's Roll
Miser's Roll
Companion's Roll**
Avenger's Roll***

scripts:         Adds/Updates Item, Ability, Effect Scripts.
scripts:         Adds/Updates Ability, Modifier, Effect Globals.
scripts/src:     Add MOD for SAVETP, PHANTOM_ROLL, and PHANTOM_ROLL_DURATION
src:             Adds/Updates Ability, Modifier, Status Effects to map
src:             Expands m_LearnedAbilities to account for new dice out of range in map/charentity.h.
src:             Adds Logic and MOD for SAVETP to Weaponskill TP routine in map/utils/battleutils.cpp.
sql:             Adds/Updates abilities, item_usable, status_effects.
sql:             Modifies Animations for Dice to Pouch instead of a Firework Animation.
sql:             The actual Animation ID for Dice usage doesn't appear to be in the index for item usage.
sql:             Refactored animations based on Youtube and assumed sequential increase of animation id by roll.
sql:             Corrected Snake Eye Animation to use "Roman I" animation.
src:             Modifies Logic for COR Roll Ability/Effect Checking to account for new Rolls.
scripts/src:     Modifies Logic for Eleven Roll to account for new Rolls Range and outlier stand alone Roll.
scripts:         Cleans up ID, Name, and Description for Legacy Dice (most had ID 5477/Warrior Die/desc instead of correct detail).
scripts/src/sql: Corrected Reference from RUNEIST_S to RUNEISTS.

notes:

  * Naturalist's Roll Item, Ability, and Effect are all implemented.
  * Naturalist's Roll Effect is empty currently due to unimplemented Enhancing Magic Duration Logic (TODO).
 ** Companion's Roll gives two effects. To maintain consistency, Roll power is 1-11, and Corresponding values for mods are in the Companion's Roll Effect instead.
*** No Trusted Community Sources have values for these abilities. Used comparable rolls to adapt values (Fighters > Avengers, Blitzers > Coursers)

Tested All Rolls against a clean build, validated effects being applied using test script to echo value of associated MOD before and after roll.
Spot checked Lucky 11 rolls against new effects.
Tested SAVETP effect via Weaponskill usage.

TODO:

Add Logic to ability scripts to account for Phantom Roll + (SOA Rings) and Phantom Roll Duration (Effect from Armor).


scripts/globals/ability.lua
```diff
@@ -469,10 +476,14 @@ function checkForElevenRoll(caster)
     local effects = caster:getStatusEffects()
     for _,effect in ipairs(effects) do
         if (effect:getType() >= EFFECT_FIGHTERS_ROLL and
```
teschnei: can't do this since naturalists and runeists arent sequential with the rest (doing it this way lets like 200 abilities pass this check).  It should probably be like (>= FIGHTERS and <= TACTICIANS) or (>= NATURALISTS and <= RUNEISTS)

teschnei: and then you don't need the runeists roll block after

src/map/status_effect_container.cpp
```diff
@@ -857,8 +857,8 @@ bool CStatusEffectContainer::ApplyBardEffect(CStatusEffect* PStatusEffect, uint8
 bool CStatusEffectContainer::ApplyCorsairEffect(CStatusEffect* PStatusEffect, uint8 maxRolls, uint8 bustDuration)
 {
     //break if not a COR roll.
-    DSP_DEBUG_BREAK_IF(!(PStatusEffect->GetStatusID() >= EFFECT_FIGHTERS_ROLL &&
-        PStatusEffect->GetStatusID() <= EFFECT_SCHOLARS_ROLL));
+    DSP_DEBUG_BREAK_IF(!((PStatusEffect->GetStatusID() >= EFFECT_FIGHTERS_ROLL &&
+        PStatusEffect->GetStatusID() <= EFFECT_NATURALISTS_ROLL) || (PStatusEffect->GetStatusID() == EFFECT_RUNEISTS_ROLL)));
```
teschnei: same here - naturalists is not sequential with the other rolls - it needs to be tacticians, and then separately check both naturalists and runeists

I think it would be easier if we just added an isCorsairRoll to... battleutils maybe? and then have them all just call that instead so that all this logic is in the same place

src/map/utils/battleutils.cpp
```diff
@@ -2210,8 +2210,16 @@ namespace battleutils
...
             if (primary)
-            {
-                PChar->addTP((int16)(((tpMultiplier * baseTp) + bonusTP) * (1.0f + 0.01f * (float)((PChar->getMod(Mod::STORETP) + getStoreTPbonusFromMerit(PChar))))));
+
+            // Check if SAVETP value is more than WS would return and use it instead if so.
+            { 
+                int16 standbyTp = ((int16)(((tpMultiplier * baseTp) + bonusTP) * (1.0f + 0.01f * (float)((PChar->getMod(Mod::STORETP) + getStoreTPbonusFromMerit(PChar))))));
+                int16 savedTp = (PChar->getMod(Mod::SAVETP));
+
+                if (savedTp >= standbyTp)
```
teschnei: this part (the block that adds the TP) needs to be outside the `damage > 0` block (line 2240 ish) or else it won't apply on misses.  additionally, you can shorten the adding to simply `PChar->addTP(std::max(savedTp, standbyTp));`

neuromancerxi: Thanks for the feedback. Will test suggestion and submit an additional commit after testing code.

scripts/globals/effects/naturalists_roll.lua
```diff
@@ -0,0 +1,26 @@
+-----------------------------------
+--
+--
+--
+-----------------------------------
+
+-----------------------------------
+-- onEffectGain Action
```
TeoTwawki: Request: don't add the little function headers. I recently mass removed them from the project because all they do is repeat the function name.

TeoTwawki: File headers good, function headers bad :+1:

scripts/globals/items/beastmaster_die.lua
```diff
@@ -1,7 +1,7 @@
 -----------------------------------------
--- ID: 5477
--- Warrior Die
--- Teaches the job ability Fighter's Roll
+-- ID: 5485
+-- Beastmaster Die
```
TeoTwawki: Thank you for fixing that

scripts/globals/status.lua
```diff
@@ -715,7 +715,7 @@ EFFECT_VOIDSTORM_II             = 596
 EFFECT_INUNDATION               = 597
 EFFECT_CASCADE                  = 598
 EFFECT_CONSUME_MANA             = 599
-EFFECT_RUNEIST_S_ROLL           = 600
+EFFECT_RUNEISTS_ROLL            = 600
```
TeoTwawki: And thank you for fixing these 2. Hates the _S_ 's

TeoTwawki:
functionality/code wise I didn't spot any script problems

scripts/globals/effects/naturalists_roll.lua
neuromancerxi: I can address that in my next PR. Since they were present in the existing cor roll effects I kept them in for consistency sake.

Issue #4026, opened 2017-08-30
Correction required for Magic Burst multiplier calculations


Whitechaser:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** **:** 30170707_0

**_Server Version_** **:** Revision is: Unknown

**_Source Branch_** **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
References: https://www.bg-wiki.com/bg/Magic_Burst#Enhancing_Magic_Bursts
https://www.bg-wiki.com/bg/Mag._Burst_Bonus

function calculateMagicBurst() (line 689) in magic.lua calculates the multipler for magic bursts via (1) getting the base multiplier dependent on the skillchain count then (2) *adding* the MB bonus from BLM AM2 spells and the additional % damage bonus from the job trait and gear, when it should be *multiplying* the value in (1) to the total value in (2).

I'm still learning how to submit things (just installed git yesterday), but if anyone could just add one more variable (something like "burstbonus", initialized to 1.00) and then multiplying that to "burst" at the end before returning it, that should implement it nicely.

Basically the 1.something gotten up to line 714 should be multiplied to a 1.something obtained by adding the (caster:getMod(MOD_MAG_BURST_BONUS) / 100) and the (caster:getMerit(blmAMIIMerit[spell:getElement()]) - 1) * 0.03 for AM2, if applicable.

PR #4620, opened 2018-03-04
Reworked and corrected magic burst calculations, simplified AMII merit implementation

Whitechaser:
Old PR for those who wish to see previous discussion: #4056 
Fix for #4026 
References: 
https://www.bg-wiki.com/bg/Magic_Burst#Enhancing_Magic_Bursts
https://www.bg-wiki.com/bg/Mag._Burst_Bonus

- Added new parameters for BLM AMII merits in their respective scripts
- Made function addBonuses()'s parameters uniform for elemental, ninjutsu, and divine nukes so NIN bonuses can be passed on the same table as BLM bonuses for their spells
- MB multiplier from gear/atma/AMII merits now calculated separately from skillchain multiplier and capped at 40% damage bonus
- Corrected multipliers for increasing number of weaponskills participating in a chain (additional 10% damage bonus per WS)
- MB bonus from gear/atma/AMII merits is now multiplied instead of simply added to the MB bonus from the skillchain
- Made function getSpellBonusAcc()'s parameters include params table from spell scripts
- Removed old code for inserting BLM AMII merits, and deleted the merit table that used to be in line 60

### DIFF
--- a/scripts/globals/magic.lua
+++ b/scripts/globals/magic.lua
@@ -741,9 +741,7 @@ function addBonuses(caster, spell, target, dmg, params)
     local affinityBonus = AffinityBonusDmg(caster, ele);
     dmg = math.floor(dmg * affinityBonus);
 
-    if (params.bonusmab == nil) then
-        params.bonusmab = 0;
-    end
+    params.bonusmab = params.bonusmab or 0;
 
     local magicDefense = getElementalDamageReduction(target, ele);
     dmg = math.floor(dmg * magicDefense);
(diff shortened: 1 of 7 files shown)


scripts/globals/magic.lua
```diff
@@ -696,57 +687,62 @@ function adjustForTarget(target,dmg,ele)
...
-function addBonuses(caster, spell, target, dmg, bonusmab)
+function addBonuses(caster, spell, target, dmg, params)
     local ele = spell:getElement();
 
     local affinityBonus = AffinityBonusDmg(caster, ele);
     dmg = math.floor(dmg * affinityBonus);
 
-    if (bonusmab == nil) then
-        bonusmab = 0;
+    if (params.bonusmab == nil) then
```
takhlaq: could also do
```lua
params.bonusmab = params.bonusmab or 0
```
instead of the if block

scripts/globals/spells/burst_ii.lua
```diff
@@ -26,6 +26,8 @@ function onSpellCast(caster, target, spell)
     spellParams.M50 = 2;
     spellParams.M100 = 2;
     spellParams.M200 = 2;
+    spellParams.AMIIburstBonus = (caster:getMerit(MERIT_BURST_II) - 1) * 0.03;
+    spellParams.AMIIaccBonus = (caster:getMerit(MERIT_BURST_II) - 1) * 5;
```
Hozu: Shouldn't these ensure that the bonuses don't become negative if the caster has no merits? ie. mobs casting AMIIs.

Whitechaser: I vaguely remember someone mentioning that mobs by default had max merits, but I don't know if that applies for job-specific ones. I'll add it in though.

davismj:
All my MB recently are hitting for 0. Particularly, Fire II on Fusion.

Issue #4223, opened 2017-11-12
Somnolence has damage calculation but doesn't subtract it from target


Whitechaser:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30170905_05

**_Server Version_** (type `!revision` in game) **:** Unknown

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
I think [this](https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/abilities/pets/somnolence.lua) is basically missing a:

`target:delHP(dmg);`
`target:updateEnmityFromDamage(pet,dmg);`

One thing I couldn't figure out though is why it says target "takes 0 points of damage" when I use it. I'm assuming the returned value is for the message? Other than that it won't be doing any damage at all.

TeoTwawki:
neither of those would effect the message. the damage number is being returned here:
https://github.com/DarkstarProject/darkstar/blob/5078b8b1b71c30af4a5fbcb9075787e5a9a3cd3f/scripts/globals/abilities/pets/somnolence.lua#L35
so that variable `dmg` would have to be zero by that point (or nil, since nil would also get treated as zero). **There is probably a console log error on the server you play on.**

PR #4621, opened 2018-03-04
Added missing lines for Somnolence damage

Whitechaser:
Fix for #4223

### DIFF
--- a/scripts/globals/abilities/pets/somnolence.lua
+++ b/scripts/globals/abilities/pets/somnolence.lua
@@ -32,5 +32,7 @@ function onPetAbility(target, pet, skill)
         target:addStatusEffect(EFFECT_WEIGHT, 50, 0, duration);
     end
 
+    target:delHP(dmg);
+    target:updateEnmityFromDamage(pet,dmg);
     return dmg;
 end


TeoTwawki:
@takhlaq @Whitechaser 
https://github.com/DarkstarProject/darkstar/blob/157cb9a619ece8dd4c9530c75794183709b5c26b/scripts/globals/magic.lua#L648
```
target:delHP(dmg);
```
That looks like finalMagicAdjustments already removes HP there and updates enmity to lines below it.

Whitechaser:
Oh! You're right! You'll have to revert this, so sorry.

The issue persists then. Somnolence isn't doing damage to mobs.

I found something just now though. You don't suppose it could be https://github.com/DarkstarProject/darkstar/blob/5078b8b1b71c30af4a5fbcb9075787e5a9a3cd3f/scripts/globals/abilities/pets/somnolence.lua#L17 not having a semicolon at then end, do you?

TeoTwawki:
semicolons are fully ignored in lua lamguage. my bet is on something breaking the script before it gets to the the dmg so it never execs that line or the dmg calc itself getting screwed

Whitechaser:
I'll try following the damage calc then. Hopefully I can find where it's going wrong.

PR #4624, opened 2018-03-05
fixed magic.lua

takhlaq:


takhlaq:
a few fields not initialised in https://github.com/DarkstarProject/darkstar/pull/4620

scripts/globals/magic.lua
```diff
@@ -537,7 +540,7 @@ function getSpellBonusAcc(caster, target, spell, params)
     end
 
     --Add acc for klimaform
-    if (caster:hasStatusEffect(EFFECT_KLIMAFORM) and (castersWeather == singleWeatherStrong[element] or castersWeather == doubleWeatherStrong[element])) then
+    if (caster:hasStatusEffect(EFFECT_KLIMAFORM) and (castersWeather == singleWeatherStrong[params.element] or castersWeather == doubleWeatherStrong[params.element])) then
```
teschnei: should we check if params.element > 0? 0 index on these arrays does not make sense

takhlaq: we should

PR #4642, opened 2018-03-10
Howling Moon and monstertpmoves tweaks

Wiggo32:
1. Added the Day/Weather bonus to Howling Moon when cast by Fenrir Prime.
-Tested on Darksday, Lightsday, and Firesday and noted proper damage adjustments based on day.

2. mobAddBonuses had an incorrect addition in the "dayWeak[ele]" section, changed to subtraction.


ShelbyZ:
If we get these changes approved, it may benefit to have these applied to other weather-effected monster moves.

Wiggo32:
Yea, I was already going to work on them. But not all at once. I'll have more time in the upcoming week. The fix for the monstertpmoves will affect the SMN avatar skills already in place and make them behave appropriately.

PR #4656, opened 2018-03-12
Implement packet handling for mounts, now fully working.

Ivaar:


src/map/packets/char_recast.cpp
```diff
@@ -45,7 +45,12 @@ CCharRecastPacket::CCharRecastPacket(CCharEntity* PChar)
     {
         uint32 recasttime = (recast.RecastTime == 0 ? 0 : ((recast.RecastTime - (uint32)(time(0) - recast.TimeStamp))));
 
-        if (recast.ID != 0)
+        if (recast.ID == 256) // borrowing this id for mount recast
```
teschnei: what do you mean by borrowing?  if this ID (256) is written to 0xFE, doesn't that mean that it is actually 256?

scripts/globals/status.lua
```diff
@@ -2333,24 +2333,25 @@ ANIMATION_MOUNT                    = 85;
...
-    BEETLE         = 10,
-    MOOGLE         = 11,
-    MAGIC_POT      = 12,
-    TULFAIRE       = 13,
-    WARMACHINE     = 14,
-    XZOMIT         = 15,
-    HIPPOGRYPH     = 16,
-    SPECTRAL_CHAIR = 17,
-    SPHEROID       = 18
+    QUEST_RAPTOR   = 1,
```
TeoTwawki: that explains the off by one we were discussing

odd that the quest one needed its own id.

src/map/packets/char_recast.cpp
Ivaar: Well theres 8 bytes for each ability and the 4th byte is ability id.
They added 8 more bytes for mount recast and 4th byte is set to 0x01
Before calling a mount its just 0x00 and after recast expires the 0x01 hangs around
I just tested it without setting 0xFF to one and it still functions like retail.

Ivaar: I figured I could shift the recast id one byte to make it match retail packet. If mount recast really is id 256 then it breaks the packets structure where one byte is used for recast id and 4 bytes of junk follow. Anyways, now that we know the client doesn't need that byte set, should I delete it?

teschnei: actually, going off what I'm seeing on windower's documentation on the packet, I think the abilityID is actually a uint16 and the recasttime is also a uint16.  Currently we're writing 4 bytes for the timestamp, and then overwriting the last byte with the ID.  But according to windower, the third byte of the timestamp is always zero (for abilities, of which.. none have a recast ID over 255).  So it seems very likely to me that a) mounts actually are recastid over 255 and b) both should write recasttime as a uint16 and then ability ID as another uint16

scripts/globals/status.lua
Ivaar: The old enums actually match what the client is sending in 0x01A. When dealing with animations and etc they pop quest raptor in after chocobo...

It might be better to leave the enums the way they were and add quest raptor only to status.lua, I think thats where it actually belongs.
Then I could just define mount_key_item_offset somewhere and add it to mounmt enum.

The quest mount has different behavior, whenever I start scripting the quest I will need access to the updated status.lua

Ivaar: Either way a 1 needs to be added to the param coming from 0x01A if mountID greater than 1, before setting the power of the mount status effect, and so far I haven't needed to use enums.

src/map/packets/char_recast.cpp
Ivaar: That's all correct, but if mount is recast 256 then they're shifting it to start at the 3rd byte instead of the 4th byte to make room for an uint16, only for mounts, while the rest of the ability ids are stored in the 4th byte.

Ivaar: Yeah there's no way ability ID is uint16, just possibly mount ID. I recall seeing the following 4 bytes full of junk.

teschnei:
Yeah I realise now the endianness would be backwards for that.  Oh well, I
guess just leave it how it is

src/map/packets/char_recast.cpp
Ivaar: ability_recasts

Ivaar: Since ability recast IDs stop at 255, 256 was a safe value to borrow unless they add more abilities, and exceed the unused IDs.

Ivaar: If they do that we have to fix this anyways, so I think its good can even remove the comment.

teschnei: yep, agree

Ivaar:
So DSP doesn't set the chocobo flag in char update packet 0x037. When you mount a chocobo on retail it sets FLAG_CHOCOBO found in mmo.h and sends pet info packet 0x067 with uint32 starting at 0x10, containing only that flag.

So far I was only able to test raptor and tiger on retail but the flags are as follows
```
chocobo = 0x00000040
raptor  = 0x00000060
tiger   = 0x00000080
```
It doesn't look like animation changes at all, with or without it. Any idea what these do?

Ivaar:
Well I mean it does this along with setting the flag in char_update packet
pet_info

teschnei:
That's char_sync.cpp packet, not pet info (windower is mislabeled in that
regard, there are multiple packets that are 0x067), if that helps any

PR #4657, opened 2018-03-12
THF Level 75 Job Abilities, Traits, Merits

nfaletra:
* Added ability mods from gear that were missing for level 75 and below abilities
* Implemented Group 2 THF Merits
* Fixed job trait list from showing group 2 merits that are not merited
* Reworked StealStatusEffect to fix crashiness and for use with Aura Steal
* Updated mob TP moves that take advantage of StealStatuEffect

**Note:**
Originally, I had planned to add an Aura Steal effect flag, but this turned out to not be necessary. Aura Steal can only steal buffs that are dispelable. StealStatusEffect can take in an effect flag(s) (default is EFFECTFLAG_DISPELABLE on the C++ side) to handle custom situations. For instance, Binary/Trinary Tap can also steal food buffs. In the future, Larceny (THF's second 1 hour ability) CAN steal 1 hour abilities from mobs (such as Invincible), but that can be handled when Larceny is implemented.


scripts/globals/effects/evasion_down.lua
```diff
@@ -3,28 +3,36 @@
...
-end;
+end
 
 -----------------------------------
 -- onEffectTick Action
 -----------------------------------
-
 function onEffectTick(target,effect)
-end;
+    -- If this is Feint, restore 10 evasion every tick
```
teschnei: how do we know this is from feint?

scripts/globals/weaponskills.lua
```diff
@@ -117,6 +117,10 @@ function doPhysicalWeaponskill(attacker, target, wsID, tp, primary, action, taCh
         if (attacker:hasStatusEffect(EFFECT_INNIN) and attacker:isBehind(target, 23)) then -- Innin acc boost attacker is behind target
             nativecrit = nativecrit + attacker:getStatusEffect(EFFECT_INNIN):getPower();
         end
+        
+        if (attacker:hasStatusEffect(EFFECT_ASSASSINS_CHARGE)) then
+            nativecrit = nativecrit + attacker:getStatusEffect(EFFECT_ASSASSINS_CHARGE):getSubPower()/100;
```
teschnei: doesn't assassin's charge already add the crithitrate mod?

scripts/globals/effects/evasion_down.lua
teschnei: I guess the fact that it has a tick rate at all.. nevermind, seems well enough (I can't think of a reason why any other evasion down would have a tick rate that isn't shedding the effect)

src/map/lua/lua_baseentity.cpp
```diff
@@ -10509,26 +10509,34 @@ inline int32 CLuaBaseEntity::dispelAllStatusEffect(lua_State *L)
 *  Notes   : Returns a Lua table with the information on the Status Effect stolen
 ************************************************************************/
 
+//DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isuserdata(L, 1));
+//DSP_DEBUG_BREAK_IF(lua_isnil(L, 2) || !lua_isnumber(L, 2));
+//
+//CLuaBaseEntity* PEntity = Lunar<CLuaBaseEntity>::check(L, 1);
+//battleutils::DoWildCardToEntity(static_cast<CCharEntity*>(m_PBaseEntity), static_cast<CCharEntity*>(PEntity->m_PBaseEntity), (uint8)lua_tointeger(L, 2));
```
teschnei: forgot to remove debug code?

src/map/lua/lua_baseentity.cpp
```diff
@@ -10509,26 +10509,34 @@ inline int32 CLuaBaseEntity::dispelAllStatusEffect(lua_State *L)
 *  Notes   : Returns a Lua table with the information on the Status Effect stolen
```
teschnei: since it no longer returns a lua table, should probably update this

src/map/trait.cpp
```diff
@@ -35,122 +35,32 @@
 *                                                                       *
 *                                                                       *
 ************************************************************************/
-
 CTrait::CTrait(uint8 id)
+    : m_id(id)
```
teschnei: not that it particularly matters, but if you're going to move these out of the constructor anyways, it's preferable to just have them initialized in the header

scripts/globals/effects/evasion_down.lua
nfaletra: Yes, exactly. It's the only thing that ticks. I can update the comment to be a little more specific.

src/map/lua/lua_baseentity.cpp
nfaletra: I don't think I wrote this. It was pulled in from a merge, but I can remove it.

scripts/globals/weaponskills.lua
nfaletra: I'll double check that, but I believe you're correct.

src/map/lua/lua_baseentity.cpp
nfaletra: Can do.

src/map/trait.cpp
nfaletra: Just want to understand your preferred standard and make sure we're on the same page.
In the past, the MIL has been faster at initializing member variables than doing it in the constructor body. I am not 100% sure that is still the case, but it's a good habit. Would you like me to just move the ctor itself into the header with an empty body?

Also, I removed the 0/null intializations. . .but I may have done that out of reflex from working in Unreal. That engine fills out ALL data with 0 at creation so I may add that back in. herp derp

scripts/globals/weaponskills.lua
nfaletra: Yep. That's how that works. Gutting this.

src/map/trait.cpp
teschnei: see http://en.cppreference.com/w/cpp/language/initializer_list for documenation, and https://github.com/DarkstarProject/darkstar/blob/master/src/map/instance.h#L83 for some examples

nfaletra: oh. inline initializers. I can do that. Not my favorite, but you're the boss. ;)

nfaletra:
Done and done. Let me know if that satisfies the standards or just merge it. ;)

src/map/trait.h
```diff
@@ -177,13 +177,13 @@ class CTrait
...
 private:
 
-    uint8   m_id;
-    uint8   m_job;
-    uint8   m_level;
-    Mod     m_mod;
-    int16   m_value;
-    uint8   m_rank;
-    uint32  m_meritId;
+    uint8   m_id = { 0 };
```
teschnei: can we use the same style as the example I linked earlier (no equals sign, basically)? just to keep it consistent

nfaletra: Yep. pushing in a sec.

PR #4659, opened 2018-03-14
Spellfix

whasf:
Fixes https://github.com/DarkstarProject/darkstar/issues/4641

Issue #4641, opened 2018-03-09
Error in Magic.lua


ReachServer:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 

**_Source Branch_** (master/stable) **:** 

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

ReachServer:
Line 749: Attempt to index local params (a nil value). Not sure which spell is being cast just yet.

zynjec:
I believe this was fixed already in this commit https://github.com/DarkstarProject/darkstar/commit/d83a2e7f4d8026c5ffd6ace8038f02784191c672

Update and see if it persists.

TeoTwawki:
if already up to date, its posisible that `params.bonusmab = params.bonusmab or 0` generates the same bs message I see in the weaponskill global where there is no "real" error but it generates the message anyway when going with the value of zero.

ReachServer:
Teo, it's possible. Zynjec, I updated yesterday, so it's the current version.

zynjec:
Is there a specific spell setting this off? Unable to reproduce on master.

ReachServer:
ya trying to figure that out, one of my players was on RDM 30, so has to be a lower spell

zynjec:
Looks like Bio is at least one spell causing the error.

ReachServer:
Yep, I just got the same result

warlord1352:
having the same exact issue. it seems to be so far dias,diagas,bios,drain,aspir,impact

takhlaq:
@Whitechaser i think this was caused by AMII stuff
if you get time to, could you pr a fix?

zynjec:
This was fixed with https://github.com/DarkstarProject/darkstar/pull/4663

PR #4663, opened 2018-03-14
Bandaid fix in addBonuses

Whitechaser:
Caused by #4620 
fix suggested in #4659 
responds to #4641 

I'm not sure what those 14 changes takhlaq mentioned as being needed in magic.lua are. It looks like the bio/drain/aspir/dia/cure (against undead)/impact issues are mainly from #4620 requiring the params table being passed by all calls to addBonuses.

teschnei pointed out it was preferable to use this bandaid fix so as not to break custom spell scripts.

Do please comment if something else is missing.

### DIFF
--- a/scripts/globals/magic.lua
+++ b/scripts/globals/magic.lua
@@ -741,6 +741,8 @@ function calculateMagicBurst(caster, spell, target, params)
 end;
 
 function addBonuses(caster, spell, target, dmg, params)
+    params = params or {};
+
     local ele = spell:getElement();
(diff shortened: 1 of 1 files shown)


takhlaq:
there were places arithmetic was performed with nil fields which iirc lua would fail to parse
14 changes i mentioned was me derping and ctrl+f'ing #4620 for `params.` not realising there were `spellParams` and other places it was `params.ass = params.ass or 0` too

PR #4668, opened 2018-03-14
Various mount fixes.

zynjec:
- Fixed mounted.lua to set initial mount music properly.
- Removed hardcoded music bit from zonein packet, sets according to animation.
- Updated entities to ignore and deaggro mounted players.
- Sets proper movement speed for mounts.
- Mount abilities/command now return proper messages.
- Mounts checks for proper level.
- Fixed mount zoning dismount conditions

### DIFF
--- a/scripts/globals/effects/mounted.lua
+++ b/scripts/globals/effects/mounted.lua
@@ -5,35 +5,23 @@
 -----------------------------------
 require("scripts/globals/status");
 
-
------------------------------------
--- onEffectGain Action
------------------------------------
-
 function onEffectGain(target,effect)
     --[[
         Retail sends a music change packet (packet ID 0x5F) in both cases.
     ]]
-    if (effect:getPower() == 0) then
-        target:ChangeMusic(5,212);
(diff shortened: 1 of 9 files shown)


takhlaq:
prolly best to add a new function 
```cpp
PChar->isMounted()
```

instead of checking animation each time

scripts/globals/effects/mounted.lua
```diff
@@ -5,35 +5,23 @@
...
-        target:ChangeMusic(5,212);
-        target:setAnimation(ANIMATION_CHOCOBO);
+
+    if effect:getPower() == 0 then
+        target:ChangeMusic(4,212)
+        target:setAnimation(ANIMATION_CHOCOBO)
     else
-        target:ChangeMusic(5,84);
-        target:setAnimation(ANIMATION_MOUNT);
+        target:ChangeMusic(4,84)
```
TeoTwawki: **that whole do we atart at zero or one thing musta got me again**


lua meme

PR #4676, opened 2018-03-16
Restrict mounts during battle

Ivaar:
Retail prevents you from mounting if you are on any mobs enmity list in the zone. 
Players can still outrun mobs and then mount, so this implementation is not perfect.
Order of checks and related messages verified from retail.

Not sure what this comment about usable mount zones is about. Version update notes and bg wiki state they are usable in the same zones. The second mount related update added mounts and chocobo's to SoA outdoor areas. So it should probably be updated in the other file as well.


TeoTwawki:
> Not sure what this comment about usable mount zones is about. Version update notes and bg wiki state they are usable in the same zones.

When I was on retail after mounts came out I was able to use the mounts in multiple area's that I could *not* use my chocobo whistle, and because of that I just want some verification on that. There may have been some changes by SE afterward since it was downright stupid to restrict one and not the other.

TeoTwawki:
So comment was probably referring to that

zynjec:
Until we audit the MISC flags (either add MISC_CHOCOBO, or add something like MISC_MOUNT), ZONETYPE_OUTDOORS is probably a bit better check for now.

Only 30 zones currently have MISC_CHOCOBO and missing a lot of zones I know for fact you can now use mounts in, vs 90 with ZONETYPE_OUTDOORS which does hit zones you can't mount is, is more accurate.

TeoTwawki:
Thanks for clearing that up

Ivaar:
http://forum.square-enix.com/ffxi/threads/50256?p=575052#post575052
http://forum.square-enix.com/ffxi/threads/51345?p=585326#post585326

TeoTwawki:
And?

Ivaar:
Found more links, should be enough verification now.
http://forum.square-enix.com/ffxi/threads/50760?p=579911#post579911
http://forum.square-enix.com/ffxi/threads/50942?p=581618#post581618
http://forum.square-enix.com/ffxi/threads/51154?p=583671#post583671

TeoTwawki:
I don't think you understand what I was saying when I mentioned verification. more links ain't it. 

Before we go making any new flags for mounts I'd want someone to **in game** verify that we'd actually need one. Also zynjec came and posted some clarification on the comment and it wasn't even the thing I was saying.

TeoTwawki:
You might be wondering why there would be any doubts "coz SE says right here, official". Well, they say "chocobo's and other mounts". You can "mount" your chocobo, and this is not the same as calling you chocobo from the whistle - I'd want confirmation the same thing happens with both method *because in the past that wasn't the case* (thought it *should* have been, so I sincerely hope that is the case *now*).

Ivaar:
Based on my retail testing, mounts and chocobo's should be the same flag. I did not travel to every outdoor zone, only the ones which were questionable

TeoTwawki:
I highly suspect that both chocobo and mount should just do the outdoor check and then exclude abyssea region.

Ivaar:
-- outdoor zones you cant mount in
Abdhaljs_Isle-Purgonorgo
Abyssea-Altepa
Abyssea-Attohwa
Abyssea-Empyreal_Paradox
Abyssea-Grauberg
Abyssea-Konschtat
Abyssea-La_Theine
Abyssea-Misareaux
Abyssea-Tahrongi
Abyssea-Uleguerand
Abyssea-Vunkerl
AlTaieu
Arrapago_Reef
Bastok-Jeuno_Airship
Diorama_Abdhaljs-Ghelsba
Escha_RuAun
Escha_ZiTah
Kazham-Jeuno_Airship
La_Vaule_[S]
Manaclipper
Nashmau
Phanauet_Channel
Open_sea_route_to_Al_Zahbi
Open_sea_route_to_Mhaura
RuAun_Gardens
San_dOria-Jeuno_Airship
Ship_bound_for_Mhaura
Ship_bound_for_Selbina
Silver_Sea_route_to_Al_Zahbi
Silver_Sea_route_to_Nashmau
Windurst-Jeuno_Airship

TeoTwawki:
I forgot about transports entirely when I said that

Ivaar:
I used sql queries zyn provided above and compared to the list of zones in the update notes, only after traveling to questionable areas where I attempted to use whistle and mount commands, results were same for both.

TeoTwawki:
If you want to use the chocobo flag, lets start by renaming the chocobo flag (since it wouldn't be just chocobo's anymore) and then see which zones need flag adjustments

Ivaar:
I was going to rename it but, giyhubs search is only returning two results for MISC_CHOCOBO and I know I seen it elsewhere

TeoTwawki:
quite possible we never use it anywhere by name, and its just there in enum to remember wtf it is

zynjec:
There really won't be many checks, without custom content you can only get ANIMATION_CHOCOBO from rental chocobo ladies, and it kicks you off when zoning in to a zone without MISC_CHOCOBO, and a lua binding for necklace.

`
./src/map/lua/lua_baseentity.cpp:12250:    lua_pushinteger(L, (m_PBaseEntity->loc.zone->CanUseMisc(MISC_CHOCOBO) ? 0 : MSGBASIC_CANT_BE_USED_IN_AREA)); //316
./src/map/zone.h:426:    MISC_CHOCOBO    = 0x0004,   // Ability to use Chocobos
./src/map/zone.cpp:855:    else if (PChar->animation == ANIMATION_CHOCOBO && !CanUseMisc(MISC_CHOCOBO))
`

zynjec:
Might also want to double check AlTaieu on retail, for some reason I remember being able to mount up there now.

Ivaar:
I checked sea, sky and a number of other areas, the conditions are the same.

Ivaar:
I verified the changes I made to zone_settings.sql, chocobo whistle and mounts are usable in the same areas

I re-examined byte 0x5B in char_update.cpp packet and I'm getting 64 instead of 1 for quest raptor, so 3 raptor animations. When I changed the mount enums and table I was thinking of these values as sub animations of ANIMATION_MOUNT. If such a list is needed chocobo mount shouldn't be included as it does not use ANIMATION_MOUNT. The ability list and npc menu packet on trade uses [mount ids](https://github.com/Windower/Resources/blob/master/lua/mounts.lua). 

Unless anyone objects my next commit will revert MOUNTS table in status.lua, and adjust MOUNTTYPE enum to reflect mounting sub animations. Do you guys want these sub-animations added to status.lua for clarity?

TeoTwawki:
>  If such a list is needed chocobo mount shouldn't be included

can just change MOUNT_CHOCOBO to MOUNT_NONE but I notice in your own link they list the chocobo at zero the same as us. /shrug

I had the list already ins status.lua, can change that as needed. (I was unsure we'd need it but covered my bases mirroring it into a table named "MOUNTS")

TeoTwawki:
> Unless anyone objects my next commit will revert MOUNTS table in status.lua, and adjust MOUNTTYPE enum to reflect mounting sub animations. Do you guys want these sub-animations added to status.lua for clarity?

Their sub-anims? I'd need to see it but if what you do works/is accurate it'll likely be fine.

src/map/packet_system.cpp
```diff
@@ -834,23 +835,36 @@ void SmallPacket0x01A(map_session_data_t* session, CCharEntity* PChar, CBasicPac
...
+                // add recast timer
+                //PChar->pushPacket(new CMessageBasicPacket(PChar, PChar, 0, 0, 202));
+                return;
             }
-            else
+            // Retail prevents mounts if a player has enmity on any mob in the zone, need a function for this
+            for (SpawnIDList_t::iterator it = PChar->SpawnMOBList.begin(); it != PChar->SpawnMOBList.end(); ++it)
             {
-                PChar->pushPacket(new CMessageBasicPacket(PChar, PChar, 20, 0, 773));
+                CMobEntity* MOB = (CMobEntity*)it->second;
```
takhlaq: PMob

Ivaar: I copied that from case 0x05, "call for help"

PR #4679, opened 2018-03-17
March 2018 items Update

Ninjistix:
- renamed a few items
- added new items
- adjusted a few flags

Note: Contains Jan/Feb changes also. sending this to motivate someone into the doing March Client. And as far as name changes go, FFXIAH has the outdated item names, BG-Wiki has the updated names listed here.


wrenffxi:
merged as is, this would undo sync between sql table and item luas, which would cause the items to become unusable in game.

sql/item_basic.sql
```diff
@@ -4260,9 +4265,9 @@ INSERT INTO `item_basic` VALUES (4525,0,'pumpkin_pie_+1','pumpkin_pie_+1',12,154
 INSERT INTO `item_basic` VALUES (4526,0,'silkworm_egg','silkworm_egg',12,26204,99,0,250);
 INSERT INTO `item_basic` VALUES (4527,0,'jug_of_marys_milk','marys_milk',12,1548,59,0,131);
 INSERT INTO `item_basic` VALUES (4528,0,'crystal_bass','crystal_bass',12,5644,51,1,0);
-INSERT INTO `item_basic` VALUES (4529,0,'rolanberry_(881_ce)','rolanberry_881',12,1548,59,0,24);
-INSERT INTO `item_basic` VALUES (4530,0,'rolanberry_(874_ce)','rolanberry_874',12,1548,59,0,24);
-INSERT INTO `item_basic` VALUES (4531,0,'rolanberry_(864_ce)','rolanberry_864',12,1548,59,0,24);
+INSERT INTO `item_basic` VALUES (4529,0,'rolanberry_(881_c.e.)','rolanberry_881',12,1548,59,0,24);
+INSERT INTO `item_basic` VALUES (4530,0,'rolanberry_(874_c.e.)','rolanberry_874',12,1548,59,0,24);
+INSERT INTO `item_basic` VALUES (4531,0,'rolanberry_(864_c.e.)','rolanberry_864',12,1548,59,0,24);
```
TeoTwawki: dont put punctuation in item names. we strip that for a reason

Ninjistix: it's weird because other items have it, so i decided to leave tho, let me strip all

Ninjistix: question what about items like this?

```
-INSERT INTO `item_basic` VALUES (9127,0,'cyllenian_abjuration_hands','cy.abjuration_hn.',1,61520,99,0,0);
-INSERT INTO `item_basic` VALUES (9128,0,'cyllenian_abjuration_legs','cy.abjuration_lg.',1,61520,99,0,0);
-INSERT INTO `item_basic` VALUES (9129,0,'cyllenian_abjuration_feet','cy.abjuration_ft.',1,61520,99,0,0);
```

that become 

```
+INSERT INTO `item_basic` VALUES (9127,0,'cyllenian_abjuration_hands','cyabjuration_hn',1,61520,99,0,0);
+INSERT INTO `item_basic` VALUES (9128,0,'cyllenian_abjuration_legs','cyabjuration_lg',1,61520,99,0,0);
+INSERT INTO `item_basic` VALUES (9129,0,'cyllenian_abjuration_feet','cyabjuration_ft',1,61520,99,0,0);
```

without the punctuation

Ninjistix:
Question? why are item scripts stored as the item name instead of the itemID? the itemID doesn't change and scripts won't be affected by name adjustments SE does in the future

If no one wants too I would gladly convert all the script names to itemID's.

Ninjistix:
Question? why are item scripts stored as the item name instead of the itemID? the itemID doesn't change and scripts won't be affected by name adjustments SE does in the future

If no one wants too I would gladly convert all the script names to itemID's.

tagban:
That would make more sense if ItemIDs never change and are never recycled (which I don't think they are)

tagban:
Arguably, hasn't SE in the past renamed a few items but didn't change their IDs? So that would make waay more sense.

Ninjistix:
my name changes are more in ref to se doing stuff like this https://www.bg-wiki.com/bg/March_2018_Version_Update_Changes/Existing_Items

takhlaq:
no we're keeping names
we're going for readability with scripts

TeoTwawki:
Naming scripts by item id is *horrible* idea. 
123456.lua <- which frickin item is that?
Plate_of_sole_sushi.lua <- I know exactly what it is. No I would have liked us to have went with the short names (probably wouldn't have needed 2 columns then..) but I don't want to deal with renaming hundreds of existing scripts and changing sql data *and the source code that treads the table* to change the convention at this point. 

Item names don't change that often. We don't have to be concerned about SE changing item names every update. Its **rare**. The last time it was when they renamed mog bonanza "koupon" items

Ninjistix:
don't merge this yet, found some incorrect item food categories, fixing them will push in a few (busy at work)

Ninjistix:
okay im done with item_basic (for now) -.-

sql/item_basic.sql
```diff
@@ -5904,8 +5904,8 @@ INSERT INTO `item_basic` VALUES (6268,0,'komanezumi','komanezumi',99,1548,63,0,0
 INSERT INTO `item_basic` VALUES (6269,0,'eminent_quiver','eminent_quiver',12,30292,99,0,0);
 INSERT INTO `item_basic` VALUES (6270,0,'eminent_bolt_quiver','em._bolt_quiver',12,30292,99,0,0);
 INSERT INTO `item_basic` VALUES (6271,0,'eminent_bullet_pouch','em._bul._pouch',12,30292,99,0,0);
-INSERT INTO `item_basic` VALUES (6272,0,'fried_popoto','fried_pototo',12,1548,54,0,0);
-INSERT INTO `item_basic` VALUES (6273,0,'fried_popoto_+1','fried_pototo_+1',12,1548,54,0,0);
+INSERT INTO `item_basic` VALUES (6272,0,'fried_pototo','fried_popoto',12,1548,54,0,0);
```
zynjec: Item 6272 is "Fried Popoto", item 6273 is "Fried Popoto +1" and the scripts have the proper names already (fried_popoto.lua and fried_popoto_+1.lua).

Ninjistix:
@zynjec  weird part is I changed the categories according to how they were listed on https://www.bg-wiki.com/bg/Auction_House#Browsing, went through each page to see what was listed under what categories since ffxiah is useless at browsing the categories. But I'll shift them back and fix them

sql/item_basic.sql
Ninjistix: strangely enough, the long name is still fried popoto, the sort name got changed to fried pototo, if i changed the names of the script ill check it and fix it
https://www.bg-wiki.com/bg/Fried_Popoto
https://www.bg-wiki.com/bg/Fried_Popoto_%2B1

zynjec: Hmm, that item seems to be a unique mess, maybe someone else can chime in.

image

image

TeoTwawki: Guessing that was a mistake on SE's part. It's been "popoto" for years and I see the multi name still has that. I vote we leave "popoto" alone for now and see if another update corrects it.

TeoTwawki:
>  weird part is I changed the categories according to how they were listed on https://www.bg-wiki.com/bg/Auction_House#Browsing, 


@Ninjistix Like I said the other day, check client itself before trusting any website. We have readily available client info for this no reason not to.

[item-xml-dump.zip](https://github.com/DarkstarProject/darkstar/files/1834413/item-xml-dump.zip)

sql/item_basic.sql
zynjec: Looks like a case of bad english translation, in japanese it's listed as popoto, and that's what potatoes have been called from the start so lets go with that.

Ninjistix:
capture

TeoTwawki:
thats wrong, he's dead jim

TeoTwawki:
don't trust the category pages i guess

sql/item_basic.sql
Ninjistix: ill revert both to popoto, and the scripts

zynjec:
Wiki categories are not the same as auction house categories, if you're going to go by sites you'll have to look at each individual page for their listing category.

eg: BGWiki shows Cursed Beverage in the drinks category because you *can* drink it while cursed, but the auction house category is still cursed items.

Ninjistix:
their site is stupid then, because the page clearly says 

> Auction House category for drinks.

Ninjistix:
refer to #4754

TeoTwawki:
would have been better to force push to same branch and update title. guessing didnt because you put the month in your branch name. but now we have to recomment on the new pr for the same problems and thats annoying

Ninjistix:
i opted not to do any category adjustments from this one, the new one is strictly the item name changes only.

TeoTwawki:
pototo

Ninjistix:
i corrected the scripts in the new one, because its actually been changed in game can provide screenshot later

TeoTwawki:
we talked about this. se has things not matching and its always been poPoto it doesnt make sense to have it half change suddenly years later. has to be mistake on a new staffers part.

TeoTwawki:
search discord for the discussion about poPoto vs poToto

TeoTwawki:
*potayto ~~potahto~~ popohto*

Issue #4023, opened 2017-08-28
Server Crash opening Yughott Chests with Illusion Time set to 0


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**30170728_2

**_Server Version_** (type `!revision` in game) **:**unknown

**_Source Branch_** (master/stable) **:**master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
1. Set CHEST_MAX_ILLUSION_TIME = 0 and CHEST_MIN_ILLUSION_TIME = 0 in settings.lua
2. Restart server
3. Open a Treasure Chest in Yughott Grotto
4. Sulk at Error message on Crash Window

So far this only happens in Yughott Grotto though I have not checked EVERY area. There are many others the chest will open with Illusion time set to 0 with no problem.

metalfiiish:
Would help immensely if you referenced one or two working chests in this report.

Wiggo32:
Sorry, didn't see this until now. Just rechecked issue, still exists. Here's what I did:
1. Brand new git pull / rebuild -> changed only the Coffer/Chest Illusion times (all 4) in settings.lua.
2. Went to Yughott Grotto -> Opened Chest -> Got: Debug Assertion Failed!
3. Restarted server -> Went to Giddeus, Maze of Shakhrami, and Ordelle's Caves Opened each chest: no error.
4. Went back to Yughott -> Opened Chest again -> no error.  :confused:
5. Opened second chest in Yughott -> Got: Debug Assertion Failed!
![yughott_chest_crash](https://user-images.githubusercontent.com/30469395/32083436-0625ab8e-ba88-11e7-848e-f28f238d4412.jpg)

Wiggo32:
Update: I just reproduced this error without changing anything in the settings.lua. Just did a brand new pull /rebuild and also completely fresh database from the default sqls. I opened a chest with one character, got an item, then opened the chest again with the same character to get the illusion message. So, opened the chest with a second character and got: Debug Assertion Failed!

teschnei:
which debug assertion?

Wiggo32:
The same one up above, sorry should have specified.

TeoTwawki:
[...]

PR #4682, opened 2018-03-18
Fix Yughott Grotto chest bug that added negative number of gil under-flowing to 999,999,999 gil..

TeoTwawki:
Probably the cause of that math random crash in #4023 but I didn't test for that. (max smaller than min, I doubt the function likes that and debug vs release mode is probably the difference between "crash" and "do undefined bs")


Wiggo32:
I can no longer get the crash in the grotto when opening chests after applying this fix. Apparantly, the invalid min,max had nothing to do with the ILLUSION times but was referring to the gil values. That explains why changing the ILLUSION times didn't affect the other areas....

PR #4685, opened 2018-03-19
Update to ID's and bump expected client version to 30180306_2

TeoTwawki:
Please report any non matching IDs if your client matches that version.


TeoTwawki:
*nudge*

PR #4687, opened 2018-03-19
Update to ID's and bump expected client version to 30180306_2

TeoTwawki:
Please report any non matching IDs if your client matches that version.

(all stables past "merges" are shown here because I based branch off stable so both PR's would have an identical commit hash, its harmless)


TeoTwawki:
*nudge*

PR #4694, opened 2018-03-22
Lower Jeuno IDs

wrenffxi:
added param to npc_utils.completeQuest to set player variable(s) to 0
moved Lower Jeuno streetlamp code into zone globals
renamed NPCIDs to MobIDs


takhlaq:
mmm dat +/- ratio 👌

PR #4697, opened 2018-03-22
Add capability to log GM command use to database.

TeoTwawki:
Functions as a minimum to bother logging, tier zero commands always ignored. Value of 2=log t2 and above but ignore t1.

~~Note: if this is merged then from here out we require a minimum of mysql 5.6,5 or MariaDB 10.0.1 as anything older doesn't allow a default value of CURRENT_TIME on datestamp columns.~~ *default removed*

***KNOWN ISSUE:*** **auto-translate bytes break the parsing, so the `!zone` command does not get logged and generates the "Failed to log GM command" message. I hope to address that soon™.**

@teschnei bless you for putting up with my bs kind sir.

### DIFF
--- a/conf/map_darkstar.conf
+++ b/conf/map_darkstar.conf
@@ -171,7 +171,13 @@ max_merit_points: 30
 #Minimum time between uses of yell command (in seconds).
 yell_cooldown: 30
 
-#Audit[logging] settings
+#Command Audit[logging] commands with lower permission than this will not be logged.
+# Zero for no logging at all. Commands given to non GMs are not logged. Autotranslate is not parsed.
+audit_gm_cmd: 0
+
+#Todo: other logging including anti-cheat messages
+
+#Chat Audit[logging] settings
 audit_chat: 0
 audit_say: 0
 audit_shout: 0
--- a/sql/audit_gm.sql
+++ b/sql/audit_gm.sql
@@ -1,0 +1,20 @@
+-- MySQL dump 10.13  Distrib 5.7.12, for Win64 (x86_64)
+--
+-- Host: localhost    Database: dspdb
+-- ------------------------------------------------------
+-- Server version	5.5.5-10.0.20-MariaDB
+
+--
+-- Table structure for table `audit_gm`
+--
+
+DROP TABLE IF EXISTS `audit_gm`;
+CREATE TABLE `audit_gm` (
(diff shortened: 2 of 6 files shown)


conf/map_darkstar.conf
```diff
@@ -171,7 +171,13 @@ max_merit_points: 30
 #Minimum time between uses of yell command (in seconds).
 yell_cooldown: 30
 
-#Audit[logging] settings
+#Command Audit[logging] commands with lower permission than this will not be logged.
+# Zero for no logging at all. Commands given to non GMs are not logged. Autotranslate is not parsed.
+audit_gm_cmd: 0
+
```
whasf: What if I wanted to create a command with permission level 0? How would I audit that?

sql/audit_gm.sql
```diff
@@ -0,0 +1,20 @@
...
+-- ------------------------------------------------------
+-- Server version	5.5.5-10.0.20-MariaDB
+
+--
+-- Table structure for table `audit_gm`
+--
+
+DROP TABLE IF EXISTS `audit_gm`;
+CREATE TABLE `audit_gm` (
+  `date_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
whasf: In your SQL in the command handler you insert a time stamp, I don't see where a timestamp wouldn't be inserted, so wouldn't the default not be necessary?

conf/map_darkstar.conf
TeoTwawki: you don't. permission zero is intentionally ignored - don't give commands to everyone in creation if you can't trust that command to be safe :)

sql/audit_gm.sql
TeoTwawki: iunno we do that with out other tables to except they default to all 0's which is deprecated

conf/map_darkstar.conf
TeoTwawki: 1 = log everything no matter how high the tier. we just don't log zero because if you were dumb enough to give something dangerous to tier zero you are beyond help :)

takhlaq:
```
Note: if this is merged then from here out we require a minimum of mysql 5.6,5 or MariaDB 10.0.1 as anything older doesn't allow a default value of CURRENT_TIME on datestamp columns.
```
couldn't we just use c `time(0)`

TeoTwawki:
@takhlaq actually i think my comment about that no longer applies aince no default now

PR #4698, opened 2018-03-23
Additional safety checks in packet_system.cpp

TeoTwawki:
Prevents selling items in SmallPacket0x085 that other bugs left reserved during the window before we fix them. I also made some consistency changes in the log messages.


TeoTwawki:
@teschnei poke me on discord after you see this,

Issue #4690, opened 2018-03-21
Magic burst does 0 damage


davismj:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 

**_Source Branch_** (master/stable) **:** 

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

I'm doing red lotus blade > raging fists (fusion) > fire II, but I've noticed that mb does 0 damage in general.

Whitechaser:
Is it only for that particular combo of weaponskills and spells? Does Fire/Banish do 0 damage as well?

davismj:
> Is it only for that particular combo of weaponskills and spells? Does Fire/Banish do 0 damage as well?

Tried with fire, same thing. Tried blizzard II on distortion, same thing. Tried water II on reverberation, same thing.

> And also are you on the latest version of DSP?

Yes. Teo's server.

davismj:
I've done some more testing. Any magic burst candidate spell will hit for 0. So, a thunder cast on a fusion will also do 0 damage.

TeoTwawki:
```lua
-- Returns a boolean if the spell's element matches the resonace given
function doesSpellElementMatchResonance(ele, resonance)
    isMatch = matches[ele + 1][resonance:getPower() + 1];
    return (isMatch ~= nil and isMatch > 0);
end
```
ele is nil

PR #4705, opened 2018-03-26
Magic Burst logic additional fix

Whitechaser:
Fix for #4690 

Thanks for spotting that.

orz


TeoTwawki:
So you know I've left #4690 open because there has to be another bug somewhere still, because this is what I did on my server and our magic burst is still zero

Whitechaser:
Is it exactly the same results as before this fix? All of the spells hitting for 0? The returned "burst" should always default to 1 now.

TeoTwawki:
Thats what he told me when I did it. I've added some prints to see wtf is going on, will check my log tomorrow and see what all the values were when he bursts.

PR #4707, opened 2018-03-26
Fix: Use correct element value when calculating magic accuracy bonus

### LINKED ISSUE #4690: Magic burst does 0 damage

davismj:
Before, it was relying on a params.element variable which was never being passed in. Rather than passing it in, it makes more sense to read the element directly from spell:getElement().

Resolves #4690
Closes #4706


TeoTwawki:
If this really was supposed to be param'd:
whoever used params.element all those times needs to comment their code explaining for exactly this reason.

PR #4712, opened 2018-04-01
Despoil

nfaletra:
Added Thief ability "Despoil"

The current understanding of Despoil is that the debuff applied to the mob is based on the item stolen. No debuff applies if the steal fails. However, we don't have a lot of info on what items cause which debuffs. I started a table with some currently known item/debuff mappings. If the table does not have an entry for a specific itemid, the script chooses an effect at random. Despoil defaults to a Steal item if a mob doesn't have a Despoil item in their drop list.

Status effect power is ballpark guesses (except for Slow, went with the normal formula for that one). The duration of the debuff has been observed to be at or around 90 seconds no matter the debuff applied.

Basically, I've laid the ground work for others to easily add other debuffs to the table as they are discovered on retail. Or a server admin can add their own.

Testing Done:
* Despoiled Land Crab Meat from a Clipper and it pulled Slow effect id from the DB
* Despoiled other items to test non-existent data in the DB


src/map/lua/lua_baseentity.cpp
```diff
@@ -13539,23 +13539,84 @@ inline int32 CLuaBaseEntity::getStealItem(lua_State *L)
...
+*  Example : despoilItem = target:getDespoilItem()
+*  Notes   : Defaults to getStealItem() if no despoil item exists
+************************************************************************/
+
+inline int32 CLuaBaseEntity::getDespoilItem(lua_State *L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB);
+
+    CMobEntity* Mob = static_cast<CMobEntity*>(m_PBaseEntity);
```
takhlaq: PMob

src/map/lua/lua_baseentity.cpp
```diff
@@ -13539,23 +13539,84 @@ inline int32 CLuaBaseEntity::getStealItem(lua_State *L)
...
     DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB);
 
-    DropList_t* DropList = itemutils::GetDropList(((CMobEntity*)m_PBaseEntity)->m_DropID);
-
-    if (!(((CMobEntity*)m_PBaseEntity)->m_ItemStolen) && (DropList != nullptr && DropList->size()))
+    CMobEntity* Mob = static_cast<CMobEntity*>(m_PBaseEntity);
+    if (Mob)
     {
-        for (uint8 i = 0; i < DropList->size(); ++i)
+        DropList_t* DropList = itemutils::GetDropList(Mob->m_DropID);
```
takhlaq: PDropList

src/map/lua/lua_baseentity.cpp
```diff
@@ -13539,23 +13539,84 @@ inline int32 CLuaBaseEntity::getStealItem(lua_State *L)
     DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
     DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB);
 
-    DropList_t* DropList = itemutils::GetDropList(((CMobEntity*)m_PBaseEntity)->m_DropID);
-
-    if (!(((CMobEntity*)m_PBaseEntity)->m_ItemStolen) && (DropList != nullptr && DropList->size()))
+    CMobEntity* Mob = static_cast<CMobEntity*>(m_PBaseEntity);
```
takhlaq: PMob

src/map/lua/lua_baseentity.cpp
```diff
@@ -13539,23 +13539,84 @@ inline int32 CLuaBaseEntity::getStealItem(lua_State *L)
...
+
+inline int32 CLuaBaseEntity::getDespoilItem(lua_State *L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB);
+
+    CMobEntity* Mob = static_cast<CMobEntity*>(m_PBaseEntity);
+    if (Mob)
+    {
+        DropList_t* DropList = itemutils::GetDropList(Mob->m_DropID);
```
takhlaq: PDropList

PR #4722, opened 2018-04-02
Table the status effects

TeoTwawki:
No startup errors. ***It would be crazy to go in game and try to extensively test every thing that ever used an effect to see if maybe one in the thousands of replaces got borked.*** So that is as far as my testing goes for this.

p.s.
it isn't changing things *"for no reason"* we really should be tabling things instead of having hundreds of global variables.


TeoTwawki:
if at all possible, I'd like to see #4723 merged first and then I rebase this to see if it finds anything.

PR #4727, opened 2018-04-02
fixed duplicate draw-in message

takhlaq:
this is retarded and should probably be moved to lua, and check draw-in time per member instead of cooldown on mob


TeoTwawki:
we should make drawins a mixin later. we're missing a  bunch of retail functionality anyway and that'd let us script the missing parts pretty easy

PR #4730, opened 2018-04-02
/WX and /MP

ShelbyZ:
- not sure why /WX not showing up for all builds
- adding go fast juice /MP


teschnei:
does it work if you set it in debug/release.props?

ShelbyZ:
Not sure why prop files aren't working, need more time.

PR #4739, opened 2018-04-03
Trusts: Implement following/engaging/reasonable stats

giroroxi:
~~There's 1 major outstanding issue, which is that you can crash the server if you zone with an engaged/previously engaged trust. I've decorated this PR with comments, perhaps someone can find out why this crashes.~~

Fixed and ready to merge.

### DIFF
--- a/src/map/ai/controllers/trust_controller.cpp
+++ b/src/map/ai/controllers/trust_controller.cpp
@@ -25,6 +25,7 @@ This file is part of DarkStar-server source code.
 
 #include "../ai_container.h"
 #include "../../status_effect_container.h"
+#include "../../ai/states/despawn_state.h"
 #include "../../entities/charentity.h"
 #include "../../entities/trustentity.h"
 #include "../../packets/char.h"
@@ -45,9 +46,25 @@ CTrustController::~CTrustController()
     POwner->allegiance = ALLEGIANCE_PLAYER;
 }
 
+void CTrustController::Despawn()
+{
+    if (POwner)
+    {
+        POwner->PMaster = nullptr;
+    }
+    CController::Despawn();
+}
+
 void CTrustController::Tick(time_point tick)
 {
     m_Tick = tick;
+
+    if (POwner->PAI->IsCurrentState<CDespawnState>())
+    {
+        // Despawn takes 3s so it's possible we'll tick in this state.
+        // If we do, we don't want to execute any logic since we just nullptr'd everything.
+        return;
+    }
 
     if (POwner->PAI->IsEngaged())
     {
--- a/src/map/ai/controllers/trust_controller.cpp
+++ b/src/map/ai/controllers/trust_controller.cpp
@@ -90,7 +90,6 @@ void CTrustController::DoCombatTick(time_point tick)
 
 void CTrustController::DoRoamTick(time_point tick)
 {
-    // FIXME: POwner->PMaster->PAI can be null. When does this get nulled?
     if (POwner->PMaster->PAI->IsEngaged())
     {
         POwner->PAI->Internal_Engage(POwner->PMaster->GetBattleTargetID());
--- a/src/map/ai/controllers/trust_controller.cpp
+++ b/src/map/ai/controllers/trust_controller.cpp
@@ -59,18 +59,11 @@ void CTrustController::Tick(time_point tick)
 {
     m_Tick = tick;
 
-    if (POwner->PAI->IsCurrentState<CDespawnState>())
-    {
-        // Despawn takes 3s so it's possible we'll tick in this state.
-        // If we do, we don't want to execute any logic since we just nullptr'd everything.
-        return;
-    }
-
     if (POwner->PAI->IsEngaged())
     {
         DoCombatTick(tick);
     }
-    else
+    else if (!POwner->isDead())
     {
         DoRoamTick(tick);
     }
--- a/src/map/ai/controllers/trust_controller.cpp
+++ b/src/map/ai/controllers/trust_controller.cpp
@@ -90,6 +90,7 @@ void CTrustController::DoCombatTick(time_point tick)
 
 void CTrustController::DoRoamTick(time_point tick)
 {
+    // FIXME: POwner->PMaster->PAI can be null. When does this get nulled?
     if (POwner->PMaster->PAI->IsEngaged())
     {
         POwner->PAI->Internal_Engage(POwner->PMaster->GetBattleTargetID());
--- a/src/map/ai/controllers/trust_controller.cpp
+++ b/src/map/ai/controllers/trust_controller.cpp
@@ -48,7 +48,7 @@ CTrustController::~CTrustController()
 
 void CTrustController::Despawn()
 {
-    if (POwner)
+    if (POwner->PMaster)
     {
         POwner->PMaster = nullptr;
     }
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -536,7 +536,9 @@ void CCharEntity::PostTick()
         {
             ForAlliance([&](auto PEntity)
             {
-                static_cast<CCharEntity*>(PEntity)->pushPacket(new CCharHealthPacket(this));
+                if (PEntity->objtype == TYPE_PC) {
+                    static_cast<CCharEntity*>(PEntity)->pushPacket(new CCharHealthPacket(this));
+                }
             });
         }
         pushPacket(new CCharUpdatePacket(this));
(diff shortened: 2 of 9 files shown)


src/map/ai/controllers/trust_controller.cpp
```diff
@@ -90,6 +90,7 @@ void CTrustController::DoCombatTick(time_point tick)
 
 void CTrustController::DoRoamTick(time_point tick)
 {
+    // FIXME: POwner->PMaster->PAI can be null. When does this get nulled?
     if (POwner->PMaster->PAI->IsEngaged())
```
giroroxi: This needs to be resolved before this can be merged.

takhlaq:
need to despawn the trust in ccharentity destructor

takhlaq:
DeletePET even

src/map/entities/trustentity.cpp
```diff
@@ -59,11 +60,26 @@ void CTrustEntity::PostTick()
...
+        }
 
-        if (PMaster && PMaster->PPet == this)
+        if (updatemask & UPDATE_HP)
         {
-            ShowDebug("CTrustSyncPacket");
-            ((CCharEntity*)PMaster)->pushPacket(new CTrustSyncPacket((CCharEntity*)PMaster, this));
+            if (PMaster->PParty != nullptr)
+            {
+                for (auto PEntity : ((CCharEntity*)PMaster)->PParty->members)
```
giroroxi: Ideally I'd use `ForAlliance([&](auto PEntity)` like the char form of this, but I couldn't get it to recognise any party members...

src/map/lua/luautils.cpp
```diff
@@ -2326,7 +2326,7 @@ namespace luautils
         PMob->objtype == TYPE_PET ? snprintf((char*)File, sizeof(File), "scripts/globals/pets/%s.lua", static_cast<CPetEntity*>(PMob)->GetScriptName().c_str()) :
             snprintf((char*)File, sizeof(File), "scripts/zones/%s/mobs/%s.lua", PMob->loc.zone->GetName(), PMob->GetName());
 
-        if (PTarget->objtype != TYPE_PET && PTarget->objtype != TYPE_MOB)
+        if (PTarget->objtype == TYPE_PC)
```
giroroxi: This fixes a crash when trusts engage a mob. This conditional should've been like this to begin with AFAICT.

src/map/utils/petutils.cpp
```diff
@@ -886,9 +1070,6 @@ namespace petutils
         PMaster->StatusEffectContainer->CopyConfrontationEffect(PTrust);
         PMaster->loc.zone->InsertPET(PTrust);
         PMaster->PParty->ReloadParty();
-
-        // Tell the client this is a trust (shows the 'Release' option on the menu).
-        PMaster->pushPacket(new CTrustSyncPacket(PMaster, PTrust));
```
giroroxi: This is done in ` src/map/entities/trustentity.cpp` now.

src/map/utils/petutils.cpp
```diff
@@ -619,6 +619,190 @@ namespace petutils
...
+        uint8 grade;
+
+        uint8 mlvl = PTrust->GetMLevel();
+        uint8 slvl = PTrust->GetSLevel();
+        JOBTYPE mjob = PTrust->GetMJob();
+        JOBTYPE sjob = PTrust->GetSJob();
+
+        uint8 race = 0;                 //Human
+
+        switch (PTrust->look.race)
```
TeoTwawki: I was going to use a mobfamily because there are trusts who don't fit player races
liek this https://github.com/TeoTwawki/darkstar/commit/6c4f2fd8e230b03a8d117b3491c5a2baa0948dfe#diff-f70a2d82e222d87eff0e2b0598b2d4b0R18

src/map/ai/controllers/trust_controller.cpp
giroroxi: Fixed in https://github.com/DarkstarProject/darkstar/pull/4739/commits/1106cc023d734f28fe1d94cabe07235161683d51

teschnei: Does the PMaster pointer still exist?  The only situation I can think of off the top of my head where PAI is null is after the entity is destroyed and you're looking at dead memory (which may or may not set the pointer to null).

If the owner is not setting its trusts PMasters to null when they are destroyed/zone out/whatever, that's probably the best place to start

giroroxi: It was because I wasn't clearing the trusts from the zone list, so when the mob tried to interact with the trust and this logic executed, it would crash. Fixed now.

teschnei: What would a mob have to do with a trusts roam tick?

I am not sure if deletePET is the right function to call here, but I can't look until I have access to an actual repository and not GitHub's shitty web search

teschnei: Oh yeah, the zone server's ZoneServer will handle removing the pet from the zone list, so you shouldn't have to (it'll go as soon as the trust's status is STATUS_DISAPPEAR, after fadeout animation and stuff).  Removing it too early will stop its AI and I think you will end up having it leak memory because the zone server didn't delete it.  You should be fine just setting its PMaster to nullptr when it leaves the zone (since they have to persist for a few seconds for the despawn animation)

teschnei: or.. just set PMaster to nullptr in Despawn.  That would work too

giroroxi: Done. https://github.com/DarkstarProject/darkstar/pull/4739/commits/76ee2a62689cc9dcd26109ca9a628c97e7de60f3

takhlaq: sorry, that was my fault for not rereading DeletePET and assuming it still frees the pet from memory (i swear it did at some point)

src/map/entities/charentity.cpp
```diff
@@ -492,6 +492,7 @@ void CCharEntity::ClearTrusts()
     for (auto trust : PTrusts)
     {
         trust->PAI->Despawn();
+        loc.zone->DeletePET(trust);
```
takhlaq: sorry i misremembered what this function does, and this line should be removed as it doesn't do what i thought it did
i had a look a while after and saw that instead of freeing the pet from memory it will only remove it from the zone's pet list, leaving it hanging around in memory

src/map/entities/charentity.cpp
```diff
@@ -536,7 +537,9 @@ void CCharEntity::PostTick()
         {
             ForAlliance([&](auto PEntity)
             {
-                static_cast<CCharEntity*>(PEntity)->pushPacket(new CCharHealthPacket(this));
+                if (PEntity->objtype == TYPE_PC) {
```
takhlaq: curly brace on new line

src/map/entities/trustentity.cpp
takhlaq: not sure why that wouldnt work since what you have atm is the equivalent of ForParty
https://github.com/DarkstarProject/darkstar/blob/master/src/map/entities/battleentity.h#L545
```cpp
PMaster->ForParty([this](auto PMember)
{
    if (PMember->objtype == TYPE_PC)
        static_cast<CCharEntity*>(PEntity)->pushPacket(new CCharHealthPacket(this));
});
```

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -45,10 +46,26 @@ CTrustController::~CTrustController()
     POwner->allegiance = ALLEGIANCE_PLAYER;
 }
 
+void CTrustController::Despawn()
+{
+    if (POwner)
```
takhlaq: typo maybe?
`if (POwner->PMaster)`

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -45,10 +46,26 @@ CTrustController::~CTrustController()
...
+        POwner->PMaster = nullptr;
+    }
+    CController::Despawn();
+}
+
 void CTrustController::Tick(time_point tick)
 {
     m_Tick = tick;
 
+    if (POwner->PAI->IsCurrentState<CDespawnState>())
```
teschnei: should just do what mob controller does instead of needing to check specifically despawn state (death state probably also does it, along with any other state we may ever add)

instead of this check, just add a check for `!PMob->isDead` to the else before RoamTick

giroroxi: Done. https://github.com/DarkstarProject/darkstar/pull/4739/commits/26bbbf89557ed574561cdbb1015c86119637499b

PR #4741, opened 2018-04-04
Nasomi Fix for Lag resulting from AH seller history searches

Wiggo32:


sql/auction_house.sql
```diff
@@ -39,6 +39,7 @@ CREATE TABLE IF NOT EXISTS `auction_house` (
   `sell_date` int(10) unsigned NOT NULL DEFAULT '0',
   PRIMARY KEY (`id`),
   KEY `itemid` (`itemid`)
```
takhlaq: looks like we're missing a comma after ```(`itemid`)```

TeoTwawki:
Build failing because it couldn't fetch packages for luacheck stuff, just fyi

TeoTwawki:
which is blocking merge, nothing in this pr is the cause.

TeoTwawki:
stale reviews should auto dismiss once requested changes are made :thinking:

PR #4751, opened 2018-04-08
removing unneeded hacks, adding bcnm Creeping Doom

EulerPath:
adding Creeping Doom.
changed the drop rates of some of the bcnm to match what is on wiki aka not 100% rate for each group.


TeoTwawki:
There is a bug in our bcnm loot handling that forces a drop even when not set to 100%, which is why those happened. But I leave you in @takhlaq's hands as he will know what will conflict with the other branch or not.

EulerPath:
I found the bug, and fixed it, that was causing the 100% drop for bcnm and pushed that some time last year.  I have only now taken the time to start to update the drop rates in the sql files.

PR #4756, opened 2018-04-10
Lieutenants Sash Effects

dwn134:


TeoTwawki:
whats retail do if used when there is no effect to remove?

dwn134:
Hmm. I havent had retail experience in over 10 years

Wiggo32:
Nothing:
sash1
We use the sash, but remove no food:
sash2

TeoTwawki:
So it *does* waste the charge then, so we won't need any `onItemCheck()` extras for a fail message. Thanks.

TeoTwawki:
err.. no, your before and after both show 99 charges. So we need..to silently exit?

dwn134:
Was it 100 when 1st used and the screen shot shows 99 now so it did use a
charge?

Wiggo32:
It *did* use a charge... rather immediately upon use before casting was finished. You can see the text right above the "Damisha uses..." message where I had *just* obtained it. Hope they didn't give me one with a charge missing already, I'd be miffed.

TeoTwawki:
ok, thanks. got confused because both screencaps has 99/100

KnowOne134:
So looks to be all in order then. Uses charge regardless of having food
effect of not

Issue #4650, opened 2018-03-11
Lunar Roar (Fenrir Prime) issues


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30180203_1

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
1. Lunar Roar when used by Fenrir Prime is supposed to remove up to 10 beneficial status effects (except reraise). DSP currently has it set up to remove only 2, which is correct for the Avatar version.
refrences: http://ffxiclopedia.wikia.com/wiki/Fenrir_Prime
https://www.youtube.com/watch?v=WcZCtReetpw
at 37:35 in the video 9 effects were removed.

2. Lunar Roar is NOT supposed to remove reraise. DSP currently has it set up to remove reraise. Use the exact same links above for references.

PR #4758, opened 2018-04-10
redesign Lunar Roar to behave as retail does

Wiggo32:
This fixes: https://github.com/DarkstarProject/darkstar/issues/4650

No real efficient way to single out Reraise without adding a flag check for effects.
This script looks through the list of "dispelable" effects and moves forward as long as reraise isn't one of them.

### DIFF
--- a/src/map/lua/lua_statuseffect.cpp
+++ b/src/map/lua/lua_statuseffect.cpp
@@ -286,6 +286,14 @@ inline int32 CLuaStatusEffect::addMod(lua_State* L)
 
 //======================================================//
 
+inline int32 CLuaStatusEffect::getFlag(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PLuaStatusEffect == nullptr);
+
+    lua_pushinteger(L, m_PLuaStatusEffect->GetFlag());
+    return 1;
+}
+
 inline int32 CLuaStatusEffect::setFlag(lua_State* L)
 {
     DSP_DEBUG_BREAK_IF(m_PLuaStatusEffect == nullptr);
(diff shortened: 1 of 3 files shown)


src/map/lua/lua_statuseffect.cpp
```diff
@@ -286,6 +286,14 @@ inline int32 CLuaStatusEffect::addMod(lua_State* L)
 
 //======================================================//
 
+inline int32 CLuaStatusEffect::getFlag(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PLuaStatusEffect == nullptr);
+
```
TeoTwawki: I *think* we have a different method to do this, but I vote we keep what wiggo added here. It makes more sense / is more intuitive to do this way when I want to know wtf flag is set.

TeoTwawki: The other way prolly being `somecrapStatusEffectByFlag()` if it exists. Teo fuzzy memory strikes again.

Wiggo32: yes, we do have hasStatusEffectByFlag and delStatusEffectsByFlag but nothing to return all the statuseffectsbyflag so that we can thumb through them for specifics (i.e. reraise in this example)
I tried going that route first.

Wiggo32: The other way I seen it being handled was to make a huge table of all pertinent effects and then cycle through the table like as in:
https://github.com/DarkstarProject/darkstar/blob/3ff2fefa3d1ffbdf2cd2cc3e3e03f85a9833b004/scripts/globals/abilities/benediction.lua#L15

TeoTwawki: Yeah I like how you did it, was point of my comment. :+1: 

:smiley:

PR #4760, opened 2018-04-11
Fixed Text ID for ITEM_OBTAINED in Southern San d'Oria

zynjec:
Previous ID was for obtained gil, rather than item.


TeoTwawki:
:hankey: my bad on that

PR #4762, opened 2018-04-11
Clean up Lunar cry

Wiggo32:
-Corrected description heading
-Reimplemented? VanadielMoonPhase (not sure why it was commented out, it works fine. Maybe it didn't in the past?)
-Added the correct message for when the mobskill is used


Wiggo32:
oops... messed up... hold off a min

PR #4764, opened 2018-04-11
Standardize mob names.  Match mob luas to table.

wrenffxi:
Applied standard to mob names:
    spaces become underscores
    periods and apostrophes are removed
Checked all mob scripts against the new table, and fixed any that didn't match.
Removed four mobs from Dynamis Beaucedine, that have been removed from retail.
Fixed a bug in Luison.lua


TeoTwawki:
I'll probably be able to review this Monday if nobody else gets to it. its big enough to require me pulling the branch local to look at.

teschnei:
I don't have time to look, but we need to make sure certain name fields
stay (thinking mostly of mob_pools with things like selh'teus)

TeoTwawki:
polname/packetname column

wrenffxi:
Anywhere I changed the polutils_name column in this PR, I verified the change against the actual POL MassExtract .xmls.  Vast majority of changes are to the mobname column.

teschnei:
I am not able to say with certainty that we use polutils_name inside the
server, it's possible we use mobname as the packet name. Need to verify

TeoTwawki:
I thought selh'teus was using packet_name in mob_pools table but I spoze we better make sure

wrenffxi:
Rebased.

TeoTwawki:
@teschnei 
https://github.com/DarkstarProject/darkstar/blob/97f78e03e50f289d56f7a27aef20d82dfc4bc364/src/map/utils/mobutils.cpp#L1344
```
PMob->name.insert(0, (const char*)Sql_GetData(SqlHandle, 1));
PMob->packetName.insert(0, (const char*)Sql_GetData(SqlHandle, 61));
```
```cpp
            if (updatemask & UPDATE_NAME)
            {
                //depending on size of name, this can be 0x20, 0x22, or 0x24
                this->size = 0x24;
                if (PMob->packetName.empty())
                    memcpy(data + (0x34), PEntity->GetName(), (PEntity->name.size() > 15 ? 15 : PEntity->name.size()));
                else
                    memcpy(data + (0x34), PMob->packetName.c_str(), (PMob->packetName.size() > 15 ? 15 : PMob->packetName.size()));
}
```
so if packetName is empty we use the name field of mob_spawn_points and if its not we're using the packet_name field of mob_pools.

The pr doesn't touch mob_pools at all, and all changes in the mob_spawn_points and script look legit to me. I pulled it down locally to reach the sql changes. A good deal of it is mobs we weren't using yet like 
`Little Wingman` >> `Little_Wingman` in abyssea and the rest were just misspelled or had extra underscores, no apostrophe deals in there.

PR #4768, opened 2018-04-15
Paralyze 2 wont overwrite para 1

dwn134:
restructured the spell to be in line with slow2, blind2
 also pickaxe meassage fix

### DIFF
--- a/scripts/globals/spells/paralyze_ii.lua
+++ b/scripts/globals/spells/paralyze_ii.lua
@@ -38,14 +38,14 @@ function onSpellCast(caster,target,spell)
         potency = potency * 2;
     end
 
-        local duration = 120;
-        local params = {};
-        params.diff = nil;
-        params.attribute = MOD_MND;
-        params.skillType = 35;
-        params.bonus = merits*2;
-        params.effect = dsp.effects.PARALYSIS;
-        local resist = applyResistanceEffect(caster, target, spell, params);
+    local duration = 120;
+    local params = {};
+    params.diff = nil;
+    params.attribute = MOD_MND;
+    params.skillType = 35;
+    params.bonus = merits*2;
+    params.effect = dsp.effects.PARALYSIS;
+    local resist = applyResistanceEffect(caster, target, spell, params);
 
     if (duration >= 90) then
         if (caster:hasStatusEffect(dsp.effects.SABOTEUR)) then
--- a/scripts/globals/spells/paralyze_ii.lua
+++ b/scripts/globals/spells/paralyze_ii.lua
@@ -14,41 +14,31 @@ function onMagicCastingCheck(caster,target,spell)
 end;
 
 function onSpellCast(caster,target,spell)
+    local merits = caster:getMerit(MERIT_PARALYZE_II);
 
-    if (target:hasStatusEffect(dsp.effects.PARALYSIS)) then -- Effect already on, do nothing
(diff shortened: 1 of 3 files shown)


scripts/globals/spells/paralyze_ii.lua
```diff
@@ -57,18 +47,21 @@ function onSpellCast(caster,target,spell)
         params.effect = dsp.effects.PARALYSIS;
         local resist = applyResistanceEffect(caster, target, spell, params);
 
+    if (duration >= 90) then
```
TeoTwawki: We're defining it as 120 then checking if its greater than 90? I see nothing that would reduce it from 120 before reaching this conditional

TeoTwawki: please fix the indent while you are here, it really throws me off trying to read this

dwn134: changed indents but i followed the same rules allready applied for blind 2 and slow 2 for the duration part, So not sure if you want to scrap that part completely of not

KnowOne134:
Anything else i can do?

scripts/globals/spells/paralyze_ii.lua
```diff
@@ -14,61 +14,54 @@ function onMagicCastingCheck(caster,target,spell)
...
+    local duration = 120;
+    local params = {};
+    params.diff = nil;
+    params.attribute = MOD_MND;
+    params.skillType = 35;
+    params.bonus = merits*2;
+    params.effect = dsp.effects.PARALYSIS;
+    local resist = applyResistanceEffect(caster, target, spell, params);
+
+    if (duration >= 90) then
```
TeoTwawki: still has the same inconsistency I pointed out before. Nothing is going to change it from 120 before it hits here, so this condition will always be the same. it will always be greater or equal to 90, because its already 120. It isn't until farther down after this check that you have resistance altering the duration.

scripts/globals/spells/paralyze_ii.lua
TeoTwawki: I dunno when those happened. Its wrong.

TeoTwawki:
```
     duration = duration * applyResistanceEffect(caster, target, spell, params);
 
     if (duration >= 90) then --Do it!
```
is what blind 2 does

scripts/globals/spells/paralyze_ii.lua
```diff
@@ -14,61 +14,52 @@ function onMagicCastingCheck(caster,target,spell)
...
-        local resist = applyResistanceEffect(caster, target, spell, params);
-
-        if (resist >= 0.5) then -- There are no quarter or less hits, if target resists more than .5 spell is resisted completely
-            if (target:addStatusEffect(dsp.effects.PARALYSIS,potency,0,duration*resist)) then
-                spell:setMsg(msgBasic.MAGIC_ENFEEB_IS);
-            else
-                -- no effect
-                spell:setMsg(msgBasic.MAGIC_NO_EFFECT);
-            end
+        if (target:addStatusEffect(dsp.effects.PARALYSIS,potency,0,duration*resist)) then
```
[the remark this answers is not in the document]
TeoTwawki: or not factor resist into duration twice actually. can't belive I missed that before posting /duncecapteo

PR #4782, opened 2018-04-19
Ghelsba IDs

wrenffxi:
Adjust [Warchief Vatgit](http://ffxiclopedia.wikia.com/wiki/Warchief_Vatgit) respawn timer to 10 min.
Save the Children NMs [don't](http://ffxiclopedia.wikia.com/wiki/Sureshot_Snatgat) [drop](http://ffxiclopedia.wikia.com/wiki/Fodderchief_Vokdek) [gil](http://ffxiclopedia.wikia.com/wiki/Strongarm_Zodvad).


scripts/zones/Ghelsba_Outpost/mobs/Colo-colo.lua
```diff
@@ -3,13 +3,18 @@
...
+require("scripts/zones/Ghelsba_Outpost/MobIDs");
+-----------------------------------
 
 function onMobSpawn(mob)
 end;
 
 function onMobEngaged(mob,target)
-    GetMobByID(17350930):updateEnmity(target);
-    GetMobByID(17350931):updateEnmity(target);
+    for i = WINGS_OF_FURY_OFFSET, WINGS_OF_FURY_OFFSET + 2 do
```
TeoTwawki: Code you are replacing looks like old code from before bcnm was properly setting Agro. We shouldn't need to update enmity at all.

wrenffxi: Even better.  Updated.

PR #4783, opened 2018-04-20
Garlaige IDs and mechanics

wrenffxi:
[Serket](http://ffxiclopedia.wikia.com/wiki/Serket): added Draw-in and 30-min Rage
[Chandelier](http://ffxiclopedia.wikia.com/wiki/Hitting_the_Marquisate) aggro mechanics.  Will self-destruct with no drop if it aggros before it is claimed.  Added 10 minute cooldown between pop attempts.


scripts/zones/Garlaige_Citadel/mobs/Chandelier.lua
```diff
@@ -0,0 +1,21 @@
...
+-- Note: Spawned for quest "Hitting the Marquisate"
+-----------------------------------
+require("scripts/zones/Garlaige_Citadel/MobIDs");
+require("scripts/globals/status");
+-----------------------------------
+
+function onMobEngaged(mob, target)
+    local ce = mob:getCE(target);
+    local ve = mob:getVE(target);
+    if (ce==0 and ve==0) then
```
TeoTwawki: are you able to actually claim it before this check fires off?

Issue #4303, opened 2017-11-29
Imperial Whitegate should not check ammo/range


aizenmyou:
**_I have:_**

- [X] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [X] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 

**_Server Version_** (type `!revision` in game) **:** 

**_Source Branch_** (master/stable) **:** 

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
This blocks "Guests of the Empire" from triggering as Naja says only remove your weapons.
Also back in the day on retail I worn bibiki shell and got this C/S.

https://github.com/DarkstarProject/darkstar/blob/9c013be68b20282255c6af4453e3aa9847ccb59c/scripts/zones/Aht_Urhgan_Whitegate/npcs/Imperial_Whitegate.lua#L29

TeoTwawki:
Are we sure its not a matter of just one of the slots needs checked instead of both?

Although checking neither is most likely, since we don't check either of those for the siren's tier quest in North Gustaberg. Unless someone knows for sure I'd guess we just check main and sub slots and leave out both as aizenmyou suggests.

teschnei:
We can also check if the slots are actually weapons

But ranged weapons don't show up on your character model anyways so I don't
think it matters much

TeoTwawki:
Adjusted in #4785, open new issue if further problems with that CS are discovered.

PR #4785, opened 2018-04-20
Issue #4303 - Removed ammo and range check

Soapycat:
Referring to this issue: https://github.com/DarkstarProject/darkstar/issues/4303

In the discussion, it was said that removing the ammo and range check is likely the best solution, as the quest only requests you to remove your weapons, not anything else. 

This is my first contribution, so I'm testing the general flow and process of this all.


Soapycat:
You're right, I totally didn't even try to test it. I forgot a parenthesis. My mistake. I pushed another commit. Should that apply correctly to this now?

teschnei:
Yep, your PR has been updated with your second commit

Issue #4786, opened 2018-04-22
Aftermath effects are not delMod


dwn134:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 
30180203_1

**_Source Branch_** (master/stable) **:** 
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
use a relic ws, gain aftermath take guttle. See that attacp went up 10% wait for aftermath to wear still have 10%. now keep doing ws's and it will keep going up 10% over and over until you zone

nfaletra:
Looks like someone changed the lua scripts to handle event "dsp.effects.LOSE" but didn't update the call in core.

https://github.com/DarkstarProject/darkstar/blob/2e38f5108d79a22bfc058505b12d2322b2837c43/scripts/globals/items/guttler.lua#L63

https://github.com/DarkstarProject/darkstar/blob/2e38f5108d79a22bfc058505b12d2322b2837c43/src/map/status_effect_container.cpp#L483

nfaletra:
https://github.com/DarkstarProject/darkstar/pull/4788

PR #4788, opened 2018-04-23
Replaced "dsp.effects.LOSE" with "EFFECT_LOSE" for lua event triggers

nfaletra:
Addresses https://github.com/DarkstarProject/darkstar/issues/4786


nfaletra:
I need to fix the icon for Mythic and Empyrean appearing, as well. Sec.

PR #4789, opened 2018-04-24
Fix empy/mythic aftermath assert

nfaletra:
Fix for aftermath for real this time.

I tested mythic and empyrean weaponskills to confirm they are applying aftermath correctly.


TeoTwawki:
> I tested 

Good minion

PR #4790, opened 2018-04-24
Drops: Sapphire Quadav in Beadeaux

davismj:
Updated to use ffxidb numbers: http://www.ffxidb.com/zones/147/sapphire-quadav.

### DIFF
--- a/sql/mob_droplist.sql
+++ b/sql/mob_droplist.sql
@@ -17969,18 +17969,18 @@ INSERT INTO `mob_droplist` VALUES (3053,0,0,1000,922,160);
 
 -- [Sapphire Quadav]
 -- Data taken from ffxidb on 4/24/2018 (http://www.ffxidb.com/zones/147/sapphire-quadav)
-INSERT INTO `mob_droplist` VALUES (3054,0,0,1000,501,0); -- Quadav Helm (0%) 
-INSERT INTO `mob_droplist` VALUES (3054,2,0,1000,748,0); -- Gold Beastcoin (0%)
+INSERT INTO `mob_droplist` VALUES (3054,0,0,1000,501,0); -- Quadav Helm (0%, was 3%) Suspected not to actually drop in retail
+INSERT INTO `mob_droplist` VALUES (3054,2,0,1000,748,0); -- Gold Beastcoin (0%) (steal)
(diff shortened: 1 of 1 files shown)


davismj:
Two reasons:

First, I think things might be added for logic reasons. For example, Serving of Herb Crawler Eggs on Sapphire Quadav. This might be a quest item or something somewhere, and I don't know  if removing it will have any affect. I'm not sure where it's being used, so I left it at 0.

Second, for maintainability, especially with respect to dubious information. Quadav Helm for example, was previously a 3% chance. FFXIDB does not list it as a drop. Rather than remove it, forcing someone to go look up the item number and what not if it turns out that it was a legitimate drop, I'd rather reduce the number to 0. If it turns out it was in fact a legitimate drop, that number can be easily tweaked.

zynjec:
`Serving of Herb Crawler Eggs` is set as dropType 4 (DROP_STEAL), and can only be obtained via steal, so that makes sense.

But something like...
```sql
INSERT INTO `mob_droplist` VALUES (3054,0,0,1000,501,0); -- Quadav Helm (0%) 
```

With a sample size of nearly 10000 on ffxidb, and it being entirely removed on FFXIclopedia, is there any functional reason to keep it there?

TeoTwawki:
> and it being entirely removed on FFXIclopedia
eh? I still see it shown. 

image

zynjec:
Whoops, late night reading and missed it!

Rest of the post still stands though, I'd trust ffxidb over people editing the wiki.

TeoTwawki:
Same

TeoTwawki:
I'm adding a small comment for future editors and then merging. I bvbelive this is correct the way @davismj has it as no single mob can drop more items than the loot pool holds and in an over 9k sample size some of these are never seen, so I think that people got their quadavs mixed up editing the wiki our current values came from. ffxidb does laughably list a water cluster so no source is perfect atm, but I think this to be more accurate than what we had. If someone wants to go parse 500-2000 saphire quadav kills in Beadeaux to try and prove any item wrong by obtaining one, the project (and **science**) appreciates you!

TeoTwawki:
p.s. 
Some of these are probably grouped too, but we don't know what groupings they would be. Usually retail allows 5 or 6 drops per mob total to prevent the treasure pool overflowing when global drops such as crystals, seals, or geodes are dropped in the same kill.

PR #4793, opened 2018-04-24
Added various Dynamis Divergence NPCs.

zynjec:
- Added Enigmatic Footprints to Bastok Mines, Windurst Walls, Southern San d'Oria, and Ru'Lude Gardens.
- Added ??? to Ru'Lude Gardens.

All position/flags/textid's accurate to retail as of April 2018 version update.

Left content_tag as null for now, do we want Divergence to have it's own tag?


TeoTwawki:
I vote we flag it as part of Rhapsodies content for now.

PR #4796, opened 2018-04-25
Fix Eagle Eye Shot message

nfaletra:
Eagle Eye Shot used to say "Player used Mercy Stroke..."
Eagle Eye Shot uses weaponskill functions to determine its damage. Those functions change the message id to a weaponskill specific one so we need to set it back to JA_DAMAGE (110) when it's done calculating damage.

### DIFF
--- a/scripts/globals/abilities/eagle_eye_shot.lua
+++ b/scripts/globals/abilities/eagle_eye_shot.lua
@@ -43,11 +43,14 @@ function onUseAbility(player,target,ability,action)
     params.enmityMult = 0.5
 
     local damage, criticalHit, tpHits, extraHits = doRangedWeaponskill(player, target, 0, params, 0, true, action)
-    action:messageID(target:getID(), msgBasic.JA_DAMAGE);
-
-    if not (tpHits + extraHits > 0) then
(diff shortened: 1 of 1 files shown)


teschnei:
What about the miss message and the shadows absorbed message?

teschnei:
maybe there's no shadows absorbed message, but at least miss

see https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/abilities/jump.lua for reference

nfaletra:
Miss is handled below. Similar to Jump.

nfaletra:
Ah I see. That's inconsistent. I fix.

teschnei:
ah.. yes.. inconsistent, not that I didn't see it entirely, yes...

nfaletra:
I was giving you the benefit of the doubt. ;)

TeoTwawki:
:thinking:

nfaletra:
WHAT TEO?! WHAT?!

TeoTwawki:
nothin. nothin at all! *hides*

nfaletra:
Also, just to confirm:
https://www.bg-wiki.com/bg/Eagle_Eye_Shot
Ignores shadows.

PR #4802, opened 2018-04-27
Indent + local variables in scripts/zones/Windurst_Waters/npcs/Chamama.lua …

TeoTwawki:
…

global vars are bad mmmkay?


TeoTwawki:
I'm so used to vars being defined before the condition hell starts. The ones I had already caught stood out because of the sore-thumb indents. Thanks for catching that.

PR #4809, opened 2018-04-28
tabled statuses

wrenffxi:
also modified !addtitle, !hastitle, and !setweather to use tables


TeoTwawki:
buh-bye `_G`!  I never liked `_G` anyway, he smelled.

PR #4812, opened 2018-04-28
tabled subEffects, effectFlags, merits, inventory, animations, behaviors, actions, skills, systems

wrenffxi:


scripts/globals/status.lua
```diff
@@ -797,45 +802,48 @@ dsp.effect =
 };
 
 ----------------------------------
--- SC masks
+-- SC masks (these are not currently used anywhere in code base. do we need them?)
```
TeoTwawki: They *used to be* for a thing removed ages ago, IIRC. I dunno if we have any future thing we might use them for but I spoze we could always re-add them then. Also doesn't hurt to just leave them commented out for now.

scripts/globals/status.lua
```diff
@@ -78,75 +78,80 @@ dsp.status =
...
+    COMPRESSION         = 7,
+    LIQUEFACATION       = 8,
+    INDURATION          = 9,
+    REVERBERATION       = 10,
+    TRANSFIXION         = 11,
+    SCISSION            = 12,
+    DETONATION          = 13,
+    IMPACTION           = 14,
+};
+dsp.sub = dsp.subEffects;
```
TeoTwawki:
```lua
dsp.subEffect
```
is superior to both sub and subEffects in my opinion, because you already depluralized (is that a word? w/e we're gonna pretend it is) the others and sub is treated as a keyword in several syntax parsers:
image

Which may possibly cause future confusion (or it might not, but just my 2 cents here)

wrenffxi: agreed. fixed.

PR #4813, opened 2018-04-30
ficks'd goto command

takhlaq:


scripts/commands/goto.lua
```diff
@@ -17,32 +17,27 @@ end;
...
-    if (targ:getZoneID() ~= player:getZoneID() or forceZone == 1) then
-        player:setPos( targ:getXPos(), targ:getYPos(), targ:getZPos(), targ:getRotPos(), targ:getZoneID() );
-    else
-        player:setPos( targ:getXPos(), targ:getYPos(), targ:getZPos(), targ:getRotPos() );
+    local targ = GetPlayerByName(target)
+    -- if we found this player, they're on the same zone server
+    -- if they're in mog house, goto them instead of setPos
+    if targ and not targ:isInMogHouse() then
+        player:setPos(targ:getXPos(), targ:getYPos(), targ:getZPos(), targ:getRotPos(), forceZone == 1 and targ:getZoneID() or nil)
+    elseif not player:gotoPlayer(target) then
```
TeoTwawki: Why not just always used `gotoPlayer()`?

takhlaq: because someone added `forceZone` param to prevent zoning, only setPos if within same zone and `gotoPlayer` just uses server message to handle everything

TeoTwawki: oooh right. goto can't move pos in same zone without a rezone. forgot.

PR #4815, opened 2018-05-01
tabled, localed, and commented out more globals vars

wrenffxi:
.


scripts/globals/effects/carol.lua
```diff
@@ -7,14 +7,14 @@
 require("scripts/globals/status");
 require("scripts/globals/magic");
 
-local resistMod = {dsp.mod.FIRERES, dsp.mod.EARTHRES, dsp.mod.WATERRES, dsp.mod.WINDRES, dsp.mod.ICERES, dsp.mod.THUNDERRES, dsp.mod.LIGHTRES, dsp.mod.DARKRES};
+local dsp.magic.resistMod = {dsp.mod.FIRERES, dsp.mod.EARTHRES, dsp.mod.WATERRES, dsp.mod.WINDRES, dsp.mod.ICERES, dsp.mod.THUNDERRES, dsp.mod.LIGHTRES, dsp.mod.DARKRES};
```
takhlaq: for local variables we don't need to namespace them under dsp, only globals (luacheck doesn't seem to agree with this bring local variable either)

ShelbyZ: Looks like a find-replace bug for the file.  Luacheck did find this if you look toward the bottom of the Travis CI failure:

`scripts/globals/effects/carol.lua:10:10: (E011) expected statement near '.'`

PR #4820, opened 2018-05-02
 Gotoh Zha the Redolent mob script + spell lists 

TeoTwawki:


TeoTwawki:
Eh, didn't think github would include comments from a line I removed here..Had we done all that talking after I opened the PR but before I changed it, they wouldn't be shown..

TeoTwawki:
rebased

PR #4827, opened 2018-05-04
Outofrange actions

ShelbyZ:
Fix issue with GetSpell/GetWeaponSkill where a forged packet could push an out of range access.  Debug break should catch for the brave souls running in debug mode.


teschnei:
why define max spell id if you don't use it for the array's size?  or just use PSpellList.size()?

ShelbyZ:
Because that would have made sense

Issue #3429, opened 2016-10-09
Afflatus Skills available to /whm subjob 


R3P0FFXI:
**_I have:_**
- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
301601831_1

**_Server Version_** (type `@revision` in game) **:**
50868e308882da87ca59798352f348186b000f2f

**_Source Branch_** (master/stable) **:**
Master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
RDM 99 / WHM 49 Afflatus solace & misery skills available to /whm job even tho they should not be according to wiki

https://www.bg-wiki.com/bg/Afflatus_Solace

https://www.bg-wiki.com/bg/Afflatus_Misery

zynjec:
This can be closed, resolved with https://github.com/DarkstarProject/darkstar/pull/4829

PR #4829, opened 2018-05-04
DRK Traits

nfaletra:
Fixed Attack bonus levels/tiers
Fixed Resist Paralyze levels/values
Added Arcana Killer tier
Added Crit. Atk. Bonus
Added Smite
Added Occult Acumen

Fixed Some Last Resort Stuff
* Added item mods for it
* Def reduction penalty is now being stored in the effect power at time of ability use so that we can remove the appropriate amount from the mod when it wears

Fixed Desperate Blows
* I tied the trait to a mod and added it to the ability haste mod in the effect script

Implemented Stalwart Soul
* Decreases the base amount of HP drained by Souleater without affecting the damage percent calculation

Implemented Muted Soul
* Enmity reduction while Souleater is active

Fixed Afflatus Solace/Misery
* Small update. This job ability is only supposed to be usable as WHM main job. Updated the entry in the ability table to reflect this.

### DIFF
--- a/scripts/globals/effects/last_resort.lua
+++ b/scripts/globals/effects/last_resort.lua
@@ -11,12 +11,11 @@ require("scripts/globals/status");
 -----------------------------------
 
 function onEffectGain(target,effect)
-    target:addMod(dsp.mod.ATTP, 15 + target:getMerit(dsp.merit.LAST_RESORT_EFFECT));
-    
-    effect:setPower(-15 - target:getMerit(dsp.merit.LAST_RESORT_EFFECT) + target:getMod(dsp.mod.LAST_RESORT_DEF_PENALTY));
-    target:addMod(dsp.mod.DEFP, effect:getPower());
-    
+    target:addMod(dsp.mod.ATTP, 25 + target:getMerit(dsp.merit.LAST_RESORT_EFFECT));
     target:addMod(dsp.mod.HASTE_ABILITY, target:getMod(dsp.mod.DESPERATE_BLOWS) + target:getMerit(dsp.merit.DESPERATE_BLOWS));
+    
+    -- Gear that affects this mod is handled by a Latent Effect because the gear must remain equipped
+    target:addMod(dsp.mod.DEFP, -25 - target:getMerit(dsp.merit.LAST_RESORT_EFFECT));
 end;
 
 -----------------------------------
@@ -31,7 +30,9 @@ end;
 -----------------------------------
 
 function onEffectLose(target,effect)
-    target:delMod(dsp.mod.ATTP, 15 + target:getMerit(dsp.merit.LAST_RESORT_EFFECT));
-    target:delMod(dsp.mod.DEFP, effect:getPower());
+    target:delMod(dsp.mod.ATTP, 25 + target:getMerit(dsp.merit.LAST_RESORT_EFFECT));
     target:delMod(dsp.mod.HASTE_ABILITY, target:getMod(dsp.mod.DESPERATE_BLOWS) + target:getMerit(dsp.merit.DESPERATE_BLOWS));
+    
+     -- Gear that affects this mod is handled by a Latent Effect because the gear must remain equipped
+    target:delMod(dsp.mod.DEFP, -25 - target:getMerit(dsp.merit.LAST_RESORT_EFFECT));
 end;
--- a/scripts/globals/effects/last_resort.lua
+++ b/scripts/globals/effects/last_resort.lua
@@ -11,9 +11,12 @@ require("scripts/globals/status");
 -----------------------------------
 
 function onEffectGain(target,effect)
-    target:addMod(dsp.mod.ATTP,15 + target:getMerit(dsp.merit.LAST_RESORT_EFFECT));
-    target:addMod(dsp.mod.DEFP,-15 - target:getMerit(dsp.merit.LAST_RESORT_EFFECT));
-    target:addMod(dsp.mod.HASTE_ABILITY, 156+effect:getPower())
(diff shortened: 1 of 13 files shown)


scripts/globals/effects/last_resort.lua
```diff
@@ -11,9 +11,12 @@ require("scripts/globals/status");
 -----------------------------------
 
 function onEffectGain(target,effect)
-    target:addMod(dsp.mod.ATTP,15 + target:getMerit(dsp.merit.LAST_RESORT_EFFECT));
-    target:addMod(dsp.mod.DEFP,-15 - target:getMerit(dsp.merit.LAST_RESORT_EFFECT));
-    target:addMod(dsp.mod.HASTE_ABILITY, 156+effect:getPower())
+    target:addMod(dsp.mod.ATTP, 15 + target:getMerit(dsp.merit.LAST_RESORT_EFFECT));
```
TeoTwawki: it appears our script is on oldschool values. BG says its 25% in modern retail and we've got 15, which I remember being correct a long time ago. ffxiclopedia's page, which is old/not up to date, says 15, and also says it effects ranged attack as well. I dunno if that's true - can't say I ever cared about my drk's ranged attack.
```
 Increases attack by 25% (64/256) and reduces defense by 25% (64/256).
Effect can be increased to ~+34.77% (89/256) attack , ~-34.77% (89/256) defense with 5/5 Last Resort Effect Merits (~1.95% per level, 5/256).
```
IIiiiiiiiiiii hates fractional craps. Ignore the fractional-ness for the moment. Just another of those things we can't get perfect.

scripts/globals/effects/last_resort.lua
```diff
@@ -11,9 +11,12 @@ require("scripts/globals/status");
...
+    
+    effect:setPower(-15 - target:getMerit(dsp.merit.LAST_RESORT_EFFECT) + target:getMod(dsp.mod.LAST_RESORT_DEF_PENALTY));
```
TeoTwawki: gonna hafta do dis differently.
```
The Abyss Sollerets need to remain equipped for this effect to remain active. 
```
also we'd normally just call the effect with the power instead of having its effect override its own power like that. or at least set power to eqqual previous power + bonus.

takhlaq: can check it's still equipped in onEffectTick, if not then remove the effect

TeoTwawki: or a latent? Wish I had finished the scriptable latents thing actually.

scripts/globals/effects/last_resort.lua
nfaletra: The fractions would be easier if we actually did 256 base. It's done in some places and not others. For example, Smite. What's also funny is HASTE_ABILITY is 1024 base. XD

scripts/globals/effects/last_resort.lua
nfaletra: I'm just going to get rid of the mod being added here and add it to where Stalwart Soul calculation happens.

nfaletra: No wait. that makes no sense. Sorry, still waking up. Let me finish my coffee first. XD

nfaletra: What about onItemCheck?

TeoTwawki: we've been trying to get rid of doing this sorta thing in the item scripts

TeoTwawki: I'd try the ontick first, see how it goes. the prob I am considering is unlikely to happen unless we change it later. testing > all

nfaletra: Definitely agree. Not sure how to keep it from feeling delayed, though.

nfaletra: I might have to move Last Resort damage taken to core.

TeoTwawki: LATENT_JOB_ABILITY_ACTIVE and then we have one for future situations to.

nfaletra: Better. Working on it now.

nfaletra: LATENT_STATUS_EFFECT_ACTIVE already exists. XD ez lyfe

nfaletra:
Also addresses: https://github.com/DarkstarProject/darkstar/issues/3429

src/map/utils/battleutils.cpp
```diff
@@ -3571,12 +3571,12 @@ namespace battleutils
...
-            m_PChar->addHP((int32)(-drainPercent * m_PChar->health.hp));
+            m_PChar->addHP((int32)(m_PChar->health.hp * -(drainPercent - m_PChar->getMod(Mod::STALWART_SOUL) * 0.001f)));
         }
         else if (m_PChar->GetSJob() == JOB_DRK &&m_PChar->health.hp >= 10 && m_PChar->StatusEffectContainer->HasStatusEffect(EFFECT_SOULEATER)) {
             //lose 10% Current HP, only HALF (5%) converted to damage
-            damage += (uint32)(m_PChar->health.hp * 0.05);
-            m_PChar->addHP((int32)(-0.1 * m_PChar->health.hp));
+            damage += (uint32)(m_PChar->health.hp * 0.05f);
+            m_PChar->addHP((int32)(m_PChar->health.hp * 0.1f));
         }
```
TeoTwawki: Shouldn't that be -0.1f ?

nfaletra: oh shit. yes.

TeoTwawki:
callin it good for the info we have

PR #4832, opened 2018-05-07
Fix BCNM for "A Thief in Norg"

davismj:


TeoTwawki:
retail exp confirms the old level as being nonsense, these would be EP to a 75

sql/bcnm_info.sql
```diff
@@ -76,7 +76,7 @@ INSERT INTO `bcnm_info` VALUES (64,144,'rank_2_mission','Shitstain',1,18,1800,25
 INSERT INTO `bcnm_info` VALUES (65,144,'worms_turn','Shitstain',1,18,1800,40,6,0,15);
 INSERT INTO `bcnm_info` VALUES (66,144,'grimshell_shocktroopers','Shitstain',1,18,1800,60,6,0,15);
 INSERT INTO `bcnm_info` VALUES (67,144,'on_my_way','Shitstain',1,18,1800,99,6,0,5);
-INSERT INTO `bcnm_info` VALUES (68,144,'thief_in_norg','Shitstain',1,18,1800,99,6,0,6);
+INSERT INTO `bcnm_info` VALUES (68,144,'thief_in_norg','Your mother',1,18,1800,99,6,0,5);
```
TeoTwawki: Someone made the enum stupid and the flag is "allow" rather than "lock"
```cpp
enum BCRULES
{
    RULES_ALLOW_SUBJOBS = 0x01,
    RULES_LOSE_EXP = 0x02,
    RULES_REMOVE_3MIN = 0x04,
    RULES_SPAWN_TREASURE_ON_WIN = 0x08,
    RULES_MAAT = 0x10
};
```
always check the enums. `#themoreyouknow`

TeoTwawki: iirc you should not lose exp here either

TeoTwawki: oh god I'm retarded! NVM!

davismj: wait, it looks like you should lose exp

PR #4833, opened 2018-05-07
bcnm(a thief in norg): Add exp loss

davismj:


TeoTwawki:
I think you were right tonremove it but we'll check - wiki might be wrong/not current with retail changes

davismj:
> You lose Experience Points if defeated.

http://ffxiclopedia.wikia.com/wiki/A_Thief_in_Norg!%3F

This should be pulled.

PR #4835, opened 2018-05-08
Pvs fixes Round #1

ShelbyZ:
- nullptr used before checking
- struct ordering not optimized for 64-bit
- missing initialization of values in constructor
- Object slicing of zmq:error_t exceptions
- Incorrect latent time checks
- Recalculation of loop conditions
- unsigned value cannot be less than zero


ShelbyZ:
All position_t usage points to an internal representation we have for a vector with direction/velocity.

PR #4843, opened 2018-05-10
tabled teleport IDs, replaced global functions in teleports.lua, fixed line endings

wrenffxi:


TeoTwawki:
:thinking: hypothetically, we should be able check if its in the table and forgo all that if/elseif'ing, but we'd need to rename some things.

pseudo code:
```lua
local exists = dsp.teleport[destination]
if exists then
    dsp.teleport.to(target, destination)
else
    -- error handling here
end
```

The calling scripts already pass in a destination, and I love chances to nuke if/elseif towers. How you feel about it?

wrenffxi:
working on this. will let you know when ready for review.

PR #4852, opened 2018-05-14
Reserve synth items after we find out if we have the right recipe

EulerPath:
when playing I noticed that when I put in a wrong recipe that I would not be able to reuse the same mats for the correct recipe.  In the code I found out that in the packet_system file it will do some checks to make sure the synth if good, but it is not until after it reserves the items that it would check if we have the correct recipe.  If we do not have the correct recipe the function would end and would not release the items from being reserved.  If there is a different way I should handle the code let me know.

As a side note as another bug related to having items set on reserve.  Odd things happen when you try to put items on AH or send items that are reserved.


TeoTwawki:
Are you using reverse merges via github to update or just had a rebasing related accident?

Also in the future you may want to use branches other than master for pull requests. Life is easier for you when master can just stay 1:1 with Darkstar's master.

teschnei:
hmm.. I was thinking that you sent a packet every time you put an item in the container, but now I realise that it's probably only one packet once you hit confirm, in which case what you said is good, and I'll agree to that.  For the AH, basically when I did reserving I made item removal not work if the item was reserved.  I fixed most places where they were removing an item and not checking if it was removed, but I guess I missed a few spots!

The changeset is good to merge, but I would like to see a branch without extra commits for the final merge

TeoTwawki:
We can likely collapse the commits using githubs merge button dropdown. I'd check but am irl busy

TeoTwawki:
Nvm demo merged as i was typing

PR #4853, opened 2018-05-15
Fallback to subjob for caster data when monsters mainjob has none.

TeoTwawki:


TeoTwawki:
..I forgot we do the stupid ranger special here too..

TeoTwawki:
Will fixit this afternoon and do new PR.sorry for the notices

TeoTwawki:
If anyone can think of an edge case not covered by what I did, comment. Else don't leave me hangin! I'll be like other contributors and sling emoji at your! I'll do it! /wrists

PR #4858, opened 2018-05-16
Update Text IDs for May 2018 Retail Update

nfaletra:
Tried to fix comments as much as I could. Only a few "guesses" were made.
Also bumping the version number in version.info


scripts/zones/Northern_San_dOria/TextIDs.lua
```diff
@@ -2,122 +2,122 @@
 
 -- General Texts
     ITEM_CANNOT_BE_OBTAINED =  6589; -- Come back after sorting your inventory.
-              ITEM_OBTAINED =  6985; -- Obtained: <<<Unknown Parameter (Type: 80) 1>>><<<Possible Special Code: 01>>><<<Possible Special Code: 05>>>
+              ITEM_OBTAINED =  6986; -- Obtained: <<<Unknown Parameter (Type: 80) 1>>><<<Possible Special Code: 01>>><<<Possible Special Code: 05>>>
                GIL_OBTAINED =  6596; -- Obtained <<<Numeric Parameter 0>>> gil.
```
TeoTwawki: double check the ID's here that the tool says did not move that are nearby ones that did - I've been having recurring issues here where it gets confused by multiple similar messages.

nfaletra: <thing type="DialogTableEntry">
    <field name="index">6986</field>
    <field name="text">Obtained: ≺Numeric Parameter 0≻ guild points.≺Prompt≻
</field>

so. . .that looks wrong.

nfaletra:
Fixed the North Sandy one. Looking for more.

scripts/zones/Northern_San_dOria/TextIDs.lua
TeoTwawki: these are usually sequential with any existing gap staying the same relative offset, so if one didn't change the others near changing is fishy (keyitem obtained moving by 1 while gil didn't = fishy). just a general thing to watch out for that has been especially prevalent in sandy zones

nfaletra:
Nothing else popped out at me.

PR #4871, opened 2018-05-18
command(addweaponskillpoints): Add addweaponskillpoints command

davismj:
!addweaponskillpoints <slot> <points> {player}


takhlaq:
@teschnei maybe we should have this in lua_item instead since we already have getWeaponSkillPoints there https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_item.h#L73

PR #4872, opened 2018-05-18
fix(weaponskill): Check for nil offhand

davismj:


TeoTwawki:
now to suppress 
> [Error] Lua::getItem: unable to find item! Slot: 1 Container: 0
 that happens after..

PR #4879, opened 2018-05-19
Add insta death to crit hits in non NM chigoe scripts. Fixes #4878

### LINKED ISSUE #4878: Chigoes don't die when critically hit
**_I have:_**

**_Client Version_** (type `/ver` in game) **:** 30180427_2

Critical Hits, Weaponskills, and certain Job abilities (Jump, Shield Bash, Weapon Bash, etc.) Will make chigoes instantly fall to the ground in retail. This aspect is missing from dsp.

TeoTwawki:
Note: if you want a mixin for it, knock yourself out and convert the existing onCriticalHit calls unrelated to this at the same time. It didn't seem worth it to me to go do that instead of this 5 second fix.

Insta death only triggers if the attack would not empty its hp bar, according to what I have been told.

----

Todo's:
 - DNC step also insta kills chigoe.
 - message displayed on death is the "falls down" message instead of regular defeat message (this might already happen, unsure).

### DIFF
--- a/scripts/zones/Abyssea-Attohwa/mobs/Murrain_Chigoe.lua
+++ b/scripts/zones/Abyssea-Attohwa/mobs/Murrain_Chigoe.lua
@@ -2,19 +2,32 @@
 -- Area:
 --  MOB: Chigoe
 -----------------------------------
-
---require("scripts/globals/titles");
 mixins = { require("scripts/mixins/families/chigoe") }
+require("scripts/globals/status")
 -----------------------------------
 
 function onMobSpawn(mob)
-end;
+end
 
 function onMobEngaged(mob, target)
-end;
+end
 
 function onMobDisengage(mob)
-end;
+end
+
+function onCriticalHit(mob)
+    if mob:getHP() > 0 then
+        mob:setMobMod(dsp.mobMod.EXP_BONUS, -100)
+        mob:setHP(0)
+    end
+end
+
+function onWeaponskillHit(mob, attacker, weaponskill)
+    if mob:getHP() > 0 then
+        mob:setMobMod(dsp.mobMod.EXP_BONUS, -100)
+        mob:setHP(0)
+    end
+end
 
 function onMobDeath(mob, player, isKiller)
-end;
+end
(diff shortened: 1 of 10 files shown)


TeoTwawki:
Not yet but one can be made easily. But  I only had time to quickfix and not to convert all our existing instances of crit and ws functions.

teschnei:
I just realised my phone badly auto-corrected my previous message. Should
have said "Maybe this would be better off being a mixin"?

TeoTwawki:
@teschnei the imps need redone for mixin to. and Gotah.

Its just we haven't migrated any yet, so I just quickly did what we do elsewhere thinking we'll do them all at once.

https://github.com/DarkstarProject/darkstar/blob/cb9fe3437ad6206d0104c15e83199bc612978f93/scripts/zones/Xarcabard_%5BS%5D/mobs/Ruly_Imp.lua#L17

https://github.com/DarkstarProject/darkstar/blob/0ab8cda20b7865e6ce74da8eee097cbce417347a/scripts/zones/Wajaom_Woodlands/mobs/Gotoh_Zha_the_Redolent.lua#L79

Both onWeaponskillHit and onCriticalHit should eventually be mixins, along with a few other lua bindings.

```cpp
int32 OnSpellPrecast(CBattleEntity* PCaster, CSpell* PSpell);                           // triggered just before casting a spell 
std::optional<SpellID> OnMonsterMagicPrepare(CBattleEntity* PCaster, CBattleEntity* PTarget);            // triggered when monster wants to use a spell on target 
int32 OnMagicHit(CBattleEntity* PCaster, CBattleEntity* PTarget, CSpell* PSpell);       // triggered when spell cast on monster  
```

takhlaq:
we can do them all at once instead of merging this and changing it later

TeoTwawki:
/shrug

I thiught somebody would like that r chigoes fixed and since we have to convert the others anyway they'd all get changed at once. Either way youbare looking at two (minimum) changes - one to convert multiple functions over to mixings (or more if per function) and another to add the same thing onto chigoes.

PR #4882, opened 2018-05-19
commands to modify skill level and craft ranks.

TeoTwawki:


scripts/commands/setcraftrank.lua
```diff
@@ -0,0 +1,56 @@
...
+        player:PrintToPlayer(string.format("%s's skillID %s rank: %u", targ:getName(), craftName, targ:getSkillRank(skillID)))
+        return
+    end
+
+    local craftRank = tonumber(tier) or dsp.craftRank[string.upper(tier)]
+    if craftRank == nil then
+        error(player, "Invalid craft rank!")
+        return
+    else
+        targ:setSkillRank(skillID, craftRank)
```
wrenffxi: maybe when targ:getName() == player:getName(), send the first message and not the second.

TeoTwawki: good idea, will edit when I get free from work

TeoTwawki: went to fix and realized I also forgot to handle nil skill/rank input while still having a player target. I could either move those to separate commands, so you'd have get and set..or do the argument shifting thing. I'll decide which and then come back to this.

TeoTwawki: I didn't want to use a bunch of checking for "is this a craft or other skill?" and "did we omit param a but include param B anyway?" so I went with a  "just works" separate commands way of handling it. someone else can simplify/merge the commands later if they want.

Issue #4559, opened 2018-02-14
Goldsmith Guild Point Exploit - Free Points!


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30171223_0

**_Server Version_** (type `!revision` in game) **:** fae4c89

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

1. Get to Novice Rank in Goldsmithing
2. Start a contract with Ellard in the Goldsmithing Guild
3. Trade Ellard the requested item/items for Guild Points
4. Max out on GP for the Day and now go sell the items because he didn't take them.

vrtra:
FYI, its not just goldsmithing

Wiggo32:
Until it gets fixed properly, the 'old' way still works good enough. Just replace this line:
https://github.com/DarkstarProject/darkstar/blob/8c941d709db70d821d626f108e696af9d67b4222/scripts/globals/crafting.lua#L309
with:
player:tradeComplete();

For anyone who wants to get their Guilds to not be exploitable for the time being.

TeoTwawki:
one downside to make the players aware of if using that workaround: 
if they have extra items the guild wouldn't normally except in the trade window, tradeComplete will take those to. (so pay attention to your trade windows kiddos)

TeoTwawki:
This condition:
```cpp
PChar->TradeContainer->getConfirmedStatus(slotID)
```
on line [3417](https://github.com/DarkstarProject/darkstar/blob/ccf839115d8cdefeb1f96f738cd915f72cae95e3/src/map/lua/lua_baseentity.cpp#L3417) is somehow not being met. So the item is not being confirmed proper before getting there even though the correct functions got called :thinking:

TeoTwawki:
I tossed some debug prints in and found confirmItem() is always returning false, and has an itemID of zero.
```
[19/Mar] [19:52:58][Info] parse: 01A | 001F 001E 0E from user: Testdummy
[19/Mar] [19:52:58][Debug] CLIENT Testdummy PERFORMING ACTION 00
[19/Mar] [19:52:59][Info] parse: 016 | 0021 0020 04 from user: Testdummy
[...]

PR #4884, opened 2018-05-20
Implements the Lamian Fang Key and Gate

Wiggo32:


scripts/zones/Caedarva_Mire/npcs/qm8.lua
```diff
@@ -0,0 +1,35 @@
...
+-- Gives Lamian Fang Key
+-----------------------------------
+package.loaded["scripts/zones/Caedarva_Mire/TextIDs"] = nil
+-----------------------------------
+require("scripts/zones/Caedarva_Mire/TextIDs")
+
+function onTrade(player,npc,trade)
+end
+
+function onTrigger(player,npc)
```
wrenffxi: with scripts/globals/npc_util required,
```
    if player:getVar("[TIMER]Lamian_Fang_Key") ~= VanadielDayOfTheYear() and npcUtil.giveItem(player, 2219) then
        player:setVar("[TIMER]Lamian_Fang_Key", VanadielDayOfTheYear()) -- Can obtain key once per vanadiel day
    else
        player:messageSpecial(NOTHING_OUT_OF_ORDINARY)
    end
end
```

scripts/zones/Arrapago_Reef/npcs/_1ie.lua
```diff
@@ -0,0 +1,66 @@
...
+-- Door: Iron Gate (Lamian Fang Key)
+-- !pos 580 -17 120
+-----------------------------------
+package.loaded["scripts/zones/Arrapago_Reef/TextIDs"] = nil
+-----------------------------------
+require("scripts/zones/Arrapago_Reef/TextIDs")
+require("scripts/globals/status")
+
+function onTrade(player,npc,trade)
+    if npc:getAnimation() == ANIMATION_CLOSE_DOOR then
```
wrenffxi: if npc:getAnimation() == dsp.anim.CLOSE_DOOR (replace all the others too)

scripts/zones/Arrapago_Reef/npcs/_1ie.lua
```diff
@@ -0,0 +1,66 @@
...
+        if trade:hasItemQty(2219,1) and trade:getItemCount() == 1 then
```
wrenffxi: with scripts/globals/npc_util required, all these trade:hasItemQty() + trade:getItemCount()s become
```
if npcUtil.tradeHas(trade, 2219) then
```

scripts/zones/Arrapago_Reef/npcs/_1ie.lua
```diff
@@ -0,0 +1,66 @@
...
+            npc:openDoor(ANIMATION_OPEN_DOOR)
+            player:showText(npc,KEY_BREAKS,2219)
+            player:tradeComplete()
```
wrenffxi: with scripts/globals/npc_util required, all these player:tradeComplete()s become
```
player:confirmTrade()
```

switching to these npc_util functions allows the player to add more items to the trade window than required.  Only take the required items will be removed from player.

scripts/zones/Arrapago_Reef/TextIDs.lua
```diff
@@ -1,23 +1,29 @@
...
+         MEMBER_TOO_FAR =  8453; -- One or more party members are too far away from the entrance.  Unable to enter area.
 
 -- Other Texts
-            YOU_NO_REQS = 7894; -- You do not meet the requirements to enter the battlefield
-        NOTHING_HAPPENS =  119; -- Nothing happens...
-               RESPONSE = 7328; -- There is no response...
+            YOU_NO_REQS =  7894; -- You do not meet the requirements to enter the battlefield
+        NOTHING_HAPPENS =   119; -- Nothing happens...
+               RESPONSE =  7328; -- There is no response...
+         DOOR_IS_LOCKED = 40225; -- The door is locked. You might be able to open it with a <item>.
```
wrenffxi: I see this as 7457 in the POLUtils extract

scripts/zones/Arrapago_Reef/TextIDs.lua
```diff
@@ -1,23 +1,29 @@
...
+        DOOR_IS_LOCKED2 = 40226; -- The door is locked. You might be able to open it with a <item> or a <item>.
```
wrenffxi: 7458

scripts/zones/Arrapago_Reef/TextIDs.lua
```diff
@@ -1,23 +1,29 @@
...
+             KEY_BREAKS = 40227; -- The <item> breaks.
```
wrenffxi: 7459

scripts/zones/Arrapago_Reef/TextIDs.lua
```diff
@@ -1,23 +1,29 @@
...
+           LOCK_SUCCESS = 40231; -- <player> successfully opened the lock with the <item>!
```
wrenffxi: 7463

scripts/zones/Arrapago_Reef/TextIDs.lua
```diff
@@ -1,23 +1,29 @@
...
+              LOCK_FAIL = 40232; -- <player> failed to open the lock with the <item>...
```
wrenffxi: 7464

scripts/zones/Arrapago_Reef/npcs/_1ie.lua
Wiggo32: yep, forgot we moving everything to tables... I'll have to fix it

scripts/zones/Arrapago_Reef/TextIDs.lua
Wiggo32: I didn't use POLUtils, I got numbers from retail capture

scripts/zones/Arrapago_Reef/TextIDs.lua
Wiggo32: I didn't use POLUtils, I got numbers from retail capture

scripts/zones/Arrapago_Reef/TextIDs.lua
Wiggo32: I didn't use POLUtils, I got numbers from retail capture

scripts/zones/Arrapago_Reef/TextIDs.lua
Wiggo32: I didn't use POLUtils, I got numbers from retail capture

scripts/zones/Arrapago_Reef/TextIDs.lua
Wiggo32: I didn't use POLUtils, I got numbers from retail capture

scripts/zones/Arrapago_Reef/npcs/_1ie.lua
Wiggo32: isn't this trade system not working correctly though? Refrenced here: https://github.com/DarkstarProject/darkstar/issues/4559
It's good I can switch it over, just wondering. Also, is there a "trade hierarchy"? Meaning when I put a Lamian Key and a Skeleton key in the trade window on retail it takes priority on the Lamian Key. Do we have that?

scripts/zones/Caedarva_Mire/npcs/qm8.lua
Wiggo32: easy enough...

scripts/zones/Arrapago_Reef/npcs/_1ie.lua
wrenffxi: trade:hasItemQty and trade:getItemCount still work, but it forces the player to trade exactly the correct items and nothing more.  using the npc_util methods allows for more lenient trading (which is usually what retail does).  You can implement hierarchy by order of checks.

```
if npcUtil.tradeHas(trade, foo1) or npcUtil.tradeHas(trade, foo2) then ...
```

if trade window contains foo1, it will be marked for removal, and taken when player:confirmTrade() is called.  foo2 will not.

if foo1 is not traded, but foo2 is, then foo2 will be taken when player:confirmTrade() is called.

scripts/zones/Arrapago_Reef/TextIDs.lua
wrenffxi: Hm, interesting.  Looks like the 7457 value is needed for player:messageSpecial(), whereas the 40225 value is needed for player:showText().  I wonder if these showText values will bonk the TextIDs batch for version update.  Will ask in Discord.

scripts/zones/Arrapago_Reef/TextIDs.lua
TeoTwawki: Retail will use a separate id for each npc where in dsp we can away with reusing them most of the time

scripts/zones/Arrapago_Reef/TextIDs.lua
TeoTwawki: Showtext is just a wrapper for message special that includes lookAt() as well. It simulates a startEvent() for a one liner

scripts/zones/Arrapago_Reef/npcs/_1ie.lua
Wiggo32: yeah, you're right retail is more lenient in that regard. my thing was i thought it wasn't working correctly. I actually had to 'unuse' that system on my server to get the guilds to not be exploitable. we should be trying to do what retail does though in the end. so, i guess i change it and then hope someone fixes it later...

scripts/zones/Arrapago_Reef/TextIDs.lua
Wiggo32: I actually had to use messageSpecial *AND* showText in the script with different messages to get it to display properly without the player's name tag indicating who is speaking. ( "Damisha: " )
I don't know if those functions are set up that way on purpose but otherwise the text would show up as if the player was speaking it.

TeoTwawki: There are some quirks I'll do my best to explain here:

if the calling entity is not a player, the calling entity will be the "speaker" for show text and it'll be an "anybody in range" message
```cpp
((CCharEntity*)m_PBaseEntity)->pushPacket(new CMessageSpecialPacket(PBaseEntity, messageID, param0, param1, param2, param3));
```
but if the calling entity is a player, then it is 
```cpp
PBaseEntity->loc.p.rotation = getangle(PBaseEntity->loc.p, m_PBaseEntity->loc.p);
```
(rotation and angle)
followed by literally just message special but with teh player (base entity) as the speaking entity.
```cpp
m_PBaseEntity->loc.zone->PushPacket(m_PBaseEntity, CHAR_INRANGE, new CMessageSpecialPacket(PBaseEntity, messageID, param0, param1, param3));
```
So the players name was there because the packet is making the player "speak" and that is probably not the correct way to get the message out (those things are usually an event or a parameter passed into a message or event) but sometimes we don't get a choice.

TeoTwawki: In any case which copy of a duplicate message you use won't matter in darkstar.

scripts/zones/Arrapago_Reef/npcs/_1ie.lua
TeoTwawki: for what its worth I tried to find why that bug was happening and didn't spot the cause

scripts/zones/Arrapago_Reef/TextIDs.lua
zynjec: Subtract 0x8000 (32768) from the MessageID you get from retail 0x2A packets for the the values used by DSP/found by POLUtils.

scripts/zones/Caedarva_Mire/npcs/qm8.lua
Wiggo32: ok so I had to change this script example given to me in order to get it working somewhat properly and while messing with this system, this is as close as I could get it to retail:
```
    if player:getVar("[TIMER]Lamian_Fang_Key") ~= VanadielDayOfTheYear() then 
        if npcUtil.giveItem(player, 2219) then
            player:setVar("[TIMER]Lamian_Fang_Key", VanadielDayOfTheYear()) -- Can obtain key once per vanadiel day
        end
    else
        player:messageSpecial(NOTHING_OUT_OF_ORDINARY)
    end
end
```
But, it still isn't right as you still get the "Obtained:" message even with the Key already in your posession. The way I scripted it originally gives the proper retail responses.

scripts/zones/Arrapago_Reef/TextIDs.lua
TeoTwawki: I shoulda have known when seeing them past the signed int16 limit. previously only whitegate has used multiple string tables.

scripts/zones/Caedarva_Mire/npcs/qm8.lua
wrenffxi: What's retail behavior when being given a rare item they already have?  I can update npcUtil.giveItem() to match.

Wiggo32: Awesome! Let me try it.

Wiggo32: It works... but... eh. I'd really like it to be retail accurate which is what i thought the goal of dsp was. the way i have it scripted originally is retail accurate. if you're interested, this is what the message from retail shows when you click on it while having a key already in your inventory:
lamian_key
get npc_utils to behave that way and i'll change it, but i don't see any reason to make it less like retail just to use the npc_util.

scripts/zones/Caedarva_Mire/npcs/qm8.lua
```diff
@@ -0,0 +1,34 @@
...
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end
+
+function onTrigger(player,npc)
+    local fangKeyTimer = player:getVar("[TIMER]Lamian Fang Key")
+
+    if fangKeyTimer < os.time() then
+        if player:getFreeSlotsCount() > 0 and player:hasItem(2219) == false then
```
zynjec:
```lua
and not player:hasItem(2219)
```

scripts/zones/Caedarva_Mire/npcs/qm8.lua
```diff
@@ -0,0 +1,34 @@
...
+            player:addItem(2219,1)
+            player:messageSpecial(ITEM_OBTAINED,2219)
+            player:setVar("[TIMER]Lamian Fang Key",getConquestTally())
```
zynjec: Not sure why `getConquestTally()` is used, this would result in one free key a conquest tally, rather than every vanadiel day, unless you've confirmed this to not be true.

`A character can obtain a free Lamian Fang Key once per Vana'diel day (resets at 0:00 game time) from a ??? in Caedarva Mire (H/I-7), in a small crevice behind a Jnun.`
http://ffxiclopedia.wikia.com/wiki/Lamian_Fang_Key

zynjec:
All the TextID's are incorrect for what DSP uses, and partly the reason why you've having to swap message types.

scripts/zones/Caedarva_Mire/npcs/qm8.lua
Wiggo32: oh crap... i dorked that... pulled old code from wrong place...

Wiggo32: Yeah, thanks for catching that... I spent the better part of the day trying and trying to get npc_utils to work before finally giving up and grabbing my old code and just didn't pay enough attention to it. it's all good now.

Wiggo32:
what else while i have everything open?

scripts/zones/Caedarva_Mire/npcs/qm8.lua
wrenffxi: [This](https://github.com/DarkstarProject/darkstar/pull/4904) should make npcUtil.giveItem behave like you want, though it only works when single item is being given (see discussion on #4889).

Wiggo32: it's not like how I want, it's like how dsp wants... we are on the same team.

Wiggo32: thank you! by the way... =)

scripts/zones/Arrapago_Reef/npcs/_1ie.lua
```diff
@@ -0,0 +1,64 @@
...
+function onTrade(player,npc,trade)
+    if npc:getAnimation() == dsp.anim.CLOSE_DOOR then
+        if npcUtil.tradeHas(trade, 2219) then
+            npc:openDoor(dsp.anim.ANIMATION_OPEN_DOOR)
+            player:messageSpecial(KEY_BREAKS,2219)
+            player:confirmTrade()
+        elseif npcUtil.tradeHas(trade, 1022) and player:getMainJob() == dsp.job.THF then -- thief's tools
+            if math.random(1,2) == 1 then -- TODO: figure out actual percentage chance to pick locks; 50% for now
+                player:messageSpecial(LOCK_SUCCESS,1022)
+                npc:openDoor(dsp.anim.ANIMATION_OPEN_DOOR)
```
wrenffxi: dsp.anim.ANIMATION_OPEN_DOOR would be dsp.anim.OPEN_DOOR, but also npc:openDoor() seems to take as its only argument time in seconds.  (if omitted, defaults to 7 seconds)

Wiggo32: yea, i did use the wrong argument... that's the argument for setAnimation... :man_facepalming:

PR #4885, opened 2018-05-20
Jnun's sleep during the day

Wiggo32:
The AnimationSub reset is in there purposefully in order to make sure you see the animation after moving more then 50 yalms from the Mob. Otherwise you don't get the sleep animation unless the mob falls asleep while within range of the player.

### DIFF
--- a/scripts/zones/Caedarva_Mire/mobs/Jnun.lua
+++ b/scripts/zones/Caedarva_Mire/mobs/Jnun.lua
@@ -1,0 +1,31 @@
+-----------------------------------
+-- Area: Caedarva Mire
+--  MOB: Jnun
+-----------------------------------
+require("scripts/globals/status")
+
+function onMobInitialize(mob)
+end
+
+function onMobSpawn(mob)
+end
+
+function onMobFight(mob, target)
+end
+
+function onMobRoam(mob)
+    if VanadielHour() >= 6 and VanadielHour() < 18 then -- Jnun's sleep during the day
+        mob:AnimationSub(0)
+        mob:AnimationSub(1)
+        mob:setMobMod(MOBMOD_NO_MOVE, 1)
+        mob:setAggressive(0)
+    elseif VanadielHour() >= 18 or VanadielHour() < 6 then -- Jnun's awake at night
+        mob:AnimationSub(1)
+        mob:AnimationSub(0)
+        mob:setMobMod(MOBMOD_NO_MOVE, 0)
+        mob:setAggressive(1)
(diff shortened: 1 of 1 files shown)


TeoTwawki:
> Otherwise you don't get the sleep animation unless the mob falls asleep while within range of the player.

shouldn't be the case, that's a bug.

teschnei:
There are lots of mobs that sleep, it should be a mixin, I think

Wiggo32:
Hmm... I also realized I needed to make an adjustment to make sure the "sleep" animation and everything along with that changed back to awake if it happened to be pulled while it was sleeping. But, now that we realize there may be something wrong with the AnimationSub function itself, I'll have to wait until we figure out how it is going to be dealt with.

scripts/zones/Caedarva_Mire/mobs/Jnun.lua
```diff
@@ -0,0 +1,31 @@
...
+end
+
+function onMobFight(mob, target)
+end
+
+function onMobRoam(mob)
+    if VanadielHour() >= 6 and VanadielHour() < 18 then -- Jnun's sleep during the day
+        mob:AnimationSub(0)
+        mob:AnimationSub(1)
+        mob:setMobMod(MOBMOD_NO_MOVE, 1)
```
wrenffxi: dsp.mobMod.NO_MOVE

scripts/zones/Caedarva_Mire/mobs/Jnun.lua
```diff
@@ -0,0 +1,31 @@
...
+        mob:setAggressive(0)
+    elseif VanadielHour() >= 18 or VanadielHour() < 6 then -- Jnun's awake at night
+        mob:AnimationSub(1)
+        mob:AnimationSub(0)
+        mob:setMobMod(MOBMOD_NO_MOVE, 0)
```
wrenffxi: dsp.mobMod.NO_MOVE

Wiggo32:
How would we do a mixin for time change? I found examples for "engage" and "disengage" and some others but not sure about time change.

zynjec:
There is a `ROAM_TICK` listener I believe, but that seems like it'd run way more than this would require.

teschnei:
Adding a listener for any event is no cost and trivial, so you can just add
one at the same place the Lua binding gets called

Wiggo32:
it's gonna be a while til i figure out listeners for this... gonna close it and come back once i have something solid...

PR #4889, opened 2018-05-22
Assault Leujaoam Cleansing changes

KnowOne134:
fixes issues with cs's not warping player or party members as well as give correct points for a completion.
1-3 players get 1000 points 4 = 900 5 = 800 6 = 700
also 1 ID was double typed causing an unwinable condition
updated the assault npc with correct purchaseable goods


scripts/zones/Aht_Urhgan_Whitegate/npcs/Yahsra.lua
```diff
@@ -55,47 +55,46 @@ function onEventFinish(player,csid,option)
...
+                itemID = 15606;
                 price = 20000;
             else
                 return;
             end
 
             player:addItem(itemID);
             player:messageSpecial(ITEM_OBTAINED,itemID);
-            player:delAssaultPoint(LEBROS_ASSAULT_POINT,price);]]
+            player:delCurrency("LEUJAOAM_ASSAULT_POINT",price);
```
zynjec: You might want to add an inventory check before attempting to adding item/taking points, if the characters inventory is full you won't get your item but points will be taken.

scripts/zones/Caedarva_Mire/Zone.lua
```diff
@@ -64,5 +64,7 @@ function onEventFinish(player,csid,option)
         player:addMission(TOAU,IN_THE_BLOOD);
     elseif (csid == 133) then -- enter instance, warp to periqia
         player:setPos(0,0,0,0,56);
+    elseif (csid == 130) then
```
davismj: `elseif csid == 130`

scripts/zones/Aht_Urhgan_Whitegate/npcs/Yahsra.lua
```diff
@@ -91,10 +91,13 @@ function onEventFinish(player,csid,option)
             else
                 return;
             end
-
-            player:addItem(itemID);
-            player:messageSpecial(ITEM_OBTAINED,itemID);
-            player:delCurrency("LEUJAOAM_ASSAULT_POINT",price);
+            if player:getFreeSlotsCount() > 0 then
```
wrenffxi: if you require scripts/globals/npc_util, lines 94-100 can be replaced with 
````
if npcUtil.giveItem(player, itemID) then
    player:delCurrency("LEUJAOAM_ASSAULT_POINT", price);
end
````

Comment (wrenffxi) [reply scripts/zones/Aht_Urhgan_Whitegate/npcs/Yahsra.lua]: Wiggo is right.  npcUtil replaces the code as written, but after checking wiki for what this NPC does, I agree that neither solution does what we want.

1. what is the correct message?

2. I don't see an easy way in LUA to tell whether a given itemId is rare or not.  This is important for cases like this, where some of the rewards are rare, and others not.

**Comment by teschnei** [comment]
I think the additem will fail and return false in the case of a rare item
that is already possessed. However, it would also fail on full inventory I
think. It may be enough to check inventory count before calling, and then
assuming addItem's failure is due to rare. Otherwise we will need to change
addItem to return a reason for failure

File: scripts/zones/Aht_Urhgan_Whitegate/npcs/Yahsra.lua (hunk 3 above)
Comment (KnowOne134) [reply scripts/zones/Aht_Urhgan_Whitegate/npcs/Yahsra.lua]: @wrenffxi  will npcUtilis work for this now? i saw you updated it. does it require any other lines?

Comment (wrenffxi) [reply scripts/zones/Aht_Urhgan_Whitegate/npcs/Yahsra.lua]: Yes.  If npcUtil.giveItem(player, itemId) where itemId is a rare item the player already possesses, it will return false, which will prevent the deletion of currency.  The code I quoted in comment above should work now.

File: scripts/zones/Aht_Urhgan_Whitegate/npcs/Yahsra.lua [hunk 4]
```diff
@@ -55,47 +56,45 @@ function onEventFinish(player,csid,option)
             local itemID = 0;
             local price = 0;
 
-            -- Copy/pasted from Famad, TODO: fill in the actual IDs/prices for Yahsra
-            --[[if (item == 1) then
-                itemID = 15972;
+            if (item == 1) then
```
takhlaq: we can reduce this if/else tree to
```lua
local items = 
{
    [1] = {itemid = 15970, price = 3000},
    [2] = {itemid = 15775, price = 5000},
    -- etc..
}

local choice = items[item]
if choice and npcUtil.giveItem(player, choice.itemid) then
    player:delCurrency("LEUJAOAM_ASSAULT_POINT", choice.price)
end
```

scripts/zones/Leujaoam_Sanctum/npcs/rune_of_release.lua
```diff
@@ -48,11 +43,10 @@ function onEventFinish(player,csid,option)
             else
                 v:setVar("AssaultPromotion", v:getVar("AssaultPromotion")+5);
             end
-
-            if (v ~= player) then
-                v:startEvent(102);
-            end
-        end
+            v:startEvent(102);
+		end
```
takhlaq: tab -> 4 spaces

zynjec:
For removing currency, there's actually a ad/delAssaultPoint, do we want to get rid of them and continue to just use delCurrency?

KnowOne134:
I'm not sure how to resolve the conflict, dont know what it wants

PR #4897, opened 2018-05-25
Fixed incorrect mobType for LoO Wyvern

Delkathus:
mobType is set to 16 for this regular mob (mobID: 17649827) which spawns it as a NM "Impossible to Gauge!".

Corrected the mobType to 0 (normal).


davismj:
Note that this pool is shared with the Wyverns in Nyzul Isle and Zhayolm Remnants. Since (most) other mobs in those zones aren't mobType 16, I believe this change is correct.

PR #4900, opened 2018-05-25
Whm traits

nfaletra:
Implemented Divine Benison
Added Shield Def. Bonus traits

Also fixed a spot in mob_spell_container that was doing a redundant check.


sql/traits.sql
```diff
@@ -479,6 +481,11 @@ INSERT INTO `traits` VALUES ('110','mag. burst bonus','13','90','2','487','7','A
 INSERT INTO `traits` VALUES ('110','mag. burst bonus','20','79','1','487','5','ABYSSEA','0');
 INSERT INTO `traits` VALUES ('110','mag. burst bonus','20','89','2','487','7','ABYSSEA','0');
 INSERT INTO `traits` VALUES ('110','mag. burst bonus','20','99','3','487','9','ABYSSEA','0');
+INSERT INTO `traits` VALUES ('111','divine benison','3','50','1','909','10','ABYSSEA','0');
```
davismj: ModId be 910, right?

Numbers should not be quoted, see https://github.com/DarkstarProject/darkstar/blob/master/CONTRIBUTING.md#sql

nfaletra: Oh yes. Forgot to update it after quick magic got merged.

I agree that numbers should not be in quotes but someone added them to the entirety of traits.sql. I lean towards consistency in these cases and I'm not going to remove them from the entire file and mess up the diff. I plan to clean up the whole file as a separate commit.

src/map/mob_spell_container.cpp
```diff
@@ -67,7 +67,7 @@ void CMobSpellContainer::AddSpell(SpellID spellId)
         m_damageList.push_back(spellId);
 
     }
-    else if(spell->isNa() || spellId == SpellID::Erase){
+    else if(spell->isNa()){
```
Hozu: If you're removing Erase from the isNa() function, please add Erase check for Divine Veil here:
https://github.com/DarkstarProject/darkstar/blob/0807c177c8b4202d073c8ff54a79145d68c538fe/src/map/utils/battleutils.cpp#L4521

Though I'm not sure what the devs would prefer.

nfaletra: isNa() checks for Erase as part of the function call. I didn’t actually remove anything except redundant code.

nfaletra: isNa() checks for Erase as part of the function call. I didn’t actually remove anything except redundant code.

Hozu: Ooh, I missed that. My bad.

sql/traits.sql
TeoTwawki: After all the current prs changing that table are done I'm reformatting it with a key for sort purposes and tonstrip the quotes off the numbers. Only holding off so you guys don't have to rebase. Just a heads up.

nfaletra: I'm holding off on more traits fixes for the time being after this one for a bit.

PR #4904, opened 2018-05-26
npcUtil.giveItem gives retail-accurate msg and returns false when player has rare item being given

wrenffxi:
can call lua_baseentity:addItem with both quantity and silent

in case that player already has rare item given by npcUtil.giveItem, will now issue ITEM_CANNOT_BE_OBTAINED per [Wiggo's retail screenshot](https://user-images.githubusercontent.com/30469395/40568353-c9f47114-603f-11e8-8a6e-9a36c2c75867.jpg).


wrenffxi:
updated

PR #4909, opened 2018-05-27
add Elemental Ores to Chocobo Digging

Wiggo32:
Elemental Ores are supposed to be available in every zone except Zilart areas. 
http://ffxi.somepage.com/chocobo/
http://ffxi.allakhazam.com/forum.html?forum=24;mid=113972218420714016;num=0;page=1
I just took the line from one of the other zones and copied the dig chance from there.
Unfortunately, there isn't much data on digging nowadays since the Elemental Ores as well as a lot of the other stuff isn't worth much anymore.


TeoTwawki:
Warning: somepage has long been known to be a frequently infested den of malware, and its not advised going there. if you do, make sure you have good protection (no javascript and or good AV).

Such a shame, because they always did have the best search features around.

PR #4913, opened 2018-05-27
Add Chocobo NPCs next to renters

Wiggo32:
I got this data from the most recent retail version... (just a couple days ago) It *Should* match with what dsp has currently because I don't think that has changed. If it doesn't, it will once dsp is up to date with the most recent retail update.


sql/npc_list.sql
```diff
@@ -8350,7 +8350,7 @@ INSERT INTO `npc_list` VALUES (17113994,'Krudsmud','Krudsmud',0,0.000,0.000,0.00
 INSERT INTO `npc_list` VALUES (17114008,'Shrouded_Piper','Shrouded Piper',0,0.000,0.000,0.000,0,40,40,0,0,0,6,27,0x0100000112101220853085408550006000700000,0,'WOTG',1);
 
 INSERT INTO `npc_list` VALUES (17114010,'Helmyre','Helmyre',64,-120.853,0.000,-152.582,20,40,40,0,1,0,0,27,0x01000304151010200A3015401850006000700000,0,'WOTG',1);
--- NC: INSERT INTO `npc_list` VALUES (17114011,'NOT_CAPTURED','     ',0,0.000,0.000,0.000,0,50,50,0,0,0,0,0,0x0000320000000000000000000000000000000000,0,NULL,0);
+INSERT INTO `npc_list` VALUES (17114011,'blank','     ',64,-122.286,-0.015,-151.853,1,50,50,0,0,0,0,2073,0x0700210000000000000001000000000000000000,32,WOTG,0);
```
takhlaq: strings in mysql need quoting, so `WOTG` should be `'WOTG'`

PR #4916, opened 2018-05-27
Stables Teleport items can be used in same Zone

Wiggo32:
https://youtu.be/pNTVaTASONM?t=2105


TeoTwawki:
that code you are removing isn't preventing it from being used in same zone at all. its checking that your character is even aware the zone exists yet. (retail keeps track of whether or not your character has in universe knowledge of a zone, and will restrict the zone form the search list if you do not).

Wiggo32:
Oh, well... whoops! Nevermind this then!

PR #4918, opened 2018-05-27
Added old magic damage formulas.

godmode:
Using  two resources and some guess work I wrote pre-abyssea magic damage formulas. It is damage only, resistance is coded the same. Some guesswork was done after much testing using the dINT +0 and +100 that SE published.

http://ffxiclopedia.wikia.com/wiki/Calculating_Magic_Damage
http://forum.square-enix.com/ffxi/threads/35228-July-9-2013-%28JST%29-Version-Update


scripts/globals/settings.lua
```diff
@@ -135,6 +135,7 @@ ABSORB_SPELL_AMOUNT = 8; -- how much of a stat gets absorbed by DRK absorb spell
 ABSORB_SPELL_TICK = 9; -- duration of 1 absorb spell tick
 SNEAK_INVIS_DURATION_MULTIPLIER = 1; -- multiplies duration of sneak,invis,deodorize to reduce player torture. 1 = retail behavior.
 USE_OLD_CURE_FORMULA = false; -- true/false. if true, uses older cure formula (3*MND + VIT + 3*(healing skill/5)) // cure 6 will use the newer formula
+USE_OLD_MAGIC_DAMAGE = true; -- true/false. if true, uses older magic damage formulas
```
Hozu: This should default to false since DSP wants to be accurate to current retail.

PR #4922, opened 2018-05-28
Rank 2 for all nations no longer capped

kaincenteno:
Mission 2-3 BCNM
The Three Kingdoms, Journey Abroad, The Emissary

### DIFF
--- a/sql/bcnm_info.sql
+++ b/sql/bcnm_info.sql
@@ -45,7 +45,7 @@ CREATE TABLE `bcnm_info` (
 LOCK TABLES `bcnm_info` WRITE;
 /*!40000 ALTER TABLE `bcnm_info` DISABLE KEYS */;
 set autocommit=0;
-INSERT INTO `bcnm_info` VALUES (0,139,'rank_2_mission','Shitstain',1,18,1800,25,6,0,5);
+INSERT INTO `bcnm_info` VALUES (0,139,'rank_2_mission','Shitstain',1,18,1800,99,6,0,5);
 INSERT INTO `bcnm_info` VALUES (1,139,'tails_of_woe','Shitstain',1,18,1800,40,6,0,15);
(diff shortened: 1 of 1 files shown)


sql/bcnm_info.sql
```diff
@@ -45,7 +45,7 @@ CREATE TABLE `bcnm_info` (
 LOCK TABLES `bcnm_info` WRITE;
 /*!40000 ALTER TABLE `bcnm_info` DISABLE KEYS */;
 set autocommit=0;
-INSERT INTO `bcnm_info` VALUES (0,139,'rank_2_mission','Shitstain',1,18,1800,25,6,0,5);
+INSERT INTO `bcnm_info` VALUES (0,139,'rank_2_mission','Shitstain',1,18,1800,99,6,0,5);
```
davismj: This is technically a level cap of 99, no uncapped.

helixhamin:
@davismj Agreed. If there was a way to toggle the change, then it would be different. 

Makes me wonder, we have:
`#Enable/disable level cap of Chains of Promathia mission battlefields stored in database.
CoP_Battle_cap: 1` in the map conf file, why not the nations battlefields?

TeoTwawki:
we did a setting for this for the CoP battlefields. At the time nation 2-3 was still capped. maybe rename the setting and combine them? old_school_mission_caps or something.

Actually, the way I did the setting was crappy. :thinking: I have an idea I will look into after work today.

TeoTwawki:
ok so, I was thinking of adding a column for what expansion a battlefield comes from, and then instead of checking the BCNM ID as the existing CoP_Battle_cap does, we could per-expansion enable/disable the caps.

TeoTwawki:
er, I no forget that. A column for "is this a mission", and then enable/disable them via checking that instead of ID

Id rather not have a half dozen settings when 1 will do.

TeoTwawki:
#4988

takhlaq:
https://github.com/DarkstarProject/darkstar/pull/4988

PR #4927, opened 2018-05-29
prevents activation for players with '0' in mod 'QUICK_MAGIC'

Wiggo32:


nfaletra:
I derped. Good catch.

PR #4932, opened 2018-05-29
Add Destrier beret (Costume Function)

Wiggo32:
Adds the costume feature to the destrier beret item which when used turns you into an egg initially.
Then, based on how long you wait as an egg you get a speed bonus while the costume remains in effect.
I wasn't able to get more than 20% even after waiting an hour as an egg. Only thing this doesn't do is keep the costume on after zoning. That's probably something that needs to be defined in core? Some costumes come off with zoning, some don't. I even was able to wear the chocobo costume into my mog house. But, there it is. @wrenffxi if you wanna streamline it or tell me how to make it prettier?

### DIFF
--- a/scripts/globals/effects/costume.lua
+++ b/scripts/globals/effects/costume.lua
@@ -4,25 +4,19 @@
 --
 -----------------------------------
 
------------------------------------
--- onEffectGain Action
------------------------------------
-
 function onEffectGain(target,effect)
-    target:costume(effect:getPower());
-end;
-
------------------------------------
--- onEffectTick Action
------------------------------------
+    target:costume(effect:getPower())
+    if effect:getPower() == 1999 then
+        target:speed(target:speed()+10) 
+    end
+end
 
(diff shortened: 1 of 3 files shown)


scripts/globals/effects/costume.lua
```diff
@@ -4,25 +4,27 @@
...
------------------------------------
+end
 
 function onEffectLose(target,effect)
-    target:costume(0);
-end;
\ No newline at end of file
+    target:costume(0)
+    if effect:getPower() == 19991 then
+        target:speed(target:speed()-5)
```
TeoTwawki: almost anyplace you think youw nat to use `speed()` you will wanna use modifier for move instead, so we can addMod/delMod on effect gain/loss. You can't trust that the speed hasn't been changed again between adding and losing the effect, could gum it up.

Wiggo32:
I'm redoing this... there's some more speed variations than just the two that I found... We'll get back to it.

PR #4938, opened 2018-05-30
fix(weapon skill points): Only add ws points from exp yielding mobs

davismj:


src/map/entities/charentity.cpp
```diff
@@ -832,7 +832,9 @@ void CCharEntity::OnWeaponSkillFinished(CWeaponSkillState& state, action_t& acti
                         }
                     }
                     // check for ws points
-                    charutils::AddWeaponSkillPoints(this, damslot, wspoints);
+                    if (charutils::GetRealExp(this->GetMLevel(), PTarget->GetMLevel()) > 0) {
```
takhlaq: curly brace new line

davismj: kill me

PR #4939, opened 2018-05-30
Fix bl._chocobo_suit rslot fix:

helixhamin:
Should be hands, legs, feet
Not Head, hands, legs, feet


helixhamin:
Meh, forgot this one when I deleted my trees...I will redo later

PR #4941, opened 2018-05-31
Fix mistake with Shark Necklace 

helixhamin:
Unfortunately, I applied the wrong latent id to shadow necklace. It should be `LATENT_WEAPON_DRAWN_MP_OVER` with value of 2 (ffxiwiki shows 0, but bg shows 2, so I defaulted to bg value) not `LATENT_WEAPON_DRAWN`

Black Chocobo Suit `rslot` corrected (head gear can be equipped, but value prevented it).

Fix StoreTP mod for Pak Corselet +1 (Was 78, which doesn't exist. StoreTP mod id is 73)

- [x] Shark necklace corrected latent
- [x] Black Chocobo suit allows head again
- [x] Pak Corelet +1 now has correct modid


TeoTwawki:
Do we want same latent trigger on both lines for that item?

helixhamin:
@TeoTwawki I can do that, but reading it very carefully, it doesn't seem to suggest that it works that way:

> Latent Effect is triggered when the player's weapon is drawn and has more than 2 MP.
> Has a hidden effect of draining 3 MP/tick while weapon is drawn.

From what I gather, assume you have Refresh for 2 MP per tick.
Weapon is drawn, at 2 MP, you lose regen, but you continue to lose 3 MP per tick.
So, next tick, +2 - 3 = -1 + 2 = 1, meaning still no regen, but you still get the MP loss.
Next tick, +2 - 3 = -1 +1 = 0
After that, it will still continue to take any MP gained by refresh effect, sanction effect, gear refresh, etc, but will not give back Regen until you hit over 2 MP.

Honestly, maybe it would work either way, because even if you stopped the MP drain, next tick it would just kick back in (assuming your MP goes above 2) and then immediately take away 3 but regen you for 1.

helixhamin:
@TeoTwawki Updated. Maybe you are right. Still not sure, since BG states:

> Latent Effect is triggered when the player's weapon is drawn and has more than 2 MP.
> Has a hidden effect of draining 3 MP/tick while weapon is drawn.

Seems like the hidden effect may be only weapon drawn, not weapon drawn + MP over, according to the exact wording, but not sure it will matter much either way. Let me know if I need to change it back (weapon drawn hidden, weapon drawn + MP over latent), or keep both as weapon drawn + MP over 2.

TeoTwawki:
Not important enough to hold off fixing the latent ID but we can get a retail test later to confirm if both effects stop when mp less than 3 but above 0

PR #4942, opened 2018-05-31
Implement sitchair functionality

### LINKED ISSUE #4182: /sitchair not implemented
**_I have:_**

30170707_0

**_Server Version_** (type `!revision` in game) **:** 
de8eab2e5da8bfad3f4e0ea98dca001a025dda9b

master

Items for the /sitchair emote seem to have been added into vendor menus, and the one craft-able one does have the correct recipe. The item however is not able to be used, and vendors do not grant you the key item when you try to purchase them. When you manually add the key items to yourself /sitchair still does not work.

http://ffxiclopedia.wikia.com/wiki/Shadow_Throne - Item not able to be used to grant KI.
http://ffxiclopedia.wikia.com/wiki//sitchair
https://www.bg-wiki.com/bg/Emotes#.2Fsitchair

MalRD:
This PR adds support for /sitchair to fix #4182. One notable aspect of this PR is that it changes the LuaUtils::OnItemCheck function to return a tuple<msgID, param1, param2> instead of a single error value. This allows a parameterized message basic packet to be pushed to the client in an error scenario, which occurs when attempting to acquire the same key item twice. 

Since I couldn't test every existing item script, I tried to maintain existing functionality. If the third value of the tuple is not provided, CItemState will fall back to getting the subID / itemID from the item based on the ITEM_SCROLL_FLAG. Since all previous OnItemCheck calls only return a single item, the third value in the tuple should always be 0, This allows already learned scrolls to continue giving the same error message. 

- [x] Add 0x113 sitchair packet handler
- [x] Add sitchair animation IDs
- [x] Add sitchair item scripts
- [x] Add sitchair item_usable records
- [x] Add sitchair message IDs
- [x] Add sitchair key item IDs
- [x] Modify LuaUtils::OnItemCheck to return msgID, param1, param2
- [x] Modify CItemState to handle tuple and avoid breaking old scripts

### DIFF
--- a/scripts/globals/items/astral_cube.lua
+++ b/scripts/globals/items/astral_cube.lua
@@ -10,7 +10,7 @@ require("scripts/globals/msg");
 local keyItemId = dsp.ki.ASTRAL_CUBE;
 
 function onItemCheck(target)
-    if (target:hasKeyItem(keyItemId)) then
+    if target:hasKeyItem(keyItemId) then
         return dsp.msg.basic.ALREADY_HAVE_KEY_ITEM,0,keyItemId
     end
     return 0
--- a/scripts/globals/items/astral_cube.lua
+++ b/scripts/globals/items/astral_cube.lua
@@ -1,0 +1,22 @@
+-----------------------------------------
+-- ID: 6413
+-- Item: Astral Cube
+-- Item Effect: Grant Astral cube key item
+-----------------------------------------
+require("scripts/globals/settings");
+require("scripts/globals/keyitems");
+require("scripts/globals/msg");
+
+local keyItemId = dsp.ki.ASTRAL_CUBE;
+
+function onItemCheck(target)
+    if (target:hasKeyItem(keyItemId)) then
+        return dsp.msg.basic.ALREADY_HAVE_KEY_ITEM,0,keyItemId
+    end
+    return 0
+end
+
+function onItemUse(target)
+    target:addKeyItem(keyItemId)
+    target:messageBasic(dsp.basic.OBTAINED_KEY_ITEM,6413,keyItemId);
+end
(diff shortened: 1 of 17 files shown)


src/map/lua/luautils.cpp
```diff
@@ -1920,13 +1920,13 @@ namespace luautils
...
 
-    int32 OnItemCheck(CBaseEntity* PTarget, CItem* PItem, ITEMCHECK param, CBaseEntity* PCaster)
+    std::tuple<int32, int32, int32> OnItemCheck(CBaseEntity* PTarget, CItem* PItem, ITEMCHECK param, CBaseEntity* PCaster)
     {
         lua_prepscript("scripts/globals/items/%s.lua", PItem->getName());
 
         if (prepFile(File, "onItemCheck"))
         {
-            return 56;
+            return std::tuple(56, 0, 0);
```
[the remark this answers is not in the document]
MalRD: Yeah, it should.

scripts/globals/items/astral_cube.lua
```diff
@@ -0,0 +1,22 @@
...
+-- Item Effect: Grant Astral cube key item
+-----------------------------------------
+require("scripts/globals/settings");
+require("scripts/globals/keyitems");
+require("scripts/globals/msg");
+
+local keyItemId = dsp.ki.ASTRAL_CUBE;
+
+function onItemCheck(target)
+    if (target:hasKeyItem(keyItemId)) then
```
takhlaq: we're killing parentheses in lua (they're optional)

`if (thing) then` -> `if thing then`

MalRD: I've sent them to a farm upstate.

PR #4943, opened 2018-05-31
[WIP] Allow purchasing multiple same CP items

davismj:
If you try to purchase multiple of the same CP item it will not let you whether or not the item is rare. It should allow this behavior.

**Test:**
- [ ] Purchasing multiple non-rare items
- [ ] Purchasing multiple rare items
- [ ] Trading EXP rings

### DIFF
--- a/scripts/globals/conquest.lua
+++ b/scripts/globals/conquest.lua
@@ -236,40 +236,27 @@ function finishConquestGuard(player,csid,option,size,inventory,guardnation)
         player:addStatusEffect(dsp.effect.SIGNET,0,0,duration) -- Grant Signet
     elseif option >= 32768 and option <= 32944 then
         for item = 1,size,3 do
-            if option == inventory[item] then
-                if player:getFreeSlotsCount() >= 1 then
-                    -- Logic to impose limits on exp bands
-                    if option >= 32933 and option <= 32935 then
-                        if checkConquestRing(player) > 0 then
-                            player:messageSpecial(CONQUEST+60,0,0,inventory[item+2])
-                            break
-                        else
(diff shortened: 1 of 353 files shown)


Wiggo32:
I forgot about this issue until you just brought it up. But, yes you should be able to purchase multiple Non-Rare Items. Found this problem out while trying to synth a HQ body piece a long time ago. They only gave me 1 try! P.S. I've done it on retail since then just to make sure I wasn't crazy.

scripts/globals/conquest.lua
```diff
@@ -3,1388 +3,1360 @@
...
-
-nationAlly = 3;
+require("scripts/globals/keyitems")
+require("scripts/globals/missions")
+require("scripts/globals/common")
+
+dsp = dsp or {}
+dsp.conquest = {}
+
+dsp.conquest.NATION_SANDORIA = 0
```
nfaletra: Should this be divided into dsp.nation, dsp.region, and dsp.conquest? You assign the same values for dsp.conquest.CONQUEST_TALLY down below and that is super gross.

Also, assuming dsp.nation/dsp.region/dsp.conquest is defined in this file, just do dsp.nation= { NATION_SANDORIA = 0, NATION_BASTOK = 1, etc. }
More C-style enum and consistent with other tabled values.

You can still continue to define your functions down below outside the initializations, too.

helixhamin:
@davismj So, should this one be closed, or readjusted?

davismj:
I need to rebase / start over.

takhlaq:
i'll close for now otherwise i'll keep rechecking this

PR #4947, opened 2018-06-01
Moghancement fix

helixhamin:
- [x] Change Moghancements to match https://github.com/DarkstarProject/darkstar/blob/fec0d41f24f8bc560c48b0f0f9448bad937e7998/src/map/items/item_furnishing.h values, instead of whatever they were doing before.
- [x] Fixed the aura value for furniture.
- [x] Added missing furniture to item_furnishing.sql and item_basic.sql.

All moghancements and aura values should now match: https://www.bg-wiki.com/bg/Category:Furnishing#Furniture_List


sql/item_basic.sql
```diff
@@ -3575,6 +3576,9 @@ INSERT INTO `item_basic` VALUES (3715,0,'pot_of_pink_clematis','pink_clematis',1
 INSERT INTO `item_basic` VALUES (3717,0,'birch_tree','birch_tree',1,28736,99,0,0);
 INSERT INTO `item_basic` VALUES (3718,0,'handful_of_adoulinian_tomatoes','ado._tomatoes',1,28752,99,0,0);
 INSERT INTO `item_basic` VALUES (3722,0,'lion_statue','lion_statue',1,61504,99,0,0);
+INSERT INTO `item_basic` VALUES (3723,0,'lilisette_statue','lilisette_statue',1,61504,99,1,0);
+INSERT INTO `item_basic` VALUES (3725,0,'cornelia_statue','cornelia_statue',1,61440,99,1,0);
+INSERT INTO `item_basic` VALUES (
```
davismj: rip

helixhamin: Hmm, wtf? Thanks for the catch, not sure what happened here...

PR #4954, opened 2018-06-03
item(bokuto): Add blindness additional effect to Bokuto / +1

davismj:


TeoTwawki:
How bout I give you access tonwork with me on my refactor instead.

I'm gutting ALL individual item scripts for additional effects and handiy the function in a global eith values passed from database instead

PR #4957, opened 2018-06-04
Crawler's Nest IDs and npcUtil.popFromQM

wrenffxi:
npcUtil.popFromQM is used to pop one or more mobs from a question mark NPC.  Optionally, it disappears the QM and assigns to each mob a despawn listener that reappears the QM after all mobs are defeated and a given time has passed (defaults to QM_RESET_TIME)


TeoTwawki:
Q: how do we handle multiple mobs spawn, but only one is spawned claimed with this?

wrenffxi:
A: We don't! 🥇

TeoTwawki:
ok then, old fashioned way remains for those rare cases lol

wrenffxi:
You could do this:
```
if npcUtil.popFromQM(player, npc, {FOO1, FOO2, FOO3}, false) then -- spawn three with no claim
    GetMobByID(FOO1):updateClaim(player) -- make first one claim
end
```

PR #4965, opened 2018-06-05
Implement STATUS_INVISIBLE and updated Shikigami Weapon.

zynjec:
STATUS_INVISIBLE (previously STATUS_3) is used for mobs that roam normally, entirely invisible to client, but visible on widescan. (eg: Shikigami Weapon, Manananggal). Also made Invisible targets invalid targets for targetfinding purposes.

Updated Shikigami Weapon to spawn using this status, and return to this status on deaggro, and new family so it's aggressive only to magic.


sql/mob_family_system.sql
```diff
@@ -535,6 +535,7 @@ INSERT INTO `mob_family_system` VALUES (498,'Doll',3,'Arcana',1,40,108,90,1,3,4,
 INSERT INTO `mob_family_system` VALUES (499,'Stray',12,'Empty',0,40,20,90,1,3,4,3,6,6,5,3,3,3,3,1,1,1,1,1,1,1,1,1,1,1,1,0,2,0);
 INSERT INTO `mob_family_system` VALUES (500,'Mokkurkalfi',3,'Arcana',1,40,130,130,2,3,3,3,3,6,5,3,2,3,3,1,1,1,1,1,1,1,1,1.125,1,1,1,7,34,0);
 INSERT INTO `mob_family_system` VALUES (501,'NioA_NioHum',3,'Arcana',1,40,108,90,1,3,4,3,6,6,5,3,3,3,3,1,1,1,1,1,1,1,1,1.25,1,1,1,2,34,0);
+INSERT INTO `mob_family_system` VALUES (502,'Shikigami_Weapon',3,'Arcana',0,40,105,120,1,3,4,3,3,3,4,3,5,3,3,1,1,1,1,1.25,0.9,0.9,0.9,0.9,0.9,1.25,0.9,3,32,0);
```
TeoTwawki: Justa  heads up (I'm not gonna make you move it) in the future remember we have some gaps we can fill in
```sql
-- 405,411-434,439-443 available for use
```
We'll have more gaps in the future when we start nuking duplicate families we don't need (they exist because we used to need an extra family ID just to have a diff skill list).

PR #4967, opened 2018-06-06
Got Drachen Mail regen effect working as per bg-wiki

areckx:
I got the field data 370 from 
```
INSERT INTO `item_mods_pet` VALUES (18547,370,3,0);     -- Ravana's Axe Pet: Adds "Regen" effect`
```
and it appears that the 370 gives regen, the next field 3 indicates that it's 3 seconds apparently

### DIFF
--- a/sql/item_mods_pet.sql
+++ b/sql/item_mods_pet.sql
@@ -96,7 +96,7 @@ INSERT INTO `item_mods_pet` VALUES (11720,28,3,3);      -- Automaton: "Magic Atk
 INSERT INTO `item_mods_pet` VALUES (11739,27,2,1);      -- Caller's Sash Avatar: Enmity +2
 INSERT INTO `item_mods_pet` VALUES (11739,28,2,1);      -- Avatar: Magic Atk. Bonus +2
 INSERT INTO `item_mods_pet` VALUES (12520,27,-3,1);     -- Evoker's Horn Avatar: Enmity -3
-INSERT INTO `item_mods_pet` VALUES (12649,370,3,2);     -- Drachen Mail Adds "Regen" effect to wyvern
+INSERT INTO `item_mods_pet` VALUES (12649,370,1,2);     -- Drachen Mail Adds "Regen" effect to wyvern
 INSERT INTO `item_mods_pet` VALUES (12650,27,-2,1);     -- Evoker's Doublet Avatar: Enmity -2
 INSERT INTO `item_mods_pet` VALUES (13974,25,5,2);      -- Drachen Fng. Gnt. Enhances wyvern's accuracy+5 (?)
 INSERT INTO `item_mods_pet` VALUES (13975,27,-2,1);     -- Evoker's Bracers Avatar: Enmity -2
--- a/sql/item_mods_pet.sql
+++ b/sql/item_mods_pet.sql
@@ -96,6 +96,7 @@ INSERT INTO `item_mods_pet` VALUES (11720,28,3,3);      -- Automaton: "Magic Atk
 INSERT INTO `item_mods_pet` VALUES (11739,27,2,1);      -- Caller's Sash Avatar: Enmity +2
 INSERT INTO `item_mods_pet` VALUES (11739,28,2,1);      -- Avatar: Magic Atk. Bonus +2
 INSERT INTO `item_mods_pet` VALUES (12520,27,-3,1);     -- Evoker's Horn Avatar: Enmity -3
+INSERT INTO `item_mods_pet` VALUES (12649,370,3,2);     -- Drachen Mail Adds "Regen" effect to wyvern
 INSERT INTO `item_mods_pet` VALUES (12650,27,-2,1);     -- Evoker's Doublet Avatar: Enmity -2
 INSERT INTO `item_mods_pet` VALUES (13974,25,5,2);      -- Drachen Fng. Gnt. Enhances wyvern's accuracy+5 (?)
 INSERT INTO `item_mods_pet` VALUES (13975,27,-2,1);     -- Evoker's Bracers Avatar: Enmity -2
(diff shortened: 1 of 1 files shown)


TeoTwawki:
from the header of file:
```sql 
 `itemId` smallint(5) unsigned NOT NULL, 
 `modId` smallint(5) unsigned NOT NULL, 
 `value` smallint(5) NOT NULL DEFAULT '0', 
 `petType` tinyint(3) unsigned NOT NULL DEFAULT '0', 
```
so
```sql
INSERT INTO `item_mods_pet` VALUES (itemId, modId, value, petType);
```
You are giving 3 hp regen per tick to *all pets of all kinds*. "2" is wyvern only.

areckx:
```
INSERT INTO `item_mods_pet` VALUES (12649,370,3,2);
```
Wait so 2 isn't for wyvern? I thought the 3 was for seconds... I am not sure where the values for these entries come from. I searched all over the code base for what values do what in these. The first one is the itemid, of course, the second one is the effect, but where do I get a list of what possible numbers to put into value and petType?

TeoTwawki:
I literally just showed you where they come form at the top of the file.

We wouldn't need to store seconds anywhere because speed of regen never changes. If you thought it was 3 seconds where did you think the size of the regen was at?

TeoTwawki:
> "2" is wyvern only.

.

> Wait so 2 isn't for wyvern? 

Quoting  a line with "0" in your opening comment made me miss you already have a "2" for the type. the rest still stands - that is not seconds and everything is right in front of you in the same file. I quoted it in my 1st comment.

helixhamin:
Hmm, neither wiki states the regen tic amount, but ffxiah does:

> Quetzalcoatl.Kainstryder
> Regen is 3 HP/tic.

`    REGEN                           = 370,`

Values seem to be correct from my analysis.

TeoTwawki:
Very unusual for neither wiki to say and it to be more than 1.--06.06.2018, 10:50, "Delaide" <notifications@github.com>:Hmm, neither wiki states the regen tic amount, but ffxiah does:

Quetzalcoatl.Kainstryder
Regen is 3 HP/tic.

REGEN = 370,
Values seem to be correct from my analysis.

sql/item_mods_pet.sql
```diff
@@ -96,6 +96,7 @@ INSERT INTO `item_mods_pet` VALUES (11720,28,3,3);      -- Automaton: "Magic Atk
 INSERT INTO `item_mods_pet` VALUES (11739,27,2,1);      -- Caller's Sash Avatar: Enmity +2
 INSERT INTO `item_mods_pet` VALUES (11739,28,2,1);      -- Avatar: Magic Atk. Bonus +2
 INSERT INTO `item_mods_pet` VALUES (12520,27,-3,1);     -- Evoker's Horn Avatar: Enmity -3
+INSERT INTO `item_mods_pet` VALUES (12649,370,3,2);     -- Drachen Mail Adds "Regen" effect to wyvern
```
helixhamin: 1 hp = 1 tic, ignore the 3 second comment on bg, since a tick is 3 seconds...

http://ffxiclopedia.wikia.com/wiki/Tick:
> The standard tick occurs every 3 seconds (Earth time). This is used to trigger the effects of DOT spells such as Poison and Choke as well as recovery effects such as Regen, Refresh, and Regain. Such effects generally last for a certain predetermined amount of ticks before wearing off.

and from BG: 
> The regen effect is 1 HP every 3 seconds.

TeoTwawki:
Ok so it does need an edit then. I'm at work right now so thanks for spotting it--06.06.2018, 11:01, "Delaide" <notifications@github.com>:@helixhamin commented on this pull request.

lol, @TeoTwawki , I made the same mistake as you, I looked up the item he put as his "example" comment, 18547, which is not the item changed here...hence the deleted comment and the update, although think I messed up and should have done single line comment, not "review" to my new comment.

In sql/item_mods_pet.sql:
> @@ -96,6 +96,7 @@ INSERT INTO `item_mods_pet` VALUES (11720,28,3,3);      -- Automaton: "Magic Atk
 INSERT INTO `item_mods_pet` VALUES (11739,27,2,1);      -- Caller's Sash Avatar: Enmity +2
 INSERT INTO `item_mods_pet` VALUES (11739,28,2,1);      -- Avatar: Magic Atk. Bonus +2
 INSERT INTO `item_mods_pet` VALUES (12520,27,-3,1);     -- Evoker's Horn Avatar: Enmity -3
+INSERT INTO `item_mods_pet` VALUES (12649,370,3,2);     -- Drachen Mail Adds "Regen" effect to wyvern

1 hp = 1 tic, ignore the 3 second comment on bg, since a tick is 3 seconds...
http://ffxiclopedia.wikia.com/wiki/Tick:

The standard tick occurs every 3 seconds (Earth time). This is used to trigger the effects of DOT spells such as Poison and Choke as well as recovery effects such as Regen, Refresh, and Regain. Such effects generally last for a certain predetermined amount of ticks before wearing off.

and from BG:

The regen effect is 1 HP every 3 seconds.

areckx:
Ok I fixed the value from 3 to 1

Yes, you did point me to the header of the file. What I mean to say is where is a list of what the values mean?

For example, in mathematics, it's no good to simply state that `x = 3`. Three what? Three days? Three minutes? Three percent?

So how am I supposed to know what the value means? I know now that the value for regen is hp/tick, but what if other effects work differently, as in percent or something? Or am I really overthinking this?

Also, it is my bad, I used a separate line as an example and didn't clearly indicate the line I was editing outside of the commit. I will strive to be clearer in future.

teschnei:
src/map/modifier.h will typically have a comment on what the magnitude of the modifiers are.  If they don't, you should be able to check any existing item with the modifier (item_mods table) and compare it with what it displays it has.

TeoTwawki:
> Or am I really overthinking this?


Yes. Things vary by what it is. Some are even fractions of 1024. But in all cases you can go where it is used or to the comment where it is defined and see the answer.

And 3 regen is obviously 3hp per tick of regen. What else could it even be that would leave that field compatible with the other things in this file? You gotta relax more areck, you psych yourself into thinking its hard and then it is. When I get a day off I'll show you few things you can use to help yourself.

areckx:
```
enum class PetModType
{
    All = 0,
    Avatar = 1,
    Wyvern = 2,
    Automaton = 3,
    Harlequin = 4,
    Valoredge = 5,
    Sharpshot = 6,
    Stormwaker = 7
};
```
This is why I should look closer at header files o_O 

Thanks

PR #4969, opened 2018-06-06
fix typos in Oldschool cures

wrenffxi:


wrenffxi:
hold review. more cure fixes incoming

PR #4971, opened 2018-06-07
Retail Update 06/07/2018

nfaletra:
Ran Text/NPC ID update tools for latest retail patch
TextIDs did not appear to shift, but I updated some of the comments that got marked.
NPCs shifted quite a bit. Either removal of event NPCs or some preparatory work for July's update

Also updated npc ids in lua scripts


TeoTwawki:
>but I updated some of the comments that got marked.

It's not the text in the lua file that matters to the tool. It is not comparing to those, its looking at what is in the xml dump produced by mass extractor and you can fix them here: https://github.com/DarkstarProject/darkstar-utils/blob/master/id-update-batch-jobs/binaries/config.properties#L606

nfaletra:
So how does anything that isn't in there work?

TeoTwawki:
variable name + number compared to mass extractor dump. I edited my comment prolly while you were still reading it.

the tool has problems on some strings with funky symbols so the config lets us manually override.

zynjec:
The tool absolutely uses those comments, it's the entire reason they're there.


https://github.com/DarkstarProject/darkstar-utils/blob/7d0f9479bfb7d3a9a88675f2cd98ccd70d15e3d2/id-update-batch-jobs/source/darkstar-batch/src/main/java/org/darkstar/batch/TextIdUpdate.java#L41-L43

https://github.com/DarkstarProject/darkstar-utils/blob/master/id-update-batch-jobs/source/darkstar-batch/src/main/java/org/darkstar/batch/TextIdUpdate.java#L198-L211

It first checks the comments, if it fails, it checks the overrides.

TeoTwawki:
¯\\_\(ツ)\_/¯
still worth an argument that the non human readable stuff is likely to be lost again anyway if we don't override. The tool trips up there a lot. Easy way to test : rerun and see if its still a "fixme". If it is we still need PR to the utils repo too.

Issue #4972, opened 2018-06-07
Utsusemi not working as intended


Delkathus:
**_I have:_**

- [X] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [X] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 

**_Source Branch_** (master/stable) **:Master** 

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Mobs with breath attacks now completely wipe Copy Image Effects. Splash Breath, Flame Breath, Radiant Breath, etc. They are to ignore the Copy Image effect and do damage.

Several physical single hit abilities like Queasyshroom are wiping the Copy Image Effect as well when it should only remove 1 shadow. Some AOE Magic abilities are not wiping the effect as well like they should.

Utsusemi: Ni also does not show up in text log when applied. Only when casting.

https://www.bg-wiki.com/bg/Copy_Image

ShelbyZ:
[Queasyshroom](https://www.bg-wiki.com/bg/Queasyshroom) - appears to be working as intended unless they made an update:

> This spell, whether it hits or misses, will remove all shadows on the target, regardless of blink or utsusemi.

There is no data to support Radiant Breath not wiping all shadows (wiki/bg).  It does however appear that Splash Breath was incorrectly wiping shadows.

DSP code around shadow wiping has not changed in quite some time, but certain moves may be incorrect and will need verification (on retail) to determine if the current choices are right or wrong.

Delkathus:
I'm trying to compile a list of abilities that are not effecting copy image as it should. Will be a day or two since there will be power disruptions on and off due to construction.

Wiggo32:
This falls under the category "not working as intended" but maybe should be it's own Issue?
When you zone with shadows already on, they cannot be overwritten.
1. Change Job to NIN
2. Cast Utsusemi: Ichi
3. Zone
4. Cast Utsusemi: Ni and get "no effect" message and no Ni shadows
Also will not let you overwrite same tier upon zone in. To get around this, you must manually remove the shadows before recasting and overwrites will go back to working as normal. (until next zone)

ShelbyZ:
Create a new issue

Delkathus:
I went through all the spells and abilities and they seem to be working as intended as for wiping shadows properly.

[...]

PR #4973, opened 2018-06-08
Utsusemi fix

ShelbyZ:
- Fix issue that Ni/San not showing message of effect being applied
- Apply fix for Splash Breath wiping shadows
- Partially addresses #4972, but needs more verification about what else is broken


Delkathus:
I'm trying to compile a list of abilities that are not effecting copy image as it should. Will be a day or two since there will be power disruptions on and off due to construction.

PR #4977, opened 2018-06-09
[WIP] Basic Campaign Implementation

### LINKED ISSUE #4960: Campaign "Region Info" doesn't show Allied Notes
**_I have:_**

**_Client Version_** (type `/ver` in game) **:** 30180427_2

1. Do some quest that adds Allied Notes
2. Check Amount under Campaign section of Region Info
3. It reports 0 no matter how much you have.
4. Check status -> currencies (shows actual amount you have)

MalRD:
This PR contains an implementation of the campaign system to support creating the correct campaign map packets  send to the client (it should also fix #4960). I'm looking for feedback on any changes that should be made to the implementation. The goal is to create a base that can be improved as we better understand the requirements to implement campaign completely.

**What this PR does**:

- Implement the 0x71 packets sent to the client for campaign map updates
- Add SQL tables for storing campaign area status and campaign nation status
- Add basic campaign map functions
  - Get / Set zone battle status
  - Get / Set zone control
  - Modify current / max fortification amount
  - Modify current / max resource amount
  - Modify influence value
  - Modify reconnaissance value
  - Modify morale
  - Modify prosperity
- Expose the campaign functions to Lua scripts

This PR does not touch campaign battles, notifications of campaign battles, troop movements, campaign ops, weekly tallies, or anything else related to campaign.


src/map/campaign_system.cpp
```diff
@@ -0,0 +1,364 @@
+/*
+===========================================================================
+
+Copyright (c) 2010-2015 Darkstar Dev Teams
```
ShelbyZ: I like how we still live in the past for copyright headers.

MalRD: I thought about updating them all, but then I decided not to.

src/map/lua/luautils.cpp
```diff
@@ -151,6 +151,19 @@ namespace luautils
         lua_register(LuaHandle, "getAbility", luautils::getAbility);
         lua_register(LuaHandle, "getSpell", luautils::getSpell);
 
+        lua_register(LuaHandle, "CampaignHasBattle", luautils::CampaignHasBattle);
```
teschnei: should these all be globals? I would imagine that any that use a zone as a parameter could be in lua_zone instead?

src/map/campaign_system.h
```diff
@@ -0,0 +1,183 @@
...
+
+enum CampaignControl : uint8
+{
+    SandoriaMask = 0x02,
+    BastokMask = 0x04,
+    WindurstMask = 0x06,
+    BeastmanMask = 0x08
+};
+
+enum CampaignZone : uint8
```
teschnei: should also have some way of getting this value from the regular zoneid.  a column in zone_settings, maybe, and then a local on the CZone

src/map/lua/luautils.cpp
MalRD: They shouldn't be globals. I wasn't sure where they'd end up, so I wrapped them in campaign.lua until a better solution arose. I'll look into moving them into lua_zone.

src/map/campaign_system.h
MalRD: I thought about it. I was going to remove the enum completely, as the core only uses it for type information, but I haven't yet confirmed if it is used in any of the troop movement messages. Once I confirm that, the plan is to change ID to be the zoneid and add a new column for sort order.

MalRD:
Thoughts and opinions: 

Campaign battles are limited to a maximum of one hour. The start time will need to be stored so the battle can be ended after that hour passes. The cool-down between battles is quick, but I'm not sure if there is a lower limit. The heroism gauge above the area increases the frequency of battles and may also impact any cool-down. 

Troop and freelancer movement is interesting. For freelancer's, we'll need to store which nation has recruited them. For both troop and freelancer, we'll need to store the dispatched status. They can't be dispatched to more than one area. I'm unsure if there's any cool-down between dispatches. 

The troop movement messages vary based on the reconnaissance level of the nations. Since these messages give you the status of a campaign battle in another zone (X returned victorious from Y), we'll have to pass the result of the battles to the relevant instances. This is where it seems like it would be beneficial to have an orchestrator serving up some messages. I don't recall if you can see the victory/defeated messages outside of the cities / zone the battle occurred in. If you can, having the message dispatch occur from a single process might be the way to go.

If we were to go that route, it may make sense to have the single instance control all campaign events. Each zone process could receive messages from the orchestrator to update the campaign state, send campaign dispatch messages, tell zones to start a battle, and etc. 

I don't believe that retail updates the data in real time. There were patch notes from when WotG was new-ish that talked about lowering the interval between region updates. In theory, that could've been reducing the amount of time that /cmap caches information locally, but I don't have any evidence to support it one way or the other. 

Allied notes appears to be the only character-specific variable in the packet, so we could cache the generation of the packet and only refresh it at an interval. This would reduce overhead of DB queries and rebuilding the packet, but it would also mean that you could zone to another server and get more out-dated information. Removing DB queries from the servers (making the orchestrator handle all DB activity) and pushing state updates to the zone processes would reduce the change of that occurring. 

Available campaign operations are based on the zones that a given nation owns. If we want retail captures for some of those instances, it would take several weeks and a decent number of people to take over the areas required to open up the operations.

Fortifications dropping to 50% reduces the current fortification value for that zone by one. Lower fortification values means that the defenders are quicker to retreat. I'm unsure what the default current / maximum values are for these. 

Resources are used to repair the fortifications and to buy (maybe) temporary items. If you have 0 resources in a zone, you can't buy temporary items and the fortifications won't be repaired by dispatched units. 

I don't know what morale and prosperity do. Had you asked me if that was a thing that ever existed, I would've said no. 

The Northlands can only be taken one zone per week. The three nations can have battles occur in them if the city doesn't own other territory. The cities cannot be taken over by beastmen tribes.

TeoTwawki:
> The cool-down between battles is quick, but I'm not sure if there is a lower limit. The heroism gauge above the area increases the frequency of battles and may also impact any cool-down.

I think any cooldown would be per military unit: Shadowhand Cuirassier's get defeated they take however long to recover. But some other unit, say Yuu Mjuu the Awakened and his Divine Sentinel's can show up immediately after..

..Or even during the same attack, overlapping and extending it - have seen this in game but there can still be a hard cap on how far a battle can be extended, I don't know. Some in game watching the clock will probably be needed at some point to tell us if we want to try and make it perfect.

I have defiantly seen the next within 1 clock tick of the previous ending, and that ought to be fairly easy to confirm - it wasn't a rare occurrence.

TeoTwawki:
LIVE! LIVE DAMN YOU!

zynjec:
Closing this for now, feel free to PR when you get back to it.

TeoTwawki:
:sob: Ducky has abandoned us to the Shadowlord's non attacking hordes :sob:

MalRD:
Hasn't even been a year and you're smashing hopes and dreams.

PR #4979, opened 2018-06-10
Item mod fixes

Jessland86:
Commented and fixed mods on Pteroslaver Armet/+1
Added mods for Pteroslaver Armet +2/3
Fixed missing Wyvern mods on Vishap Brais +1/2/3
Removed mods from Empyrian weapons that are not retail accurate. (Weaponskill damage buffs)
There is no documentation I could find anywhere in reference to Empy weaponskill damage bonus as a hidden effect on the weapons.

PR #4981, opened 2018-06-10
Added regen effect for drachen +1 and vishap(+1,+2,+3)

areckx:
Also corrected an entry in item_armor.sql where 23445 was given the name "vishap_mail_+3" instead of "spae._coat_+3"


areckx:
Fixed item ids and resolved conflicts with #4979

PR #4984, opened 2018-06-11
Added Endark to proper vendor NPC.

Jessland86:


helixhamin:
Appears correct, items check out with ffxiah in decimal.

PR #4987, opened 2018-06-13
Fixed subskills for some bullets

nfaletra:
Subskill was not set correctly for these bullets in the item_weapon table. Updated them to 1 for SUBSKILL_GUN.


teschnei:
🔫

PR #4988, opened 2018-06-13
change setting that enabled/disabled CoP mission caps into a setting for all mission BCNM for all expansions.

TeoTwawki:


helixhamin:
I approve this change! (Tired of trying to do 2-3 capped, as a solo player...)

src/map/battlefield.cpp
```diff
@@ -235,33 +235,25 @@ uint8 CBattlefield::getPlayerMainLevel()
...
+{
+    //adjust player's level to the appropriate cap and remove buffs
     if (m_PlayerList.size() == 0)
     {
         ShowWarning("battlefield:getPlayerMainLevel - No players in battlefield!\n");
         return;
     }
-    uint8 cap = getLevelCap();
+
+    uint8 cap = std::clamp<uint8>(getLevelCap(), 0, 99);
```
santssoft: cap should be between 1 and 99 instead of 0 and 99?

sql/bcnm_info.sql
```diff
@@ -45,262 +46,262 @@ CREATE TABLE `bcnm_info` (
...
-INSERT INTO `bcnm_info` VALUES (33,140,'holy_crest','Shitstain',1,18,600,75,6,0,7);
-INSERT INTO `bcnm_info` VALUES (34,140,'wings_of_fury','Shitstain',1,18,900,20,3,34,15);
-INSERT INTO `bcnm_info` VALUES (35,140,'petrifying_pair','Shitstain',1,18,900,30,3,35,15);
-INSERT INTO `bcnm_info` VALUES (36,140,'toadal_recall','Shitstain',1,18,1800,30,6,36,15);
-INSERT INTO `bcnm_info` VALUES (37,140,'mirror_mirror','Shitstain',1,18,900,40,3,0,7);
-INSERT INTO `bcnm_info` VALUES (64,144,'rank_2_mission','Shitstain',1,18,1800,25,6,0,5);
-INSERT INTO `bcnm_info` VALUES (65,144,'worms_turn','Shitstain',1,18,1800,40,6,0,15);
-INSERT INTO `bcnm_info` VALUES (66,144,'grimshell_shocktroopers','Shitstain',1,18,1800,60,6,0,15);
-INSERT INTO `bcnm_info` VALUES (67,144,'on_my_way','Shitstain',1,18,1800,99,6,0,5);
-INSERT INTO `bcnm_info` VALUES (68,144,'thief_in_norg','Your mother',1,18,1800,99,6,0,5);
```
davismj: wtf

TeoTwawki: It was demolish. Except "your mother" I think was Autkast.

src/map/battlefield.cpp
TeoTwawki: no because a zero cap is valid.

99 cap gets used to strip buff 0 gets used for uncapped that doesn't strip buffs

sql/bcnm_info.sql
davismj: Represent

TeoTwawki:
rebased to resolve conflict

PR #4989, opened 2018-06-13
Crawler's Nest and Zvahl Baileys IDs

wrenffxi:
identify Dark Spark torch by ID offset rather than coordinates

store "spawned from Torch number" as mob local variable rather than player variable (since player variable will only be set on the spawning player, not anyone else in party).

~**Note**: I assume clicking the wrong torch respawns Dark Spark only because that's how it was coded in the existing LUA.  I cannot find anything to verify this on the 'net.  I see "click the same torch used to spawn Dark Spark to receive KI" -- but no indication of what happens if you click the wrong torch.~

style guide


TeoTwawki:
if memory serves, on retail wrong torch simply has no response at all.

wrenffxi:
All right.  That makes more sense.  I'll adjust this PR tonight.

PR #4993, opened 2018-06-15
Cleaned up NPC Shops (F-L)

zynjec:


scripts/zones/Lower_Jeuno/npcs/Ghebi_Damomohe.lua
```diff
@@ -6,91 +6,81 @@
...
-            --Gray Chip (Ex) - Diremite Stalker (at Tower near Ranguemont), Diremite Assaulter, Diremite
-            --Cyan Chip (Ex) - Treasure Chest Mimics
-            player:startEvent(52);
+    if player:getQuestStatus(JEUNO,TENSHODO_MEMBERSHIP) ~= QUEST_COMPLETED and npcUtil.tradeHas(trade, 548) then
+        -- Finish Quest: Tenshodo Membership (Invitation)
+        player:startEvent(108)
+    elseif player:getCurrentMission(COP) == DARKNESS_NAMED and
+           not player:hasKeyItem(dsp.ki.PSOXJA_PASS) and
+           player:getVar("PXPassGetGems") == 1 and
+           npcUtil.tradeHas(trade, 1692) or npcUtil.tradeHas(trade, 1694) or npcUtil.tradeHas(trade, 1693) then
```
wrenffxi: I think AND is processed before OR in order of operations, so you'd want to put parentheses around this group of three tradeHas's

zynjec: Ah you're right, good catch.

PR #4994, opened 2018-06-15
fix wyvern doing multi ws after dismiss/call

areckx:
[15/Jun] [15:32:49][Error] luautils::onMobDeath: scripts/globals/pets/wyvern.lua:157: attempt to index local 'player' (a nil value)

set to 'master'

### DIFF
--- a/scripts/globals/pets/wyvern.lua
+++ b/scripts/globals/pets/wyvern.lua
@@ -154,8 +154,7 @@ end;
 
 function onMobDeath(mob, player)
     local master = mob:getMaster();
-    local pet = master:getPet()
-    local numLvls = pet:getLocalVar("level_Ups")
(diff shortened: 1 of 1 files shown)


scripts/globals/pets/wyvern.lua
```diff
@@ -154,7 +154,7 @@ end;
 
 function onMobDeath(mob, player)
     local master = mob:getMaster();
-    local pet = player:getPet()
+    local pet = master:getPet()
```
TeoTwawki: I think we should just be dropping that line and changing the next to 
```lua
local numLvls = mob:getLocalVar("level_Ups")
```
since in this context the "mob" is the wyvern.

Otherwise we are effectively getting master from the mob then get the mob from the master right after.

PR #4997, opened 2018-06-17
Metalworks Elevators

areckx:
https://github.com/DarkstarProject/darkstar/blob/7c23efa653cd3f78b45c8500f42b4e09d6fc579d/src/map/lua/lua_baseentity.cpp#L1892


areckx:
Testing for awhile. For some reason the value 1 will result in a zero during the check or something like that. elevators.sql seems to not be used at all, so that should be removed. looking into why 6 seems to be working (or I'm guessing any value above 1) because what I think is happening is that the 1 is being subtracted by 1 and making it 0, resulting in isPermanent check to set value to 0

Wiggo32:
What is the issue you are trying to address with this? And what does this change do to gameplay?

helixhamin:
https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Metalworks/npcs/_6lt.lua

Wiggo32:
does deleting those values from the database resolve the issue with the elevators? meaning, do they now work flawlessly as they're supposed to?

TeoTwawki:
the elevators sql table isn't even used anymore, its a relic.

Wiggo32:
awww... i though he might have fixed it finally... so we probably shouldn't have that table in our database then?

TeoTwawki:
I've got fix incoming tonight

PR #5000, opened 2018-06-18
Casting animation for taru students in Orastery

Wiggo32:


Wiggo32:
redo cuz im a bananahead...

PR #5004, opened 2018-06-18
spawn mechanics, 2hrs, additional effects for beastman kings

wrenffxi:


scripts/zones/Qulun_Dome/mobs/ZaDha_Adamantking.lua
```diff
@@ -1,62 +1,59 @@
...
-    local nqId = mob:getID() - 1;
-    local wait = 72 * 3600;
-    SetServerVariable("[POP]Za_Dha_Adamantking", os.time() + wait); -- 3 days
-    SetServerVariable("[PH]Za_Dha_Adamantking", 0);
-    DisallowRespawn(mob:getID(), true);
-    DisallowRespawn(nqId, false);
-    UpdateNMSpawnPoint(nqId);
-    GetMobByID(nqId):setRespawnTime(math.random(75600,86400));
-end;
+    local nqId = mob:getID() - 1
```
ShelbyZ: We should consider creating a helper method to handle these style of pops:

- set server vars for pop/ph
- disallow respawns
- update spawn points
- set respawn time

**In a later set of changes** and port those back to all other recent updates.

PR #5006, opened 2018-06-19
Elevator updates

TeoTwawki:
Forth Ghelsba working proper right down to the winch animations.

Metalworks still screwy but without crisscrossed "door" toggles now (retail packet confirmation of NPC IDs involved).

I think the duration doesn't line up with the platforms animation time, so every so many cycles its being changed again while mid animation resulting in the platform changing direction in the middle of a trip and the "doors" going out of sync. If I am right, that will need a change here: 
https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_baseentity.cpp#L1895
so that we can specify duration's

I won't be able to test that theory till the weekend. Incremental improvemenst ppl, please don't leave this sitting just for that.


areckx:
Closing #4997 man I would never have gotten this far! lol

I can see now the elevator animation speed issue. Looking into it.

scripts/zones/Fort_Ghelsba/npcs/_3x2.lua
```diff
@@ -0,0 +1,36 @@
...
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end;
+
+function onTrigger(player,npc)
+    -- Animate lever
+    npc:openDoor(1)
+
+    -- Animate lever's Gear
```
takhlaq:
```lua
local myID = npc:getID()
local lever = GetNPCByID(myID + 2)
local winch = GetNPCByID(myID - 2)
```

scripts/zones/Fort_Ghelsba/npcs/_3x2.lua
```diff
@@ -0,0 +1,36 @@
...
+    if GetNPCByID(npc:getID()+2):getAnimation() == 8 then
+        GetNPCByID(npc:getID()+2):setAnimation(9)
```
takhlaq: dsp.animation.WHATEVER

scripts/zones/Fort_Ghelsba/npcs/_3x1.lua
```diff
@@ -0,0 +1,36 @@
...
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end;
+
+function onTrigger(player,npc)
+    -- Animate lever
+    npc:openDoor(1)
+
+    -- Animate lever's Gear
```
takhlaq: same here

TeoTwawki:
Fix'd

PR #5020, opened 2018-06-22
Fixing luarocks issue Shadow Lord comparison

ShelbyZ:
7d01 --> 0x07d01

### DIFF
--- a/scripts/zones/Throne_Room/bcnms/shadow_lord_battle.lua
+++ b/scripts/zones/Throne_Room/bcnms/shadow_lord_battle.lua
@@ -51,7 +51,7 @@ end;
 
 function onEventFinish(player,csid,option)
     -- print("bc finish csid "..csid.." and option "..option);
-    if (csid == 7d01) then
+    if (csid == 0x07d01) then
         if (player:getCurrentMission(player:getNation()) == 15 and player:getVar("MissionStatus") == 3) then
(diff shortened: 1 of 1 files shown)


scripts/zones/Throne_Room/bcnms/shadow_lord_battle.lua
```diff
@@ -51,7 +51,7 @@ end;
 
 function onEventFinish(player,csid,option)
     -- print("bc finish csid "..csid.." and option "..option);
-    if (csid == 7d01) then
+    if (csid == 0x07d01) then
```
TeoTwawki: This should have been fixed when we merges master back to bcnmcrap last. its means something wasn't merged right (or at all, I was sure we did a back merge already..) which has me concerned some other things are missing now.

zynjec: I think last time demo merged in master, he just picked bcnmcrap's scripts for all bcnms, so this problem probably doesn't extend past bcnm scripts.

TeoTwawki: I had worked on merging them correctly and pr'd it to bcnm crap though... no reason it would have been wrecked after. That behavior will rebreak a lot of stuff in the future when we finally merge bcnm crap to master which was why I did that for you :-(

TeoTwawki: I'll compare shiz to try and spot the breaks iver the weekend and pr to bcnm crap again

PR #5031, opened 2018-06-25
Cleaned up NPC Shops (M-N)

zynjec:


scripts/zones/Bastok_Markets/npcs/Oggodett.lua
```diff
@@ -24,11 +22,11 @@ function onTrigger(player,npc)
             629,    43,    -- Millioncorn
             4415,  111,    -- Roasted Corn
             4505,   90,    -- Sunflower Seeds
-            841,    36     -- Yagudo Feather
+            841,    36,     -- Yagudo Feather
         }
 
         player:showText(npc, OGGODETT_OPEN_DIALOG)
-        dsp.shop.general(player, stock, BASTOK)
+        dsp.shop.general(player, stock, dsp.nation.BASTOK)
```
wrenffxi: unlike dsp.shop.nation, the dsp.shop.general function doesn't take a nation as its third parameter.  it takes logId (scripts/globals/log_ids.lua), which we haven't tabled yet.  So BASTOK is actually correct.

zynjec: Oh yikes, I'll revert the general shops today then.

Serves me right for not looking at the actual function.

PR #5042, opened 2018-06-29
Cleaned up NPC Shops (R-U)

zynjec:
Dabih Jajalioh & Macchi Gazlitah need core work to track amount of gil earned through sales for occasional goods.


scripts/zones/Tavnazian_Safehold/npcs/Caiphimonride.lua
```diff
@@ -3,33 +3,36 @@
...
-    18214,51905}    --Voulge (COP Chapter 4 Needed; not implemented yet)
-
-    dsp.shop.general(player, stock);
-end;
+        16450, 1867,    -- Dagger
+        16566, 8478,    -- Longsword
+        17335,    8,    -- Rusty Bolt
+    }
+
+    if player:getCurrentMission(COP) >= SHELTERING_DOUBT then
```
wrenffxi: I remember that you want to do mission progress checks with player:hasCompletedMission() instead of >= {MissionName}, because sometimes the player's mission value is set to 255 between missions or something.  I had to correct an old PR similarly here:

https://github.com/DarkstarProject/darkstar/pull/4129/commits/5c130075922ca5d9e910ca787cf59f9ed5312249

zynjec: I can do some testing with it, but if that's the case, there should never be a scenario where COP is without a mission to return 255.

It starts on THE_RITES_OF_LIFE (1), and will always sit at THE_LAST_VERSE (94) even once the story is completed.

zynjec: I tried to break this with GM commands, even deleting everything in the log still returns 0, so I think we're safe here.

wrenffxi: Sweet.  Thanks for double-checking.

PR #5043, opened 2018-06-29
Added 2hrs to remaining Dynamis mobs

wrenffxi:
Removed several unneeded scripts from Dynamis Bastok


Wiggo32:
didn't check them all but a lot of those first NM's are Dynamis Jeuno NM's... don't think they should be in the Dynamis Bastok folder.

wrenffxi:
Oh god.  Thanks, will fix shortly.

scripts/zones/Dynamis-Bastok/mobs/Vanguard_Militant.lua
```diff
@@ -1,18 +1,9 @@
 -----------------------------------
 -- Area: Dynamis Bastok
---  MOB: Vanguard Militant
+--  MOB: Vanguard_Militant
```
TeoTwawki: why adding the underscore? I don't mind it when we fix complete wrong ones by copying out of the sql but don't see a reason to put it in when the name was otherwise accurate. Find&replace regex happen here?

wrenffxi: Heh. I wish it were find/replace.  It was more like ... I got Dynamis Bastok and Dynamis Jeuno folders mixed up somehow (see above) and deleted all the Vanguard scripts from Bastok's folder.  So I copied them all over from another Dynamis folder and changed the Area to Dynamis Bastok all sneaky-like.

teschnei:
hmm.. this will be interesting to merge into bcnmcrap

wrenffxi:
wouldn't think it'd be too hard.  95% of it's adding scripts for mobs that didn't have them, and the other 5% is adding a single line with the job special mixin to existing scripts.

teschnei:
yeah, but they'll all have scripts in bcnmcrap that are already using a
mixin so you have to combine the mixins

which on it's own isn't hard, but there's like... a million scripts

wrenffxi:
note: combined mixins in bcnmcrap in separate PR

teschnei:
this good on my end, @TeoTwawki

TeoTwawki:
> this good on my end, @TeoTwawki

why no no just approve den. ya silly goof.

scripts/zones/Dynamis-Beaucedine/mobs/BeZhe_Keeprazer.lua
```diff
@@ -0,0 +1,14 @@
+-----------------------------------
+-- Area: Dynamis Beaucedine
+--  MOB: BeZhe_Keeprazer
+-----------------------------------
+mixins = {require("scripts/mixins/job_special")}
+require("scripts/globals/status")
+-----------------------------------
+
+function onMobSpawn(mob)
+    mob:setLocalVar("mainSpec", dsp.jsa.ASTRAL_FLOW_MAAT)
```
TeoTwawki: not sure if right animation for these but I think thats a problem that predates this pr

PR #5045, opened 2018-06-30
Goblin Wolfman not Spawning at Scrawled Writing

Delkathus:
Goblin Wolfman will spawn in his default spawn pos, then move to the Scrawled Writing resulting in sitting in the default spawn location because the setPos is used before the mob is spawned.

Ideally it's best to use setSpawn to spawn at the onTrade location instead of spawning it else where and using setPos to relocate it.

### DIFF
--- a/scripts/zones/Oldton_Movalpolos/mobs/Goblin_Wolfman.lua
+++ b/scripts/zones/Oldton_Movalpolos/mobs/Goblin_Wolfman.lua
@@ -1,0 +1,20 @@
+-----------------------------------
+-- Area: Oldton Movalpolos
+--  MOB: Goblin Wolfman
+-----------------------------------
+require("scripts/zones/Oldton_Movalpolos/MobIDs");
+require("scripts/globals/settings");
+require("scripts/globals/status");
+-----------------------------------
+
+function onMobInitialize(mob)
+mob:setMobMod(dsp.mobMod.IDLE_DESPAWN, 55);
+
+function onMobSpawn(mob)
+end;
+
+function onMobDeath(mob, player, isKiller)
+end;
+
+function onMobDespawn(mob)
+end;
--- a/scripts/zones/Oldton_Movalpolos/mobs/Goblin_Wolfman.lua
+++ b/scripts/zones/Oldton_Movalpolos/mobs/Goblin_Wolfman.lua
@@ -8,7 +8,7 @@ require("scripts/globals/status");
 -----------------------------------
 
 function onMobInitialize(mob)
-    mob:setMobMod(dsp.mobMod.IDLE_DESPAWN, 55);
+    mob:setMobMod(dsp.mobMod.IDLE_DESPAWN, 180);
 end;
 
 function onMobSpawn(mob)
--- a/scripts/zones/Oldton_Movalpolos/mobs/Goblin_Wolfman.lua
+++ b/scripts/zones/Oldton_Movalpolos/mobs/Goblin_Wolfman.lua
@@ -8,7 +8,8 @@ require("scripts/globals/status");
 -----------------------------------
 
 function onMobInitialize(mob)
-mob:setMobMod(dsp.mobMod.IDLE_DESPAWN, 55);
+    mob:setMobMod(dsp.mobMod.IDLE_DESPAWN, 55);
+end;
 
 function onMobSpawn(mob)
 end;
--- a/scripts/zones/Oldton_Movalpolos/npcs/Scrawled_Writing.lua
+++ b/scripts/zones/Oldton_Movalpolos/npcs/Scrawled_Writing.lua
@@ -14,7 +14,7 @@ function onTrade(player,npc,trade)
         local x = npc:getXPos();
         local y = npc:getYPos();
         local z = npc:getZPos();
-        mob:setPos(x-1,y,z);
+        mob:setSpawn(x-1,y,z)
         SpawnMob(GOBLIN_WOLFMAN):updateClaim(player);
     end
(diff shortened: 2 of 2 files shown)


TeoTwawki:
what that really changes is where "home" is in the event he kills the player and has to walk back. as for when he spawns, that will be the same place. if he's sitting at the default location, he was spawned but not killed and has not depopped.

we've a mod mob to make things depop, he'll either need a script added to set it in `onMobInitialize()` or it added to the mob_spawn_mods table: https://github.com/DarkstarProject/darkstar/blob/db327fa75c69d5cfbd063ca501d3877f2b93930a/scripts/globals/status.lua#L2042

TeoTwawki:
The change you are making here is still correct, just if the problem is him being someplace else because he didn't depop, you'd also need the mobmod too

Delkathus:
Ah I see, so then it would be ideal to have the pos set during the onMobInitialize in Goblin_Wolfman.lua in relation to where the Scrawled Writing is.

Would this be the best method then for any mob that is spawned via NPC?

TeoTwawki:
> Ah I see, so then it would be ideal to have the pos set during the onMobInitialize in Goblin_Wolfman.lua in relation to where the Scrawled Writing is.

no. what you did here already is exactly right. but we **also** need to make sure that if a player spawns him and does not win, he spawns himself after a set time like retail does. This prevents a 2nd player from trying to pop him at a different location and thinking he is broken.

TeoTwawki:
> Ah I see, so then it would be ideal to have the pos set during the onMobInitialize in Goblin_Wolfman.lua in relation to where the Scrawled Writing is.

no. what you did here already is exactly right. but we **also** need to make sure that if a player spawns him and does not win, he despawns himself after a set time like retail does. This prevents a 2nd player from trying to pop him at a different location and thinking he is broken.

Delkathus:
LOL wow, i'm running on fumes here and miss read that. Time to goto sleep. XD

Delkathus:
Added the despawn mod to Goblin Wolfman.

scripts/zones/Oldton_Movalpolos/mobs/Goblin_Wolfman.lua
```diff
@@ -0,0 +1,20 @@
...
+-- Area: Oldton Movalpolos
+--  MOB: Goblin Wolfman
+-----------------------------------
+require("scripts/zones/Oldton_Movalpolos/MobIDs");
+require("scripts/globals/settings");
+require("scripts/globals/status");
+-----------------------------------
+
+function onMobInitialize(mob)
+mob:setMobMod(dsp.mobMod.IDLE_DESPAWN, 55);
```
TeoTwawki: gotta indent
```lua
stuff
    things inside stuff
end
```
4 spaces per level of indention.

```lua
function example(argumnet1, argument2)
    if conditions then
        if otherConditions then
            doThings
        end
    end
end
```

and make sure the value for that mobmod is retail accurate. I don't thin 55 seconds is correct. its usually 3 or 5 minutes

scripts/zones/Oldton_Movalpolos/mobs/Goblin_Wolfman.lua
```diff
@@ -0,0 +1,21 @@
...
+-- Area: Oldton Movalpolos
+--  MOB: Goblin Wolfman
+-----------------------------------
+require("scripts/zones/Oldton_Movalpolos/MobIDs");
+require("scripts/globals/settings");
+require("scripts/globals/status");
+-----------------------------------
+
+function onMobInitialize(mob)
+    mob:setMobMod(dsp.mobMod.IDLE_DESPAWN, 55);
```
TeoTwawki: make sure the value for that mobmod is retail accurate. I don't think 55 seconds is correct. its usually 3 or 5 minutes

Delkathus:
Changed it to 3 minutes, couldn't find any sources of a exact time frame for despawn other than what you said of 3-5 minutes. Can be tweaked if necessary.

TeoTwawki:
alright. merging as is.

scripts/zones/Oldton_Movalpolos/mobs/Goblin_Wolfman.lua
```diff
@@ -0,0 +1,21 @@
+-----------------------------------
+-- Area: Oldton Movalpolos
+--  MOB: Goblin Wolfman
+-----------------------------------
+require("scripts/zones/Oldton_Movalpolos/MobIDs");
+require("scripts/globals/settings");
```
TeoTwawki: it'll get taken care of later but just so you know, we don't actually have to require settings or mobid's in the wolfman script because he's not using anything form them. he only needed status because thats where dsp.mobMod.IDLE_DESPAWN got defined

Delkathus: I was working on a bunch of custom stuff for lol's and it became a habit to put that in XD

Delkathus: I planned on making the Scrawled Writing to work as intended of spawning in a random location (3 places iicr). I can make the correction then.

TeoTwawki: we have a utils function for shuffling to diff positions I think.

the way to do it is despawn npc when NM spawns (right in the trade function) and choose a new random pos. retail has it hide right when you spawn him. the random location list includes the current location so sometimes it looks as if it "didn't move".

PR #5046, opened 2018-06-30
[bcnmcrap] Dynamis 2HRs

wrenffxi:
here is the job_special mixin added to the existing mixin on dynamis mobs


scripts/zones/Dynamis-Valkurm/mobs/Vanguard_Prelate.lua
```diff
@@ -1,21 +1,14 @@
...
+--  MOB: Vanguard_Prelate
 -----------------------------------
------------------------------------
-require("scripts/globals/status")
-mixins = { require("scripts/mixins/dynamis_beastmen") }
-
------------------------------------
--- onMobSpawn Action
+mixins = {require("scripts/mixins/dynamis_beastmen")}
+mixins = {require("scripts/mixins/job_special")}
```
teschnei: this not going to work like that, it'll replace the dynamis mixin with the 2hour one (which is why I said the merge was gonna be annoying heh).  So far I only got the dynamis listener on Dynamis Valkurm mobs, but it's probably ready to be added to all of them at this point.. but you'd have to also set the currency var if someone were to do that

teschnei:
You can, but they have to be in the same variable and separated by a comma
(it's an array basically)

scripts/zones/Dynamis-Valkurm/mobs/Vanguard_Prelate.lua
TeoTwawki:
```lua
mixins = {require("scripts/mixins/dynamis_beastmen"), require("scripts/mixins/job_special")}
```

PR #5050, opened 2018-07-02
misc edits and add 2 blu spells

KnowOne134:


scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,48 @@
...
+        params.dex_wsc = 0.0
+        params.vit_wsc = 0.0
+        params.agi_wsc = 0.0
+        params.int_wsc = 0.2
+        params.mnd_wsc = 0.0
+        params.chr_wsc = 0.0
+    damage = BluePhysicalSpell(caster, target, spell, params)
+    damage = BlueFinalAdjustments(caster, target, spell, damage, params)
+
+    if (damage > 0) then
```
TeoTwawki: with no resist check you are creating a 100% sleep rate for any mob that isn't outright immune.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,48 @@
...
+        local typeEffect = dsp.effect.SLEEP_I
+        target:delStatusEffect(typeEffect)
```
TeoTwawki: does it always overwrite even if the effect its replacing is stronger? because thats what you have happening here,

KnowOne134: Well it would overwrite becuase it does damage 1st then a sleep effect. Its the only physical sleep spell in the game

TeoTwawki: that just makes me realize the delstatus was even more redundant than I thought it was. since damage wakes you and this only applies sleep if you were hit/damaged by it that means there will never be a status to delete when delStatusEffect gets called.

if you do zero damage but landed it, will retail sleep? this script won't because checking for dmg > 0

probably want to check for hit instead of dmg value but if it must be > 0 dmg then we do what shockwave does. I recall there being a core issue with the order of operation, if thats not sorted out already then no script will work for this yet anyway. I *think* your other blue spell is ok but ima wait and let someone else look at that

KnowOne134:
The order of oporations is why i think i had the delstatus in there. I had made this spell maybr a year ago. Cleaning out my out right now

scripts/zones/Bastok_Markets/npcs/Degenhard.lua
```diff
@@ -5,49 +5,42 @@
...
-end;
+end
 
 function onTrigger(player,npc)
-    local BareBones = player:getQuestStatus(BASTOK,THE_BARE_BONES);
-    if (player:getVar("BeatAroundTheBushin") == 3) then
-        player:startEvent(342);
-    elseif (BareBones == 0) then
-        player:startEvent(256);
+    local BareBones = player:getQuestStatus(BASTOK,THE_BARE_BONES)
```
zynjec: Unused variable.

scripts/zones/Boneyard_Gully/bcnms/like_the_wind.lua
```diff
@@ -35,9 +35,9 @@ end;
 function onBcnmLeave(player,instance,leavecode)
 
     if (leavecode == 2) then -- play end CS. Need time and battle id for record keeping + storage
-        player:startEvent(32001,1,1,1,instance:getTimeInside(),1,0,0);
+        player:startEvent(32002,1,1,1,instance:getTimeInside(),1,0,0);    
     elseif (leavecode == 4) then
-        player:startEvent(32002);
+        player:startEvent(32003);
```
zynjec: `32003` appears to be the exit menu. Don't think this is right.

image

scripts/zones/Boneyard_Gully/bcnms/like_the_wind.lua
```diff
@@ -35,7 +35,8 @@ end;
 function onBcnmLeave(player,instance,leavecode)
 
     if (leavecode == 2) then -- play end CS. Need time and battle id for record keeping + storage
-        player:startEvent(32001,1,1,1,instance:getTimeInside(),1,0,0);
+        player:startEvent(32002,1,1,1,instance:getTimeInside(),1,0,0);
+        player:setVar("ENM_Win",1)
```
TeoTwawki: I'm 9000% sure we shouldn't be doing this with that variable. leavecode should already be indicating if we won or not.

TeoTwawki:
image

In the future, please keep it 1 thing to 1 pull request. the more you mix into one the longer it takes to get merged, Especially if more things get added to it. This looked like it should have been 4 prs: fish, blue spells, enm, and a quest. all unrelated.

KnowOne134:
i can break this down to multiple pulls and close this one when done. what to do that?

TeoTwawki:
that would help certainly if you are willing and able to. some parts could be merged fairly quickly while the others get reviewed.

PR #5051, opened 2018-07-03
feat(mob_skills): Add skill chain effects to blood pacts

davismj:
I'm using the pre-2016 behavior where skill chain blood pacts do not consume TP. First, there is no reliable data about how TP affects skill chain blood pacts that I could find. Second, it's a better design decision, as it needlessly interferes with the player's ability to save tp for defensive blood pacts.

I used bg-wiki for all skillchain data: https://www.bg-wiki.com/bg/Blood_Pact:_Rage

### DIFF
--- a/sql/mob_skills.sql
+++ b/sql/mob_skills.sql
@@ -650,67 +650,67 @@ INSERT INTO `mob_skills` VALUES (827,499,'lightning_blade',0,7,2000,0,1,0,0,0,0,
 INSERT INTO `mob_skills` VALUES (828,500,'water_blade',0,7,2000,0,1,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (829,501,'great_wheel',1,7.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (830,502,'light_blade',0,30.0,2000,1500,4,0,0,0,0,0,0);
-INSERT INTO `mob_skills` VALUES (831,513,'moonlit_charge',0,10.0,512,3000,4,0,0,7,0,0,0);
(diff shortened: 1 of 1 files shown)


Wiggo32:
any reason as to why you're removing the knockback effect from "moonlit charge" ? just curious if that interferes with chains or something...

nfaletra:
Couldn't find anything about knockback for Moonlit Charge. Don't think it's supposed to be there.

davismj:
Yea just fixin someones typo.

Wiggo32:
fenrir certainly does have a knockback affect with moonlit charge.
https://www.youtube.com/watch?v=WcZCtReetpw&feature=youtu.be

Hozu:
Maybe the Prime version of the mobskill has knockback, but the player version doesn't? I figured the player version should have been the one checked.

Timestamp for the knockback in that video is 43:40, btw.

zynjec:
Wiggo can double check for us to confirm, but Hozu seems to be correct.

The prime version has knock back, the player version does not. This video I found very clearly shows no knockback for the player version:
https://youtu.be/9pPiYth8DwE?t=297

Wiggo32:
Yes, you are correct, there is no knockback effect that happens on mobs. Although, even on dsp this knockback setting does NOT produce a knockback effect on mobs. Leaving it on like it is produces a knockback to players when hit from fenrir prime but no knockback when avatar fenrir hits a mob... just as retail does.

Wiggo32:
oh, thought i pasted the vid "at current time" ... lol my bad =)

TeoTwawki:
player/non player versions of things are usually different ID'd duplicates. Do these share ID, and if they do is there a duplicate readily available?

davismj:
Made the changes as per Wiggo's suggestion.

PR #5052, opened 2018-07-03
Adding in Entity ID for drop gate, allowing the switch to work.

Snugglesffxi:
this should now allow the switch at !pos -56 45 40 160 in Den of rancor to work.


TeoTwawki:
> Text ID

You mean entity ID (its an npc's ID)

:-)

Snugglesffxi:
oh yes my bad! sorry :)

TeoTwawki:
if you intended to change the commit message, you can't do that from github. The pull request title is its own thing. Only a git client can do that (via the amend feature, followed by a force push).
https://help.github.com/articles/changing-a-commit-message/

After work today I can take care of that on your behalf before I merge, if you need, But this is a good opportunity to learn a useful feature. :wink:

Snugglesffxi:
I think I managed to actually do it :) this commit is not worth nearly as much as this valuable skill, thank you!!

TeoTwawki:
revert won't help. it doesn't remove commits. Looks like you forgot to force push to overwrite, then pulled back the other commit so wound up with both commits instead of replacing it.

Snugglesffxi:
yeah im just gonna go ahead and make a correct one, this was an awesome learning experience though :) so thank you!

TeoTwawki:
it didn't require killing the PR or branch, I could have helped you after I got out of work. :(

Snugglesffxi:
np i'll put up a new one and fix it correctly, that branch wasnt the one i wanted to use anyway i have another one!

TeoTwawki:
k, if you run into anymore trouble remember we have a discord server to, someone is sure to help.

PR #5056, opened 2018-07-04
AH Cleanup

zynjec:
- Removed all items with FLAG_EX from AH.
- All unlisted items are now using category 0 instead of 99 to avoid potential collisions in the future, however unlikely.


TeoTwawki:
makin me pull changes local to view them, on muh only day off on a holiday -sniffles-

meanie.

TeoTwawki:
pls pls pleeeeease break things up into multiple commits when you get big ones like this in the future

Issue #4752, opened 2018-04-08
Conquest system math error


VincentHoshino:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 
30171223_0

**_Source Branch_** (master/stable) **:** 
stable

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Conquest point gain/loss being calculated incorrectly with negative values

map\conquest_system.cpp

auto loss = std::min<int>(points * influences[i] / 5000 - influences[nation], influences[i]);
 should be  
auto loss = std::min<int>(points * influences[i] / (5000 - influences[nation]), influences[i]);
 
Issue discovered by Ranzera

Issue #1416, opened 2015-05-09
Conquest


whasf:
Underflows all over in GainInfluencePoints()
Hopefully someone with more knowledge in C++ data types can fix it.

whasf:
This is still happening as of latest rev

whasf:
i reset conquest_system to imported table, login.. kill a mandy in east saru, and the values get all messed up still

whasf:
@z16

takhlaq:
i think this was fixed at some point but don't care to login and test
@whasf can you close if this is no longer an issue

PR #5057, opened 2018-07-04
Fixed order of operations in UpdateInfluencePoints

zynjec:
Fixes https://github.com/DarkstarProject/darkstar/issues/4752
Fixes https://github.com/DarkstarProject/darkstar/issues/1416


zynjec:
Think this also fixes https://github.com/DarkstarProject/darkstar/issues/1416

PR #5061, opened 2018-07-05
Sealion/Coral crest door in Sacrarium now has correct event/messages.

wrenffxi:


scripts/zones/Sacrarium/npcs/Small_Keyhole.lua
```diff
@@ -3,35 +3,42 @@
...
 
 function onEventUpdate(player,csid,option)
+    if csid == 100 and option == 2 then
+        GetNPCByID(SMALL_KEYHOLE):setLocalVar("canTradeSecondKey", 0)
+    end
 end
 
 function onEventFinish(player,csid,option)
-end;
+    if csid == 100 then
```
TeoTwawki: shouldn't we also have an option code checked here? this will break the key always, even if the user gets kicked out of the event. (I'm genuinely uncertain of the retail behavior, been too many years)

wrenffxi: unsure re: retail. the event is about 2.5 seconds long, and supposed to break the coral key regardless of what happens with the other sublock.

TeoTwawki: does the event let you escape key out?

wrenffxi: Escape key does nothing during event.

TeoTwawki: k we're good then

PR #5063, opened 2018-07-06
PW and lamps immune to gravity bind sleep

wrenffxi:
https://rukenshin.livejournal.com/17133.html
"Gravity, Bind, and Sleep were never successfully landed on a PW form the entire attempt."
"[on lamps,] Gravity, bind, and sleep never worked (even with ES)"


TeoTwawki:
While I am pretty sure it is indeed immune, I just want to put out there that you can easily tell immunity apart from super high resistance by the fact they generate 2 different messages in the client on modern retail. Many things we used to think were immune in 75 cap turned out to be just insanely high resistance that lv 99 players were able to overcome. I'm not sure when the 2 diff messages happened in ffxi's history (if it was always 2 messages, then man past players musta been really bad about noticing details).

PR #5070, opened 2018-07-08
Add setting for shared blood pact timer

godmode:


teschnei:
isn't this achieved by changing one number in the database? does it need a setting?

godmode:
It’s not. Each blood pact rage uses a recastid and each ward uses their own. You could change them all to 173 or 174 whatever it was and that would fix everything except updating the client that they can’t use either recast now and it would be server side check only.

PR #5071, opened 2018-07-08
Be able to fish more items

nuno-pires-cruz:
When FishingChance <= 20, you could only get quested items. With this change, you're now able to get the other possible items.

### DIFF
--- a/src/map/utils/fishingutils.cpp
+++ b/src/map/utils/fishingutils.cpp
@@ -192,8 +192,16 @@ bool CheckFisherLuck(CCharEntity* PChar)
 
 		if( ret != SQL_ERROR && Sql_NumRows(SqlHandle) != 0)
 		{
+            // array to store fish ids that i can get
+            std::vector<int32> fishIDs((int32)Sql_NumRows(SqlHandle));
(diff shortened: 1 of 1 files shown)


TeoTwawki:
ignore me, browser error

TeoTwawki:
but below were still comparing zero to another number and expecting it to be equal or less?
```cpp
 			int32 FisherLuck = 0;
             int32 FishingChance = dsprand::GetRandomNumber(1000);
 
 			while(Sql_NextRow(SqlHandle) == SQL_SUCCESS) 
 			{
 				FisherLuck += Sql_GetIntData(SqlHandle,7);
 
				if (FishingChance <= FisherLuck)
```

nuno-pires-cruz:
Hi,

In that part of the code, `FisherLuck` is always different from zero. Because he ads 0 + the value from table record ( witch is always > 0 , because of the sql statement "`AND lure.luck != 0`").

That part of the code, is still the original. If you remove that random and the verification, it would mean that the player would always catch something.

TeoTwawki:
ooooh I didn't see the += adding the sql row's value to it, my bad

PR #5077, opened 2018-07-09
fix for players unable to progress Three Paths with Cid.

wrenffxi:


scripts/zones/Metalworks/npcs/Cid.lua
```diff
@@ -220,13 +231,6 @@ function onEventFinish(player,csid,option)
         else
             player:messageSpecial(ITEM_CANNOT_BE_OBTAINED,13570)
         end
-    -- complete chapter "tree path"
-    elseif (csid == 853 or csid == 854 or csid == 855) then
```
TeoTwawki: *Looks at coded conditionals*

*looks again*

How did anybody ever complete this?

*triple take* 

thats just

*stares*

it would go into the 1st matching conditional. always. those should have been separate conditional tries not another `elseif`

PR #5083, opened 2018-07-14
Cleanup [Ab]

wrenffxi:
No logic changes.  Just removing unneeded requires and cosmetic / style guide changes.

edit: there is one logic change.  I removed the script for [Murrain Chigoe](https://ffxiclopedia.wikia.com/wiki/Murrain_Chigoe), since it does not need the chigoe mixin.


TeoTwawki:
> edit: there is one logic change. I removed the script for Murrain Chigoe, since it does not need the chigoe mixin.

it does need a script to later have its en-disease added though. even if we're not adding it right now I'd rather we just remove the mixin and leave a not so someone else doesn't try to put said mixin back.

TeoTwawki:
note\*

PR #5085, opened 2018-07-14
tabled rewards for lower jeuno treasure coffer

wrenffxi:
I've declared war on giant elseif blocks.  This battle was fought in Lower Jeuno.
elseifs before: 460
elseifs after: 2

### DIFF
--- a/scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
+++ b/scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
@@ -4,1552 +4,537 @@
 -- Type: Add-on NPC
 -- !pos 41.169 3.899 -51.005 245
 -----------------------------------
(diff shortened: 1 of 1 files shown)


TeoTwawki:
welcome to the front line soldier.

TeoTwawki:
I meant to change these to a weighted random forever ago, and never got around to it. Thank you for storming the trenches.  There is just one teensy problem with nexus cape though. You can't npc util's both instances of it. A change I had pending (needed the other crap tabled first, file to big) is that it needs to use `addUsedItem()` so that on acquisition its cooldown is at max if its been dropped and re-obtained, otherwise it is exploitable.

TeoTwawki:
if you incorporate that into the utils function in some way you can, up to you

wrenffxi:
Updated PR.  The coffer menu now shows/hides options based on expansions enabled, and whether player has previously received the Nexus Cape.  Reissues are give via addUsedItem().

PR #5086, opened 2018-07-15
Add sound effects to 3 mage circles + cleanup

Wiggo32:
With my discovery of entity animation packets I've found some missing sound effects that go along with the colored circles for the Sealed Portal.
I also tightened up the regions for the circles as best I could using a "boxed" region.
And adjusted the timing of everything to more closely match how retail does it.
Reference: https://youtu.be/9P1Y8lWEZNk

### DIFF
--- a/scripts/zones/Inner_Horutoto_Ruins/Zone.lua
+++ b/scripts/zones/Inner_Horutoto_Ruins/Zone.lua
@@ -3,118 +3,125 @@
 -- Zone: Inner_Horutoto_Ruins (192)
 --
 -----------------------------------
-package.loaded["scripts/zones/Inner_Horutoto_Ruins/TextIDs"] = nil;
+package.loaded["scripts/zones/Inner_Horutoto_Ruins/TextIDs"] = nil
 -----------------------------------
-require("scripts/zones/Inner_Horutoto_Ruins/TextIDs");
-require("scripts/zones/Inner_Horutoto_Ruins/MobIDs");
-require("scripts/globals/status");
-require("scripts/globals/zone");
+require("scripts/zones/Inner_Horutoto_Ruins/TextIDs")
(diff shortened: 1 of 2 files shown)


wrenffxi:
Not sure if you're aware of this already, but the registerRegion binding accepts [circular regions](https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_zone.cpp#L98-L106):
```zone:registerRegion(regionId, centerOfCircleX, radiusOfCircle, centerOfCircleZ, 0, 0, 0)```

That said, I've never understood how this circular region can work when you eliminate the vertical-coordinate.  Like, does it create a column of infinite height?  I feel like you'd trigger the region from rooms way beneath or above the intended location.  Maybe it's only useful for maps that don't have rooms stacks atop each other.

**Comment by TeoTwawki** [comment]
>  Like, does it create a column of infinite height

yes. so you'd have to check vertical position if there are any rooms above or below that spot, if using circle regions.

**Comment by Wiggo32** [comment]
Whatever ya'll want... it don't really matter to me. I basically did this for the sound effects, and noticed you could activate one of the circles from standing outside the circle... so... tightened it up.
```
Not sure if you're aware of this already, but the registerRegion binding accepts circular regions:
zone:registerRegion(regionId, centerOfCircleX, radiusOfCircle, centerOfCircleZ, 0, 0, 0)
```
What are the three zero's for at the end? I'm thinking that's just what dictates whether it's a circle or a box (lack of 2nd coordinate)?

wrenffxi:
```I'm thinking that's just what dictates whether it's a circle or a box (lack of 2nd coordinate)```
Correct.

I'm sure box is fine here, because the trigger area is so small.  Was more of a "this also exists" comment than recommendation to change anything.

PR #5090, opened 2018-07-16
cleaned up Yurim script

wrenffxi:
The elseif war has reached the Tavnazian Archipelago.  Today we saved a good scientist.
elseifs before: 61
elseifs after: 7

Added apple pie hints and cheat protection for quest "In the Name of Science."

### DIFF
--- a/scripts/zones/Tavnazian_Safehold/npcs/Yurim.lua
+++ b/scripts/zones/Tavnazian_Safehold/npcs/Yurim.lua
@@ -1,301 +1,132 @@
 -----------------------------------
 -- Area: Tavnazian Safehold
 --  NPC: Yurim
--- Standard Info NPC
+-- !pos 83.697 -25.000 3.250 26
 -----------------------------------
-package.loaded["scripts/zones/Tavnazian_Safehold/TextIDs"] = nil;
+package.loaded["scripts/zones/Tavnazian_Safehold/TextIDs"] = nil
 -----------------------------------
-require("scripts/zones/Tavnazian_Safehold/TextIDs");
-require("scripts/globals/keyitems");
-require("scripts/globals/quests");
+require("scripts/zones/Tavnazian_Safehold/TextIDs")
+require("scripts/globals/missions")
+require("scripts/globals/npc_util")
+require("scripts/globals/quests")
+-----------------------------------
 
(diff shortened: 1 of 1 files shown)


scripts/zones/Tavnazian_Safehold/npcs/Yurim.lua
```diff
@@ -1,301 +1,132 @@
...
-            player:messageSpecial(ITEM_CANNOT_BE_OBTAINED,itemtarget);
+    if csid == 524 then
+        player:addQuest(OTHER_AREAS_LOG, IN_THE_NAME_OF_SCIENCE)
+    elseif csid == 531 then
+        player:confirmTrade()
+    elseif csid == 526 then
+        player:confirmTrade()
+    elseif csid == 529 then
+        local itemInProgress = player:getVar("NAME_OF_SCIENCE_target")
+        local itemComplete   = player:getVar("NAME_OF_SCIENCE_complete")
```
TeoTwawki: we should be able to accomplish this using getLocalVar/setLocalVar

wrenffxi: Changed the anti-cheat one, but the in-progress var must remain as setVar since player could log out after trading the base items.

scripts/zones/Tavnazian_Safehold/npcs/Yurim.lua
```diff
@@ -119,10 +119,11 @@ function onEventFinish(player,csid,option)
         player:confirmTrade()
     elseif csid == 529 then
         local itemInProgress = player:getVar("NAME_OF_SCIENCE_target")
```
TeoTwawki: This one too? Should not need to persist through zonage

wrenffxi: That one must remain getVar.  It's a two-step process: trade the base item + chip, then (optionally) trade an apple pie, then trade the organs.  Player could logout after the first step, so a record of the thingee they're upgrading should persist.

TeoTwawki: oh ok.

TeoTwawki: the apple pie thing is new to me, didn't catch that

TeoTwawki:
if you encounter NPCs that trust the option code like this one did, give them the same exploit fix

PR #5097, opened 2018-07-17
Add some Coeurl skills

Wiggo32:
I added:
-Petrifactive Breath
-Frenzied Rage
-Pounce
by taking similar .lua's for mobskills already existing and changing them a bit to reflect the new mobskills
my models were: (Petribreath, Sharp Strike, and Big Scissors)

I changed:
-Charged Whisker (dmgmod)
From researching websites as well as retail testing, this is what *I felt* was the best number to put in there for more retail realistic play. It's nothing more than a best guess because I did not spend enough time to really get a most accurate number. If we don't like this I can just put it back.

I've put these in place ahead of working on some of the BCNM's I've been capturing for when they're ready to be implemented. The main inspiration for these was "Wild Wild Whiskers" BCNM which you can watch the videos in the BCNM channel on discord.

### DIFF
--- a/scripts/globals/mobskills/charged_whisker.lua
+++ b/scripts/globals/mobskills/charged_whisker.lua
@@ -2,21 +2,19 @@
 -- Charged Whisker
 -- Deals Lightning damage to enemies within area of effect.
 ---------------------------------------------------
-
-require("scripts/globals/settings");
-require("scripts/globals/status");
-require("scripts/globals/monstertpmoves");
-
+require("scripts/globals/monstertpmoves")
+require("scripts/globals/settings")
+require("scripts/globals/status")
 ---------------------------------------------------
 
 function onMobSkillCheck(target,mob,skill)
-    return 0;
-end;
+    return 0
+end
 
 function onMobWeaponSkill(target, mob, skill)
-    local dmgmod = 1;
-    local info = MobMagicalMove(mob,target,skill,mob:getWeaponDmg()*2.8,dsp.magic.ele.THUNDER,dmgmod,TP_MAB_BONUS,1);
-    local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,MOBSKILL_MAGICAL,MOBPARAM_THUNDER,MOBPARAM_WIPE_SHADOWS);
-    target:delHP(dmg);
-    return dmg;
-end;
+    local dmgmod = 2.5
+    local info = MobMagicalMove(mob,target,skill,mob:getWeaponDmg()*2.8,dsp.magic.ele.THUNDER,dmgmod,TP_MAB_BONUS,1)
+    local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,MOBSKILL_MAGICAL,MOBPARAM_THUNDER,MOBPARAM_WIPE_SHADOWS)
+    target:delHP(dmg)
+    return dmg
+end
--- a/scripts/globals/mobskills/frenzied_rage.lua
+++ b/scripts/globals/mobskills/frenzied_rage.lua
@@ -1,0 +1,27 @@
+---------------------------------------------
+-- Frenzied Rage
+--
+-- Description: Attack Boost
+-- Type: Enhancing
+-- Utsusemi/Blink absorb: N/A
+-- Range: Self
+-- Notes: 20% Attack Boost.
+---------------------------------------------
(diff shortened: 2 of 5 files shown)


scripts/globals/mobskills/charged_whisker.lua
```diff
@@ -2,21 +2,19 @@
...
+end
 
 function onMobWeaponSkill(target, mob, skill)
-    local dmgmod = 1;
-    local info = MobMagicalMove(mob,target,skill,mob:getWeaponDmg()*2.8,dsp.magic.ele.THUNDER,dmgmod,TP_MAB_BONUS,1);
-    local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,MOBSKILL_MAGICAL,MOBPARAM_THUNDER,MOBPARAM_WIPE_SHADOWS);
-    target:delHP(dmg);
-    return dmg;
-end;
+    local dmgmod = 2.5
```
TeoTwawki: > -Charged Whisker (dmgmod)
From researching websites as well as retail testing, this is what I felt was the best number to put in there for more retail realistic play. It's nothing more than a best guess because I did not spend enough time to really get a most accurate number. If we don't like this I can just put it back.

@teschnei I got no idea how hard that should hit or what the actual retail math is we are simulating in these variables (other than "one of them is its weapon and another is its str or int or wtf ever")

you review it :stuck_out_tongue:

scripts/globals/mobskills/petrifactive_breath.lua
```diff
@@ -0,0 +1,25 @@
...
+require("scripts/globals/settings")
+require("scripts/globals/status")
+---------------------------------------------
+
+function onMobSkillCheck(target,mob,skill)
+    return 0
+end
+
+function onMobWeaponSkill(target, mob, skill)
+    local typeEffect = dsp.effect.PETRIFICATION
```
TeoTwawki: same here, can skip the var

scripts/globals/mobskills/charged_whisker.lua
teschnei: all of our mobskills are guesses anyways, I don't think it's a problem - he did exactly what we normally do with new mobskills

TeoTwawki: :+1:

scripts/globals/mobskills/frenzied_rage.lua
```diff
@@ -0,0 +1,27 @@
...
+require("scripts/globals/settings")
+require("scripts/globals/status")
+---------------------------------------------
+
+function onMobSkillCheck(target,mob,skill)
+    return 0
+end
+
+function onMobWeaponSkill(target, mob, skill)
+    local power = 20
```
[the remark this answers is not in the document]
Wiggo32: Oh, ok. I figured it was more of a "let the reader know" what the param was for instead of it being a magic number. But, I can nix them from now on... It will save some space. =)

scripts/globals/mobskills/petrifactive_breath.lua
Wiggo32: This one though... if you can't figure out what this is without the var then... idk

I will get rid of those in the future. 👍

scripts/globals/mobskills/charged_whisker.lua
Wiggo32: if someone has a better answer or would like something different, I will change it to whatever it needs to be

scripts/globals/mobskills/frenzied_rage.lua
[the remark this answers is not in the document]
TeoTwawki: my line of thinking here is they are not required to be spelled out because anyone futzing with the function they are fed into should know that parameter/argument of the function is power just by looking at it:
https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/monstertpmoves.lua#L719
```lua
function MobBuffMove(mob, typeEffect, power, tick, duration)
```
(anytime we don't know what the functions arguments actually are, we should be looking at where the function is defined for that answer anyways)

if they are likely to be frequently edited (like everything in settings.lua), changed further along in the script, or would make a line to wide to read without excessive right/left scrolling, those are good indications it should be made a variable in my opinion.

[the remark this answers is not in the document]
Wiggo32: yes, you are correct! Roger that.

PR #5098, opened 2018-07-17
Implemented CSpecialReleasePacket

zynjec:
Implements 0x10E.

Required for Windurst 9-2 (and currently, the only thing captured using it).

### DIFF
--- a/src/map/packets/release_special.cpp
+++ b/src/map/packets/release_special.cpp
@@ -1,0 +1,33 @@
+/*
+===========================================================================
+
+Copyright (c) 2010-2018 Darkstar Dev Teams
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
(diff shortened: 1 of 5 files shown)


src/map/packets/release_special.h
```diff
@@ -0,0 +1,47 @@
...
+#include "../../common/cbasetypes.h"
+
+#include "basic.h"
+
+
+/************************************************************************
+*																		*
+*  																		*
+*																		*
+************************************************************************/
```
teschnei: just get rid of this empty block, i guess

src/map/packets/release_special.h
```diff
@@ -0,0 +1,47 @@
...
+
+This file is part of DarkStar-server source code.
+
+===========================================================================
+*/
+
+#ifndef _CRELEASE_SPECIALPACKET_H
+#define _CRELEASE_SPECIALPACKET_H
+
+#include "../../common/cbasetypes.h"
```
teschnei: is this header needed?

src/map/packets/release_special.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see http://www.gnu.org/licenses/
+
+This file is part of DarkStar-server source code.
+
+===========================================================================
+*/
+
+#include "../../common/socket.h"
```
teschnei: shouldn't need this right? since it's using ref instead of wbuf

src/map/packets/release_special.h
zynjec: Nope, removed now.

src/map/packets/release_special.cpp
zynjec: Also nope!

PR #5101, opened 2018-07-18
Updated NPC Amalasanda

zynjec:
- No longer a guild NPC. (Sends regular NPC shop item packets on retail, doesn't close, static prices, static stock)
- Updated with stock from April 2018 update.


zynjec:
Even if they want old stock, I think its more appropriate it be sent as a regular shop now, since that's what is sent on retail now.

I think the project itself should focus on what the game is like currently, and servers can adjust/add options themselves as it fits their community.

TeoTwawki:
Well yeah no disagreement there I was just thinking about leaving the other in commented out form alongside it.

Suprised this was merged yet I aproved last night lost my connection and then forgot about it.

PR #5104, opened 2018-07-18
Script cleanup (indentation, local vars) and Meret fix

wrenffxi:
Capitalized Leypoint script
Added cheat prevention to Meret
fixed lots of bad indentation
localed lots of variables


scripts/zones/Tavnazian_Safehold/npcs/Resauchamet.lua
```diff
@@ -8,14 +8,14 @@ function onTrade(player,npc,trade)
...
- local n=0;
-  if (kill< 8) then n=0;
-  elseif (kill< 15) then n=1;
-  elseif (kill< 50) then n=2;
-  elseif (kill>= 50) then n=3;
-  end
-player:startEvent(355,n);
+    local kill=player:getVar("FOMOR_HATE");
+    local n=0;
+    if (kill< 8) then n=0;
```
TeoTwawki: tripped me up a sec till I realized its collapsed inline

TeoTwawki: I'm going to revisit this eventually. seems like there should be a better way to handle than if-else'ing it

PR #5105, opened 2018-07-18
Changed skill list ID for Kreutzet

### LINKED ISSUE #5103: Kreutzet does not spam Stormwind
**_I have:_**

- [] checked the commit log to see if my issue has been resolved since my server was last updated

I was able to successfully blink tank Kreutzet. He used Stormwind but it wasn't used three times in succession and wasn't particularly dangerous. It's possible but unlikely that I just lucked out.

zynjec:
Fixes #5103 

Bandaid fix until we make listeners per-discord, skill list 1004 is for NM Rocs, keeps the same skills but with different id's, and a correct animation time for Stormwind.


TeoTwawki:
ya know, I think this might be my fault. I recall correcting mistakes in simurg/roc and some non nm greater birds :thinking: might have accidentally affected kreu

PR #5107, opened 2018-07-18
[WIP] Cover.lua

Dynas13:
```lua
function onAbilityCheck(player,target,ability)
    if (target == nil or target:getID() == player:getID() or not target:isPC()) then
        return dsp.msg.basic.CANNOT_ON_THAT_TARG, 0;
    else
        return 0, 0;
    end
end;

--formula from https://www.bg-wiki.com/bg/Cover
local baseDuration = 15;
local bonusTime = utils.clamp(math.floor((player:getMod(dsp.mod.VIT) + player:getMod(dsp.mod.MND) - target:getMod(dsp.mod.VIT) * 2) / 4), 0, 15)
local duration = baseDuration + bonusTime;

player:addStatusEffect(player,dsp.effect.COVER,0,0,duration)
  if player:physicalDmgTaken >= 0 then
    player.addMod(dsp.mod.Enmity, +200)
    target:lowerEnmity(target, 10)
  end
end
```


Dynas13:
Very work in progress, not sure on damage taken variables.

Wiggo32:
Just a thought:
Not sure if the effect power on cover is used much for anything other than to say "you have it".
Maybe it's used to indicate whether or not the PLD is wearing the gallant coronet which allows magical attacks to be 'covered'.
But, maybe we can also pass the player id in there to be able to tell which PLD the 'cover' belongs to in the case of more than 1 in party, so the right one gets the enmity.

Dynas13:
IIRC a second Cover does not overwrite an existing Cover which is why The <player>'s attempt to cover has no effect exists in message.log

TeoTwawki:
image

:confused:

Dynas13:
Basically I meant to make this a completely new file that doesn't exist yet. Where would I commit this from?

KnowOne134:
Fork the dsp github. Close to your pc make a new branch based on master then you can change anything you want when complete send in a pull request

PR #5111, opened 2018-07-19
removed double options of regen

KnowOne134:


TeoTwawki:
someday, I'll either remember wtf I was supposed to do with get those scriptable latents going, or talk someone else into getting that done >.>

food duration would be pretty easy to implement right now. Either by modifying duration in *all the food scripts* to check for it or by going "is this food?" in the core and then adding the bonus there when we apply effect.

PR #5115, opened 2018-07-19
Correct Anims -Seether Mob Abilities + Wanion Script

Wiggo32:
Pulled some animation ids from retail while sitting at a seether for a couple hours. Also, noted empty crush to have a slight knockback effect (I guessed at 2 seems close to me - someone else can correct me if they watch the vid after it uploads)

Wanion- lots of things changed
1. The GIANT list of effects in the effects table... it would probably be nice if there were an ailments effect flag but for lack of that, I didn't know how else to do it but to continue the table that was already in place.
2. I'm sure there are more ailments that wanion transfers (i.e. songs) but I haven't confirmed that so I left them off. I can surely add anything anyone wants to tell me they know that should be on the list.
3. When effects are transferred the duration is also transferred.
4. There are no messages for wanion (either success or miss) other than the message(s) of what effect(s) the mob has LOST. Which is handled nicely with the delStatusEffect()

This PR fixes the issue which caused crashes with clients and also debug breaks with dsp, which I cannot currently find in the issue tracker for some reason... ???

### DIFF
--- a/scripts/globals/mobskills/wanion.lua
+++ b/scripts/globals/mobskills/wanion.lua
@@ -1,43 +1,31 @@
 ---------------------------------------------
 -- Wanion
--- AoE of all status ailments it has.
+-- Transfers all ailments the Seether itself has to players in AoE range.
 ---------------------------------------------
-require("scripts/globals/monstertpmoves");
-require("scripts/globals/settings");
-require("scripts/globals/status");
-require("scripts/globals/msg");
+require("scripts/globals/monstertpmoves")
+require("scripts/globals/status")
+require("scripts/globals/msg")
 ---------------------------------------------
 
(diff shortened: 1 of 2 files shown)


Wiggo32:
https://youtu.be/v4jyObjV3jo

Hozu:
Might be able to check the erasable flag + a few more statuses to make the list of effects checked shorter. But then it could transfer stuff that it might not be able to on retail. I guess that would mean more retail checking for things like songs.

TeoTwawki:
> This PR fixes the issue which caused crashes with clients and also debug breaks with dsp, which I cannot currently find in the issue tracker for some reason... ???

probably badly named issue titles and/or someone closed their issue to soon. I remember legion opening at least 1 report related to this but their dump was unusable when it was looked at.

Issue #2736, opened 2016-02-08
Vrtra is capable of being pulled above ground without drawing in


Kubisnaxx:
People can pull him upstairs out of the room/range for blood aggro

Kubisnaxx:
near level 10 mobs

Hozu:
Interesting, last I heard Vrtra got stuck on the open door to exit its area.

Deadwing888:
This is one of several mobs on the "needs specialized draw-in regions" list.

For Vrtra draw in should be single target and go off when Vrtra's target is > ~20' away or outside of the door to the room. It's also going to be fairly critical that the draw in does not get people stuck under the floor the way Tiamat's still will.

PR #5116, opened 2018-07-19
Prevent exploiting mobs with navmesh over impassible terrain

Wiggo32:
I've been running this for about a month and I'm very happy with the results, you guys can try it out and see if you like it.

What this does:
This forces a mob to kind of 'wallhack' through ALL areas of terrain that were previously impassible for them.
Before this, the only thing people could do to avoid this exploiting was to disable navmesh and that caused mobs to be able to aggro through walls and just walk straight at players with no regard to obstacles. With this, the mob will follow the path 'around' terrain as it normally would with navmesh enabled but, when it finds itself becoming 'stuck' it just simply walks through the area that was an issue and then continues to track players around terrain as normal.

No more nuking lizards in crawler's nest from the top ledge while they just stare up at you. No more standing in "safe spots" next to walls to get away from the mobs attacks. No more exploiting...

### DIFF
--- a/src/map/ai/helpers/pathfind.cpp
+++ b/src/map/ai/helpers/pathfind.cpp
@@ -381,6 +381,7 @@ bool CPathFind::FindClosestPath(const position_t& start, const position_t& end)
 {
     m_points = m_PTarget->loc.zone->m_navMesh->findPath(start, end);
     m_currentPoint = 0;
+    m_points.push_back(end);  // this prevents exploits with navmesh / impassible terrain
 
     if (m_points.empty())
     {
--- a/src/map/ai/helpers/pathfind.cpp
+++ b/src/map/ai/helpers/pathfind.cpp
@@ -383,11 +383,13 @@ bool CPathFind::FindClosestPath(const position_t& start, const position_t& end)
     m_currentPoint = 0;
     m_points.push_back(end);  // this prevents exploits with navmesh / impassible terrain
 
+/* this check requirement is never met as intended since m_points are never empty when mob has a path
     if (m_points.empty())
     {
         // this is a trick to make mobs go up / down impassible terrain
         m_points.push_back(end);
     }
+*/
 
     return true;
 }


xipies:
Any possible regressions that could cause issues like #2736?

Wiggo32:
This is only affecting a mob's tracking ability enabling a temp wallhack when they become 'stuck'. So, any issues that were already in place prior to this will remain in place. Meaning if you used to be able to pull a mob up through the floor, you would still be able to pull said mob up through the floor. There have also been reports of certain 'problem' areas in dsp where mobs would aggro through the floor. This would also not change that. Those problem mobs would still aggro through the floor.
The types of cases I have found with this PR are things like the groundskeepers that were standing on the blue lights (on the elemental islands) will now come off of that tiny ledge and chase you if you walk in front of them, whereas before, that ledge prevented them from moving at all. 
What I would suggest is try it.  It's just 1 line of code to change, try it out and see if there are any downsides to this, because the only one I have found so far is the inability to cheat the game out of free exp.

TeoTwawki:
```cpp
bool CPathFind::FindClosestPath(const position_t& start, const position_t& end) 
{ 
    m_points = m_PTarget->loc.zone->m_navMesh->findPath(start, end); 
    m_currentPoint = 0; 
    m_points.push_back(end);  // this prevents exploits with navmesh / impassible terrain 


    if (m_points.empty()) 
    { 
        // this is a trick to make mobs go up / down impassible terrain 
        m_points.push_back(end); 
    } 


    return true; 
} 
```

look at the full function. you've just bypassed the check for if it has anymore navpoints and caused that pushback to always happen. that check is supposed to handle this, and it would be better to determine why it doesn't work that to just skip it altogether like that.

Wiggo32:
Yea, I thought about deleting that if statement because I couldn't figure out why it didn't work other than when the mob gets "stuck" it still has points so it is never "empty" as the check is defining there... I just found that by bypassing it altogether, it performs as it should by:
1. Having mobs follow a path around obstacles as per navmesh and
2. Only wallhacking for a brief moment *IF* if happens to get "stuck".
That's about all I have on that one... just figured it might be useful. Pathfinding needs some redoing at some point, maybe it's best to just wait on that and see if the problem can be handled there? Idk... Whatever you guys want...

TeoTwawki:
For now lets just 
/*
 comment that block out
*/
with a note about why we are doing so

PR #5118, opened 2018-07-20
[WIP] Implement Adventuring Fellow

Wiggo32:
There's a lot to do for this, and I'm setting it here so people can tell me if there's a better way to do things.  I basically took Teo's table mockup and modified it for my useage based on my retail captures and web info.

Since most of the bits are packed and it's been getting more and more complicated working on it, I figured I better get some feedback before it get's too crazy. 

So far, I have everything working up to the completion of the first quest "Unlisted Qualities" I can write/get all necessary values from the db table and put them where they need to be as params. 

Using the system I have laid out here, I have Rendezvous points *mostly* working there's still just a couple things I need to figure out with them. Next on the list is to finish up the quest line all the way to the bcnm fight and Tactics Pearl reward.

So, any feedback or guidance is appreciated... =)


scripts/zones/Upper_Jeuno/npcs/Bheem.lua
```diff
@@ -1,23 +1,82 @@
...
+    local B,K,OUT,I,D=16,"0123456789ABCDEF","",0
+    while IN>0 do
+        I=I+1
+        IN,D=math.floor(IN/B),math.mod(IN,B)+1
+        OUT=string.sub(K,D,D)..OUT
+    end
+    return OUT
+end
 
+function fellowParam(player)
```
teschnei: this needs to be done with bit ops instead of this - it is incredibly brittle and way more complex than it needs to be by assembling a number via a string and hoping the interpreter converts it right

teschnei: it is simply 4 bit.and (and 3 bit.lshifts) for the entirety of this function (and you can get rid of DECtoHEX)

teschnei: on that note, a giant `if/elseif` chain is usually a good sign to consider a table instead

Wiggo32: the table part is easy... but, for the bit shifts... i'd have to convert the values to binary first right? is there something i can look at already that uses bit ops on a bit packed param?

src/map/packets/char_health.cpp
```diff
@@ -51,4 +51,27 @@ CCharHealthPacket::CCharHealthPacket(CCharEntity* PChar)
         ref<uint8>(0x22) = PChar->GetSJob();
         ref<uint8>(0x23) = PChar->GetSLevel();
     }
-}
\ No newline at end of file
+}
+// NPCFELLOW --------------------------------------------vv
+CCharHealthPacket::CCharHealthPacket(CFellowEntity* PFellow)
```
teschnei: this is the exact same as the other constructor, just change the parameter and use that one

Wiggo32: ummm... it looks like you did this already...

src/map/packets/char_health.h
```diff
@@ -34,14 +34,13 @@
...
-class CCharEntity;
+class CBattleEntity;
 class CFellowEntity;  // NPCFELLOW
 class CCharHealthPacket : public CBasicPacket
 {
 public:
 
-	CCharHealthPacket(CCharEntity* PChar);
-    CCharHealthPacket(CFellowEntity* PFellow);  // NPCFELLOW
+	CCharHealthPacket(CBattleEntity* PChar);
```
Wiggo32: Should this be:
```   
CCharHealthPacket(CBattleEntity* PEntity);
```
to match the one in char_health.cpp?

teschnei: you can change it if you want, it's just the name so it doesn't particularly matter but it's probably better to change it

zynjec:
Going to close this for now, feel free to PR it again once completed.

PR #5119, opened 2018-07-20
[WIP] Cover.lua

Dynas13:
MrSent in discord added that we might need to swap players id in the mobs enmityList with the cover target while the effect is on then do updateTarget


Dynas13:
TODO:
Add transfer physical damage from target to player on ability use.

Dynas13:
Yup just updated the semicolon. Conversation on troubleshooting forum says a lot of work will also have to go into the core.

PR #5126, opened 2018-07-21
Implement PrintToArea, an extension of the custom end of PrintToPlayer

TeoTwawki:
Unlike PrintToPlayer, this targets *groups* of players via same methods normal chat uses but does *not* have to match the chat type.

You can now have mobs that make server wide claim announcements, as one example.
```lua
function onMobEngaged(mob, target)
    if target:isPC() then
        local TEXT = string.format("%s has engaged in battle with the notorious monster, %s!", target:getName(), mob:getName())
        target:PrintToArea(TEXT, dsp.msg.channel.SYSTEM_3, dsp.msg.area.SYSTEM)
    elseif target:isPet() then
        local TEXT = string.format("%s has engaged in battle with the notorious monster, %s!", target:getMaster():getName(), mob:getName())
        target:getMaster():PrintToArea(TEXT, dsp.msg.channel.SYSTEM_3, dsp.msg.area.SYSTEM)
    end
end
```
This would print a **yellow** system text message to all zones.

Yes this is custom. So is PrintToPlayer. If we don't want to put this in master I understand, but I feel its worth it. We previously said the same thing about PrintToPlayers message types. By having this commonly desired addition in our main trunk we can have fewer people turning their fork into a train-wreck and saying our updates broke their server.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -288,6 +288,53 @@ inline int32 CLuaBaseEntity::PrintToPlayer(lua_State* L)
 }
 
 /************************************************************************
+*  Function: PrintToArea()
+*  Purpose : version of PrintToPlayer that passes to messageserver
+*  Example : player:PrintToArea("Im a real boy!", dsp.msg.channel.SHOUT, dsp.msg.area.SYSTEM, "Pinocchio");
+*          : would print a shout type message from Pinocchio to the entire server
+************************************************************************/
+
+inline int32 CLuaBaseEntity::PrintToArea(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isstring(L, 1));
+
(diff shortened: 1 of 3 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -287,6 +287,53 @@ inline int32 CLuaBaseEntity::PrintToPlayer(lua_State* L)
...
+inline int32 CLuaBaseEntity::PrintToArea(lua_State* L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isstring(L, 1));
+
+    CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
+
+    // see scripts\globals\msg.lua or src\map\packets\chat_message.h for values
+    CHAT_MESSAGE_TYPE messageLook = (lua_isnil(L, 2) || !lua_isnumber(L, 2)) ? MESSAGE_SYSTEM_1 : (CHAT_MESSAGE_TYPE)lua_tointeger(L, 2);
```
takhlaq: `messageLook`wat

TeoTwawki: what the message "looks like". In this case a shout might not really be a shout (can send it server wide), so I have named the var to reflect then.

PR #5139, opened 2018-07-23
npcUtil.popFromQM should use FORCE_SPAWN_QM_RESET_TIME

wrenffxi:


TeoTwawki:
we annoyingly have both in trunk

https://github.com/DarkstarProject/darkstar/search?q=QM_RESET_TIME&type=Code

https://github.com/DarkstarProject/darkstar/blob/47229b04fc379b0c0c307f64ab5101cb60904f50/scripts/globals/settings.lua#L114

```lua
QM_RESET_TIME = 300; -- Default time (in seconds) you have from killing ???-pop mission NMs to click again and get key item, until ??? resets.
```

https://github.com/DarkstarProject/darkstar/search?q=FORCE_SPAWN_QM_RESET_TIME&unscoped_q=FORCE_SPAWN_QM_RESET_TIME&type=Code

https://github.com/DarkstarProject/darkstar/blob/47229b04fc379b0c0c307f64ab5101cb60904f50/scripts/globals/settings.lua#L161

```lua
FORCE_SPAWN_QM_RESET_TIME = 300; -- Number of seconds the ??? remains hidden for after the despawning of the mob it force spawns.
```

..These either need better naming and a really good case for being 2 diff things, or merged into one.

I don't see a reason to have 1 just for mission NM and one for all the other NM if the retail default value is identical.

zynjec:
I think just killing QM_REST_TIME is fine, personally.

I'm not actually sure there are any actual "???" that stay spawned while you're fighting the resulting monster, but even if there is, that can just be handled based on where logic is put.

TeoTwawki:
Sounds good to me too

PR #5143, opened 2018-07-24
synth(scorpion subligar): Add scorpion subligar desynth

davismj:


zynjec:
Your PR has a syntax error.

ERROR 1136 (21S01) at line 4652: Column count doesn't match value count at row 1

davismj:
fixed, thanks @zynjec

PR #5145, opened 2018-07-26
tabled conquest data and functions

wrenffxi:
* removed lots of redundant code
* better messaging while purchasing CP items
* conquest-related npcIDs moved into respective zone MobIDs luas
* OUTPOST and HOME_NATION are now teleports
* added chairs to CP rewards
* more thorough checks for CP purchases: level, player rank, nation rank, conquest victor when purchasing from foreign guards

I'm not certain my luaUtils.cpp code is correct.  I had to adjust it because setRegionalConquestOverseers moved from global to dsp.conquest table.

~edit: adding a couple things to this PR.~ All set!


src/map/lua/luautils.cpp
```diff
@@ -562,11 +562,30 @@ namespace luautils
...
         {
             lua_pop(LuaHandle, 1);
-            ShowError("luautils::SetRegionalConquestOverseers: undefined procedure SetRegionalConquestOverseers\n");
+            ShowError("luautils::SetRegionalConquestOverseers: undefined global dsp\n");
+            return -1;
+        }
+
+        int TableIndex = 0;
+        TableIndex = lua_gettop(LuaHandle);
+        lua_getfield(LuaHandle,TableIndex,"conquest");
```
teschnei: should be able to just do -1 for TableIndex, since you just pushed it to the top (-1 = top)

src/map/lua/luautils.cpp
```diff
@@ -562,11 +562,30 @@ namespace luautils
...
+        if (lua_isnil(LuaHandle, -1))
+        {
+            lua_pop(LuaHandle, 1);
```
teschnei: probably need to pop 2 here: one for lua_getglobal("dsp"), and one for lua_getfield("conquest")

src/map/lua/luautils.cpp
```diff
@@ -562,11 +562,30 @@ namespace luautils
...
+            ShowError("luautils::SetRegionalConquestOverseers: undefined field dsp.conquest\n");
+            return -1;
+        }
+
+        TableIndex = lua_gettop(LuaHandle);
+        lua_getfield(LuaHandle,TableIndex,"setRegionalConquestOverseers");
```
teschnei: same thing here again, -1 for index and remove gettop

src/map/lua/luautils.cpp
```diff
@@ -562,11 +562,30 @@ namespace luautils
...
+        if (lua_isnil(LuaHandle, -1))
+        {
+            lua_pop(LuaHandle, 1);
```
teschnei: and popping 3 here

src/map/lua/luautils.cpp
```diff
@@ -562,11 +562,30 @@ namespace luautils
...
+            ShowError("luautils::SetRegionalConquestOverseers: undefined procedure dsp.conquest.setRegionalConquestOverseers\n");
```
teschnei: i can't comment down far enough so i'll put it on this line, but I'm pretty sure you'll also need to pop 2 extras after the lua_pcall (ie. pop 3 if it fails instead of 1, and pop 2 if it succeeds).  Easy way to test: print lua_gettop before doing anything in this function, and then print it again after everything is done - since it pcall doesn't return any values, it should be the same before and after

teschnei:
yeah, that does look right to me, thanks

TeoTwawki:
@wrenffxi I didn't dig into your changes to see, did you happen to kill that minor exploit I told you about in discord that the version before your PR had as part of this change?

wrenffxi:
Yes.

PR #5157, opened 2018-08-02
global Text and Mob IDs converted to "zones" table

wrenffxi:
per Discord discussion.


zynjec:
One reservation with this, have you tested the textid tool against your format?

I would hold off on merging until we've confirmed it works, or there is a tool made to work with the new format, as that'd effectively kill updates.

wrenffxi:
The IDs.lua files created in this PR match the format of the existing Salvage zone very closely -- see https://github.com/DarkstarProject/darkstar/pull/5157/commits/f4a17fd4ba92a1375d5c58899d58775a964e5d5f -- so I assume that if the textID tool had no trouble converting those files, it'll have no trouble with these.  But no, haven't looked at the tool specifically.

zynjec:
The current tool does not support the IDs.lua/Salvage format files, so this will more than likely break ID updating entirely.

Should probably look at tool support before committing to master.

wrenffxi:
rebased, and also fixed a bug -- dsp.conq.setRegionalConquestOverseers was expecting two arguments, but only getting one from core.  So I made it work with one.

scripts/globals/conquest.lua
```diff
@@ -58,7 +58,8 @@ end
 -- (LOCAL) outposts
 -----------------------------------
 
-local outposts = {
+local outposts =
```
TeoTwawki: > they can't be teleported to, but are considered for conquest tally

actually tulia has a legit teleport it just goes unused on retail. but the NPC event for outpost tele's will display and use it (and we gave it a pos for warpage on DSP enabled by a setting I think).

wrenffxi:
rebased, and added Tu'lia to possible outpost teleport destinations.

scripts/commands/additem.lua
```diff
@@ -16,9 +16,7 @@ end;
 
 function onTrigger(player, itemId, quantity, aug0, aug0val, aug1, aug1val, aug2, aug2val, aug3, aug3val, trialId)
     -- Load needed text ids for players current zone..
-    local TextIDs = "scripts/zones/" .. player:getZoneName() .. "/TextIDs";
-    package.loaded[TextIDs] = nil;
-    require(TextIDs);
+    local ID = zones[player:getZoneID()]
```
teschnei: I think these (ie. all global functions that are getting IDs) should all be `local ID = require("scripts/zones/" .. player:getZoneName() .. "/IDs.lua")` right? if you zone into a zone and immediately `!addItem`, there's no guarantee the text ids table was loaded (it would be relying on any other npc in the zone who'd happen to load them before)

wrenffxi: The IDs.lua are required by every zone's Zone.lua file, so full table will be built at server start.

teschnei: hmm.. that's good enough then, I think

teschnei:
now.. I kind of wanted to get this in before Teo's formatting, but he merged it first.  @TeoTwawki - can you resolve the conflicts? I assume you can just use wren's on every file, and then rerun whatever formatter you were using (if even needed), yeah?

teschnei:
ah, there'll be some other stuff though from the windurst mission we merged earlier too, actually..

TeoTwawki:
ALL YOUR reBASE ARE BELONG TO US

sorry about that

wrenffxi:
rebased again.  converted to zones table all luas added since original PR.  hopefully.

wrenffxi:
rebased. converted files added by 5177.

wrenffxi:
rebased. converted files added by 5184.

wrenffxi:
Rebased.  Converted files added by 5205.

scripts/zones/Leujaoam_Sanctum/IDs.lua
```diff
@@ -1,118 +1,22 @@
...
-            LEUJAOAM_WORM9 = 17059849,
-            LEUJAOAM_WORM10 = 17059850,
-            LEUJAOAM_WORM11 = 17059851,
-            LEUJAOAM_WORM12 = 17059852,
-            LEUJAOAM_WORM13 = 17059853,
-            LEUJAOAM_WORM14 = 17059854,
-            LEUJAOAM_WORM15 = 17059855,
-        }
+    mob =
+    {
```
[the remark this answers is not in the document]
wrenffxi: They're fixed in a later commit in this same PR.

wrenffxi:
rebased. converted files from 5219. fixed a calculation in conquest.lua

wrenffxi:
I've written my own code to update text ids -- it's not Java, so can't contribute it to the dsp-utils) -- and it's showing that tons of our text ID comments don't match what comes out of the Mass Extract, even after regex replacing the tags/special characters into something human-readable.

I've committed suggested changes to the A% zones here.  Is this something you want for all zones, or shall I revert the last commit?

wrenffxi:
rebased. converted files added by 5229.

wrenffxi:
rebased. converted files from 5190.

PR #5158, opened 2018-08-05
Windurst Mission 9-2 Moon Reading completed

ffxijuggalo:
Here it is folks! The final Windurst mission to be able to obtain rank 10. This was a joint effort involving several different ideas/suggestions/"do it this way or it's not going to work"s from the dsp community. Without your help, this wouldn't have been possible for me to even attempt. So I wanted to thank every one of you for that. It's been a long time coming and hopefully I didn't mess things up too bad.


wrenffxi:
I think your editor is set to tabs instead of 4-spaces.

ffxijuggalo:
Found the issue in Notepad++, thanks for pointing that out wren. I'll get them all converted and updated to here just as soon as I can.

ffxijuggalo:
Spacing issue should be corrected now.

PR #5162, opened 2018-08-07
stop error spam during conquest tally on multi-server setups

wrenffxi:


scripts/globals/conquest.lua
```diff
@@ -860,26 +860,32 @@ end
...
-                    npc:setStatus(dsp.status.DISAPPEAR)
-                end
+            -- update the npcs
+            local owner = GetRegionOwner(region)
+            for _, v in pairs(npcs) do
+                local npc = GetNPCByID(base + v.offset)
 
-                if v.nation == dsp.nation.OTHER then
-                    if owner ~= dsp.nation.BEASTMEN then
+                if (npc ~= nil) then
```
takhlaq: no parens
'if npc then` would work here also

wrenffxi: fixed

PR #5163, opened 2018-08-07
-- Added Chocobo Jig II to abilities.sql, added script for ability

EpicTaru:
-- Added Chocobo Jig II to abilities.sql
-- Added script for the ability

### DIFF
--- a/scripts/globals/abilities/chocobo_jig_ii.lua
+++ b/scripts/globals/abilities/chocobo_jig_ii.lua
@@ -1,0 +1,31 @@
+-----------------------------------
+-- Ability: Chocobo jig ii
+-- Increases Movement Speed to party members within the area of effect.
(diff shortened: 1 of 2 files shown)


scripts/globals/abilities/chocobo_jig_ii.lua
```diff
@@ -0,0 +1,31 @@
...
+-- TP Required: 0
+-- Recast Time: 1:00
+-- Duration: 2:00
+-----------------------------------
+ 
+require("scripts/globals/settings");
+require("scripts/globals/status");
+
+-----------------------------------
+-- onAbilityCheck
```
TeoTwawki: please stop adding these function headers that do nothing but repeat the function names, I stripped them all out awhile back and with ppl adding them again we'll have to do it again

PR #5166, opened 2018-08-08
SMN Traits

nfaletra:
Fix Resist Slow levels/values
Implement Blood Boon
Implement Stout Servant

Also added Stout Servant for BST and PUP


kaincenteno:
yay 😸  summoner skills thank you for contribution @nfaletra

PR #5169, opened 2018-08-09
add HP/MP recovery moves for Dyna Statues

WiggoCaptures:
This was put together using data given to us by MrSent... I checked packets from captures and they indeed are the correct Animations for the skills. I basically copied the already existing dice roll from the goblin ability since it does the same thing (well, except for the message id's, those I got from packets)

This will just get the skills in and ready to be used, Someone else is going to have to figure out how to best implement this into the statue's lua. The problem I've been having is getting the mob to perform the skill once it has died. The best I could do was to make it unkillable at 1% until after it did the move but that's wrong. But, except for the mob hanging around at 1% for a few ticks... the ability itself looks and behaves beautifully.


teschnei:
Can you post a retail 0x028 packet of it being used? Want to make sure the
statue is the source of the skill

Wiggo32:
Here's one: https://pastebin.com/K1D1di2J
From this vid: https://youtu.be/g06TADEn9vg?t=1372

sql/mob_skills.sql
```diff
@@ -943,8 +943,8 @@ INSERT INTO `mob_skills` VALUES (1120,69,'10000_needles',1,10.0,2000,1500,4,0,0,
 INSERT INTO `mob_skills` VALUES (1121,770,'eagle_eye_shot',0,25.0,2000,0,4,2,0,0,0,0,0); -- yagudo move
 INSERT INTO `mob_skills` VALUES (1122,771,'eagle_eye_shot',0,25.0,2000,0,4,2,0,0,0,0,0); -- quadav move
 -- INSERT INTO `mob_skills` VALUES (1123,867,'ore_toss',0,7.0,2000,1500,4,0,0,0,0,0,0);
--- INSERT INTO `mob_skills` VALUES (1124,868,'#868',0,7.0,2000,1500,4,0,0,0,0,0,0);
--- INSERT INTO `mob_skills` VALUES (1125,869,'#869',0,7.0,2000,1500,4,0,0,0,0,0,0);
+INSERT INTO `mob_skills` VALUES (1124,772,'regain_hp',1,25.0,2000,0,4,0,0,0,0,0,0); -- Dyna Statues
+INSERT INTO `mob_skills` VALUES (1125,773,'regain_mp',1,25.0,2000,0,4,0,0,0,0,0,0); -- Dyna Statues
```
Hozu: Judging by the Benediction skill, the mob_valid_targets value should be 1 rather than 4. IIRC 4 is for targeting players, so as-is the skills would be used on the players.

Issue #5172, opened 2018-08-09
haste needs a rework


TeoTwawki:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 
n/a

**_Source Branch_** (master/stable) **:** 
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
DB values for haste are inconsistent. people frequently get them wrong, "correct" values that were actually correct, and parts of the core aren't even calculating it the way it was when the values were originally added to the item mods table. **We need one single consistent calculation that is more intuitive** where if you want 3% haste you just put a value of 3 of the mod, and the core does math to get the proper fraction.

```
[3:40 PM] Zynjec:
        // Cap haste at appropriate levels.
        int16 hasteMagic = std::clamp<int16>(getMod(Mod::HASTE_MAGIC), -448, 448);
        int16 hasteAbility = std::clamp<int16>(getMod(Mod::HASTE_ABILITY), -256, 256);
        int16 hasteGear = std::clamp<int16>(getMod(Mod::HASTE_GEAR), -256, 256);
        WeaponDelay = (uint16)(WeaponDelay * ((1024.0f - hasteMagic - hasteAbility - hasteGear) / 1024.0f));
```
```
[3:43 PM] Zynjec:
int16 haste = PEntity->getMod(Mod::HASTE_MAGIC) + PEntity->getMod(Mod::HASTE_GEAR);

recast = (int32)(recast * ((1024.f - haste) / 1024.f));
```
```
[3:51 PM] Zynjec:
select * from item_mods where modid = 384 and value = 31
Affected rows: 0  Found rows: 4  Warnings: 0  Duration for 1 query: 0.016 sec.
```
```
[3:52 PM] Zynjec:
select * from item_mods where modid = 384 and value = 41;
Affected rows: 0  Found rows: 4 
```
```
[3:52 PM] Zynjec:
select * from item_mods where modid = 384 and value = 51;
Affected rows: 0  Found rows: 251 
```
We've been trying to use Nths of 1024th, truncated to the percent in the item helptext which matched the fraction most pages of bg wiki and ffxiclopedia describe.

This sucks.

nfaletra:
[...]

PR #5174, opened 2018-08-11
Haste Rewrite

nfaletra:
Addresses: https://github.com/DarkstarProject/darkstar/issues/5172

Turned haste mods into a 0-10000 scale. The reason for the large scale was to get two decimal places in haste calculations without using floating point numbers. After the exact percentages are calculated, it then scales and truncates the appropriate variable (whether it be weapon delay or recast timers).

I formatted most lua files to be up to date with current style guide-ness. Avoided the formatting the entirety of larger files. Makes it a little hard to see the haste changes. Sorry about that.

Slow calculations were also affected since they are handled as an inverse Haste. That math has also been updated to a 0-10000 scale.

Merit values for Desperate Blows Effect and Haste Samba Effect were also updated.

Did some minor testing via spell casts that apply the correct effects (haste/slow) and equipped gear with haste mods to observe weapon delay being affected.


nfaletra:
Still doing more testing.

src/map/entities/battleentity.cpp
```diff
@@ -256,10 +256,12 @@ int16 CBattleEntity::GetWeaponDelay(bool tp)
     if (!tp)
     {
         // Cap haste at appropriate levels.
-        int16 hasteMagic = std::clamp<int16>(getMod(Mod::HASTE_MAGIC), -448, 448);
-        int16 hasteAbility = std::clamp<int16>(getMod(Mod::HASTE_ABILITY), -256, 256);
-        int16 hasteGear = std::clamp<int16>(getMod(Mod::HASTE_GEAR), -256, 256);
-        WeaponDelay = (uint16)(WeaponDelay * ((1024.0f - hasteMagic - hasteAbility - hasteGear) / 1024.0f));
+        int16 hasteMagic = std::clamp<int16>(getMod(Mod::HASTE_MAGIC), -4375, 4375); // 43.75% cap
+        int16 hasteAbility = std::clamp<int16>(getMod(Mod::HASTE_ABILITY), -2500, 2500); // 25% cap
```
TeoTwawki: https://github.com/DarkstarProject/darkstar/blob/master/src/map/utils/battleutils.cpp#L4253

was looking to do the type of thing we have going on with pdt

TeoTwawki: @teschnei per convo in discord the thing pdt does may in fact be placebo, dumping this on you, I didn't check and I don't want to, I don't have time to find a years old commit and see exactly what happened along the way to present day.

nfaletra:
Went though all the BLU spells that were affected. Slows/Haste values being applied. No script errors.

scripts/globals/abilities/despoil.lua
```diff
@@ -20,68 +20,68 @@ local despoilDebuffs =
 }
 
 function onAbilityCheck(player, target, ability)
-    if (player:getFreeSlotsCount() == 0) then
-        return dsp.msg.basic.FULL_INVENTORY, 0;
+    if not player:getFreeSlotsCount() then
```
teschnei: 0 is true in lua - unless you changed getFreeSlotsCount to return nil instead of 0, this is not gonna work

scripts/globals/abilities/pets/hastega.lua
```diff
@@ -1,26 +1,26 @@
...
     local duration = 180 + bonusTime;
     print(bonusTime)
     print(duration)
 
     -- Garuda's Hastega is a weird exception and uses 153 instead of 15%
     -- That's why it overwrites some things regular haste won't.
-    target:addStatusEffect(dsp.effect.HASTE,153,0,duration);
-    skill:setMsg(dsp.msg.basic.SKILL_GAIN_EFFECT);
-    return dsp.effect.HASTE;
+    target:addStatusEffect(dsp.effect.HASTE, 1500, 0, duration)
```
teschnei: 1530?

scripts/globals/effects/overdrive.lua
```diff
@@ -27,23 +24,15 @@ function onEffectGain(target,effect)
...
------------------------------------
--- onEffectLose Action
------------------------------------
-
-function onEffectLose(target,effect)
+function onEffectLose(target, effect)
     target:delMod(dsp.mod.OVERLOAD_THRESH, 5000)
     local pet = target:getPet()
-    if pet and pet:getLocalVar("overdrive") ~= 0 then
+    if pet and not pet:getLocalVar("overdrive") then
```
teschnei: also can't use `not` instead of 0 here

scripts/globals/effects/overload.lua
```diff
@@ -3,38 +3,27 @@
...
 
------------------------------------
--- onEffectLose Action
------------------------------------
-
-function onEffectLose(target,effect)
+function onEffectLose(target, effect)
     local pet = target:getPet()
-    if pet and pet:getLocalVar("overload") ~= 0 then
+    if pet and not pet:getLocalVar("overload") then
```
teschnei: and here

scripts/globals/effects/weakness.lua
```diff
@@ -3,47 +3,36 @@
...
     --reduce HP and MP by the power amount. Add 100% slow
     --NOTE: The power amount dictates the amount to REDUCE MAX VALUES BY. E.g. Power=75 means 'reduce max hp/mp by 75%'
-    target:addMod(dsp.mod.HPP,-75);
-    target:addMod(dsp.mod.MPP,-75);
+    target:addMod(dsp.mod.HPP, -75)
+    target:addMod(dsp.mod.MPP, -75)
 
-    -- This is wrong, and players should be given a fixed 100% slow in the core so that all haste is ignored, but I am le tired.
-    target:addMod(dsp.mod.HASTE_MAGIC,-1024);
+    -- 100% Slow - FIXME: This kinda doesn't work since magic haste is clamped [-43.75%, 43.75%]
```
teschnei: should be clamped [-100, 43.75] probably

scripts/globals/abilities/pets/hastega.lua
teschnei: remove the comment about 1530, then

scripts/globals/effects/weakness.lua
nfaletra: You're saying that Slow from magic goes to -100%?

scripts/globals/abilities/pets/hastega.lua
nfaletra: Word. I'll do some research to find out if that's actually true, too.

scripts/globals/effects/weakness.lua
teschnei: don't know, it's either that or weakness slow is some special uncapped haste - do we know either way for sure?

scripts/globals/abilities/pets/hastega.lua
nfaletra: Hnnng.

According to BG-wiki. Haste and Hastega are both labeled 15%, but then include 150/1024 and 153/1024 respectively. This is annoying because neither of those values are 15%. They're about 14.6% and 14.9%. The whole point of this was to get accurate values while having human readable percentages in the DB for gear. I'm leaning towards adjusting the sources of magic haste to be this accurate. (So hastega would become 1494). What do you think?

scripts/globals/effects/weakness.lua
nfaletra: no, I don't think we do. But I think it's safe to clamp it further negative just because there aren't enough magic slow sources to push you that far without weakness.

scripts/globals/abilities/pets/hastega.lua
teschnei: fine with me

TeoTwawki:
death to doing math on fractions. long live decimal adjusted percentages

PR #5176, opened 2018-08-11
Added Linkshell Concierge functionality

ZeromusXYZ:


scripts/zones/Bastok_Markets/npcs/Linkshell_Concierge.lua
```diff
@@ -0,0 +1,58 @@
...
+require("scripts/globals/concierge");
+require("scripts/globals/settings");
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end;
+
+function onTrigger(player,npc)
+
+    if (npc:getID() == 17739958) then -- the one on the right
```
TeoTwawki: instead of using NPC ID's ins cript link this, its preferred to either:
A) rename the npc's to tell them apart in the sql and then use a script for each. This method future proofs the script against client updates changing the ID and we just have to update the sql.
-OR-
B) put the IDs in the zones mobIDs.lua require that in teh npc script and call teh variable here instead of the ID. That keeps all the IDs in one file so we can more easily verify/update them when they shift (they move nearly every client update, its very annoying). This method requires us to keep on top of ID updates in both script *and* the sql but avoids extra scripts.

ZeromusXYZ: Which one would you prefer / is most commonly used in DSP ?
Editing the sql some more, and adding files, or adding to mobids ?

TeoTwawki: I think its worth doing (A) but will merge either one

scripts/zones/Eastern_Adoulin/npcs/Linkshell_Concierge_B.lua
```diff
@@ -0,0 +1,50 @@
...
+end;
+
+function onTrigger(player,npc)
+    LinkShellConciergeEventTrigger(player,npc,1,dsp.lsconciergetype.EVERYTHING);
+end;
+
+function onEventUpdate(player,csid,option)
+    local lsCmd = option % 0x00000008 ;
+    local res = LinkShellConciergeEventUpdate(player,csid,option);
+    if ((lsCmd == 5) and (res == 98)) then -- dummy value for error
```
TeoTwawki: don't need the extra parenthesis aroudn conditions
```lua
if lsCmd == 5 and res == 98 then
```
will work

scripts/zones/Eastern_Adoulin/npcs/Linkshell_Concierge_B.lua
```diff
@@ -0,0 +1,50 @@
...
+        player:messageSpecial(ITEM_CANNOT_BE_OBTAINED,0x203)
+        player:release(); -- probably not the best way to do this
+    end
+end;
+
+function onEventFinish(player,csid,option)
+
+    local res = LinkShellConciergeEventFinish(player,csid,option);
+
+    if (res == 0) then
```
TeoTwawki: same for the rest of the conditionals

scripts/zones/GM_Home/npcs/Linkshell_Concierge.lua
```diff
@@ -0,0 +1,55 @@
...
+
+function onEventUpdate(player,csid,option)
+    LinkShellConciergeEventUpdate(player,csid,option)
+end
+
+function onEventUpdate(player,csid,option)
+    local lsCmd = option % 0x00000008 
+    local res = LinkShellConciergeEventUpdate(player,csid,option)
+    if lsCmd == 5 and res == 98 then -- dummy value for error
+        player:messageSpecial(ITEM_CANNOT_BE_OBTAINED,0x203)
```
TeoTwawki: add in decimal form to the zones textIDs.lua instead of leaving the raw message ID in file like this. these shift in most client updates and it'll break until we run the textIDs update batch.

TeoTwawki: derp I dunno why I read that backwards ignore me

src/map/packets/linkshell_concierge.cpp
```diff
@@ -0,0 +1,227 @@
...
+		ret = Sql_Query(SqlHandle, qStr.c_str());
+
+		if (ret != SQL_ERROR && Sql_NumRows(SqlHandle) != 0 && Sql_NextRow(SqlHandle) == SQL_SUCCESS)
+        {
+			// Valid results
+			int i = 0;
+			while ((ret == SQL_SUCCESS) && (i < 4))
+			{
+                
+				uint16 lsID = Sql_GetUIntData(SqlHandle, 2);
```
TeoTwawki: tabs to spaces please. 4 spaces per indent level.

win32/vcxproj/DSGame-server.vcxproj
```diff
@@ -23,6 +23,7 @@
     <ProjectGuid>{89A19910-84DB-454D-9C7F-0A47A293867B}</ProjectGuid>
     <RootNamespace>DSGameserver</RootNamespace>
     <Keyword>Win32Proj</Keyword>
+    <WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>
```
TeoTwawki: this is uneeded and was probably auto generated by having vs itself modify file while being used on win10 machine so it applied the default for your setup.

TeoTwawki: when omitted the build just uses whatever you are "on" anyway.

scripts/zones/GM_Home/npcs/Linkshell_Concierge.lua
```diff
@@ -0,0 +1,55 @@
...
+        player:release() -- probably not the best way to do this
```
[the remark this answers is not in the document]
ZeromusXYZ: Here's the thing with this one. Client sends 3 packets to the server, where only the first seems to contain related data when you "buy" a pearl. I doesn't seem to wait for any kind of response, and will just display the normal text for when you buy a pearl, even if I were to not send any data back.
Guess I'm going to have to go and grab a capture to see what it does when your inventory is full.
That's why I just triggered a release right after it sends the first buy data, and the server says, no more room.

src/map/packets/linkshell_concierge.cpp
TeoTwawki: vs2017 has editorconfig support so you can automate it without fiddling with the vs settings https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328

src/map/packets/linkshell_concierge.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+#include "linkshell_concierge.h"
+
+#include "../entities/charentity.h"
+#include "../linkshell.h"
+#include "../../common/utils.h"
+#include "../../map/utils/zoneutils.h"
+
+
+CLinkshellConciergePacket::CLinkshellConciergePacket(CCharEntity* PChar, uint32 conciergeID, uint32 csid, uint32 option, uint8 conciergeShareSettings)
```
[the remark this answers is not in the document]
ZeromusXYZ: So ... basically try and put as much as I can in the lua scripts and lua_baseentity.cpp then ?

teschnei:
For the db calls and packet assembly, the linkshell namespace in
src/map/linkshell is probably good

PR #5178, opened 2018-08-12
Sugaar NM Pashhow Marshlands

n0pax:
Attached please find all the code necessary to make Sugaar, the peiste NM in Pashhow Marshlands [S] fully functioning.  Credit to KnowOne of ERA for assisting with everything and teaching me how to do it.  Additional credit to the KV code which was used as a template to create the on hit silence.


sql/mob_pools.sql
```diff
@@ -4737,7 +4737,7 @@ INSERT INTO `mob_pools` VALUES (4676,'Eald_narche_2','Eald_narche_2',394,0x00000
 INSERT INTO `mob_pools` VALUES (4677,'Achuka','Achuka',461,0x0000D30900000000000000000000000000000000,1,1,7,240,100,0,0,0,1,0,0,0,1,131,0,0,0,1,0,461);
 INSERT INTO `mob_pools` VALUES (4678,'Hurkan','Hurkan',460,0x0000140A00000000000000000000000000000000,1,1,7,240,100,0,0,0,1,0,0,0,1,129,0,0,0,1,0,460);
 INSERT INTO `mob_pools` VALUES (4679,'Ocythoe','Ocythoe',471,0x0000130800000000000000000000000000000000,4,4,6,240,100,0,1,0,0,2,0,0,7,1695,0,0,2,1,0,471);
-INSERT INTO `mob_pools` VALUES (4680,'Sugaar','Sugaar',192,0x0000DF0700000000000000000000000000000000,1,1,7,240,100,0,1,0,0,0,7,0,7,155,0,0,0,1,0,192);
+INSERT INTO `mob_pools` VALUES (4680,'Sugaar','Sugaar',192,0x0000DF0700000000000000000000000000000000,1,1,7,240,100,0,1,0,0,2,7,0,7,155,0,0,0,1,0,192);
```
TeoTwawki: I think this NM also needs a non generic skill list, looks like it only uses [Grim Glower](http://ffxiclopedia.wikia.com/wiki/Grim_Glower) (I don't think we have that working yet) and [Torpefying Charge.](http://ffxiclopedia.wikia.com/wiki/Torpefying_Charge)

If I'm wrong there let me know - I've never met this NM on retail, so I only have [sh\*\*WikiSays](http://ffxiclopedia.wikia.com/wiki/Sugaar) to go on

sql/mob_groups.sql
```diff
@@ -11608,7 +11608,7 @@ INSERT INTO `mob_groups` VALUES (13721,4736,81,0,0,2535,0,5000,90,92,0);
 INSERT INTO `mob_groups` VALUES (13722,4883,89,7200,0,0,8000,5000,70,72,0);
 INSERT INTO `mob_groups` VALUES (13723,4869,89,7200,0,4519,13000,0,80,80,0);
 INSERT INTO `mob_groups` VALUES (13724,4679,89,0,128,0,0,9999,90,95,0);
-INSERT INTO `mob_groups` VALUES (13725,4680,90,7200,128,4520,12000,9999,80,80,0);
+INSERT INTO `mob_groups` VALUES (13725,4680,90,7200,32,4520,12000,9999,80,80,0);
```
TeoTwawki: if its placeholdered then respawn time here should be zero

scripts/zones/Pashhow_Marshlands_[S]/mobs/Sugaar.lua
```diff
@@ -0,0 +1,36 @@
...
+-- Area: Pashhow Marshlands [S] (90)
+--  NM:  Sugaar
+-- !pos -416.496,24.11,-441.589,132
+-----------------------------------
+require("scripts/globals/status");
+require("scripts/globals/magic");
+require("scripts/globals/msg");
+
+function onMobSpawn(mob)
+    mob:setMod(dsp.mod.STORETP, 20);
```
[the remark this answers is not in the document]
Comment (n0pax) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Sugaar.lua]: No source, I can remove it np if that is preferred.  Sorry I should have put this in the notes but the NM seems to be a larger threat and has minimal online information so I figured it would be accurate to give it some to increase its strength.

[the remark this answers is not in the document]
Comment (TeoTwawki) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Sugaar.lua]: if its spammy its more likely to be regain than storetp. we'd need to test in retail to determine if it has it (if it still gains tp without being fed any, its got regain for sure).

sql/mob_pools.sql
n0pax: I never encountered it on retail either so I am unsure, but my thought process here was the person on the wiki was explicitly noting those because they are the "big threats" it has and it has normal moves too, but they just didn't note them.

scripts/zones/Pashhow_Marshlands_[S]/mobs/Sugaar.lua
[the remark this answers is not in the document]
Comment (TeoTwawki) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Sugaar.lua]: for now remove till we know if it has it.

sql/mob_groups.sql
n0pax: This is my first time ever doing any of this so I am very raw.  The person helping me create the mob said changing this from 128 to 32 signifies it's a lottery pop instead of a scripted pop.  Should the 32 be changed to a 0?

sql/mob_pools.sql
TeoTwawki: usually they'll says more like "uses standard peiste moves" if it has the full moveset and lists specific ones if those are all it has. most pages will just list them all even if that amounts to all the moves that mob family has anyway. NMs often have fewer movebs but the ones they have often hit harder.

We can get someone to check retail if you use discord https://discord.gg/sX9JgDT

sql/mob_groups.sql
TeoTwawki: no thats the spawn *type*, I'm talking the spawn *time*.

scripts/zones/Pashhow_Marshlands_[S]/mobs/Sugaar.lua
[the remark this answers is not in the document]
Comment (n0pax) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Sugaar.lua]: Ok that change should be done assuming I pushed it properly.

sql/mob_groups.sql
n0pax: Thanks, that change should also be live now assuming I pushed it properly.

sql/mob_pools.sql
n0pax: Thanks, request posted on discord and once we get that information I'll do my best to adjust accordingly.

TeoTwawki:
one mor ething ( really need to go to bed haha)

strip out the semicolons (they do nothing in lua) and extra ()'s

we haven't fully cleane dup whats in trunk yet but we're moving to eliminate them.

so
```lua
function stuff(arguments)
    local thing = whahtever
    if thing then
        doIt
    end
end
```
instead of
```lua
function stuff(arguments)
    local thing = whahtever;
    if (thing) then
        doIt
    end
end;
```
just as an example

sql/mob_pools.sql
KnowOne134: From the wiki page here it looks to use all tp moves but Oppresive Gaze. 
The tp move Grim Glower is a sunanimation gaze attack that while in effect if you are face to face you become petrified the entire time to sub is active. I dont believe this can be coded yet. 
http://ffxiclopedia.wikia.com/wiki/Category:Peistes

TeoTwawki: thats the family page, it doesn't say anything specific to this one NM other than its name where it lives and what it drops.

Wiggo32: this should be able to be coded... why not? we have scripts that can check to see if you're facing mob for other "gaze type" moves... it's not like the petrification just "comes off" once it stops looking at you... you get petrified for a set time ~15sec and then once it wears you get it back if you're still looking at it's glowing eyes... should be doable

Wiggo32: prolly onmobfight i'd guess
if mob has subanimation GloweyEyesThingy and mob isfacing target and mob target isfacing mob then give petrification... is where i would start

TeoTwawki: No not onMobFight we'd do the same thing we need to support aura's except check both entities are facing and within a cone. So a large part of this happens in core which is why ppl haven't done it yet.

TeoTwawki: Its honestly not real hard but ppl who know how haven't had both time and interest at the same time yet

Wiggo32: https://youtu.be/dfCXNfkOObo
I did this with dsp master just the way it is... no core changes... only lua scripting in onMobFight

scripts/zones/Pashhow_Marshlands_[S]/mobs/Virulent_Peiste.lua
```diff
@@ -0,0 +1,14 @@
...
+-- Note: PH for Sugaar
+-----------------------------------
+require("scripts/zones/Pashhow_Marshlands_[S]/MobIDs.lua");
+require("scripts/globals/mobs");
+
+function onMobDeath(mob, player, isKiller)
+end;
+
+function onMobDespawn(mob)
+    phOnDespawn(mob,Sugaar_PH,15,28800); -- 15% spawn chance and 8 hour respawn are guesses as no information can be found online regarding it
```
Comment (Wiggo32) [inline scripts/zones/Pashhow_Marshlands_[S]/mobs/Virulent_Peiste.lua]: 15% seems kinda high based on how long it took me to pop it on retail (during the course of 2 days)
and remember, with all those PH's in the MobID's that's acutally more than 15% cause its 15% for every PH when on retail you only have 1 PH (which I also found the correct one in the vid)
https://youtu.be/1zB-nDvt0tA

zynjec:
Going to close this for now, feel free PR again once it's rebased and proper abilities list.

PR #5181, opened 2018-08-13
changed physical to magical

KnowOne134:


takhlaq:
can you rebase with discoid in the commit title since `changed physical to magical` doesn't specify what exactly was changed

TeoTwawki:
All ya hadda do was git amend

PR #5184, opened 2018-08-13
Ankabut

Caladius:
This is the PR to implement Ankabut in North Gustaburg [S}.

Information pulled from:
http://ffxiclopedia.wikia.com/wiki/Ankabut


sql/mob_groups.sql
```diff
@@ -12542,6 +12542,8 @@ INSERT INTO `mob_groups` VALUES (14631,830,22,1056,0,591,0,0,58,60,0);  -- Promy
 INSERT INTO `mob_groups` VALUES (14632,5842,122,7200,0,0,0,0,77,79,0); -- Nargun
 INSERT INTO `mob_groups` VALUES (14633,5843,61,0,32,4599,0,0,80,82,0);  -- Ignamoth
 
+INSERT INTO `mob_groups` VALUES (14634,2002,88,3600,128,4600,5000,0,50,53,0) -- Ankabut
```
nfaletra: First, need a semicolon.

Second, can't use the same mob_pool as Huge Spider. They are not flagged as NMs and I'm sure the aggro isn't set up correctly either. You need a new mob_pool for Ankabut.

scripts/zones/North_Gustaberg_[S]/mobs/Huge_Spider.lua
```diff
@@ -0,0 +1,15 @@
+-----------------------------------
+-- Area: North Gustaberg [S]
+--  MOB: Huge Spider
+-- Note: Place holder for Ankabut
+-----------------------------------
+require("scripts/zones/North_Gustaberg_[S]/MobIDs")
+require("scripts/globals/mobs")
+
+function onMobDeath(mob, player, isKiller)
+
```
Comment (TeoTwawki) [inline scripts/zones/North_Gustaberg_[S]/mobs/Huge_Spider.lua]: excess newline. remove.
```lua
end
```

sql/mob_spawn_points.sql
```diff
@@ -70634,5 +70634,3 @@ INSERT INTO `mob_spawn_points` VALUES (17977406,'Darkness','Darkness','12987','0
 -- ------------------------------------------------------------
 -- Dynamis-Jeuno [D] (Zone 297)
 -- ------------------------------------------------------------
-
```
TeoTwawki: didn't need to trim that, its unrelated to the NM you are adding

sql/mob_groups.sql
TeoTwawki: this, but also make sure not to add a respawn time since its a lottery pop and will be handled by the script end.

TeoTwawki: hp and mp is also normally calculated (the core will do it base don its job and mob family for you) and not hard set unless we have retail data telling its it also has this amount

sql/mob_spawn_points.sql
Caladius: i actually had never touched that file, unsure why it said i made a change to it (among others)

TeoTwawki:
you accidentally committed your `.DS_Store` file. you can add that to the git ignore list if you wish to prevent it from slipping in again in the future.

Caladius:
Oops, thanks Teo!

I'll get that added going forward =(

TeoTwawki:
its probably ok to commit the change to the projects ignore list actually.
https://github.com/DarkstarProject/darkstar/blob/master/.gitignore

Caladius:
Should i just add a new line for it at the end of the file?

sql/mob_pools.sql
```diff
@@ -5860,6 +5860,7 @@ INSERT INTO `mob_pools` VALUES (5840,'Champion_Crawler','Champion_Crawler',79,0x
 INSERT INTO `mob_pools` VALUES (5841,'Hovering_Hotpot','Hovering_Hotpot',175,0x00009C0100000000000000000000000000000000,5,1,7,240,100,0,1,0,0,2,0,0,0,0,0,0,36,1,0,998);
 INSERT INTO `mob_pools` VALUES (5842,'Nargun','Nargun',135,0x0000B00100000000000000000000000000000000,1,1,11,320,100,0,1,0,0,2,0,32,0,0,0,0,0,1,0,1005);
 INSERT INTO `mob_pools` VALUES (5843,'Ignamoth','Ignamoth',253,0x0000100700000000000000000000000000000000,1,1,7,240,100,0,1,1,1,2,39,0,311,131,0,0,0,1,0,1006);
+INSERT INTO `mob_pools` VALUES (5844,'Ankabut','Ankabut',235,0x0000360100000000000000000000000000000000,1,1,7,240,100,1,0,0,1,0,0,0,1,131,0,0,0,1,0,235);
```
nfaletra: mobType needs to be 2 for Notorious Monster.
Does not link, according to both bg-wiki and ffxiclopedia.
Both wikis also mention that it only uses Acid Spray so this needs to use a different skill list that only includes Acid Spray.

TeoTwawki:
> Should i just add a new line for it at the end of the file?

if you are talking about the ignor elist entry, `User-specific files` seems like the appropriate section

PR #5185, opened 2018-08-14
Chakra rework to use mods

nfaletra:
- Gutted hardcoded item id checks in Chakra ability script
- Added mod for VIT mult
- Added mod for status effect removal (mask based)
- Did not update cooldown or merits to current retail values (3 min cooldown and 6 seconds per merit I think?)
- Math not updated to take into account the "function of the player's level" (This was updated in August 2017. There is literally no info about how the level affects recovered HP)


nfaletra:
Rebased CAUSE TEO!

scripts/globals/abilities/chakra.lua
```diff
@@ -8,48 +8,38 @@
...
-    if (player:hasStatusEffect(dsp.effect.BLINDNESS)) then
-        player:delStatusEffect(dsp.effect.BLINDNESS)
-    end
-
-    if ((body == 12639) or (body == 14474)) then -- Temple Cyclas (+1) equipped
-        if (player:hasStatusEffect(dsp.effect.PARALYSIS)) then
-            player:delStatusEffect(dsp.effect.PARALYSIS)
+function onUseAbility(player, target, ability)
+    local chakraRemoval = player:getMod(dsp.mod.CHAKRA_REMOVAL)
+    if chakraRemoval then
```
takhlaq: ~= 0, probably

scripts/globals/abilities/chakra.lua
```diff
@@ -8,48 +8,38 @@
...
-        multi = multi + 0.6
-    end
-
-    local recover = (multi * vit)
-    player:setHP((hp + recover))
+    local recover = player:getStat(dsp.mod.VIT) * (2 + player:getMod(dsp.mod.CHAKRA_MULT) / 10) -- TODO: Figure out "function of level" addition (August 2017 update)
+    player:setHP(player:getHP() + recover)
 
+    local merits = player:getMerit(dsp.merit.INVIGORATE)
     if (merits >= 1) then
```
takhlaq: can get rid of these parentheses too

scripts/globals/abilities/chakra.lua
nfaletra: Still want it to remove Poison and Blindness.

nfaletra: Probably doesn't need to be checked at all, actually.

nfaletra: Actually we want it to run if it's 0 for Blindness and Poison removal. So check can probably go away.

PR #5192, opened 2018-08-16
Augments for haste rewrite

nfaletra:
Correct me if I'm wrong, but these were super busted before, right?


TeoTwawki:
you totally right. super busted.

PR #5193, opened 2018-08-16
Fix msg for Familiar

WiggoCaptures:
This was displaying: The <mob> uses Acid Mist. 
Now it displays: The <mob> uses Familiar.


TeoTwawki:
We've a mistake in our documentation at the same line ([108](https://github.com/DarkstarProject/darkstar/blob/master/documentation/message.log#L108)) too:
```
108 - The <player> uses .. Pet's powers increase!
109 - The <player> uses .. Pet's powers increase!
```
it got duplicated, which probably lead to this mistake in scripts before we moved it to global.

zynjec:
Unless I'm not understanding the issue correctly, I'm unable to reproduce on master.

Using Familiar yields me:
image

zynjec:
Adding to that, it seems both of those messages in documentation are accurate.

image

TeoTwawki:
huh. ok then.

zynjec:
For whatever reason, one is an ability parameter, and another is weapon skill parameter, despite never being passed a WS.

image

TeoTwawki:
wierd

WiggoCaptures:
that fix was for mob's using familiar... go to dynamis bastok and get your beasttender to pop 2hr... read what it says

TeoTwawki:
oh hey, mobs have same index as ws for mobskills. so thats why there are 2

TeoTwawki:
so we're going to want both in the global.

WiggoCaptures:
but... i suppose the messages are used for both mob and player... so maybe there's a different way it should be handled?

zynjec:
A bit more detail would have helped then.

Either way, this fix isn't appropriate, as it'd break the player message.

TeoTwawki:
FAMILIAR_MOB
FAMILIAR_PC

or something

zynjec:
I did forget that mob abilities were in the weapon skill index!

So we just need something like:

```lua
FAMILIAR_PC = 108,
FAMILIAR_NPC = 109,
```

and then update `scripts/globals/mobskills/familiar.lua` to use `dsp.msg.basic.FAMILIAR_NPC`

TeoTwawki:
I'd say `_MOB` not `_NPC`, but yeah

WiggoCaptures:
I'd also have to update wherever it calls FAMILIAR_PC too right?

TeoTwawki:
yes

zynjec:
```$ git grep dsp.msg.basic.FAMILIAR
scripts/globals/abilities/familiar.lua:28:    ability:setMsg(dsp.msg.basic.FAMILIAR)
scripts/globals/mobskills/familiar.lua:24:    skill:setMsg(dsp.msg.basic.FAMILIAR```

PR #5196, opened 2018-08-17
increase size of readable packets

WiggoCaptures:
from @ZeromusXYZ#0849


ZeromusXYZ:
You know I didn't actually test that properly right? But if it works, it works ...

Wiggo32:
ahhh, ok... if someone wants to test this then, it works fine for me. or give me a specific test to do and I'll do it and post results...

WiggoCaptures:
closing this as I am not completely sure it is finished... plus I think the other one i'm submitting that modifies lua_baseentity is more important.

PR #5204, opened 2018-08-19
fix for messageSpecial packet

Omnione:
this is to fix the messageSpecial packet being able to take an additional entity used in the message if not the base entity sending the message,

from my testing this does not seem to intefere with exsisting uses of messageSpecial or showText, but it meets the needs of special cases when the entity needs to be set.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -180,24 +180,36 @@ inline int32 CLuaBaseEntity::showText(lua_State *L)
 
     CLuaBaseEntity* PLuaBaseEntity = Lunar<CLuaBaseEntity>::check(L, 1);
 
+    bool rotateNPC = true;
+
+    if (!lua_isnil(L, 9) && lua_isboolean(L, 9))
+        rotateNPC = lua_toboolean(L, 9);
+
     if (PLuaBaseEntity != nullptr)
     {
         CBaseEntity* PBaseEntity = PLuaBaseEntity->GetBaseEntity();
         if (PBaseEntity->objtype == TYPE_NPC)
         {
             PBaseEntity->m_TargID = m_PBaseEntity->targid;
-            PBaseEntity->loc.p.rotation = getangle(PBaseEntity->loc.p, m_PBaseEntity->loc.p);
+            if (rotateNPC)
+            {
+                PBaseEntity->loc.p.rotation = getangle(PBaseEntity->loc.p, m_PBaseEntity->loc.p);
 
-            PBaseEntity->loc.zone->PushPacket(
-                PBaseEntity,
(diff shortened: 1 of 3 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -180,25 +180,37 @@ inline int32 CLuaBaseEntity::showText(lua_State *L)
...
     if (PLuaBaseEntity != nullptr)
     {
         CBaseEntity* PBaseEntity = PLuaBaseEntity->GetBaseEntity();
         if (PBaseEntity->objtype == TYPE_NPC)
         {
             PBaseEntity->m_TargID = m_PBaseEntity->targid;
-            PBaseEntity->loc.p.rotation = getangle(PBaseEntity->loc.p, m_PBaseEntity->loc.p);
+            if (rotateNPC)
+            {
+                PBaseEntity->loc.p.rotation = getangle(PBaseEntity->loc.p, m_PBaseEntity->loc.p);
```
TeoTwawki: we have another function to do this already. It doesn't need to be in messageSpecial. use `lookAt()`

Omnione: that was allready it there i just added a check to bypass it as some instances you dont want the npc rotating, as the boxes do when a message is displayed

TeoTwawki: /facepalms sorry mybad. sometimes this project gives me a migraine.

Omnione: i feel the pain lol

TeoTwawki: actually this makes them turn their head but not their body. I guess boxes are "just a head". ShowText() annoyingly turns the body in its code.

teschnei:
are there any use cases of message_special showing a name that isn't the entity doing the talking?

zynjec:
I'd have to go back and look, but I'm pretty sure Abyssea Pyrix do that for loot distribution.

The actor is the chest, but the ascii name is the person receiving loot.

TeoTwawki:
lets verify ita actually using a 0x2A packet for that

zynjec:
Looking through my stuff, pretty much everything in Abyssea use 0x2A.

- Time remaining on chest (actor pyxis, acsii name person receiving message)
- Message announcing someone transferred all the items to treasure pool (actor pyxis, ascii name character that opened it).
- Someone taking item from pyxis (actor pyxis, ascii name of person looting)

TeoTwawki:
k, thanks for looking

whasf:
What are we doing with this??

zynjec:
Going to close this for now, if you're still around Omni we can work on this packet some.

PR #5208, opened 2018-08-22
Correct some things with outpost warps to TULIA and TAVNAZIANARCH

TeoTwawki:
~~The selection exists in retail but the player will never see it or use it since they cannot unlock it.~~

**Todo: rework this into a setting to enable warp rather than unlocking them at char creation.**

~~easy review link: https://github.com/DarkstarProject/darkstar/commit/183bd6df6d9ab3760db0990f9508edc7d2d6fed0?w=1~~

~~it's a 2 line edit.~~


TeoTwawki:
if anyone gets board and does my Todo before I do, I intend to do:
`=1` to enable all retail attainable outpost warps minus tavnazia (because CoP)
`=2` to enable all retail including Tavnazia
and `=3` to enable all even if unattainable on retail. if its in the NPC's event in the client, it works.

wrenffxi:
Reminder that this PR -- as currently implemented -- is included in open PR #5157.  Although I don't have a level or gil/cp requirement for the Tulia warp.

TeoTwawki:
If yours merges first I'll rebase this onto it to add the fee (I used the fee we had before the change to using tables).

I didn't realize wren had this in the other pr when I opened, I kept think they fixed it in their conquest pr and then was confused when I couldn't find it in master.

TeoTwawki:
Rebased

**Easy review link:**
https://github.com/DarkstarProject/darkstar/pull/5208/files?w=1

PR #5218, opened 2018-09-03
Re-Insert players to a BCNM in case of D'C

WiggoCaptures:
This was the best method that I was able to find for re-applying the Level Cap Sync after a Disconnect from a BCNM. Every BCNM capable zone.lua would have to have this 'afterZoneIn' code and If this is accepted, I will do the rest of them.


WiggoCaptures:
Demo already incorporated this and we need to find a more efficient way to handle re-sync... so closing

PR #5225, opened 2018-09-12
Changing verlock behavior to allow new greater than or equal behavior

ShelbyZ:
- Rework verlock to an integer
- Add new state allowing greater than or equal matching
- Update console input to allow toggling between states
- Update handling of new verlock

### DIFF
--- a/src/login/login.cpp
+++ b/src/login/login.cpp
@@ -457,6 +457,12 @@ int32 version_info_read(const char *fileName)
         else if (strcmp(w1, "VER_LOCK") == 0)
         {
             version_info.ver_lock = std::atoi(w2);
+
+            if (version_info.ver_lock > 2 || version_info.ver_lock < 0)
+            {
+                ShowError("ver_lock not within bounds (0..2) was %i, defaulting to 1\r\n", version_info.ver_lock);
+                version_info.ver_lock = 1;
+            }
         }
     }
     fclose(fp);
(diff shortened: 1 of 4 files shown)


ShelbyZ:
@KnowOne134 - this might be close to inline with what you were suggesting

TeoTwawki:
This 3rd option was the original behavior when I created it, which @teschnei killed because people kept reporting "bugs" from their version being old.

TeoTwawki:
Also I assume changing bools into numbers and then using the numbers as a bool isn't wanted, since that's what this was and it got changed from that..We do that all over the core I don't know why \\\_(o.O)\_/

KnowOne134:
Yes thats what i was exactly looking for. We do keep the server updated amd version locked but there is a few week transition before I find the time to do the updates ussually. So something that lets me lock out the lower ver but let in what it was and let in the new players also would be very helpfull. And give my current players time when i announve uts time to update to newest version

src/login/login.cpp
```diff
@@ -131,8 +131,23 @@ int32 do_init(int32 argc, char** argv)
...
+                        switch (version_info.ver_lock)
+                        {
+                            case 0:
+                                value = "disabled";
+                                break;
+                            case 1:
+                                value = "enabled - strict";
+                                break;
+                            case 2:
+                                value = "enabled - greather than or equal";
```
takhlaq: that's pretty greath

PR #5230, opened 2018-09-15
Add item

teschnei:


src/map/items/item_armor.cpp
```diff
@@ -255,6 +255,21 @@ void CItemArmor::LoadAugment(uint8 slot, uint16 augment)
     ref<uint16>(m_extra, 2 + (slot * 2)) = augment;
 }
 
+void CItemArmor::PushAugment(uint16 type, uint8 value)
```
takhlaq: `bool PushAugment` in case there's no free slot maybe? idk

src/map/items/item_armor.cpp
```diff
@@ -255,6 +255,21 @@ void CItemArmor::LoadAugment(uint8 slot, uint16 augment)
...
+{
+    uint8 slot = 0;
+    uint16 augment = ref<uint16>(m_extra, 2 + (slot * 2));
+    while (augment != 0 && slot < 4)
```
takhlaq: i guess we should enum max augment slot somewhere in case it changed in the future

src/map/lua/lua_baseentity.cpp
```diff
@@ -3070,90 +3070,166 @@ inline int32 CLuaBaseEntity::addItem(lua_State *L)
...
+                {
+                    lua_getfield(L, 1, "trial");
+                    uint16 trial = (uint16)lua_tointeger(L, -1);
+                    if (trial != 0) ((CItemArmor*)PItem)->setTrialNumber(trial);
+                    lua_getfield(L, 1, "augments");
+                    if (lua_istable(L,-1))
+                    {
+                        auto table = lua_gettop(L);
+                        lua_pushnil(L);
+                        while(lua_next(L,table) != 0)
```
takhlaq: `while (`
space ur face

PR #5231, opened 2018-09-16
Fix small copy paste mistake in a Print Binding

TeoTwawki:


src/map/lua/lua_baseentity.cpp
```diff
@@ -310,7 +310,7 @@ inline int32 CLuaBaseEntity::PrintToArea(lua_State* L)
 
     if (messageRange == 0) // All zones world wide
     {
-        message::send(MSG_CHAT_SERVMES, 0, 0, new CChatMessagePacket(PChar, MESSAGE_SYSTEM_1, (char*)lua_tostring(L, 1), name));
+        message::send(MSG_CHAT_SERVMES, 0, 0, new CChatMessagePacket(PChar, messageLook, (char*)lua_tostring(L, 1), name));
```
nfaletra: Can we change the char* cast to an actual std::string ctor? Not sure why it needs to strip the const from the lua_tostring call in the first place. Perhaps I'm missing something.

TeoTwawki: no idea I just copied that from where it was already done

zynjec: OOF, i'll PR a cleanup for both later.

PR #5232, opened 2018-09-17
Added support for monster drop groups

jmcmorris:
Added support for monster drop groups and updated the monster scripts and droplist entries that could obviously benefit from them. This is achieved by using the unused "GroupId" and GroupRate database entries.

DropList is now more than just a list of items. It is also a list of item groups. Each group has a GroupRate which is the likelihood for that group to drop an item (much like DropRate for single items). Treasure Hunter works on the GroupRate and not the individual items inside of a group. When an item is in a group its DropRate is used as a weight for how likely it is to be chosen and the sum of the DropRates in a group should be 1000.

This has been lightly tested and stepped through with Mee Deggi, normal mobs, along with Steal and Despoil. This is my first contribution here so please let me know what I can do better. Thanks!


src/map/lua/luautils.cpp
```diff
@@ -3924,11 +3924,11 @@ namespace luautils
     {
         DropList_t* DropList = itemutils::GetDropList((uint16)lua_tointeger(L, 1));
```
nfaletra: Can we add a pointer guard here? It currently crashes if called with an invalid Drop ID. Not your fault, just something to clean up while you're in here. :)

src/map/utils/itemutils.cpp
```diff
@@ -472,15 +482,27 @@ namespace itemutils
...
+                uint16 ItemID = (uint16)Sql_GetIntData(SqlHandle, 1);
+                uint8 DropType = (uint8)Sql_GetIntData(SqlHandle, 2);
+                uint16 DropRate = (uint16)Sql_GetIntData(SqlHandle, 3);
 
-                g_pDropList[DropID]->push_back(DropItem);
+                if (DropType == DROP_GROUPED)
+                {
+                    uint8 GroupId = (uint8)Sql_GetIntData(SqlHandle, 4);
+                    uint16 GroupRate = (uint16)Sql_GetIntData(SqlHandle, 5);
+                    //This requires the database to have the groups in order
```
teschnei: should this silently fail if it isn't?

src/map/lua/luautils.cpp
jmcmorris: Sure, I can do that.

TeoTwawki:
NotFinishingWhatHeStartedInATimelyManner Teo thanks you in the 3rd person for this contribution :smiley:

src/map/utils/itemutils.cpp
jmcmorris: Yeah, I was originally going to do that then I was wondering if there is any sort of validation on the database elsewhere. I think the best solution would be to have some test scripts that validate the database.

Assuming that you always use all of the groupids from 0 then it could emplace_back groups until it gets to the correct groupid with a while loop. I meant to make this change but couldn't remember it before...

src/map/lua/luautils.cpp
jmcmorris: Fixed this in my second commit

takhlaq:
thanks!

PR #5235, opened 2018-09-18
Limit number of items a mob can drop to the treasure pool size

jmcmorris:
Monsters were able to drop more than 10 items which could result in the treasure pool being filled instantly and some items being randomly dropped to players before they had a chance to lot. With the new drop groups this should not be as much of a problem once the groups are adjusted in mob_droplist.sql. This will ignore any drops after 10 which means that perhaps we want to have the better drops to be listed first so they have the highest priority. I do not know if we want to ever allow mobs to drop over 10 items or not but if we dont then...

I wrote two makeshift unit tests and one of them would be useful for tracking down the potential offenders by checking for the maximum number of possible items - [commit with unit tests](https://github.com/jmcmorris/darkstar/commit/b3ea3863bb27240d3e4c4bf9e3aae30a11051d6c). Spoiler! There are a lot of offenders right now - 752 to be precise.

The majority of this change is honestly refactoring the code to make it easier to stop dropping items. This was done by moving all of the other mob death related logic to `CMobEntity::HandleDeath()` and then have that call `CMobEntity::DropItems()`. I was not sure if the macro I use fits with the styling of the source and if it does not then I can change it.

**Tested lightly** to ensure drops were working but did not try it with a mob that drops 10+ items.

### DIFF
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -784,7 +784,7 @@ void CMobEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
     }
 }
 
-void CMobEntity::DropItems()
+void CMobEntity::HandleDeath()
 {
     CCharEntity* PChar = (CCharEntity*)GetEntity(m_OwnerID.targid, TYPE_PC);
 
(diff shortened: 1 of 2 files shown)


src/map/entities/mobentity.cpp
```diff
@@ -803,150 +803,160 @@ void CMobEntity::DropItems()
...
+    else
+    {
+        luautils::OnMobDeath(this, nullptr);
+    }
+}
+
+void CMobEntity::DropItems(CCharEntity* PChar) {
+    //Limit number of items that can drop to the treasure pool size
+    uint8 dropCount = 0;
+    #define ADD_ITEM_TO_POOL(ItemID) { PChar->PTreasurePool->AddItem(ItemID, this); if( ++dropCount == TREASUREPOOL_SIZE ) { return; } }
```
teschnei: this is definitely no good, but it should be trivial to replace it with a lambda

jmcmorris: Yeah, I wasn't sure and originally had it exploded out in every instance. Unfortunately a lambda wouldn't work because I want to return from `DropItems()` if the dropcount is the treasure pool size. It is easy enough to explode it out though.

teschnei: the lambda can just return a bool and you can do `if (lambda()) return;`

teschnei: oh sorry, I don't see edits in email, but yeah you got it

jmcmorris: Not sure if you got notified but I made those changes :)

src/map/entities/mobentity.cpp
```diff
@@ -803,150 +803,179 @@ void CMobEntity::DropItems()
...
+    }
+    else
+    {
+        luautils::OnMobDeath(this, nullptr);
+    }
+}
+
+void CMobEntity::DropItems(CCharEntity* PChar) {
+    //Adds an item to the treasure pool and returns true if the pool has been filled
+    auto AddItemToPool = [this, PChar](uint16 ItemID, uint8 dropCount) -> bool {
```
teschnei: you don't need the `-> bool` part

src/map/entities/mobentity.cpp
```diff
@@ -803,150 +803,179 @@ void CMobEntity::DropItems()
...
-
-        // NOTE: this is called for all alliance / party members!
-        luautils::OnMobDeath(this, PChar);
-
-    }
-    else
-    {
-        luautils::OnMobDeath(this, nullptr);
     }
+    #undef ADD_ITEM_TO_POOL
```
teschnei: this should be removed

src/map/entities/mobentity.cpp
```diff
@@ -784,7 +784,7 @@ void CMobEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
     }
 }
 
-void CMobEntity::DropItems()
+void CMobEntity::HandleDeath()
```
teschnei: can we change the name of this function? It gets queued from `m_dropItemTime` milliseconds after dying, so it's not strictly "handling death" - mostly the distribution of stuff (minus kill all status effects, but that could probably get moved back into `Die`)
something like DistributeRewards, or whatever is probably good

src/map/entities/mobentity.cpp
jmcmorris: Ah thanks, this is my first foray into c++ lambdas - I've been stuck using pre-c++11 code. I didn't know the return value would be inferred.

PR #5236, opened 2018-09-18
Fixed issue with unlotted treasure pool items being distributed incorrectly

jmcmorris:
Fixed an issue where an unlotted rare item in the treasure pool would be randomly given to a player even if they passed or already have that rare item. The code was mostly there it just was under the invalid `highestInfo` conditional that rarely (if ever) is triggered. I moved that logic below so it is triggered if no lotters or the highestInfo sanity check failed. While at it I also fixed some formatting issues.

Tested in the following ways:
- Treasure pool with rare item and having one player with the item and one player without. Verified would always go to the correct player.
- Both players already have the item and it would be lost
- Both players lot on the item and it would go to the player with highest lot
- One player lot without having the item
- Both players not have the item and one player pass and it would go to the player who did not pass

### DIFF
--- a/src/map/treasure_pool.cpp
+++ b/src/map/treasure_pool.cpp
@@ -392,95 +392,82 @@ void CTreasurePool::CheckTreasureItem(time_point tick, uint8 SlotID)
             highestInfo.lot = 0;
             highestInfo.member = nullptr;
 
-            for(uint8 i = 0; i < m_PoolItems[SlotID].Lotters.size(); i++){
+            for(uint8 i = 0; i < m_PoolItems[SlotID].Lotters.size(); i++)
+            {
                 LotInfo curInfo = m_PoolItems[SlotID].Lotters[i];
-                if(curInfo.lot > highestInfo.lot){
+                if(curInfo.lot > highestInfo.lot)
+                {
                     highestInfo = curInfo;
                 }
             }
 
             //sanity check
-            if(highestInfo.member != nullptr && highestInfo.lot!=0){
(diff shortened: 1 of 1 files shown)


TeoTwawki:
if this also fixed the "one guy didn't pass and it fell to the dirt even thought I lotted" error then you just became every servers hero

jmcmorris:
@TeoTwawki Is that also a problem? I don't see that in the code. I have ran into the issue where only one guy would have a rare item (such as Damsefly Worm) and anyone else would want it but it would try to give it to that one guy and then the item would fall to the ground. Sad times.

TeoTwawki:
we've had a longstanding issue where on retail it would fall to the highest lot even if one person passed, but on darkstar it just acts like nobody lotted and will try to toss it at someone who passed or already had it.. resulting in lost item.

KnowOne134:
It indeed still a problem

jmcmorris:
I'll look into that and see if I can get it fixed for this PR then

jmcmorris:
I read through the code more thoroughly and from what I can see it looks like it was fixed. I tested this specific case you mentioned where one player lots and the other passes and it always falls to the player that lotted.

KnowOne134:
Its not always its a wierd sometimes bug. Also happens if someone disbands party or leaves zone

TeoTwawki:
> specific case you mentioned where one player lots and the other passes and it always falls to the player that 

no its when they _*don't*_ pass and the time limit is reached. and its not a weird sometimes bug - if its not happening its fixed. There is a completely different weird sometimes bug..Which is probably an index problem.

jmcmorris:
Refactored the code so that it is much cleaner and then tested once more

PR #5237, opened 2018-09-20
Player death message fix

jmcmorris:
This fixes the issue where a player can be sent the wrong death message. It currently uses the target but if the player is not currently targeting the monster that kills them then it will either say that the wrong monster killed them or "falls to the ground".

The fix here is to track which entity was the last one to deal damage to the character and clear it whenever the player takes neutral damage such as from poison.

One thing I'm not crazy about is how the magic case is handled. Magic from monsters deal damage by calling `CLuaBaseEntity::delHP()` but that does not pass in the attacker and is called by hundreds of scripts. It would be somewhat trival to do it for monster scripts but would greatly bloat this PR and have a much higher chance for conflicts. I'm open to whatever.

Tested this by dying to melee attack, magic attack, spikes and poison. All but the last now show "defeated by" message and the last shows "falls to the ground".

Thanks to @Wiggo32 for helping get the correct death spikes message.

### DIFF
--- a/scripts/globals/spells/death.lua
+++ b/scripts/globals/spells/death.lua
@@ -17,8 +17,6 @@ function onSpellCast(caster,target,spell)
         return 0
     end
 
-    -- falls to the ground
-    spell:setMsg(dsp.msg.basic.FALL_TO_GROUND)
     target:setHP(0)
 
     return 0
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -73,6 +73,7 @@ CBattleEntity::CBattleEntity()
     PPet = nullptr;
     PParty = nullptr;
     PMaster = nullptr;
+    PLastAttacker = nullptr;
 
     StatusEffectContainer = std::make_unique<CStatusEffectContainer>(this);
     PRecastContainer = std::make_unique<CRecastContainer>(this);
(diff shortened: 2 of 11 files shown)


TeoTwawki:
> by calling CLuaBaseEntity::delHP() but that does not pass in the attacker and is called by hundreds of scripts.

At some future point I think we'll have to make a function to put taking damage into so that we aren't directly calling delHP like that

jmcmorris:
I can make that change if you'd like but it is going to touch nearly 700 files - most of which are in mobskills. Not entirely sure it fits within the scope of this PR but I see it going either way.

teschnei:
okay, it looks pretty good, but.. why's `PLastAttacker` in CCharEntity only? don't we want that in all battle entities anyways?

jmcmorris:
I originally had it in CBattleEntity but since I was only using it in CCharEntity I decided to keep it in there. I can always revert that change if you think it will be useful for CBattleEntity.

teschnei:
yeah? mobs do the same thing (falls to the ground vs defeated by x), don't they?

jmcmorris:
You're right that mobs need to be fixed at least a little bit. Right now if a mob dies with no target (player died but mob was poisoned) then it does not display any message. This should definitely say "falls to the ground".

What happens in retail if a mob dies from poison but does have a target? Should it say "falls to the ground" or "defeated by"?

jmcmorris:
@wiggo32 was kind enough to test this on retail. It is "falls to the ground". I will update the code.

teschnei:
It is falls to the ground, yes. It's also falls to the ground for any
script that sets HP to 0, such as chlevnik and chigoes

teschnei:
one last thing - setHP should clear LastAttacker.  The situations I described earlier (chelvnik and chigoes) would (or are, I forget) use setHP to "kill" them

jmcmorris:
Haha yup, I was just reviewing things and remembered that. Testing out my fix now. :)

jmcmorris:
@teschnei Thanks for pushing on this and helping get it to a good state - same for my other PRs

PR #5238, opened 2018-09-20
Resolved issue with players death timer being set incorrectly while zoning

jmcmorris:
Whenever the player would zone we would set their death time to how long they were dead for not how much more time they have. So the quicker you zoned after death the faster you would be automatically sent back to your homepoint.

This also fixes an issue where we were not sending the death timer correctly in the char_update packet. While digging through the auto HP issue for hours I noticed that retail always sends max death time while alive and sends the time remaining while dead. It is worth noting that retail also sends the char_update with an updated death time every 15~ seconds. My guess is that this is to help keep the client time synced.

This was tested with countless deaths and tractoring. After tractoring your timer correctly persists.

@Wiggo32 helped significantly through the process of coming up with these changes.

### DIFF
--- a/src/map/entities/charentity.h
+++ b/src/map/entities/charentity.h
@@ -39,6 +39,7 @@ This file is part of DarkStar-server source code.
 #define MAX_QUESTID     256
 #define MAX_MISSIONAREA	 15
 #define MAX_MISSIONID    226
+#define PLAYER_DEATH_COUNTER_TIME 60min
 
 struct jobs_t
 {
--- a/src/map/entities/charentity.h
+++ b/src/map/entities/charentity.h
@@ -39,7 +39,6 @@ This file is part of DarkStar-server source code.
 #define MAX_QUESTID     256
 #define MAX_MISSIONAREA	 15
 #define MAX_MISSIONID    226
-#define PLAYER_DEATH_COUNTER_TIME 60min
 
 struct jobs_t
 {
@@ -266,7 +265,6 @@ public:
     uint16            m_Costum;                     // карнавальный костюм персонажа (модель)
     uint16			  m_Monstrosity;				// Monstrosity model ID
     uint32			  m_AHHistoryTimestamp;			// Timestamp when last asked to view history
-    uint32            m_DeathCounter;               // Counter when you last died. This is set when you first login
     uint32            m_DeathTimestamp;             // Timestamp when death counter has been saved to database
 
     uint8			  m_hasTractor;					// checks if player has tractor already
@@ -299,6 +297,8 @@ public:
     bool              m_EffectsChanged;
     time_point        m_LastSynthTime;
 
+    static constexpr duration death_duration = 60min;
+
     int16 addTP(int16 tp) override;
     int32 addHP(int32 hp) override;
     int32 addMP(int32 mp) override;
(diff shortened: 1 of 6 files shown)


src/map/entities/charentity.h
```diff
@@ -39,6 +39,7 @@ This file is part of DarkStar-server source code.
 #define MAX_QUESTID     256
 #define MAX_MISSIONAREA	 15
 #define MAX_MISSIONID    226
+#define PLAYER_DEATH_COUNTER_TIME 60min
```
teschnei: replace define with constexpr

src/map/packets/char_update.cpp
```diff
@@ -90,8 +92,20 @@ CCharUpdatePacket::CCharUpdatePacket(CCharEntity* PChar)
     
     ref<uint8>(0x36) = flag;
 
-    if (!PChar->isDead() || PChar->m_DeathCounter == 0) //prevent this packet from resetting the homepoint timer after tractor
+    //This contains the amount of time remaining before the player is forced back to homepoint while dead
+    if (PChar->PAI->IsCurrentState<CDeathState>())
```
teschnei: PChar->isDead()?

src/map/packets/char_update.cpp
```diff
@@ -90,8 +92,20 @@ CCharUpdatePacket::CCharUpdatePacket(CCharEntity* PChar)
...
+    {
+        //Update the player death counter and timestamp and send the amount of time remaining
+        uint32 currentTime = (uint32)time(nullptr);
+        PChar->m_DeathCounter += (currentTime - PChar->m_DeathTimestamp);
+        PChar->m_DeathTimestamp = currentTime;
```
teschnei: real lost here - why is a packet updating member info? why is the death timestamp being constantly overwritten with current time?
Maybe it's because the way we're currently dealing with death timers (two member variables) is real wrong, but I guess this presents an opportunity to fix it

src/map/packets/char_update.cpp
teschnei: if a character is zoning, that means the entire ccharentity is about to get deleted and remade somewhere else (and I imagine client side it all gets overwritten again with the zone in packet too), or are you talking about zoning *in*?

src/map/packets/char_update.cpp
jmcmorris: You bring up a good point. It is wrong for the packet to be updating member info. This could be changed to the following so it is not updating member info

```c++
uint32 currentTime = (uint32)time(nullptr);
int32 counter = PChar->m_DeathCounter + (currentTime - PChar->m_DeathTimestamp);
ref<uint32>(0x3C) = 0x0003A020 - (60 * counter);
```

I can look into consolidating these two variables into a single one though.

src/map/packets/char_update.cpp
jmcmorris: I think it happens for both zoning out and in. If I recall correctly this happens when being tractored. I can double check that though.

teschnei:
That would send the right timestamp though? Being dead or being
status_disappear.. it's the same for being tractored

jmcmorris:
@teschnei Ok, I did some testing. If I use `PChar->isDead()` then it is only being sent when zoning out and not while zoning in. However, it does mean that while zoning in it will send this packet with incorrect data (as I said, retail always sends `0x0003A020` while you're alive).

While looking at refactoring the two death variables into one I will see if I can simplify this so we can use `PChar->isDead()` or maybe not even needing a conditional in here at all.

teschnei:
we probably don't need the conditional at all,, I'm guessing it can just check the death timestamp instead (i.e if timestamp is zero, then they're alive and send the full counter, otherwise figure out the real counter)

jmcmorris:
Much cleaner now! Thanks for the discussion and direction. 👍

Issue #5239, opened 2018-09-20
Item Signatures cut off at certain lengths/ending with certain characters


nfaletra:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30180904_0

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Character names that are 7 or 15 characters long, depending on the character they end in, could be cutoff by 1 character during the DecodeStringSignature process.

My character name "Autkast" appears as expected on signed items directly after they are created. However, when the owner of said item (could be traded or bazaared) logs out and logs back in, causing the signature to be reloaded from the DB, it appears as "Autkas".

After much debugging and bitwise math (my brain hurts), I have discovered that the issue lies in calculating the length of the original string (before it is 'encoded') in the DecodeStringSignature function here:

`
auto length = std::min<size_t>(15u, (strlen((const char*)signature) * 8) / 6);
`

To explain what is happening, the last 2 bits of the encoded string are 0 and are shoved into the 6th byte of signature. The way strlen works is it reads characters until it hits a null. In this case the 6th byte is null as far as it's concerned and returns 5 for strlen. Then the rest of the math happens and we end up with a length of 6. This is incorrect and the last 6 bits that are packed into the signature are ignored.

I have a few solutions (some are gross) but I wanted to open a discussion here rather in discord because of the amount of information I've collected (oh and Zynjec helped. His server now has a million items signed by Autkas.)

1. Add 1 to length if the math comes out to 6 or 14 (gross)
2. Store the encoded string into the DB and never decode again (also gross)
3. Add 1 to what strlen returns ALWAYS (not as gross)
4. Always loop through the max signature length of 15 characters (preferred)

Thoughts? All this bitwise math started to make my head hurt. If there's some voodoo way to make sure those last bits are always read in, I don't know it, but this decode happens so infrequently that I don't think it's a bad thing to just loop through the full 15 bitpacked bytes.

teschnei:
We used to do 2 but I changed it so that the db could be understood more
clearly. It's probably fine to do 4, we know the max length is 15 anyways
so it's not like we risk an overflow

nfaletra:
[...]

PR #5240, opened 2018-09-20
Fix item signatures getting cut off

nfaletra:
Addresses: https://github.com/DarkstarProject/darkstar/issues/5239

### DIFF
--- a/src/common/utils.cpp
+++ b/src/common/utils.cpp
@@ -33,6 +33,8 @@
 #	include <intrin.h>
 #endif
 
+#define MAX_ITEM_SIGNATURE_LENGNTH 15
+
 /************************************************************************
 *																		*
 *																		*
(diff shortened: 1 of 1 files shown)


src/common/utils.cpp
```diff
@@ -33,6 +33,8 @@
 #	include <intrin.h>
 #endif
 
+#define MAX_ITEM_SIGNATURE_LENGTH 15
```
teschnei: it's not the max item signature length, it's just the maximum character name length, isn't it? so maybe this should be somewhere related to that and renamed?

also, please favor constexpr over `define`

nfaletra: like constexpr size_t GetPlayerNameLengt() { return 15; } ? I feel bad that I've never used constexpr before. It's still compile time from what I understand?

Switching that over and cleaning up other hardcoded places that use 15 for player name length. A lot of them are redundant because CCharEntity::SetName(...) already limits the number of characters that get copied.

teschnei: just `constexpr size_t max_name_length = 15;` should suffice

src/common/utils.h
```diff
@@ -29,6 +29,8 @@
 #include "../common/cbasetypes.h"
 #include "../common/mmo.h"
 
+constexpr size_t GetPlayerNameLength() { return 15; }
```
teschnei: i'm not sure if it should explicitly say "player" - technically, the size is also 15 for every entity in the packets isn't it?  up for discussion, that one

but either way this should just be a variable instead of a function `constexpr size_t max_name_length = 15;`

TeoTwawki: mobs and npc's can have longer names, unless we shove phone text through player chat packets

teschnei:
Yes, their real name is longer, but that doesn't go through the packet
(which is the only part I'm concerned with here)

src/common/utils.h
zynjec: For packet purposes, they're all limited to 15 characters + null terminator, anything more gets truncated.

teschnei: sure, but not `Get` because it won't be a function

nfaletra: Do you want me to prefix it with anything to signify that it's a constant?

nfaletra: Also, please correct me if I'm wrong, but packets are cleared out with 0 before being constructed, yes?

teschnei: packets are zero initialised, yep
I don't think it needs a prefix, the type is constexpr and that's probably good enough

nfaletra: ez lyfe. I'm not going to attempt to limit any packets that are using EXPLICITLY a CCharEntity name because it's already limiting the length to 15 characters. Anything else, e.g. BaseEntity, packetname, etc, I will make sure it limits the amount of data being copied into the packet if it was doing so before.

Auction house is a weird exception. Not sure why it was 16, but it uses CCharEntity so letting the limit in SetName handle it.

PR #5241, opened 2018-09-21
Fix issue with a cancelled raise preventing future raises and tractors

jmcmorris:
The raise flag was only being cleared when raised which meant that if the player declined a raise they would not receive anymore raise or tractor menus until they homepointed. Now we clear it when the player cancels the menu.

Tested with cancelling a raise and casting another raise then cancelling that and casting tractor on the character then cancelling that and then casting raise and accepting finally.


wrenffxi:
just wanted to say thanks for the contributions.  much appreciated.

PR #5245, opened 2018-09-22
Resolved death auto hp issue (and friends)

jmcmorris:
The death auto HP issue is now finally understood and fixed. This is after a long night of digging through decompiled code, using cheat engine to debug breakpoints on the client, and carefully reviewing the incoming packets on retail with great help from @atom0s.

The underlying issue is that char_update (`0x37`) should be sending in `0x40` the VanaTime of when the player will be forced to homepoint while dead. While alive this value is about 3960 (66 minutes) above the client's VanaTime and keeps increasing. Once the player dies this value stops moving and decreases at the same rate the value at `0x3C` in char_update packet increases. The problem is that dsp would always send the current VanaTime regardless if the character is alive or dead and it would not be offset by any amount.

Digging deeper we found the client compares `0x40` to its locally calculated VanaTime and if it is less than it would set that time as the "send to homepoint" time. Since this time had already passed it would fail the initial check and immediately send the character back to their homepoint. This appeared randomly because sometimes the conditional would fail and it would instead use the value in `0x3C` to calculate the send to homepoint time which would be a correct time.

Along the way several other discrepancies were noticed which I addressed here as well:
* The char_update packet is being sent whenever the character moves but this is not necessary and not the case on retail
* The char_update packet should be sent while the character is dead at some interval. This interval is about 16 seconds and seems to be used to keep the client's death times in sync.
* We were sending an unnecessary homepoint menu packet on death (retail does not do this)
* There was an issue with the death timestamp being set incorrectly while zoning in

While working on this I also looked over the zone_in packet to see if that also needs the "send to homepoint" VanaTime but it does not.

More data can be provided such as the decompiled code or retail packets.

This fix was tested with countless sacrificial deaths.

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -1629,6 +1629,15 @@ int32 CCharEntity::GetSecondsElapsedSinceDeath()
     return m_DeathTimestamp > 0 ? (uint32)time(nullptr) - m_DeathTimestamp : 0;
 }
 
+int32 CCharEntity::GetTimeRemainingUntilDeathHomepoint()
+{
+    //0x0003A020 is 60 * 3960 and 3960 is 66 minutes in seconds
+    //The client uses this time as the maximum amount of time for death
+    //We convert the elapsed death time to this total time and subtract it which gives us the remaining time to a forced homepoint
+    //Once the returned value here reaches below 360 then the client with force homepoint the character
+    return 0x0003A020 - (60 * GetSecondsElapsedSinceDeath());
+}
+
 void CCharEntity::TrackArrowUsageForScavenge(CItemWeapon* PAmmo)
 {
     // Check if local has been set yet
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -464,6 +464,17 @@ bool CCharEntity::ReloadParty()
     return m_reloadParty;
 }
 
+void CCharEntity::Tick(time_point tick)
+{
+    CBattleEntity::Tick(tick);
+    if (m_DeathTimestamp > 0 && tick >= m_deathSyncTime)
+    {
+        //Send an update packet at a regular interval to keep the player's death variables synced
+        updatemask |= UPDATE_STATUS;
+        m_deathSyncTime = tick + death_update_frequency;
+    }
+}
+
 void CCharEntity::PostTick()
 {
     CBattleEntity::PostTick();
@@ -1597,6 +1608,7 @@ void CCharEntity::Die()
 
 void CCharEntity::Die(duration _duration)
 {
+    m_deathSyncTime = server_clock::now() + death_update_frequency;
     PAI->ClearStateStack();
     PAI->Internal_Die(_duration);
     pushPacket(new CRaiseTractorMenuPacket(this, TYPE_HOMEPOINT));
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -519,7 +519,9 @@ void CCharEntity::PostTick()
                 static_cast<CCharEntity*>(PEntity)->pushPacket(new CCharHealthPacket(this));
             });
         }
-        pushPacket(new CCharUpdatePacket(this));
+        //Do not send an update packet when only the position has change
+        if (updatemask ^ UPDATE_POS)
+            pushPacket(new CCharUpdatePacket(this));
         updatemask = 0;
     }
 }
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -1613,7 +1613,6 @@ void CCharEntity::Die(duration _duration)
     m_deathSyncTime = server_clock::now() + death_update_frequency;
     PAI->ClearStateStack();
     PAI->Internal_Die(_duration);
-    pushPacket(new CRaiseTractorMenuPacket(this, TYPE_HOMEPOINT));
 
     // reraise modifiers
     if (this->getMod(Mod::RERAISE_I) > 0)
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -469,7 +469,7 @@ void CCharEntity::Tick(time_point tick)
     CBattleEntity::Tick(tick);
     if (m_DeathTimestamp > 0 && tick >= m_deathSyncTime)
     {
-        //Send an update packet at a regular interval to keep the player's death variables synced
+        // Send an update packet at a regular interval to keep the player's death variables synced
         updatemask |= UPDATE_STATUS;
         m_deathSyncTime = tick + death_update_frequency;
     }
@@ -519,7 +519,7 @@ void CCharEntity::PostTick()
                 static_cast<CCharEntity*>(PEntity)->pushPacket(new CCharHealthPacket(this));
             });
         }
-        //Do not send an update packet when only the position has change
+        // Do not send an update packet when only the position has change
         if (updatemask ^ UPDATE_POS)
             pushPacket(new CCharUpdatePacket(this));
         updatemask = 0;
@@ -1644,10 +1644,10 @@ int32 CCharEntity::GetSecondsElapsedSinceDeath()
 
 int32 CCharEntity::GetTimeRemainingUntilDeathHomepoint()
 {
-    //0x0003A020 is 60 * 3960 and 3960 is 66 minutes in seconds
-    //The client uses this time as the maximum amount of time for death
-    //We convert the elapsed death time to this total time and subtract it which gives us the remaining time to a forced homepoint
-    //Once the returned value here reaches below 360 then the client with force homepoint the character
+    // 0x0003A020 is 60 * 3960 and 3960 is 66 minutes in seconds
+    // The client uses this time as the maximum amount of time for death
+    // We convert the elapsed death time to this total time and subtract it which gives us the remaining time to a forced homepoint
+    // Once the returned value here reaches below 360 then the client with force homepoint the character
     return 0x0003A020 - (60 * GetSecondsElapsedSinceDeath());
 }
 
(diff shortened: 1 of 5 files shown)


jmcmorris:
Haha yeah, that function name is quite long. I was wanting to make it super descriptive since it is fairly complex and very specific. The one thing I do like about the hex is that is exactly what you see in the packets but then again most people wont be looking at the packets when looking at this code. I can see it going either way.

waterlgndx:
yeah i definitely think the comment should remain so if someone searches
the hex it comes up.  since it appears in a packet it might be best to
leave it in there ( and it's a typical value coming up in packets ).  I
doubt it's something that is likely to change in a version update.

atom0s:
Here is the reversing work I did for this to help @jmcmorris get this fixed.

First thing we looked at was the outgoing 0x1A handler to see if something was causing the client to automatically homepoint, which was not the case:

```cpp
signed int __cdecl sub_10077120(__int16 a1, int a2, __int16 a3, int a4, int a5, int a6, int a7, int a8)
{
  int v8; // eax

  v8 = sub_100F90F0(0x1A, a8, 0);
  if ( !v8 )
    return -1;
  *(_WORD *)(v8 + 8) = a1;
  *(_DWORD *)(v8 + 4) = a2;
  *(_WORD *)(v8 + 10) = a3;
  *(_DWORD *)(v8 + 12) = a4;
  *(_DWORD *)(v8 + 16) = a5;
  *(_DWORD *)(v8 + 20) = a6;
  *(_DWORD *)(v8 + 24) = a7;
  sub_100F9200(v8, 28, 0);
  return 0;
}
```

Next was to dig backward to see what invoked the 0x1A packet and for the specific action of homepointing, which was action 0x0B. Which yeilds this switch case:

```cpp
        case 2:
          if ( (unsigned __int8)sub_10096510() )
            v22 = sub_100770F0(
                    *(_DWORD *)(a1 + 116),
                    *(_DWORD *)(a1 + 120),
                    0xB,
                    2 - (*(_DWORD *)(dword_10565840 + 36) != 1),
                    0);
          else
            v22 = sub_100770F0(*(_DWORD *)(a1 + 116), *(_DWORD *)(a1 + 120), 0xB, 0, 0);
          if ( !v22 )
          {
            *(_DWORD *)(a2 + 4) = 3;
            *(_DWORD *)(a2 + 16) = 600;
          }
          return;
```

Here, the client is testing a few things. The first call is a check for if the client is dead or on an invalid job.

```cpp
bool sub_10096510()
{
  bool result; // al

  if ( dword_104CD388 && *(_WORD *)(dword_104CD388 + 0x3B002) )
    result = 1;
  else
    result = (_BYTE)dword_10472C48 == 0x17;
  return result;
}
```

The packet is being invoked with:

```cpp
sub_100770F0(*(_DWORD *)(a1 + 116), *(_DWORD *)(a1 + 120), 0xB, 0, 0);
```

a1+116 = player_target_index
a1+120 = player_server_id

These are onnly called when your client clicks the option to homepoint so it wasn't the location needed to determine what caused the auto-homepointing, but it was needed to look into first to ensure extra bits weren't being overlooked.

Next step was to look at the packet we knew held the information about your death time, 0x37. Inside of that packet handler is the timer information set here:

```cpp
  if ( *(_DWORD *)(a3 + 60) > 0 )
  {
    v54 = sub_10100550();
    v55 = *(_DWORD *)(a3 + 64);
    if ( v55 > v54 )
    {
      *(_DWORD *)(a1 + 0x3B028) = v55;
      return 1;
    }
    *(_DWORD *)(a1 + 0x3B028) = v54 + *(_DWORD *)(a3 + 60) / 60;
  }
```
sub_10100550 is a function used to get the clients synced time with the server. This statement is checking if the client and server times are out of sync to correct them as well as to update the death timing. This func is:

```cpp
signed int sub_100D9250()
{
  unsigned int v0; // eax
  unsigned int *v1; // edi
  unsigned int v2; // esi
  signed int result; // eax
  int v4; // edi
  signed int v5; // edx
  int v6; // ecx

  v0 = timeGetTime();
  v1 = (unsigned int *)dword_10480400;
  v2 = v0;
  result = sub_100D9230(-1, *(_DWORD *)dword_10480400, v0);
  *v1 = v2;
  if ( result >= 0 )
  {
    v4 = dword_10480400;
    v5 = *(_DWORD *)(dword_10480400 + 20);
    if ( v5 && result > 20 )
    {
      v6 = *(_DWORD *)(dword_10480400 + 20);
      if ( v5 >= 0 )
      {
        if ( v5 > 20 )
          v6 = 20;
        if ( v6 > result )
          v6 = result;
      }
      else
      {
        if ( v5 < -20 )
          v6 = -20;
        if ( v6 < -result )
          v6 = -result;
      }
      result += v6;
      *(_DWORD *)(dword_10480400 + 20) = v5 - v6;
      v4 = dword_10480400;
    }
    *(_DWORD *)(v4 + 16) += result;
    for ( result = dword_10480400; *(_DWORD *)(dword_10480400 + 16) >= 0x3E8u; result = dword_10480400 )
    {
      ++*(_DWORD *)(result + 12);
      *(_DWORD *)(dword_10480400 + 16) -= 1000;
    }
  }
  return result;
}
```

Another chunk of code used to test and homepoint is:
```cpp
else if ( !(unsigned __int8)sub_10118390(dword_10565780) )
{
  v16 = sub_10100550();
  if ( *(_DWORD *)(a2 + 12) <= v16 )
  {
    if ( !(unsigned __int8)sub_101C2440(1, 0) )
      goto LABEL_83;
    *(_DWORD *)(a2 + 4) = 3;
    *(_DWORD *)(a2 + 16) = 600;
  }
  else if ( !(unsigned __int8)sub_101C2440(1, 0) || *(_DWORD *)(a2 + 12) < v16 + 300 )
  {
    sub_10129D00(*(_DWORD *)(a2 + 12));
  }
}
```

This is used to determine if the client is in ballista when they died, which is checked via:
```cpp
bool __cdecl sub_101C2440(char a1, char a2)
{
  int v2; // eax
  int v3; // ecx
  unsigned __int8 v4; // cl
  unsigned __int8 v5; // al
  bool result; // al

  v2 = sub_101C25E0();
  result = 0;
  if ( v2 )
  {
    v3 = *(_DWORD *)(v2 + 0x1CC);
    if ( v3 & 0x400 )
    {
      if ( !a1 || (v4 = *(_BYTE *)(v2 + 0x1FB), v4 < 8u) || v4 > 9u )
      {
        if ( !a2 || (v5 = *(_BYTE *)(v2 + 0x1FB), v5 < 0x28u) || v5 > 0x2Bu )
          result = 1;
      }
    }
  }
  return result;
}
```

In the same func with the switch case mentioned earlier, the timer data is set and handled with:
```cpp
  if ( !*(_DWORD *)(a2 + 12) )
  {
    v3 = *(_DWORD *)(dword_104CD388 + 241704);
    if ( v3 > 360 )
      *(_DWORD *)(a2 + 12) = v3 - 360;
  }
```

a2 is a static address in the client, currently:
(int)&dword_104735F0

This is used and set when the client is seen as dead:
```cpp
bool __thiscall sub_10096220(_DWORD *this)
{
  int v1; // eax

  v1 = this[91]; // the player entity server status
  return v1 == 2 || v1 == 3; // 2 = dead, 3 = dead in combat
}
```

Next step was to dump this data being set from the packets on retail, specifically the two timers at offsets 0x3C and 0x40 in the 0x37 packet:


To figure out how these values were being compared to and linked to the other information, the time sync function's return value became an important clue. I wrote a quick Cheat Engine auto-assembler script and addon to dump and read the information each time the 0x37 packet was received:

```
FFXiMain.dll+100550 - E8 0B8FFDFF           - call FFXiMain.dll+D9460
FFXiMain.dll+100555 - 03 05 E8E05305        - add eax,[FFXiMain.dll+4CE0E8] { [-415.02] }
FFXiMain.dll+10055B - C3                    - ret 
```
```
[ENABLE]
alloc(cave,1024)
alloc(time,4)
registersymbol(time)
label(rethere)

cave:
    add eax, [FFXiMain.dll+4CE0E8]
    mov [time], eax // store the time return..
    jmp rethere

FFXiMain.dll+100555:
    jmp cave
    nop
rethere:


[DISABLE]
FFXiMain.dll+100555:
    add eax,[FFXiMain.dll+4CE0E8]
dealloc(cave)
dealloc(time)
unregistersymbol(time)
```

Then in a simple Ashita addon:
```lua
_addon.author = 'atom0s';
_addon.version = '1.0';
_addon.name = 'test';

ashita.register_event('incoming_packet', function(id, size, data)
    if (id == 0x37) then
        local t1 = struct.unpack('L', data, 0x3C + 1);
        local t2 = struct.unpack('L', data, 0x40 + 1);
        local t3 = ashita.memory.read_uint32(0x037B0400); -- ce 'time' address
        local t4 = t2 - t3;
        print(string.format('0x3C: %08X (%d) || 0x40: %08X (%d) || time: %08X (%d) || dif: %d', t1, t1, t2, t2, t3, t3, t4));
    end
    return false;
end);
```
Replacing the time address with the current one allocated in CE when using the asm script above.

This then showed us what was happening with the timers on retail:


We can see that the times have a ~66min difference on purpose from the server. Which is the 3960 value (or 3960 * 60 = 0x3A020) used in the packet on DSP. 

So knowing this it was time to die and see what happens. 


Bingo! The missing key. as soon as the client dies, the time value is locked and resent ~15-16 seconds in 0x40 spot of the packet. This then uses 0x3C to count down the difference, and you can see the time diff on the right counting down from the 3960 value. 

Cheers to @jmcmorris for asking for help with this and giving me a fun night of reversing and working with him to fix this. :) (And for letting me use his retail acct for the needed info.)

waterlgndx:
@atom0s that's pretty intense, I like it lol

whasf:
Thank you for the details! This has been quite an elusive bug that required quite a lot of digging.

maxtherabbit:
nice work! I always knew it was the times in char update, but I honestly never would have guessed it needed a vana time offset

gweivyth:
Awesome work guys.  Thanks so much for this.  I know that it was an unsolved mystery for many years in the DSP community.  <3

PR #5248, opened 2018-09-23
Black coffin zone in out and instance complete changes

KnowOne134:
issues with party members that have previously completed not being able to enter. Problem with beating the instince the mobs would re agro and attack canceling CS. the end cs zoning you into the reef to be stuck in a cs while mobs killed you and players that have previous beat the mission stuck in instance.


KnowOne134:
rebased for global IDs

PR #5253, opened 2018-09-28
[bcnmcrap] Add dynamis_beastmen mixin and coins to non-dreamlands beastmen

wrenffxi:


teschnei:
for build failures, it looks like both from when whasf merged master in - the lua checks passed and this is lua only, so it's good

this looks good, thanks a bunch for doing that.  there's still work left to be done for cities, but this is one of the most boring parts done

PR #5257, opened 2018-09-29
Moghancements

jmcmorris:
This is the initial implementation of moghancements which supports all but 3 moghancements: 
- `MOGHANCEMENT_GARDENING`
- `MOGHANCEMENT_FISHING_ITEM`
- Part of `MEGA_MOGLIFICATION_FISHING`

While loading a character the current moghancement is determined by looking for any moghancement key items. The moghancement effect is applied through a modifier. Since modifiers do not persist we do not worry about removing its effect since the only time a player's current moghancement can change is while in a Mog House. Several new modifiers were added in to support these.

After the character finishes laying out furniture in their MH the client sends a `0x0FA` (Place furniture) packet with no item. At this point the character's current moghancement is recalculated based on the dominant element, highest item aura and placement order. This is all outlined on [ffxiclopedia moghancements](http://ffxiclopedia.wikia.com/wiki/Moghancements).

Placement order is tracked inside the item furnishing extra data as a single byte. Whenever an item is placed it updates the placed furniture orders so that the most recently placed item is `order = 0`.

As far as the strength of moghancements go, it appears that originally it was speculated that the aura strength affects the moghancement but I never found any data to back this up. There are also several moghancements that have no corresponding verified strength. I made a note on the ones that I took the liberty to assign a value. I did my best to give it a strength that makes it somewhat worthwhile but not overpowered by any means.

I tested this with several different moghancements making sure that the calculated one was correct and applied as well as when no moghancement is selected it removes the key item. Tested having `MOGHANCEMENT_ALCHEMY` active outside of the MH and having it properly grant +1 Alchemy Skill.

**I am still figuring out the coding style for DSP so feel free to suggest styling/code changes.**


src/map/packets/message_special.h
```diff
@@ -28,6 +28,11 @@
 
 #include "basic.h"
 
+enum MSGSPECIAL_ID : uint16
+{
+    MSGSPECIAL_KEYITEM_OBTAINED = 6390, /* Obtained key item: <item>. */
```
nfaletra: These text IDs are per zone. Gonna have to figure out a different way to handle this. Maybe a helper that calls to npcUtil.giveKeyItem? Not sure the correct way to handle C++ -> Lua.

zynjec: This would be a good situation to use `luautils::GetTextIDVariable`, but it'll need a bit of work since it wasn't updated with the tabling of IDs.

src/map/packet_system.cpp
```diff
@@ -5198,8 +5201,45 @@ void SmallPacket0x0FA(map_session_data_t* session, CCharEntity* PChar, CBasicPac
         PItem->setLevel(level);
         PItem->setRotation(rotation);
 
+        // Update installed furniture placement orders
+        // First we place the furniture into placed items using the order number as the index
+        static constexpr uint8 containers[2] = { LOC_MOGSAFE, LOC_MOGSAFE2 };
+        CItemFurnishing* placedItems[MAX_CONTAINER_SIZE * 2] = { nullptr };
+        for (int containerIndex = 0; containerIndex < 2; ++containerIndex)
```
teschnei: drop `containers` and just replace this line with
```
for (auto containerid : {LOC_MOGSAFE, LOC_MOGSAFE2})
```

src/map/packet_system.cpp
```diff
@@ -5198,8 +5201,45 @@ void SmallPacket0x0FA(map_session_data_t* session, CCharEntity* PChar, CBasicPac
         PItem->setLevel(level);
         PItem->setRotation(rotation);
 
+        // Update installed furniture placement orders
+        // First we place the furniture into placed items using the order number as the index
+        static constexpr uint8 containers[2] = { LOC_MOGSAFE, LOC_MOGSAFE2 };
+        CItemFurnishing* placedItems[MAX_CONTAINER_SIZE * 2] = { nullptr };
```
teschnei: `std::array`

src/map/entities/charentity.cpp
```diff
@@ -1651,6 +1658,340 @@ int32 CCharEntity::GetTimeRemainingUntilDeathHomepoint()
     return 0x0003A020 - (60 * GetSecondsElapsedSinceDeath());
 }
 
+void CCharEntity::ApplyCurrentMoghancement()
+{
+    // Determines the current moghancement and sets it which applies the correct modifiers
+    std::bitset<512U>& moghancementKeyItems = keys.tables[1].keyList;
```
teschnei: unused variable?

src/map/entities/charentity.cpp
```diff
@@ -1651,6 +1658,340 @@ int32 CCharEntity::GetTimeRemainingUntilDeathHomepoint()
...
+    return m_moghancementID == moghancementID;
+}
+
+void CCharEntity::UpdateMoghancement()
+{
+    static constexpr uint8 containers[2] = { LOC_MOGSAFE, LOC_MOGSAFE2 };
+
+    // Add up all of the installed furniture auras
+    uint16 elements[8] = { 0 };
+    for (int containerIndex = 0; containerIndex < 2; ++containerIndex)
```
teschnei: same thing here with `containers`

src/map/entities/charentity.cpp
```diff
@@ -1651,6 +1658,340 @@ int32 CCharEntity::GetTimeRemainingUntilDeathHomepoint()
...
+{
+    return m_moghancementID == moghancementID;
+}
+
+void CCharEntity::UpdateMoghancement()
+{
+    static constexpr uint8 containers[2] = { LOC_MOGSAFE, LOC_MOGSAFE2 };
+
+    // Add up all of the installed furniture auras
+    uint16 elements[8] = { 0 };
```
teschnei: this also can be a `std::array`

src/map/items/item_furnishing.h
```diff
@@ -100,6 +100,67 @@ enum MOGHOUSE_AURA
     MOGLIFICATION_RESIST_CURSE         = 2855  // Slightly improves your resistance to curse effects
 };
 
+static const std::initializer_list<MOGHANCEMENT_TYPE> MoghancementList =
```
teschnei: i *really* don't like this - it's even only used in one single place (loading the moghancement). honestly, I think it'd just be better to store the current moghancement elsewhere in the db so we don't have to iterate all these keyitems on every load

teschnei:
in general: no c-style arrays please - they have a tendency to make bugs just by the volume of less experienced developers the code base sometimes gets, so I want to make it as foolproof as possible

src/map/packets/message_special.h
TeoTwawki: _suddenly remebers the textids used for abyssea cruor_

😅 Don't mind me

jmcmorris:
Sure, happy to move away from c style arrays. I honestly never use them myself but the places I looked here I saw them being used so thought that was the norm. Thanks for the feedback. I'll get that stuff fixed up.

teschnei:
It is a slowly moving effort to get everything better ;)

src/map/items/item_furnishing.h
jmcmorris: Did you have any places in mind? New column in chars table?

teschnei: chars_storage maybe? or chars, they both probably work fine

src/map/packets/message_special.h
jmcmorris: Ah ok, thanks for letting me know. This technically is always in the MH but I'll look at merging in current master and trying to use this mentioned `luautils::GetTextIDVariable().

teschnei:
Ah, but the moghouse is technically whichever zone you entered from

src/map/items/item_furnishing.h
TeoTwawki: as long as its not a char_vars I won't murder anybody :wink:

TeoTwawki:
@jmcmorris need a rebase due to a conflict

PR #5258, opened 2018-09-30
HELM rewrite

wrenffxi:
Big thanks to Wiggo for retail captures and information.

combined all four HELM activities into scripts/globals/helm.lua
tabled all the HELM data
drop rates taken from wiki when available
standardized settings
all point NPCs are now visible (previously only one was, except for mining)
point NPCs can share a location (verified by Wiggo)
removed redundant code for Olduum quest
fixed many ID-shifted point NPCs
moved point NPCs into global zones table
no more server variables


Wiggo32:
i think we're missing a few points here... all 6 "Excavation Point" NPC's are supposed to be up all at the same time and they rotate between about 10-15 points (or so) per zone... If you model excavation according to how Mining is done currently on dsp, it will be closer to retail... and I can get retail points for ya, I just haven't gotten them all yet.

wrenffxi:
That's very helpful, thanks.  My plan is to merge all the HELM activities into one global, since they seem to reuse a lot of code, and knowing that Excavation works similar to Mining 2.0 makes me more determined to do so.  I assume it's the same for logging and harvesting as well -- that all NPCs are up, and rotate among a larger pool of locations?

TeoTwawki:
Something like `globals\helm.lua` and porting id's to their respective zones id.lua's {yes please}

Mining excavation logging and harvesting all do pretty much the same thing, just different items involved. Clamming is the only thing that differs. And goldfish scooping (doesn't exist yet) is basically the old fishing system before minigames got involved. so the 4 "letter" activities that spell helm can all go to 1 script easily.

wrenffxi:
That's exactly what I'd planned to do with this PR originally, until I saw that Mining and Excavating had different systems (currently coded) for moving/appearing the points.  Now that I know they're the same on retail, I'll go back to that -- the only holdup is not knowing the pool of possible locations for all the activities (and no retail access to get 'em).

TeoTwawki:
what happened was somebody fixed mining and didn't touch the other file.

Wiggo32:
i have a LOT of retail points already logged... i'll get them to you if your plan is to fix all this crap... that was something i had on the back burner but there's just so much... but, yea, I can get you what all I have and you can make it look pretty cause I'm sure it's a mess at this point, albeit retail accurate...

wrenffxi:
There we go.  Updated original comment.  Ready for review.

Wiggo32:
Wow, that was super fast! One comment, I know it's gonna be a while before we get ALL the retail points (if ever) but, for the ones that have only 3 or so point locations to move between, maybe we should make something up (from guesstimating wiki maps) to add a few more. Otherwise, the points don't have many places to move about and the potential to mine/log/harvest the same spot for days without moving might exist.

TeoTwawki:
@wiggo if you want to pr some reasonable guestimates and we can correct the pos later, I'd be down for that

PR #5261, opened 2018-10-01
Conquest Teleporter fixes

nfaletra:
Conquest Teleporters were sending incorrect values for some of their event params.

I did some retail captures and with some sleuthing help from @zynjec (happy?), we discovered the correct values for the params.
* The first 4 params are region masks for each nation (including beastmen). If the nation owns that region, the bit is flipped OFF.
* The 5th param is a bit flag corresponding to the current nation/region (19th bit turned on for Sandy, etc.)
** My educated guess is it uses this bit to know which of the masks to check to get accurate messages to the player, i.e. "This region is no longer under our nations control."
* 6th param is always 0 in your home nation (the existing code for talking to conquest teleport NPCs outside of your home nation was correct)
* 7th param player level (already correct)
* 8th param is the list of regions that should populate the menu. Prior to this fix, it was always showing every region you had completed a supply run for (saved in the DB). It was not doing a level check to make sure you are high enough level to teleport to said region. I confirmed that retail DOES do this check and limits what regions the player can choose from based on current level.

I also renamed some functions to more accurately describe what they are doing and did some refactoring to clean it all up. If there are any styling preferences I missed, please let me know.


nfaletra:
Standby. I think the nation flag part is wrong. Fixing.

nfaletra:
Aight it's working all hunky dory now.

PR #5264, opened 2018-10-01
Fix GetTextIDVariable

nfaletra:
GetTextIDVariable needed updating after TextIDs were tabled


wrenffxi:
No matter how many times I read up on how to get LUA variables into C++, it doesn't click for me.

teschnei:
http://pgl.yoyo.org/luai/i/3.7+Functions+and+Types is *extremely* useful for figuring it out - all you really gotta know is that everything is implemented on a stack, and then from there it's just "x pops from stack/y pushes to stack"

PR #5266, opened 2018-10-03
Resolved issue with items traded for guild points not being consumed

### LINKED ISSUE #4559: Goldsmith Guild Point Exploit - Free Points!

jmcmorris:
There were two issues here with the code.
1. The correct signature is `confirmItem(itemID, amount)` but it was being passed in `i` which I guess was intended as the `slotID`. This lead to `confirmItem` never confirming any items.
2. `confirmItem` iterates over the entire trade container looking for `amount` of the provided `itemID`. The problem is that it was iterating over the trade container outside of that looking for each item. So if 2+ guild items were traded at the same time it would reconfirm the first one over and over.

The fix is to just add up the total number of items we want to confirm and then call that with the correct parameters. This fixes #4559.

Tested this with various trades with the goldsmith guild point npc.

### DIFF
--- a/scripts/globals/crafting.lua
+++ b/scripts/globals/crafting.lua
@@ -292,22 +292,22 @@ function unionRepresentativeTriggerFinish(player, option, target, guildID, curre
 end
 
 function unionRepresentativeTrade(player, npc, trade, csid, guildID)
-    local gpItem, remainingPoints = player:getCurrentGPItem(guildID);
+    local gpItem, remainingPoints = player:getCurrentGPItem(guildID)
     if (player:getVar('[GUILD]currentGuild') - 1 == guildID) then
         if remainingPoints == 0 then
-            player:messageText(npc, NO_MORE_GP_ELIGIBLE);
+            player:messageText(npc, NO_MORE_GP_ELIGIBLE)
         else
-            local totalPoints = 0;
-            for i=0,8,1 do
-                local items, points = player:addGuildPoints(guildID,i)
+            local totalPoints = 0
+            for i = 0, 8 do
+                local items, points = player:addGuildPoints(guildID, i)
                 if items ~= 0 and points ~= 0 then
-                    totalPoints = totalPoints + points;
-                    trade:confirmItem(i, items);
+                    totalPoints = totalPoints + points
+                    trade:confirmSlot(i, items)
                 end
             end
             if (totalPoints > 0) then
-                player:confirmTrade();
-                player:startEvent(csid,totalPoints);
+                player:confirmTrade()
+                player:startEvent(csid,totalPoints)
             end
         end
     end
--- a/scripts/globals/crafting.lua
+++ b/scripts/globals/crafting.lua
@@ -297,15 +297,16 @@ function unionRepresentativeTrade(player, npc, trade, csid, guildID)
         if remainingPoints == 0 then
             player:messageText(npc, NO_MORE_GP_ELIGIBLE);
         else
+            local totalItems = 0;
             local totalPoints = 0;
             for i=0,8,1 do
                 local items, points = player:addGuildPoints(guildID,i)
-                if items ~= 0 and points ~= 0 then
-                    totalPoints = totalPoints + points;
-                    trade:confirmItem(i, items);
-                end
+                totalItems = totalItems + items;
+                totalPoints = totalPoints + points;
             end
-            if (totalPoints > 0) then
+
+            if (totalItems > 0 and totalPoints > 0) then
+                trade:confirmItem(gpItem, totalItems);
                 player:confirmTrade();
                 player:startEvent(csid,totalPoints);
             end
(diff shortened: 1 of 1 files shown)


nfaletra:
Trading more items than points remaining will consume the extra items still.

Example:

Bone Ring gives 302 guild points.
I am allowed to receive 606 more guild points.
Trade 4 Bone Rings and get 606 guild points.
All 4 Rings are consumed despite only 3 being required to hit the cap of 606 (2 @ 302gp each = 604, the 3rd @ 2gp to hit 606 cap).

Is this how it behaves in retail?

Obviously, this is still better than not consuming the item, but I don't remember how retail handled this situation.

scripts/globals/crafting.lua
```diff
@@ -297,15 +297,16 @@ function unionRepresentativeTrade(player, npc, trade, csid, guildID)
         if remainingPoints == 0 then
             player:messageText(npc, NO_MORE_GP_ELIGIBLE);
         else
+            local totalItems = 0;
```
nfaletra: Can you format for the new standards? Just the function you are working in is fine, rather than the whole file.

Remove semicolons, space out function variables etc.

scripts/globals/crafting.lua
```diff
@@ -297,15 +297,16 @@ function unionRepresentativeTrade(player, npc, trade, csid, guildID)
...
             local totalPoints = 0;
             for i=0,8,1 do
```
nfaletra: `for i = 0, 8 do
end`

the trailing 1 is unnecessary. It's the default increment.

TeoTwawki:
Retail will consume items for partial points
example:  
item worth 600 but you only have 300 to go before cap, so that last item only gets you 300.
but trade that contains multiple of that 600 point item when you have 300 to go will only consume one

nfaletra:
I know how the partial points work. But in my example, only 3/4 items give points. First 2 were full points, 3rd was partial, and it consumed the 4th willy-nilly. That's the edge case I'm wondering about.

TeoTwawki:
I thought I was clearer than I was I guess: it *won't* consume beyond the cap unless its a partial situation like I described.

nfaletra:
k cool. I'm thinking that should be a separate ticket/PR?

TeoTwawki:
unless he wants to tackle it at the same time or someone from dev wants to tack it onto this, sure.

teschnei:
@nfaletra is that from memory or tested on this branch? player:AddGuildPoints is supposed to return the amount of items consumed so that confirmItem will only confirm the amount of items that get consumed.  There could be a bug in AddGuildPoints, but it was written to handle that case

teschnei:
Can you test this with HQ items, and with mixing HQ items with NQ items?  I think it will probably not work

nfaletra:
The items ~= and points ~= check was removed. That could be the cause.

nfaletra:
A mix of NQ/HQ items works fine.
Putting that check back in prevents items that give 0 points (over the max allowed) from being consumed.
HQ items specifically are not being consumed at all. Still looking into that.

teschnei:
> A mix of NQ/HQ items works fine.
Did it actually consume the HQ items?

It's because the confirmItem only confirms the "gpitem" instead of the actual item in the slot.

teschnei:
I think the proper change is to keep the old way, but just fix the confirmItem parameter to get the itemID from the slot instead of incorrectly passing the slot ID.

jmcmorris:
I will admit I did not consider HQ items which definitely changes things. I looked at this is the only place we use `player:addGuildPoints()`. How do you feel about me making `addGuildPoints` also confirm the trade as well? So the calling function would look like this:

```lua
function unionRepresentativeTrade(player, npc, trade, csid, guildID)
    local gpItem, remainingPoints = player:getCurrentGPItem(guildID);
    if (player:getVar('[GUILD]currentGuild') - 1 == guildID) then
        if remainingPoints == 0 then
            player:messageText(npc, NO_MORE_GP_ELIGIBLE);
        else
            local points = player:addGuildPoints(guildID)
            if (points > 0) then
                player:startEvent(csid,totalPoints)
            end
        end
    end
end

teschnei:
After talking with @nfaletra, the proper fix is probably just changing `confirmItem` to `confirmSlot`

jmcmorris:
I agree using `confirmSlot` is definitely the simplest change. Is there a reason why it is preferred to have this specialized function `addGuildPoints` only do half of the transaction and require the script to finish it? This feels wrong to me.

teschnei:
I dunno, it does feel wrong to me to be doing that kind of stuff in the core (confirming trades, etc).  obviously I'd rather have it *all* be in script but we can't really do that with guild stuff all in the core.  for now, I'd rather keep as much in script as possible, even if it's a specialized function only used in one place.

PR #5268, opened 2018-10-03
sagheera and wilhelm rewrite

wrenffxi:
removed armor_upgrade global (none of its contents were shared by more than one npc)
neatened the limbus and af/relic upgrade tables in their respective npcs
added cheat protection
added missing messages to Sagheera
added missing parameter to Sagheera's menu event (case when player is waiting for AF but it's not yet ready)


TeoTwawki:
```
file deleted
```
wren you give me the bestest presents :heart:

PR #5269, opened 2018-10-03
added !gotoid command

wrenffxi:
Dunno if this interests you, but I've found it really helpful for both debugging code and helping players with things.

```!gotoid <mob or NPC id>```

Send GM to mob or NPC if it can be found.
If the entity is not up, send to last known coordinates.

### DIFF
--- a/scripts/commands/gotoid.lua
+++ b/scripts/commands/gotoid.lua
@@ -1,0 +1,71 @@
+---------------------------------------------------------------------------------------------------
+-- func: gotoid
+-- desc: Go to given mob or npc ID
+---------------------------------------------------------------------------------------------------
+require("scripts/globals/status")
+
+cmdprops =
+{
+    permission = 1,
+    parameters = "i"
+}
+
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!gotoid <mobId|npcId>")
+end
+
+function onTrigger(player, target)
+
+    -- validate npc
+    if not target or target == 0 then
+        error(player, "You must enter a mob or NPC ID.")
+        return
+    end
+
+    -- is entity up?
+    local isUp = false
+    local targ = GetNPCByID(target)
+    local pos0 = false
+    
+    if targ then
+        local pos = targ:getPos()
+        pos0 = (pos.x == 0 and pos.y == 0 and pos.z == 0)
+        if targ:getStatus() == dsp.status.NORMAL and not pos0 then
+            isUp = true
+        end
+    else
+        targ = GetMobByID(target)
+        if targ then
+            local pos = targ:getPos()
+            pos0 = (pos.x == 0 and pos.y == 0 and pos.z == 0)
+            if targ:isSpawned() and not pos0 then
+                isUp = true
+            end
+        end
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by whasf** [comment]
Does this work on targets that are on a different game server?

**Comment by TeoTwawki** [comment]
I like

**Comment by KnowOne134** [comment]
Ive made this on my server also. But doesnt work cross servers

**Comment by wrenffxi** [comment]
I don't know much about multi-game server setups, but I'm guessing this would not work across them.

**Comment by KnowOne134** [comment]
Not much does lol

**Comment by TeoTwawki** [comment]
would have to fetch pos from database or else use message server for it to go across servers.

**Comment by takhlaq** [comment]
can you have it use CLuaBaseEntity::GoToEntity

**Comment by wrenffxi** [comment]
Updated.

**Comment by KnowOne134** [comment]
Awesome. Now if we could get nexus cape on board to find party leader across server maps also

**Comment by TeoTwawki** [comment]
> Awesome. Now if we could get nexus cape on board to find party leader across server maps also

yeah problem there being getting the name across servers, I think. I used gotoplayer for it. will need to try getting the id from the party data and passing it to gotoentity instead but haven't had the time to look into that again. maybe somebody really awesome could take that on so we can close #5088 *hint hint*

File: scripts/commands/gotoid.lua [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+        -- display message
+        if isUp then
+            player:PrintToPlayer(string.format("Going to %s (%i).", targ:getName(), targ:getID()))
+        else
+            player:PrintToPlayer(string.format("%s (%i) is not currently up. Going to last known coordinates.", targ:getName(), targ:getID()))
+        end
+
+        -- half a second later, go.  this delay gives time for previous message to appear
+        player:timer(500, function(player)
+            player:setPos(targ:getXPos(), targ:getYPos(), targ:getZPos(), targ:getRotPos(), gotoZone)
```
TeoTwawki: `player:goToEntity(target)` actually initiates the position change for you core side already
https://github.com/DarkstarProject/darkstar/blob/09b3ddfe5f24190203a2ceb225c479bcc699d045/src/map/message.cpp#L439

wrenffxi: target is the ID (it's the number passed in by the player).  targ is the entity.

TeoTwawki: Ok. Just need to rework your message placement and kill the setpos then

wrenffxi: the messages and setpos are there purposefully.  If it can't locate the entity ID on current map server, it uses goToEntity.  If it finds them on the current map server, it setPos'es you to the x/y/z or the x/y/z/zone if in different zone.  If the mob/npc was spawned but currently isn't, it sends you to last known coordinates.  If the entity's coordinates are 0/0/0 it doesn't send you there.

TeoTwawki: My bad I totally misunderstood what you were doing there, didn't realize you wantes to avoid a load screen for same zone and avoid pos 0/0/0, I sort of skim read the changes coz cell phone I'm sorry

PR #5274, opened 2018-10-04
Begin migrating non currency things like teleports out of char_point

TeoTwawki:
Migrations instructions:
you copy from the old table to the new before deleting the columns. It ain't hard to query.

Testing: outpost unlock and use works, I assume the rest does as well since its such a simple change.


RunAwayDSP:
so making a new table to do same thing because "points" can be broken down into different categories?

KnowOne134:
Whats benefit of this besides make char table smaller but adding another table to import?

TeoTwawki:
char_points has gotten huge, and we have some things in char_vars I want to see moved here later as well.

teschnei:
also, char_points should be directly analogous with the currency packets - the things moving out in this PR aren't in them

TeoTwawki:
Oddly enough, SaveCharPoints saved only things there were not at all points. Unless you take Points to mean "Locations" as in "warp points" and exclude the obviously point and currency things.

hilts-vaughan:
Newcomer just reading some PRs to get used to how things work...

Is there a reason this project doesn't use something like Liquibase to wrap those migrations?

TeoTwawki:
well honestly we shouldn't be using raw db table dumps in the first place, but mostly lack of anyone familiar with things like that. I'm not at all fond of our existing migrations.

In a perfect world we'd be generating the database tables and any missing/new fields at first run.

takhlaq:
leaving this til we decide on how to handle migrations

TeoTwawki:
> leaving this til we decide on how to handle migrations 

in this particular case, its not worth writing a migration. you can latterly copy you existing table rename it and then drop the unused fields. (or leave them - the query won't care)

wrenffxi:
This is the DB update procedure I sent to my server's admin

```
after the typical database build (which doesn't touch any of the sensitive tables),

1. build the char_unlocks table by running once: sql/char_unlocks.sql

2. move columns from char_points into char_unlocks:

   INSERT INTO char_unlocks (charid, sandoria_supply, bastok_supply, windurst_supply, runic_portal, maw, past_sandoria_tp, past_bastok_tp, past_windurst_tp)
   SELECT charid, sandoria_supply, bastok_supply, windurst_supply, runic_portal, maw, past_sandoria_tp, past_bastok_tp, past_windurst_tp FROM char_points
   
3. remove the no-longer-needed columns from char_points:

   ALTER TABLE char_points
   DROP COLUMN sandoria_supply,
   DROP COLUMN bastok_supply,
   DROP COLUMN windurst_supply,
   DROP COLUMN runic_portal,
   DROP COLUMN maw,
   DROP COLUMN past_sandoria_tp,
   DROP COLUMN past_bastok_tp,
   DROP COLUMN past_windurst_tp
```

PR #5276, opened 2018-10-05
Resolve issue with item in inventory becoming unusable

jmcmorris:
If a player rests while using an item it will clear the AI state leaving the item in a locked state since it was not technically `m_interrupted`. The fix here is to treat the item as being interrupted if we are cleaning up the state before it finished using the item.


jmcmorris:
@nfaletra Thanks, I should get in the habit of looking for these fixes in #issues.

PR #5278, opened 2018-10-05
Things of Valor rewrite

wrenffxi:
* Combined fieldsofvalor, groundsofvalor, regimeinfo, and regimereward into one global.
* Tabled all the global vars.
* Moved textIDs into zones table.
* Fixed an issue where players would get locked in cutscene when examining a large page number from book.
* Added "not enough tabs" message to books.
* Made many effect removals silent (when effects are being immediately reapplied).

This touches a ton of files, but other than globals/valor.lua, all the edits fall into one of three categories:

1. Field Manuals and Grounds Tomes.  Removed the hard-coded zone and regime data from the book luas.
2. Mobs who are regime targets.  Added one extra parameter to distinguish fields from grounds.
3. IDs.lua in regime zones.  Added the REGIME_REGISTERED textId, from which other messages are offset.

If you look at a couple of each, you've seen it all; same changes repeated in all the other files.


TeoTwawki:
Jeeze wren, and its not even my birthday or xmas or nothing!

One thing I request on this is naming the global something other than valor since it will eventually be expanded to handle hunts regime's and dominion regimes as well. Those will use nearly the same code, just like grounds and fields did.

TeoTwawki:
in this case all the objects we're interacting with will be in the same server instance so I think we're good whasf.

PR #5279, opened 2018-10-05
Old school barspell duration

WiggoCaptures:
Adds the calculation for barspell duration before the update.
https://www.bg-wiki.com/bg/Category:Barspell

Also, adds the setting for a 'generalized' old school switch which I've been starting to covert things over to use.

Also, in an attempt to organize the settings a bit we could clump the 'old school' related stuff into its own area? maybe? and another thought... should these settings be tabled?

### DIFF
--- a/scripts/globals/settings.lua
+++ b/scripts/globals/settings.lua
@@ -28,6 +28,9 @@ ENABLE_ROV     = 0;
 
 ENABLE_VOIDWATCH = 0; -- Not an expansion, but has its own storyline.
 
+-- Era "Old School" related settings
+OLD_SCHOOL_ENABLED = false -- General wide range Old School setting. Default: false (current retail)
+
 -- FIELDS OF VALOR/Grounds of Valor settings
 ENABLE_FIELD_MANUALS  = 1; -- Enables Fields of Valor
 ENABLE_GROUNDS_TOMES  = 1; -- Enables Grounds of Valor
--- a/scripts/globals/spells/barspell.lua
+++ b/scripts/globals/spells/barspell.lua
@@ -1,8 +1,11 @@
 -----------------------------------------
 -- Implementation of Bar-spells
 -----------------------------------------
-require("scripts/globals/magic")
+require("scripts/globals/settings")
 require("scripts/globals/status")
+require("scripts/globals/magic")
+require("scripts/globals/utils")
+-----------------------------------------
 
 function calculateBarspellPower(caster,enhanceSkill)
     local meritBonus = caster:getMerit(dsp.merit.BAR_SPELL_EFFECT)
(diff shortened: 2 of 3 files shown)


godmode:
@zynjec My 2 cents FWIW: I’ve structured Edens code contribution style to model our PRs to our private repo based on the fact that we will eventually submit back to DSP. While some things are small and could be left out of master it would kind of become a guessing game if one setting should be left out. Ideally, it would be nice to have a more maintainable settings system though. We’d be happy to look into an overhaul in the future once we’ve got a few things under wraps on our new server. Obviously we don’t get to vote on these things but I’d like to see more settings in one form or another.

zynjec:
There is simply too many moving parts, targeted "era's", personal preference, etc to really make any scaleable system.

There is already an insane amount of bloat from previously allowed non-retail stuff, and "old school" settings, and at a certain point I think it's best for the project to focus on one thing, and have people tailor it to their preference.

TeoTwawki:
this is why I wish we had some sort of modular system to make ppl home brewing settings simpler, without putting them in the main project. but we don't, so ppl need to do their own alterations for the little things and we have to get pickier about what things we make into settings.

like, we could break the the current settings into several categorized files and put all the oldschool stuff into its own file (have talked about doing this to settinsg.lua before) buuuuut that would still require maintaining all the bloat that gets added to the places the setting(s) get used in.

We actually did at one point have a setting for some similar things durations before. They got removed. To be fair, most were going unused by then because refactors had removed them or they were never even implemented beyond placeholders. https://github.com/DarkstarProject/darkstar/commit/94ce635802dc3a7259755bd47f4aa109c429514c

ibm2431:
Been going through my notification backlog. Appending a note that if DSP decides not to implement non-current retail behavior, `OLD_SCHOOL_ENABLED` currently exists in the code base, added in #5219. So if this is declined, the reference should be removed there as well.

Personal opinion time: QoL and balance changes happen for a reason. 

But _if_ DSP starts branching code depending on a setting (ie: baked in, instead of a module system), the checks should be against a _date_ instead of one true/false flag. "Old school" is impossible to define. What'd be better is to determine which version update (or even year/month) resulted in a given change, and do a `if SERVER_ERA_DATE < 2012_01 then doOldBehavior()`. Then server operators can decide "when" they want their server to be.

TeoTwawki:
> "Old school" is impossible to define

that's why its bad to try and put it all under one setting. its to ambiguous when that may mean different things to different people. I'd rather see an old school file containing all those settings to adjust thinsg that change din era than to gate content base don date. My own server has a mix of different things from different era's and I may even wind up enabling wotg before toau, because blue mage is absurdly unbalanced from the inaccuracies of its spell scripts and nobody seems interested in working on that.

zynjec:
Going to close this for now, while we figure out if we even want era-settings in master, or how we'd approach it in another branch.

@ibm2431 Thanks for pointing that out, we can see about cleaning it up in another PR.

PR #5281, opened 2018-10-06
scripts: Fixed an issue where the WHM AF1 weapon NM spawning even during the day

hilts-vaughan:
I think the ??? is supposed to respawn in most cases technically but it does not, at least not  while I was testing. It does not do this on the Nasomi server either (which may have diverged) 

This adds a check at least so that we can prevent it from spawning even if the ??? is there.

First time contributor, so open to guidance.

### DIFF
--- a/scripts/zones/Valkurm_Dunes/npcs/qm1.lua
+++ b/scripts/zones/Valkurm_Dunes/npcs/qm1.lua
@@ -13,14 +13,14 @@ function onTrade(player,npc,trade)
 end;
 
 function onTrigger(player,npc)
+    local TOTD = VanadielTOTD();
 
-    if (player:getFreeSlotsCount() > 0 and player:hasItem(503) == false) then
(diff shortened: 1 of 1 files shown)


TeoTwawki:
>This adds a check at least so that we can prevent it from spawning even if the ??? is there.

You aren't preventing any spawning of anything here. you are preventing an item acquisition.

What is the actual problem you intended to fix?

TeoTwawki:
I don't think you are even working with the correct script here.

hilts-vaughan:
This is actually a brain failure, this is the wrong script. There is another here iirc that is used for Messenger from the Beyond -- and this is for something else (see the header)

The issue _should_ have been that the NPC cannot be spawned during the day but it can be right now. This is what I get for trying to "clean up" after.

TeoTwawki:
I think this is where you want to look:
https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Valkurm_Dunes/npcs/qm2.lua#L14

feel free to open new PR and don't beat yourself up

Issue #5069, opened 2018-07-07
A Boy's Nightmare (pld AF2)


rheinecatsin:
DemiUrge server.

Fish wouldn't spawn by fishing. 

GM had it spawn via console command. (even though that issue was supposedly resolved last year)

Killed it, dropped loot.

Traded loot to Albeche. He oriented me to fisherman in Selbina.

Zaldon  won't accept loot. Talked with him or others before. Tried to trade on same job as I killed NM or as PLD. Nothing happens if I try to trade him the Odontotyrannus.

whasf:
Should open this on demiurge's GitHub, this is for darkstar itself

TeoTwawki:
I asked them to because its not server specific

whasf:
oh, well then! :)

rheinecatsin:
worked, thanks.

TeoTwawki:
```lua
function onTrade(player,npc,trade)
    -- "Flyers for Regine" conditional script
    local FlyerForRegine = player:getQuestStatus(SANDORIA,FLYERS_FOR_REGINE);

    if (player:getQuestStatus(SANDORIA, FATHER_AND_SON) == QUEST_COMPLETED and player:getVar("returnedAilbecheRod") ~= 1) then
        if (trade:hasItemQty(17391,1) == true and trade:getItemCount() == 1) then
            player:startEvent(61); -- Finish Quest "Father and Son" (part2) (trading fishing rod)
        end
    end

    if (player:getVar("aBoysDreamCS") >= 3) then
        if (trade:hasItemQty(17001,1) == true and trade:getItemCount() == 1 and player:hasItem(4562) == false) then
            player:startEvent(15); -- During Quest "A Boy's Dream" (trading bug) madame ?
        elseif (trade:hasItemQty(4562,1) == true and trade:getItemCount() == 1) then
            player:startEvent(47); -- During Quest "A Boy's Dream" (trading odontotyrannus)
        end
    end

    if (FlyerForRegine == 1) then
        local count = trade:getItemCount();
        local MagicFlyer = trade:hasItemQty(532,1);
        if (MagicFlyer == true and count == 1) then
            player:messageSpecial(FLYER_REFUSED);
        end
    end
end;
```
^ if someone wants to try and rework this, there it is.

TeoTwawki:
I especially dislike how this continues to execute code to look at more conditionals after a trade is triggering one (and you obviously can't have 2 different trades at once..)

[...]

PR #5282, opened 2018-10-06
quest: A Boy's Dream fixes where having the fish and bug at the same …

hilts-vaughan:
There are a few issues with the quest. See below, this is just  a WIP PR so that I can get some feedback before I go ahead and do some cleanup while I setup DSP. 

**I still need to setup DSP locally to test, so doing that right after this. The problems I found are...**

1. If you have the fish and the bug in your inventory at the same time, they will not accept the bug. But the bug CS is needed in order for the fish CS to count for state credit. Otherwise, you can trade the fish but you end up in this weird state where you are told to go to Zaldon and you are unable to actually trade the fish. (https://github.com/DarkstarProject/darkstar/issues/5069)

2. There were some greater than checks which made no sense, since getting into that state would indicate some  kind of optional step but none of those steps were optional.

I cleaned up a few of these things and serialized the bug and the fish portion so you must do both. If one is truly optional or the ordering is not important, we can rejig it. I do not have a retail subscription so I cannot say one way or the other if this is right. I asked in Discord to see if anyone is able to confirm or deny.

### DIFF
--- a/scripts/zones/Northern_San_dOria/npcs/Ailbeche.lua
+++ b/scripts/zones/Northern_San_dOria/npcs/Ailbeche.lua
@@ -11,26 +11,39 @@ require("scripts/globals/settings");
 require("scripts/globals/keyitems");
 require("scripts/globals/titles");
 require("scripts/globals/quests");
+
 require("scripts/globals/shop");
 -----------------------------------
 
-function onTrade(player,npc,trade)
-    -- "Flyers for Regine" conditional script
-    local FlyerForRegine = player:getQuestStatus(SANDORIA,FLYERS_FOR_REGINE);
+-- Helper constants in order to keep it straight about what part
+-- of the quest we are working on
+BOYS_DREAM_WILL_BEGIN_ACCEPTING_BUG = 3
+BOYS_DREAM_HAS_TRADED_BUG = 4
+BOYS_DREAM_VISIT_ZALDON_FOR_BOOTS_TRADED_FISH = 5
 
+function onTrade(player,npc,trade)
     if (player:getQuestStatus(SANDORIA, FATHER_AND_SON) == QUEST_COMPLETED and player:getVar("returnedAilbecheRod") ~= 1) then
         if (trade:hasItemQty(17391,1) == true and trade:getItemCount() == 1) then
             player:startEvent(61); -- Finish Quest "Father and Son" (part2) (trading fishing rod)
+            return
         end
     end
 
-    if (player:getVar("aBoysDreamCS") >= 3) then
-        if (trade:hasItemQty(17001,1) == true and trade:getItemCount() == 1 and player:hasItem(4562) == false) then
+    aBoysDreamCS = player:getVar("aBoysDreamCS");
+    
+    if (aBoysDreamCS >= BOYS_DREAM_WILL_BEGIN_ACCEPTING_BUG) then
+        -- If you have been told to visit Zaldon already (traded in the fish), your chance to get the CS is over with now
+        if (aBoysDreamCS == BOYS_DREAM_WILL_BEGIN_ACCEPTING_BUG and trade:hasItemQty(17001,1) == true and trade:getItemCount() == 1) then
             player:startEvent(15); -- During Quest "A Boy's Dream" (trading bug) madame ?
-        elseif (trade:hasItemQty(4562,1) == true and trade:getItemCount() == 1) then
+            return
+        elseif (aBoysDreamCS == BOYS_DREAM_HAS_TRADED_BUG and trade:hasItemQty(4562,1) == true and trade:getItemCount() == 1) then
             player:startEvent(47); -- During Quest "A Boy's Dream" (trading odontotyrannus)
+            return
         end
     end
+    
+    -- "Flyers for Regine" conditional script
+    local FlyerForRegine = player:getQuestStatus(SANDORIA,FLYERS_FOR_REGINE);
 
     if (FlyerForRegine == 1) then
         local count = trade:getItemCount();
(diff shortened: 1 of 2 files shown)


scripts/zones/Northern_San_dOria/npcs/Ailbeche.lua
```diff
@@ -11,26 +11,39 @@ require("scripts/globals/settings");
 require("scripts/globals/keyitems");
 require("scripts/globals/titles");
 require("scripts/globals/quests");
+
```
TeoTwawki: Don't do that

scripts/zones/Northern_San_dOria/npcs/Ailbeche.lua
```diff
@@ -11,26 +11,39 @@ require("scripts/globals/settings");
...
 require("scripts/globals/shop");
 -----------------------------------
 
-function onTrade(player,npc,trade)
-    -- "Flyers for Regine" conditional script
-    local FlyerForRegine = player:getQuestStatus(SANDORIA,FLYERS_FOR_REGINE);
+-- Helper constants in order to keep it straight about what part
+-- of the quest we are working on
+BOYS_DREAM_WILL_BEGIN_ACCEPTING_BUG = 3
```
TeoTwawki: 1. Global vars are bad.
2. Please use comments to explain the steps when needed, rather than turning the sql variable value into multiple lua variables..

scripts/zones/Northern_San_dOria/npcs/Ailbeche.lua
```diff
@@ -11,26 +11,39 @@ require("scripts/globals/settings");
...
+BOYS_DREAM_HAS_TRADED_BUG = 4
+BOYS_DREAM_VISIT_ZALDON_FOR_BOOTS_TRADED_FISH = 5
 
+function onTrade(player,npc,trade)
     if (player:getQuestStatus(SANDORIA, FATHER_AND_SON) == QUEST_COMPLETED and player:getVar("returnedAilbecheRod") ~= 1) then
         if (trade:hasItemQty(17391,1) == true and trade:getItemCount() == 1) then
             player:startEvent(61); -- Finish Quest "Father and Son" (part2) (trading fishing rod)
+            return
```
TeoTwawki: there shouldn't be a return here it will kill the event you just started and put an error in the console log for the map server

scripts/zones/Northern_San_dOria/npcs/Ailbeche.lua
```diff
@@ -11,26 +11,39 @@ require("scripts/globals/settings");
...
-    if (player:getVar("aBoysDreamCS") >= 3) then
-        if (trade:hasItemQty(17001,1) == true and trade:getItemCount() == 1 and player:hasItem(4562) == false) then
+    aBoysDreamCS = player:getVar("aBoysDreamCS");
+    
+    if (aBoysDreamCS >= BOYS_DREAM_WILL_BEGIN_ACCEPTING_BUG) then
+        -- If you have been told to visit Zaldon already (traded in the fish), your chance to get the CS is over with now
+        if (aBoysDreamCS == BOYS_DREAM_WILL_BEGIN_ACCEPTING_BUG and trade:hasItemQty(17001,1) == true and trade:getItemCount() == 1) then
             player:startEvent(15); -- During Quest "A Boy's Dream" (trading bug) madame ?
-        elseif (trade:hasItemQty(4562,1) == true and trade:getItemCount() == 1) then
+            return
```
TeoTwawki: again, there shouldn't be a return here

scripts/zones/Northern_San_dOria/npcs/Ailbeche.lua
```diff
@@ -11,26 +11,39 @@ require("scripts/globals/settings");
...
+        elseif (aBoysDreamCS == BOYS_DREAM_HAS_TRADED_BUG and trade:hasItemQty(4562,1) == true and trade:getItemCount() == 1) then
             player:startEvent(47); -- During Quest "A Boy's Dream" (trading odontotyrannus)
+            return
```
TeoTwawki: you get the idea

hilts-vaughan:
Great, thanks for the feedback. I just got DSP running yesterday so I'll be able to actually test and catch some of this stuff.

TeoTwawki:
We may need to fix the quest rewrite branch or retarget to master to test it. quest rewrite looks like it has broken a global file atm.

scripts/zones/Northern_San_dOria/npcs/Ailbeche.lua
hilts-vaughan: Is there a reason? The variables are more self describing. The fact that they're globals sucks but otherwise, it's definitely easier and less redundant, no?

That being said, there is consistency to value.

scripts/zones/Northern_San_dOria/npcs/Ailbeche.lua
hilts-vaughan: Why is this the case then? Judging by how it looks, one would think it's an async operation.

Is there a place in the C++ code I can read to understand the flow better?

takhlaq:
ye though would be nice to get quests in the new format once i get an example up

hilts-vaughan:
What's the status of that branch? Where should I be putting stuff?

TeoTwawki:
-nudge-
@takhlaq 
what does he need to do here?

@hilts-vaughan 
are you still around for this?

hilts-vaughan:
I am here. Just tell me what I need to do.

hilts-vaughan:
Where am I putting this again, though? Master or the  rewrite branch? Can someone tell me? :)

takhlaq:
master

Bizzle-Dapp:
Was this ever fixed? Interested as one of my players is having issues fishing up Odontotyrannus. 
I can't seem to find any other merged branch relating to this quest, everything seems to be closed without a clear reasoning...?

PR #5283, opened 2018-10-06
scripts: Fixed an issue where the WHM AF1 weapon NM  appeared at night

hilts-vaughan:
I think the ??? is supposed to despawn at night in most cases technically but it does not, at least not while I was testing. It does not do this on the Nasomi server either (which may have diverged)

This adds a check at least so that we can prevent it from spawning the NM even if the ??? is there.

First time contributor, so open to guidance.

### DIFF
--- a/scripts/zones/Valkurm_Dunes/npcs/qm2.lua
+++ b/scripts/zones/Valkurm_Dunes/npcs/qm2.lua
@@ -14,9 +14,10 @@ function onTrade(player,npc,trade)
 end;
 
 function onTrigger(player,npc)
-    local TOTD = VanadielTOTD();
-
-    if (TOTD == dsp.time.NIGHT and not GetMobByID(MARCHELUTE):isSpawned() and player:getQuestStatus(SANDORIA,MESSENGER_FROM_BEYOND) == QUEST_ACCEPTED and not player:hasItem(1096)) then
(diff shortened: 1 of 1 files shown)


hilts-vaughan:
OK, this should be the right file. Setting up to test as well.

scripts/zones/Valkurm_Dunes/npcs/qm2.lua
```diff
@@ -14,7 +14,10 @@ function onTrade(player,npc,trade)
 end;
 
 function onTrigger(player,npc)
-    if (not GetMobByID(MARCHELUTE):isSpawned() and player:getQuestStatus(SANDORIA,MESSENGER_FROM_BEYOND) == QUEST_ACCEPTED and not player:hasItem(1096)) then
+    local TOTD = VanadielTOTD();
+    local isNight = TOTD == dsp.time.NIGHT
+    
+    if (isNight and not GetMobByID(MARCHELUTE):isSpawned() and player:getQuestStatus(SANDORIA,MESSENGER_FROM_BEYOND) == QUEST_ACCEPTED and not player:hasItem(1096)) then
```
wrenffxi: ```if VanadielTOTD() == dsp.time.NIGHT``` works and is just as intuitive to read. No need to make extra variables that are only used once.

hilts-vaughan: Sure, fixed.

PR #5285, opened 2018-10-06
Update Kreutzet Weather Spawn Logic

WiggoCaptures:
- Cleaned up the files with new style changes
- Updated tabled vars so they now work
- Updated weather spawn logic so the NM only spawns while wind weather is active and it's window is open. Achieved by adding a "cooldown" LocalVar which checks to make sure enough time has passed and a DisallowRespawn flag to make sure the mob doesn't accidently pop with no weather and then immediately despawn resetting its cooldown to 9-12 hrs.

ref: http://ffxiclopedia.wikia.com/wiki/Kreutzet

### DIFF
--- a/scripts/zones/Cape_Teriggan/Zone.lua
+++ b/scripts/zones/Cape_Teriggan/Zone.lua
@@ -4,56 +4,59 @@
 --
 -----------------------------------
 local ID = require("scripts/zones/Cape_Teriggan/IDs")
-require("scripts/globals/icanheararainbow");
-require("scripts/globals/conquest");
-require("scripts/globals/weather");
-require("scripts/globals/zone");
+-----------------------------------
+require("scripts/globals/icanheararainbow")
+require("scripts/globals/conquest")
+require("scripts/globals/weather")
+require("scripts/globals/zone")
 -----------------------------------
 
 function onInitialize(zone)
-    UpdateNMSpawnPoint(ID.mob.KREUTZET);
-    GetMobByID(ID.mob.KREUTZET):setRespawnTime(math.random(900, 10800));
+    UpdateNMSpawnPoint(KREUTZET)
+    GetMobByID(KREUTZET):setRespawnTime(math.random(32400,43200)) -- 9 to 12 hours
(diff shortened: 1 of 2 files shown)


scripts/zones/Cape_Teriggan/Zone.lua
```diff
@@ -4,56 +4,59 @@
...
     end
-end;
+end
 
 function onZoneWeatherChange(weather)
-    if (GetMobAction(ID.mob.KREUTZET) == 24 and (weather == dsp.weather.WIND or weather == dsp.weather.GALES)) then
-        SpawnMob(ID.mob.KREUTZET); -- Kreutzet
-    elseif (GetMobAction(ID.mob.KREUTZET) == 16 and (weather ~= dsp.weather.WIND and weather ~= dsp.weather.GALES)) then
-        DespawnMob(ID.mob.KREUTZET);
+    if GetMobAction(ID.mob.KREUTZET) == 0 and os.time() > GetMobByID(ID.mob.KREUTZET):getLocalVar("cooldown") and
```
TeoTwawki: we should replace `GetMobAction()` checks (it is deprecated) with `isSpawned()` checks. You'll have to do a `GetMobByID()` to use it though.

scripts/zones/Cape_Teriggan/Zone.lua
```diff
@@ -14,7 +14,7 @@ require("scripts/globals/zone")
 function onInitialize(zone)
     UpdateNMSpawnPoint(ID.mob.KREUTZET)
     GetMobByID(ID.mob.KREUTZET):setRespawnTime(math.random(32400,43200)) -- 9 to 12 hours
```
TeoTwawki: This is a good time to go ahead and make qa local variable for this so that the core doesn't have to do the lookup multiple times here.

```lua
local mob = GetMobByID(ID.mob.KREUTZET)
```

Wiggo32: I did it in both spots... that's good, i need to start learning to pay attention to performance now.

PR #5286, opened 2018-10-06
nuke deprecated functions

TeoTwawki:
Hey look, I actually remembered to finish something I started!

canUseMisc has now taken over for the multiple canUseThing functions. one function to rule all the misc flags.


whasf:
why remove the not, just add another one

TeoTwawki:
its not not right?

PR #5287, opened 2018-10-06
level cap/sync should be taken into account when charming mobs

WiggoCaptures:
When doing things like a lvl cap 20 bcnm we shouldn't be comparing charm lvl as if we are 75 to calculate charm chance.

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -4098,9 +4098,9 @@ namespace battleutils
             {
                 charmerBSTlevel = charmerBRDlevel;
             }
-            if (charmerBSTlevel > PCharmer->GetMLevel()) // takes level cap/sync into account
+            if (charmerBSTlevel > charmerLvl) // takes level cap/sync into account
             {
-                charmerBSTlevel = PCharmer->GetMLevel();
+                charmerBSTlevel = charmerLvl;
             }
         }
         else if (PCharmer->objtype == TYPE_MOB)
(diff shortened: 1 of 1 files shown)


teschnei:
Should probably check that main job is actually bst

Wiggo32:
But... we'd want All jobs to be restricted from using BST Highest Level Obtained... not just BST... no?

src/map/utils/battleutils.cpp
```diff
@@ -4098,6 +4098,10 @@ namespace battleutils
             {
                 charmerBSTlevel = charmerBRDlevel;
             }
+            if (charmerBSTlevel > PCharmer->GetMLevel()) // takes level cap/sync into account
```
TeoTwawki: `PCharmer->GetMLevel())` is already contained in the variable `charmerLvl`

this already gets factored into charm chance if you look at whats happening with the variable `base`.

Wiggo32:
@teschnei We don't want to check for BST main job there like the check does for BRD up above it because every job (which has the ability to use FULL BST lvl for charm calculation regardless of BST being a subjob or not) should be restricted upon having a level cap restriction. Otherwise, people would just go into "Wings of Fury" as RDM/BST and still be able to charm the bats with a charm rate as if using a lvl 75 BST. (or whatever lvl your BST has currently achieved)

teschnei:
okay, charm works more strangely than I thought - can you remove the comment, though? this fix also fixes people levelling as /bst (same problem), but that case has nothing to do with level sync/cap, so the comment is kind of misleading

Wiggo32:
comment removed

PR #5294, opened 2018-10-10
Charutils: Rewrote some of the russian comments into English

hilts-vaughan:
While trying to understand some of the other stuff, I found a few things I wish I had known and could just read in English. Since it appears you guys are converting it as you see it, here is  what I see and understand as-is. I am not an expert so I tried to touch only things I was looking at in detail.


TeoTwawki:
Feel free to do more of this and also kill any tabs fixing alignment problems caused by tab2space along the way :+1:

PR #5300, opened 2018-10-11
SetMobLevel (for abyssea)

TeoTwawki:
ability to set monster levels in script - for use in abyssea mixin

I thought it better to go ahead and make a new function rather than a conditional on the existing player specific function. Can change that on request but the sub level version is 100% player only at this time so I figured may as well be specific.

Todo: move cruor stuff out of charutils and into a mixin. The reason it was in the core was I didn't want to individually script mobs (this was before mixins existed) and now we know each mobs is going to have a script+mixin ***anyways*** for lights level changes size changes etc. so we may as well do cruor there too.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -12425,9 +12425,9 @@ inline int32 CLuaBaseEntity::removeAllManeuvers(lua_State* L)
 }
 /************************************************************************
 *  Function: setMobLevel()
-*  Purpose : Updates the level of the monsters level
+*  Purpose : Updates the monsters level and recalculates stats
 *  Example : mob:setMobLevel(125)
-*  Notes   : does not refill mobs hp/mp
+*  Notes   : CalculateStats will refill mobs hp/mp as well
 ************************************************************************/
 
 inline int32 CLuaBaseEntity::setMobLevel(lua_State *L)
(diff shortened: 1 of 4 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -12423,6 +12423,27 @@ inline int32 CLuaBaseEntity::removeAllManeuvers(lua_State* L)
...
+************************************************************************/
+
+inline int32 CLuaBaseEntity::setMobLevel(lua_State *L)
+{
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB);
+
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
+
+    CMobEntity * PMob = (CMobEntity*)m_PBaseEntity;
```
nfaletra: Should still be checking for nullptr here to prevent crashing in release? Either way, we should stay away from c-style casting on complex types.

can either use dynamic_cast<CMobEntity*> to check for nullptr or static_cast<CMobEntity*> if we don't care.

TeoTwawki: ah rightm bad copy pasta. will fix where I got it from too

TeoTwawki:
https://github.com/DarkstarProject/darkstar/pull/5300/files?w=1

PR #5315, opened 2018-10-14
Maint mode

ShelbyZ:
- Adding conf/maint.conf to contain maintenance mode configuration
- Updating console input thread for lobby server to support multiple input params
- adding maint_mode option to control change maintenance mode value (0 - disabled/default, 1 - enabled)
- when value is changed via console input it is saved to the above configuration file
- while maintenance mode is active:
  - character list will be filter to only contain GM characters
  - if the character list is empty they will be kicked out with error 3321 - character parameters are incorrect
  - character creation is disabled - will result in error 3314 - failed to register with name server
  - if a player was idle on the character list with a non-GM character and attempts login they will be kicked out with error 3321 - character parameters are incorrect (note character list is not update because the client doesn't re-request the character list)

**Still TODO**:
- consider adding logging of who attempts login during maintenance and keep count
- add GM command to put the system into maint_mode 1, set a timer and during intervals send messages updating the countdown to maintenance, once interval has passed force logout all remaining characters

### DIFF
--- a/src/login/login.cpp
+++ b/src/login/login.cpp
@@ -28,9 +28,13 @@
 
 #include <stdio.h>
 #include <stdlib.h>
-#include <string.h>
+#include <string>
+#include <sstream>
+#include <vector>
+#include <limits>
 #include <thread>
 #include <iostream>
+#include <functional>
 
 #ifdef WIN32
 #include <io.h>
@@ -46,10 +50,12 @@
 
 const char* LOGIN_CONF_FILENAME = nullptr;
 const char* VERSION_INFO_FILENAME = nullptr;
+const char* MAINT_CONF_FILENAME = nullptr;
 
 //lan_config_t   lan_config;        // lan settings
 login_config_t login_config;    //main settings
 version_info_t version_info;
+maint_config_t maint_config;
 
 Sql_t *SqlHandle = nullptr;
 std::thread messageThread;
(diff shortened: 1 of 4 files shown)


src/login/login.cpp
```diff
@@ -323,182 +551,67 @@ int32 login_config_read(const char *cfgName)
...
-{
-    login_config.login_data_ip = "127.0.0.1";
-    login_config.login_data_port = 54230;
-    login_config.login_view_ip = "127.0.0.1";
-    login_config.login_view_port = 54001;
-    login_config.login_auth_ip = "127.0.0.1";
-    login_config.login_auth_port = 54231;
 
-    login_config.servername = "DarkStar";
+    fp = fopen(fileName, "w");
```
takhlaq: wanna check fp != nullptr again

src/login/login.cpp
```diff
@@ -305,15 +357,191 @@ int parse_console(char *buf)
...
+void maint_config_default()
+{
+    maint_config.maint_mode = 0;
+}
+
+std::string maint_config_write(const char* key)
+{
+    if (strcmp(key, "MAINT_MODE") == 0)
+    {
+        return std::to_string(maint_config.maint_mode).c_str();
```
takhlaq: dont need to reconstruct from `.c_str()`

PR #5320, opened 2018-10-16
Chocobo Digging rewrite

wrenffxi:
Added DIGGING_RATE setting.
Moved all digging rewards into the global script.
Added weighted reward selection (weight do not need to sum to 1000).

### DIFF
--- a/scripts/globals/chocobo_digging.lua
+++ b/scripts/globals/chocobo_digging.lua
@@ -1,256 +1,997 @@
--- chocobo_digging.lua
+-----------------------------------
+-- Chocobo Digging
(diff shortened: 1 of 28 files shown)


TeoTwawki:
in muh zone files, pwning muh customz with minor edit conflicts...

wrenffxi:
I didn't remove chance to skillup on failed dig.  The skillup check is outside of the pass/fail conditional.

Xaver-Entropia:
rechecked, >< my bad wren.

TeoTwawki:
Ok good. I'm on mobile at work so I couldn't even see the code today

PR #5323, opened 2018-10-18
add desynth recipe for spinel earring/oak cudgel +1

WiggoCaptures:
http://ffxiclopedia.wikia.com/wiki/Spinel_Earring?oldid=67885


WiggoCaptures:
also, found a bunch of broken synths that were recently added and decided to fix those as well...
can't have 0's at the end of desynth's its not going to give you anything but a strange message if you happen to hit that tier. if the item doesn't have a reported HQ4 for example, then it should just be filled in with whatever the HQ3 tier result is.

WiggoCaptures:
added oak cudgel +1 as well: http://ffxiclopedia.wikia.com/wiki/Oak_Cudgel_%2B1
sorry for so many revisions... wish i found these all at once...

PR #5326, opened 2018-10-20
Chasseur's Bottes/+1

EpicTaru:
--Added Quick Draw +22/+25 Mod (ModID 411)
--Added comments to both items
--Fixed my farkup!


nfaletra:
https://www.bg-wiki.com/bg/Chass._Bottes_%2B1

Mod 411 is Quick Draw base damage increase. That is not what these boots modify based on that page.

PR #5328, opened 2018-10-21
added imp mixin

wrenffxi:
I think this new CRITICAL_TAKE listener will be useful for chigoes too.


TeoTwawki:
> I think this new CRITICAL_TAKE listener will be useful for chigoes too.

yes. Got a few other situations to work out for chiggo's to be 100% accurate though. dnc steps, ws stuff.

PR #5333, opened 2018-10-22
wrong item id in header

KnowOne134:


wrenffxi:
Good catch.  The 4447 on line 23 should also be changed.

KnowOne134:
Better catch lol. I was just looking for an id other day and this script popped up. I never even looked at the rest :forehead smack:

PR #5346, opened 2018-10-26
tabled homepoint functions

wrenffxi:


TeoTwawki:
Want a emurder player vars.. baaaaadly.

Would you be so kind as to move those to char_unlocks table? Its ok if you don't wanna just thought I'd ask

wrenffxi:
I'm not at all familiar with char_unlocks. what are the bindings i'd use to read and write homepoints there?

TeoTwawki:
Its same as when you save the ourposts data with some minor cpp change

I moved non currency things out of char_points to form cha_unlocks

TeoTwawki:
ah yeah..the markers are reading player vars...

TeoTwawki:
ima just merg this, we're work on ebil charvar to sql blob conversion later

PR #5350, opened 2018-10-27
Bloody Bones lottery pop conditions

wrenffxi:


scripts/zones/Arrapago_Reef/mobs/Draugar_Servant.lua
```diff
@@ -0,0 +1,15 @@
...
+-----------------------------------
+local ID = require("scripts/zones/Arrapago_Reef/IDs")
+require("scripts/globals/mobs")
+-----------------------------------
+
+function onMobDeath(mob, player, isKiller)
+end
+
+function onMobDespawn(mob)
+    phOnDespawn(mob, ID.mob.BLOODY_BONES_PH, 20, 108000) -- 30 hours
```
TeoTwawki: huh.

crap wiki: 
> Spawns every ~~24 hours.~~ We camped him for 23 straight pops, his spawn time seemed to be 30+ hours, the longest taking 41 hours.

bg wiki:
> every 21-24 hours 

I guess he's such a low spawn rate that it can take forever after the window opens..

Juts my opinion but I'd open the window at 21 and then leave the spawn rate lower than 20 (maybe 5) to try and match that.

wrenffxi: changed to 21 hours with 5% chance

PR #5354, opened 2018-10-28
neaten player global

wrenffxi:


TeoTwawki:
not keen on the function re-ordering but other than that looks fine

wrenffxi:
I think it makes sense to have the local functions above the global ones, since the global functions usually call the locals, and not the other way around.  The way the code is now,

```
function A
  B()
  -- whole bunch of other stuff
end

function B()
end
```

you see B() being called but you haven't yet seen it, so you don't know if it's in the same file, or being called from one of the requires, whereas

```
local function A()
end

function B()
  A()
  -- whole bunch of other stuff
end
```

now when you see A() being called, you've already seen it and you know it's a local function.

TeoTwawki:
I suppose. And that is whats been done elsewhere (never cared for it elsewhere either, I always bottom'd them in my own custom code, purely aesthetic lua itself doesn't care). Didn't even occur to me charCreate was a local function to be honest.

PR #5359, opened 2018-10-31
telepoints: style guide and trade functions

wrenffxi:
added !pos coordinates to header comment
localed some variables
trade functions -> npcUtil
style guide
death to if boolean == false


scripts/zones/Eastern_Altepa_Desert/npcs/Telepoint.lua
```diff
@@ -1,40 +1,34 @@
...
+    if trade:getItemCount() == 1 and item >= 4096 and item <= 4103 and npcUtil.giveItem(player, 613) then
+        player:tradeComplete()
     end
-
-end;
+end
 
 function onTrigger(player,npc)
 
     if (player:hasKeyItem(dsp.ki.ALTEPA_GATE_CRYSTAL) == false) then
```
zynjec: One slipped through.

wrenffxi: kill it with fire

PR #5363, opened 2018-11-01
popFromQM can handle mobs popped from moving QMs

wrenffxi:
npcUtil.popFromQM can now handle mobs that spawn from moving QMs (Phantom and Morion Worm examples)
parameters tabled


scripts/globals/npc_util.lua
```diff
@@ -56,18 +66,31 @@ function npcUtil.popFromQM(player, qm, mobId, claim, hideDuration)
...
     -- hide qm
-    if hideDuration then
+    if params.hide > 0 then
         qm:setStatus(dsp.status.DISAPPEAR)
     end
 
     -- spawn mobs and give each a listener that will show QM after they are all dead
     for _, mob in pairs(mobs) do
+        -- choose random position uniformly from within radius
+        if params.radius and type(params.radius == "number") then
```
takhlaq:
```suggestion
        if params.radius and type(params.radius) == 'number' then
```

PR #5368, opened 2018-11-05
global id salvage cells

KnowOne134:


scripts/globals/items/castellanus_cell.lua
```diff
@@ -1,4 +1,8 @@
 -----------------------------------------
+-- Castellanus Cell
+-- ID 5366
+-- Unlocks head and neck equipment
+-----------------------------------------
 require("scripts/globals/status")
```
takhlaq: +require("scripts/globals/zone")

scripts/globals/items/castellanus_cell.lua
```diff
@@ -21,6 +25,5 @@ function onItemUse(target)
     if (newpower > 0) then
         target:addStatusEffectEx(dsp.effect.ENCUMBRANCE_I, dsp.effect.ENCUMBRANCE_I, newpower, 0, 0)
     end
-    target:messageText(target, 7209)
-end
-
+    target:messageText(target, zones[target:getZoneID()].text.CELL_OFFSET + 1)
```
takhlaq: dsp.zones

scripts/globals/items/castellanus_cell.lua
KnowOne134: tried it your way and got a console error, but my way works. took the example from and file in the global section that uses ID from the current zone.
image

PR #5369, opened 2018-11-05
Style Guide this one particular food line that appears 721 times

wrenffxi:


TeoTwawki:
eventually ima rework that to be 1000% less retarded a way to handle non-itemID'd food effects.. *stupid subID craps*

PR #5371, opened 2018-11-05
fix grouped loot in mob_droplist.sql

wrenffxi:
Currently this happens when you kill KB (and a few other NMs with grouped drops):

prevent this from happening


TeoTwawki:
somebody forgot to make non zero group when they nuked the scripted version

PR #5373, opened 2018-11-06
remove no-longer-needed NPC ID prefixes

wrenffxi:
now that IDs are tabled by zone, we don't need these prefixes to avoid conflicts


wrenffxi:
Nope, but I think many of them were given prefixes "just in case / I don't feel like searching the code base for possible conflicts"

PR #5385, opened 2018-11-10
fix bug in homepoint.lua that would blackscreen people using hp teleport

wrenffxi:


wrenffxi:
hold review. still gettin' blackscreened on cross-zone teleports.  will fix and add here.

PR #5399, opened 2018-11-15
Maat Fight Rework

jmcmorris:
Discovered there were several issues with the Maat fight for G5.
- Maat's subjob was MNK giving him many MNK traits
- Maat's combat skill type was always H2H regardless of the player's job
- Maat would never use his 2h special
- Many Maat's messages were not used properly or at all
- Prefix wasn't set so messages would say "The Maat" instead of just "Maat"

One of the main changes here is that I added in a Maat mixin to handle nearly all of his logic in a single place. The other major change is in `mobutils.cpp` where Maat is setup. The combat delays are based off of common combat delays for the corresponding skill type.

This was tested thoroughly and even balance tested with the help of several Eden server community members.


KnowOne134:
Possibly use the same core code to setup the mob in the future for the enm "Pulling the strings"

scripts/mixins/families/maat.lua
```diff
@@ -0,0 +1,113 @@
...
+            player = GetPlayerByID(engagedID)
+            if player:getHP() == 0 then
+                local ID = require("scripts/zones/"..mob:getZoneName().."/IDs")
+                mob:showText(mob, ID.text.LOOKS_LIKE_YOU_WERENT_READY)
+            end
+        end
+    end)
+
+    mob:addListener("COMBAT_TICK", "MAAT_CTICK", function(mob)
+        local defaultAbility = {
```
nfaletra: Curly brace on new line

scripts/mixins/families/maat.lua
```diff
@@ -0,0 +1,113 @@
...
+            [dsp.job.BRD] = dsp.jsa.SOUL_VOICE_MAAT,
+            [dsp.job.RNG] = dsp.jsa.EES_MAAT,
+            [dsp.job.SAM] = dsp.jsa.MEIKYO_SHISUI_MAAT,
+            [dsp.job.NIN] = dsp.jsa.MIJIN_GAKURE_MAAT,
+            [dsp.job.DRG] = dsp.jsa.CALL_WYVERN_MAAT,
+            [dsp.job.SMN] = dsp.jsa.ASTRAL_FLOW_MAAT,
+        }
+
+        if mob:getHPP() < mob:getLocalVar("specialThreshold") then
+            local ID = require("scripts/zones/"..mob:getZoneName().."/IDs")
```
nfaletra: You can get IDs from the zones global instead of doing string concats.
`local ID = zones[mob:getZoneID()]`

TeoTwawki:
Also: darkstar straight up does maat entirely wrong. if you look at the dats (or the captures) you can see retail actually has multiple versions of him. a diff mob ID with diff job & stats. [darkstar has been doing this insane copy of the player to try and rebuild him on the fly.](https://github.com/DarkstarProject/darkstar/blob/09b3ddfe5f24190203a2ceb225c479bcc699d045/src/map/utils/mobutils.h#L76) and there is nothing "special" (lookin at you, dumb `CheckMaatFights()` function in the global )about his battlefield really. the testimony is just his "orb" its really not as that complicated.

**I've just been waiting for bcnm rewrite to get merged** before I push a rewrite of his entire mess.

scripts/mixins/families/maat.lua
jmcmorris: Thanks for pointing that out. I was hoping there was a better way to do this!

PR #5401, opened 2018-11-16
Cor AF2 Against all odds

KnowOne134:


scripts/zones/The_Ashu_Talif/mobs/Yazquhl.lua
```diff
@@ -0,0 +1,38 @@
...
+local ID = require("scripts/zones/The_Ashu_Talif/IDs")
+mixins = {require("scripts/mixins/job_special")}
+-----------------------------------
+
+function onMobSpawn(mob)
+    mob:addMod(dsp.mod.SLEEPRES, 150)
+    mob:addMod(dsp.mod.SILENCERES, 150)
+end
+
+function onMobFight(mob,target)
```
wrenffxi: this shouldn't go in onMobFight, or it'll add the listener over and over during combat.

you can use onMobWeaponskill() in a mob script the same way you use it in a mobskill script.  The core calls both.

KnowOne134: Ok i was copying the script from the black coffin captain its the same way but doesnt work

PR #5402, opened 2018-11-17
Fixes maat stuck in 2 hour casting loop

kaincenteno:
For some reason if specialThreshold is set to -1, Maat ends up stuck recasting 2 hour ability once his HPP gets below the original specialThreshold value . This only gets fixed if the server is restarted as Maat will be found still casting 2 hour ability indefinitely.

### DIFF
--- a/scripts/mixins/families/maat.lua
+++ b/scripts/mixins/families/maat.lua
@@ -69,7 +69,7 @@ g_mixins.maat = function(mob)
             local ID = zones[mob:getZoneID()]
             mob:messageText(mob, ID.text.NOW_THAT_IM_WARMED_UP)
             mob:useMobAbility(defaultAbility[mob:getMainJob()])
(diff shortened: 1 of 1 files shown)


nfaletra:
`inline int32 CLuaBaseEntity::setLocalVar(lua_State* L)
{
    DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isstring(L, 1));
    DSP_DEBUG_BREAK_IF(lua_isnil(L, 2) || !lua_isnumber(L, 2));

    const char* var = lua_tostring(L, 1);
    auto val = (uint32)lua_tointeger(L, 2);

    m_PBaseEntity->SetLocalVar(var, val);

    return 0;
}`

This is why. Can't set local vars to negative values. It casts to unsigned and underflows to INT_MAX. Therefore the threshold check is always true.

PR #5404, opened 2018-11-17
November 2018 Retail Update

nfaletra:
* All Text IDs shifted around the 6000+ range
* No currently coded mobs/npcs shifted


TeoTwawki:
I'm surprised SE hasn't made noob zones shift from holiday NPCs yet. guess December will do that

PR #5406, opened 2018-11-18
Resolved issue with Rose Garden Spawning

jmcmorris:
When Voluptuous Vilma would be spawned it would enable the Rose Garden PH to spawn as well by calling the `Rose_Garden:onMobDespawn`. This is resolved by checking the `timeToGrow` variable and only setting Rose Garden PH to spawn if Vilma did not spawn.

Bonus: Cleaned up related scripts and made Rose Garden and Vilma spawn in the exact place the previous one was at so that it looks more like it is evolving.

### DIFF
--- a/scripts/zones/Yuhtunga_Jungle/mobs/Overgrown_Rose.lua
+++ b/scripts/zones/Yuhtunga_Jungle/mobs/Overgrown_Rose.lua
@@ -4,22 +4,16 @@
 -----------------------------------
 local ID = require("scripts/zones/Yuhtunga_Jungle/IDs")
 
-function disturbMob(mob)
+function onMobSpawn(mob)
     if mob:getID() == ID.mob.ROSE_GARDEN_PH then
         mob:setLocalVar("timeToGrow", os.time() + math.random(36000,37800)) -- 10:00:00 to 10:30:00
     end
 end
 
-function onMobSpawn(mob)
-    disturbMob(mob)
-end
-
-function onMobEngaged(mob, target)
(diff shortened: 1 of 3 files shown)


scripts/zones/Yuhtunga_Jungle/mobs/Overgrown_Rose.lua
```diff
@@ -2,38 +2,39 @@
...
+end
 
 function onMobRoam(mob)
     -- Rose Garden PH has been left alone for 10.25 hours
-    if (mob:getID() == ID.mob.ROSE_GARDEN_PH and os.time() > mob:getLocalVar("timeToGrow")) then
-        DisallowRespawn(ID.mob.ROSE_GARDEN_PH, true);
-        DespawnMob(ID.mob.ROSE_GARDEN_PH);
-        DisallowRespawn(ID.mob.ROSE_GARDEN, false);
-        SpawnMob(ID.mob.ROSE_GARDEN);
+    if mob:gtID() == ID.mob.ROSE_GARDEN_PH and os.time() > mob:getLocalVar("timeToGrow") then
```
godmode: Typo

jmcmorris: Thanks, not sure how that happened...

jmcmorris:
@ShelbyZ That is a good suggestion. When I started this I didn't think I was focused on just fixing the issue but you're right that I should also clean up the code logic as well. I can do that!

PR #5408, opened 2018-11-18
rage mode mixin

wrenffxi:


TeoTwawki:
oh..I just realized a thing, will catch you on irc

TeoTwawki:
10x instead of 10% to match existing core function, and we can deal with gradual rage later if you want.

PR #5409, opened 2018-11-19
Quick Draw Additions and Cleanup

nfaletra:
* Fixed Quick Draw damage calculation (was multiplying ranged weapon dmg by 2 and then adding ammo dmg)
* Added Quick Draw Magic Accuracy mod (added mod to Death Penalty)
* Removed marksmanship skill from the accuracy check
* Quick Draw should ignore shadows, gutted that nonsense
* Cleaned up and formatted the *_shot.lua scripts
* Quick Draw doesn't give the target TP

Made sure all shots still work. Light shot still sleeps. Dark shot still dispels.

### DIFF
--- a/scripts/globals/abilities/dark_shot.lua
+++ b/scripts/globals/abilities/dark_shot.lua
@@ -3,30 +3,29 @@
 -- Consumes a Dark Card to enhance dark-based debuffs. Additional effect: Dark-based Dispel
 -- Bio Effect: Attack Down Effect +5% and DoT + 3
 -----------------------------------
-require("scripts/globals/settings")
-require("scripts/globals/status")
 require("scripts/globals/magic")
+require("scripts/globals/status")
 -----------------------------------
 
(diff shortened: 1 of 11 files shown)


Hozu:
Hi, I added the TP gain for the user to Quick Draw somewhat recently because the BG wiki page for it says it it does that. Is this not actually the case on current retail?

https://www.bg-wiki.com/bg/Quick_Draw

nfaletra:
TP gain for user is handled in the takeWeaponskillDamage call. Not sure what you mean. All I did was set the target tp mult to 0 so the target of the Quick Draw doesn't gain TP from the damage taken.

Hozu:
Ooh, I misread that, my bad. Though I thought takeWeaponSkillDamage has a multiplier for TP gain for the target, which I thought I had set to 0. I feel like I had that part working when I tested that, but it was too long ago for me to remember properly.

PR #5412, opened 2018-11-21
mixin for colibri who copy spells

wrenffxi:


Brierre:
Super sweet, was just looking for this!

PR #5418, opened 2018-11-24
Promyvion global and portal randomizing

wrenffxi:
moved all the promyvion MR/Stray/Stream logic into a global

the correct exit to next floor is now picked randomly at server spawn, and randomized again after the portal opens, rather than having it be a chance based on number of remaining MRs.

moved spawn mods into sql

style in separate commit

### DIFF
--- a/scripts/globals/promyvion.lua
+++ b/scripts/globals/promyvion.lua
@@ -1,0 +1,153 @@
+require("scripts/zones/Promyvion-Dem/IDs")
+require("scripts/zones/Promyvion-Holla/IDs")
+require("scripts/zones/Promyvion-Mea/IDs")
+require("scripts/zones/Promyvion-Vahzl/IDs")
+require("scripts/globals/status")
+------------------------------------
+
+dsp = dsp or {}
+dsp.promyvion = dsp.promyvion or {}
+
+------------------------------------
+-- LOCAL FUNCTIONS
+------------------------------------
+
+local function maxFloor(ID)
+    local m = 0
+
+    for _, v in pairs(ID.mob.MEMORY_RECEPTACLES) do
+        m = math.max(m, v[1])
+    end
+
+    return m
+end
+
+local function randomizeFloorExit(ID, floor)
+    local streams = {}
+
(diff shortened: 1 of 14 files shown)


scripts/globals/promyvion.lua
```diff
@@ -0,0 +1,153 @@
...
+        mother:AnimationSub(1)
+    end
+end
+
+dsp.promyvion.receptacleOnFight = function(mob, target)
+    local ID = zones[mob:getZoneID()]
+    local mobId = mob:getID()
+    local numStrays = ID.mob.MEMORY_RECEPTACLES[mobId][2]
+
+    -- keep pets linked
```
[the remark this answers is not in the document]
TeoTwawki: They do start off with initial agro on whoever has main hate on the receptacle at spawn time, if it is engaged. Otherwise they just hang around waiting for somebody to look at them wrong and if they are already fighting somebody they don't change target.

sql/mob_pool_mods.sql
```diff
@@ -292,6 +292,9 @@ INSERT INTO `mob_pool_mods` VALUES (4836,62,1,1); -- Maat (BLM) NO_STANDBACK
 INSERT INTO `mob_pool_mods` VALUES (4837,62,1,1); -- Maat (RNG) NO_STANDBACK
 INSERT INTO `mob_pool_mods` VALUES (5403,62,1,1); -- Maat (NIN) NO_STANDBACK
 
+INSERT INTO `mob_pool_mods` VALUES (3784,2,-1,1); -- Stray GIL_MAX: don't drop gil
+INSERT INTO `mob_pool_mods` VALUES (2614,368,100,0); -- Memory Receptacle regain 10%
```
TeoTwawki: Do these work form onMobInitialize? the regain I know _can_ but not sure about the gil mod. There is a preference for doing them in script if they already have a mob script because we've had some discussion on getting rid of the sql tables holding them eventually.

wrenffxi: yes it works. just tested.

PR #5420, opened 2018-11-27
add optional target argument for release command.

TeoTwawki:
Normally we'd just force zone someone stuck in cs but there is one situation where we'd want to avoid that: stuck at BCNNM entrance. Zoning the player would cancel/fail the BCNNM.

### DIFF
--- a/scripts/commands/release.lua
+++ b/scripts/commands/release.lua
@@ -6,9 +6,25 @@
 cmdprops =
 {
     permission = 1,
-    parameters = ""
-};
+    parameters = "s"
+}
 
-function onTrigger(player)
-    player:release();
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!release {name}")
+end
+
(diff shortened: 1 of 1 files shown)


TeoTwawki:
p.s. this **intentionally** does not use `getCursorTarget()` as I feel it would be too easy to accidently fire off on the wrong target

scripts/commands/release.lua
```diff
@@ -6,9 +6,25 @@
...
-    player:release();
-end
\ No newline at end of file
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!release {name}")
+end
+
+function onTrigger(player, name)
+    local target
```
nfaletra: Initialize to nil, plox

TeoTwawki: That is initialized as nil. When you don't specify thats what lua does. i fot told to stop doing = nil there for this reason.

PR #5429, opened 2018-12-02
strays don't reward XP

wrenffxi:
retail evidence (volume warning!)
https://www.youtube.com/watch?v=oEBb5bU2sM4#t=2m45s

### DIFF
--- a/sql/mob_pool_mods.sql
+++ b/sql/mob_pool_mods.sql
@@ -293,6 +293,7 @@ INSERT INTO `mob_pool_mods` VALUES (4837,62,1,1); -- Maat (RNG) NO_STANDBACK
 INSERT INTO `mob_pool_mods` VALUES (5403,62,1,1); -- Maat (NIN) NO_STANDBACK
 
 INSERT INTO `mob_pool_mods` VALUES (3784,2,-1,1); -- Stray GIL_MAX: don't drop gil
+INSERT INTO `mob_pool_mods` VALUES (3784,28,-100,1); -- Stray EXP_BONUS: don't give XP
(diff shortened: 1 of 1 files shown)


TeoTwawki:
```sql
INSERT INTO `mob_pool_mods` VALUES (3784,2,-1,1); -- Stray GIL_MAX: don't drop gil
INSERT INTO `mob_pool_mods` VALUES (3784,28,-100,1); -- Stray EXP_BONUS: don't give XP
INSERT INTO `mob_pool_mods` VALUES (2614,368,100,0); -- Memory Receptacle regain 10%
```
you know I'm just gonna move all 3 of these to lua if you don't, before merging

zynjec:
Why would you put it in lua?

This is literally what this table is there for, instead of having to copy and paste those 3 mods to strays across 4 zones.

TeoTwawki:
This is literally a thing we're trying to get rid of and has been mentioned multiple times, why would you keep expanding it?

zynjec:
Maybe for one off mobs, but this is a family across multiple zones, applying 3 mods for 4 different scripts is a lot messier.

TeoTwawki:
I disagree Like I said were intending for this table to not exist in the future. I was told be dear leader we prefered to lua them. 3 isn't a lot of scripts to add 1 line to.

TeoTwawki:
fk it merge if you want, I don't care enough to die on such a small hill

nfaletra:
Per discussion on discord, mob_pool_mods isn't going anywhere. We're trying to avoid having sql handle mob mods on an individual mob.

PR #5431, opened 2018-12-02
Turned signet defense and evasion boost into a latent

jmcmorris:
Now Signet DEF and EVA boosts are only gained while engaged to an EM or lower. According to https://ffxiclopedia.fandom.com/wiki/Signet#Signet_Bonuses


jmcmorris:
This latent (and all other related effects) should only be in effect for the corresponding region areas. I will make that change.

src/map/latent_effect_container.cpp
```diff
@@ -712,6 +722,18 @@ bool CLatentEffectContainer::ProcessLatentEffect(CLatentEffect& latentEffect)
     case LATENT_WEAPON_SHEATHED:
         expression = m_POwner->animation != ANIMATION_ATTACK;
         break;
+    case LATENT_TARGET_EVEN_MATCH_LOWER:
+    {
+        if (m_POwner->GetBattleTargetID()) {
```
nfaletra: { on newline

nfaletra: Not sure why there's a check on GetBattleTargetID when we check the pointer later. Could this be simplified to just checking the pointer? Unless I'm missing something.

jmcmorris:
I decided to pull all of the remaining necessary changes into this PR.

I also went ahead and added in a check for if we are attempting to get an entity with ID 0 and quickly return `nullptr`. This is assuming that no entity will ever have an ID 0 and if that is not true then I will revert that assumption.

PR #5433, opened 2018-12-03
Last item in inventory container not being iterated on in a few scenarios

jmcmorris:
There is a `CItemContainer::ForEachItem` which could be used here instead but is only used in one place. Using it might complicate the code more than it is worth; however, if desired I can refactor the code to use that.


jmcmorris:
Noting that the build failing here has been fixed in my other PR

teschnei:
probably worth using ForEachItem so this doesn't happen again in the future ;)

PR #5434, opened 2018-12-03
Percolator pup attachment has wrong element cost

DivByXI:
Percolator costs 2 water element points on the server, the client expects 1 water element point resulting in loss of synchronization and client issues.


DivByXI:
percolator

PR #5437, opened 2018-12-04
review regimes obtained in zone A from zone B without crashiness

wrenffxi:
fix so that regime global won't crash when checked across type or zone
standardize player var names with [regime] prefix
to cleanup no-longer-used player variables:
```DELETE FROM char_vars WHERE varname LIKE 'fov_%';```

for #5419

### DIFF
--- a/scripts/globals/regimes.lua
+++ b/scripts/globals/regimes.lua
@@ -925,18 +925,33 @@ local regimeInfo = {
 }
 
 local function getPageByRegimeId(regimeType, zoneId, regimeId)
-    local pages = regimeInfo[regimeType].zone[zoneId].page
-    for _, v in pairs(pages) do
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/globals/regimes.lua [hunk 1]
```diff
@@ -987,6 +994,7 @@ end
 dsp.regime.bookOnEventUpdate = function(player, option, regimeType)
     local pRegime = player:getVar("fov_regimeid")
     local zoneId = player:getZoneID()
+    local page = getPageByRegimeId(regimeType, player:getVar("fov_zoneid"), pRegime)
```
nfaletra: This is still going to die on the regimeType if you have FoV active and try to review it on a GoV

wrenffxi: good catch. updated again.

PR #5444, opened 2018-12-07
Damage Taken math was seriously busted

nfaletra:
I have no idea why floor was being used for floating points. These are percent based mods. Flooring them makes no sense. Also multiplying by 256 and then dividing by 256 cancels out. Hopefully the compiler was optimizing it out.

Anyway, the reason this started becoming an obvious issue was because I added PDT II. So resist was becoming negative after capped PDT was flooring to 0.

I went through and fixed the other damage types as the math needed similar treatment.


TeoTwawki:
A previous pr aledgedly was going to made it so we input a % and get the x/y fractions retail uses which is how this disaster happened, pretty sure.  In any case thank you for fixing it.

TeoTwawki:
P.s. Should we in the near future switch ober to 4 digits to accurately reflect the half percent possibilities or is that only going to matter for capping?

PR #5446, opened 2018-12-08
convert two NMs to 2hr mixin

wrenffxi:


scripts/zones/Beadeaux_[S]/mobs/DaDha_Hundredmask.lua
```diff
@@ -1,25 +1,16 @@
...
-    mob:addMod(dsp.mod.GRAVITYRES, 50);
-end;
-
-function onMobFight( mob, target )
-    if (mob:getHPP() < mob:getLocalVar("HpTrigger") and mob:getLocalVar("perfect_dodge") == 0) then
-        mob:useMobAbility(1013);
-        mob:setLocalVar("perfect_dodge", 1);
-    end
-end;
+    mob:setMod(dsp.mod.TRIPLE_ATTACK, 25)
```
Comment (TeoTwawki) [inline scripts/zones/Beadeaux_[S]/mobs/DaDha_Hundredmask.lua]: merging coz it was already present but now that I see it highlighted I wonder if mods here are legit retail or someoen didn't realize job traits apply to mobs /shrug

PR #5454, opened 2018-12-09
add mixin for tripple Gear type mobs add in the TP moves

KnowOne134:
captured animations from retail for single all time, triple gears in 3 2 and 1 gear form


KnowOne134:
Did a full capture of reg gears and armed gears and they both share these tp moves, the single style use a different set of moves and this reflects all of that. I did fight armed gears for an hour without seeing heavy armature,,  :sad face:

PR #5456, opened 2018-12-10
Completable quest Old Man and the Harpoon with new quest system

ibm2431:
@takhlaq @ShelbyZ @wrenffxi @TeoTwawki

So I decided to take a stab at getting a basic quest working via a newer quest system, using tables which are all stored within a single lua file. I don't know exactly what the plans are for this, but tried to intuit based off what was already there and a conversation I had with @takhlaq over a year ago.

I chose 'The Old Man and the Harpoon' because it's incredibly simple, and as an Adoulin quest, suitable as a testing bed since I imagine I might be the only one who ever enters the zone. It involves just two NPCs, and can be completed like so:
```
!zone 256
!addmission ADOULIN LIFE_ON_THE_FRONTIER
!pos 92 32 152
Talk to Jorin
!pos 84 0 -60
Talk to Shipilolo
!pos 92 32 152
Talk to Jorin again, done
```
(Note: The message IDs on this branch are out of date, so the KI message will probably be gil obtained, and the bayld obtained message will probably be blank. They are firing though.)

Some changes were made to the global quests.lua to get it working, which brings me to my first question:
**I have no idea what the `cycle`, `needsCycling`, etc are meant for.** I commented them out from this because they weren't working (trying to ipairs over tables with non-integer keys) and I didn't need them. Whoever is responsible for them, please feel free to chime in!

The second issue is, at the moment, this **locks out anything that isn't part of the quest**, due to how `dsp.quest.check` is now currently executing functions. The problem stems from it needing to return something to prevent multiple events from firing - see Jorin.lua. At the moment I have `dsp.quest.check` just return `true` after invoking an appropriate `onTrigger` function from the quest table. This prevents the standard dialogue event from firing while he's rambling about his harpoon. Because if he's not talking about the quest, he's "supposed" to revert to his usual standard dialogue. 

Problem is, this setup has the `onTrigger` function from the quest table always fire - it kind of has to, since we need to check the status of the quest each time the player talks to the NPC. But if there's nothing for it to do (ie: the quest has been completed), `dsp.quest.check` will still return true after invoking the quest's associated `onTrigger` function, so Jorin won't reach his standard dialogue. Or in Shipilolo's case, she won't reach her other quest triggers.

I don't know the best way around this. My only idea at the moment is changing `onTrigger`s from firing `startEvents` themselves to simply return just the event ID as an integer. Then, have `dsp.quest.check` fire the event, and only return `true` to the NPC script if it did so.

Despite that issue though, **the quest is complete and in a single file** - with the exceptions of the single references to check for it in the two NPCs' `onTrigger`s and `onEventFinish`.

In the future, the references in the NPCs' `onEventFinish` might be removable with a master table linking CS IDs to what quest they're associated to, and core calling the quest file directly upon finish / update, bypassing the NPC.

I do think there will still have to be references in `onTrigger`s so we can direct which missions / quests take priority over others. But at least they'll be a single line, which will handle all the scripting from a unified location.

Comments, questions, edits welcome. I was absent when this branch was being shaped, so I'm probably a bit out of the loop if these issues already have answers.

### DIFF
--- a/scripts/globals/npc_util.lua
+++ b/scripts/globals/npc_util.lua
@@ -237,7 +237,7 @@ end
             fame = 120,                 -- fame defaults to 30 if not set
             bayld = 500,
             gil = 200,
-            xp = 1000,
+            exp = 1000,
             title = dsp.title.ENTRANCE_DENIED,
             var = {"foo1", "foo2"}      -- variable(s) to set to 0. string or table
         })
@@ -279,8 +279,8 @@ function npcUtil.completeQuest(player, area, quest, params)
         player:messageSpecial(ID.text.BAYLD_OBTAINED, params["bayld"] * BAYLD_RATE)
     end
 
-    if params["xp"] ~= nil and type(params["xp"]) == "number" then
-        player:addExp(params["xp"] * EXP_RATE)
+    if params["exp"] ~= nil and type(params["exp"]) == "number" then
+        player:addExp(params["exp"] * EXP_RATE)
     end
 
     if params["title"] ~= nil then
--- a/scripts/globals/quests.lua
+++ b/scripts/globals/quests.lua
@@ -460,8 +460,8 @@ local function handleQuestVar(entity, quest, varname, val, get)
     end
 
     local var, vartype
-    if quest.vars.main == varname then
-        var = quest.vars.main
+    if quest.vars.stage == varname then
+        var = quest.vars.stage
         vartype = dsp.quests.enums.var_types.CHAR_VAR
     else
         var = quest.vars.additional[varname]
@@ -469,7 +469,7 @@ local function handleQuestVar(entity, quest, varname, val, get)
     end
 
     if not var then
-        ret.message = " unable to find "..varname.." for quest: "..quest.name.." (logid: "..quest.log_id..")"
+        ret.message = " unable to find "..varname.." for quest: "..quest.name.." (log_id: "..quest.log_id..")"
     else
         if vartype == dsp.quests.enums.var_types.CHAR_VAR then
             if get then
@@ -520,23 +520,39 @@ dsp.quests.getVar = function(entity, quest, varname, val)
     end
 end
 
+dsp.quests.setStage = function(entity, quest, val)
+    local message = "dsp.quests.setStage "
+    return dsp.quests.setVar(entity, quest, quest.vars.stage, val)
+end
+
+dsp.quests.getStage = function(entity, quest)
+    local message = "dsp.quests.getStage "
+    return dsp.quests.getVar(entity, quest, quest.vars.stage)
+end
+
 dsp.quests.complete = function(player, quest, reward_set)
+    local message = "dsp.quests.complete "
     if quest then
         if reward_set == nil then
             reward_set = quest.rewards.sets[1]
         end
         if quest.rewards and reward_set then
             -- todo: check inventory (including stack space), award items, return false if cant complete
-            return npcUtil.completeQuest(player, quest.area, quest.questid, reward_set)
-        end
+            local rewards_given = npcUtil.completeQuest(player, quest.area, quest.quest_id, reward_set)
+            if rewards_given then
+                -- clear stage CHAR_VAR if shouldnt be preserved
+                if not quest.vars.preserve_main_on_complete then
+                    player:setVar(quest.vars.stage, 0)
+                end
+                for name, var in pairs(quest.vars.additional) do
+                    if not var.preserve_on_complete then
+                        handleQuestVar(player, quest, name, 0, "dsp.quests.complete ", nil)
+                    end
+                end
 
-        -- clear main CHAR_VAR if shouldnt be preserved
-        if not quest.vars.preserve_main_on_complete then
-            player:setVar(quest.vars.main, 0)
-        end
-        for name, var in pairs(quest.vars.additional) do
-            if not var.preserve_on_complete then
-                handleQuestVar(player, quest, name, 0, "dsp.quests.complete ", nil)
+                return true
+            else
+                error(message.. "Unable to give quest rewards.")
             end
         end
     end
--- a/scripts/globals/quests.lua
+++ b/scripts/globals/quests.lua
@@ -1,4 +1,5 @@
 require("scripts/globals/log_ids")
+require("scripts/globals/npc_util")
 require("scripts/globals/zone")
 
 dsp = dsp or {}
@@ -12,6 +13,17 @@ dsp.quests.enums =
         JEUNO     = 3, OTHER_AREAS = 4,  OUTLANDS = 5,
         AHT_URGAN = 6, CRYSTAL_WAR = 7,  ABYSSEA  = 8,
         ADOULIN   = 9, COALITION   = 10
+    },
+
+    fame_areas =
+    {
+        SANDORIA          =  0, BASTOK           =  1, WINDURST           =  2,
+        JEUNO             =  3, SELBINA          =  4, MHAURA             =  2,
+        RABAO             =  4, KAZHAM           =  2, NORG               =  5,
+        ABYSSEA_KONSCHTAT =  6, ABYSSEA_TAHRONGI =  7, ABYSSEA_LATHEINE   =  8,
+        ABYSSEA_MISAREAUX =  9, ABYSSEA_VUNKERL  = 10, ABYSSEA_ATTOHWA    = 11,
+        ABYSSEA_ALTEPA    = 12, ABYSSEA_GRAUBERG = 13, ABYSSEA_ULEGUERAND = 14,
+        ADOULIN = 15
     },
 
     var_types =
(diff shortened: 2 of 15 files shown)


takhlaq:
>I have no idea what the cycle, needsCycling, etc are meant for. I commented them out from this because they weren't working (trying to ipairs over tables with non-integer keys) and I didn't need them. Whoever is responsible for them, please feel free to chime in!

idea was to cycle through any npcs which had multiple quest dialogues which the player met conditions for, unfortunately i didnt finish or test that part


the basic idea is this:
npc script:
```
require("scripts/globals/quests")
local quests = 
{
    require("scripts/globals/area/quest_name")
}

function onTrigger(npc, player)
    if not dsp.quests.onTrigger(npc, player, quest_table) then
        -- handle stuff if we didnt trigger a quest event
    end
end

-- etc for onTrigger/onEvent
```
then the quest script would return true on any events it has handled (see my commit pushed to your branch)

takhlaq:
> Oh, another thing, which is mostly a style question. In the quest file, for the cs tables, I currently have them sorted by CS ID. However, I'm wondering if we should consider sorting them by the order in which they're completed in? If they were, a scripter could read the event finishes in chronological completion order, which would make the quest easier to follow if you're looking at it "as a whole".

im still not sure on this.
on one hand, we can divide the quest into "stages" (similarly to ffxiv), that way we can check just the "stage" instead of keyitems and other player vars, however im not quite sure how that would play out since we'd have to refactor every quest with all their vars instead of just moving them from the npc script and into the self-contained quest script
it would make tracking it a lot easier though

takhlaq:
@ibm2431 i've pushed a commit to your branch handling quest dialogue cycling, as well as returning true on events handled (there's probably a much cleaner way of doing this, but i can't seem to think of anything at the moment)

i was able to progress each stage of the quest and receive rewards with the correct messages (i never update my client) as well as fall-back to default dialogue
image

takhlaq:
if we were to go the quest stages route, we may be able to build a dependency table for quest requirements e.g.

```lua
local quest_reqs =
{
    quests_missions =
    { 
        { quest = { require("path/to/quest/script"), stage = x }, mission = { require("path/to/mission"), stage = x } }
    },
    trade = { {item, qty} },
    keyitems = {...},
    fame = { {area, level} },
    -- etc..
}
```

ibm2431:
I was considering a requirements table, actually! We could also add in a fancy `!setreqs` command too which sets everything for the invoking player to meet the requirements for the quest for testing/debug purposes. So if the requirements call for you to be on/past Adoulin mission 'Life on the Frontier', it'll read the requirements table and set your mission to that if you aren't already, instead of you having to check / do it manually.

And naturally, a `!meetreqs <player>` command which would say what requirement someone is missing should a player say, "The NPC isn't giving me this quest, it must be be broken!" "No, you just don't meet the requirements."

I was also toying with the idea of a `!resetquest` command too which would reset everything associated with the quest to their initial values. This would require some additional tracking of things touched by the quest that aren't vars, like key items.

Regarding staging, I was considering saying last night, "Perhaps with the 'main' var we have, as a style preference, we should be setting and checking that, instead of checking for KIs. When writing and testing quests, we read a list of steps and treat it stage by stage anyway, so maybe our code should better reflect that."

On staging/refactoring: I'm assuming that while at first we might be able to get this "new" way into master alongside the "old" way (while deprecating "old" way), people will get unhappy if the "old" way stays around for too long due to the added confusion of multiple methods. So chances are we'll have to organize an effort go through all NPCs / quests we have and "convert" them over to the "new" way.

I'm in the camp that if we'd be going through all the quest code anyway, a little bit of extra refactoring wouldn't hurt. And if we organize the effort well enough, we could probably get up-to-date / verified information on what quests work on DSP and which still require work. As I think the [quest page on the wiki](https://wiki.dspt.info/index.php/Quests) was last updated in 2013.

I can already feel @TeoTwawki's blank stare for me pushing the idea though. But hey! It's the perfect excuse to refactor Halver! Doesn't that sound _just grand_?!

One thing I'll say now though before we get too far into it though: Even though this branch is called quest_rewrite and all our functions use the term quest, I don't know why missions can't be done in a similar manner. The game itself makes little distinction between the two (it literally sends the current TOAU mission and quest logs in the same individual packet). The gate guard system of the three cities will make those require special functions still, but for example if SE renamed Adoulin Missions to Adoulin Main Quest, no one would have been any wiser.

I'll do a quest or two for Shipilolo and see if I can break your cycling code.

TeoTwawki:
honestly I feel old and outdated and it's time to pass the torch to someone else to find small hills to die on :stuck_out_tongue_winking_eye: 

I don't even recognize the quest code anymore (or anything using utils, though wren made me hate utils a lot less with some refactoring), so I'll just have to pray y'all make it easy to understand when I have time to actually try and implement new things again someday :smile:

TeoTwawki:
p.s. halver actually isn't to terrible anymore. unfortunately patient zero spread his namesake "halver's disease" to many other npc's before he was cured.

takhlaq:
that would be great @ibm2431
also yeah, missions need refactoring similarly too

TeoTwawki:
> also yeah, missions need refactoring similarly too

**DEATH** to `FinishMissionTimeline()` !!!

ibm2431:
I tossed the rest of Shipilolo's quests into quest scripts. Since when I started working on Adoulin I was doing it on a "npcs-by-zone" basis, and got halted due to lacking an event dump for Eastern Adoulin, 'Fertile Ground' and 'Wayward Waypoints' were never fully implemented. I was keeping tabs on which quests were still "hanging", but I guess I can just toss it as "todos" in the quest file now so they don't get lost. One advantage of a centralized file I guess.

Another advantage is we can toss in all the steps to complete the quest in as well, so someone testing it doesn't need to open a wiki. We can also spot-check the comments to make sure a quest is 100% implemented in DSP, without having to look through all the scripting. I went ahead and put some in so we could see what it looks like. Feel free to say you hate comments though!

**'A Certain Substitute Patrolman' was fully implemented in DSP** though, since all _eight_ NPCs involved are in Western Adoulin. **I converted it over to the newer system, and it seems to work fine.** Having everything in one file really shines with these sorts of quests where you're running around talking to eight different people, instead of having it all spread out among their NPC files. We now have two quests on this new system.

**Neither Shipilolo, nor any of the other 8 NPCs touched by this thus far show any issues.**

When converting it, I also did it "stage-like" by using the main var (which I renamed to stage, since when talking about quests, we ask/say what "stage" of the quest we're on, not what "main"). Also tossed in some getStage and setStage shorthand functions to quickly get/set this (I like them, but do admit they might be too lazy, so I'm not _that_ married to them). I also tweaked 'The Old Man and the Harpoon' a bit to work based off of this as well instead of Key Item checks.

When putting these quest files together, I put in some "requirements" tables. I hope/intend to make use of these later, but didn't start the functions now due to fear of commit/PR creep. Also "temporary" tables, which we can use in calls to `dsp.quests.completeQuest` to make sure all remnants of a completed quest are cleaned up, or in that `!resetquest` command I mentioned before.

Unorganized thoughts while working on this:
1) It might be better to split `onTriggers`, `onTrades` into separate tables, and do the same for `onEventFinish` and `onEventUpdate`. We're rarely going to need to have all of them defined for each trigger / event, and by splitting them up we could save a couple lines per trigger / event, improve readability, and reduce tab-depth by a level:
```
this_quest.npcs =
{
    [dsp.zone.WESTERN_ADOULIN] =
    {
        ["Zaoso"] =
        {
            onTrigger = function(player, npc)
                if dsp.quests.getStage(player, this_quest) == 1 then
                    player:startEvent(2553) -- Progresses Quest: 'A Certain Substitute Patrolman'
                    return true
                end
            end
        },
        ["Clemmar"] =
        {
            onTrigger = function(player, npc)
                if dsp.quests.getStage(player, this_quest) == 2 then
                    player:startEvent(2554) -- Progresses Quest: 'A Certain Substitute Patrolman'
                    return true
                end
            end
        }
```
would become:
```
this_quest.onTriggers =
{
    [dsp.zone.WESTERN_ADOULIN] =
    {
        ["Zaoso"] = function(player, npc)
            if dsp.quests.getStage(player, this_quest) == 1 then
                player:startEvent(2553) -- Progresses Quest: 'A Certain Substitute Patrolman'
                return true
            end
        end,
        ["Clemmar"] = function(player, npc)
            if dsp.quests.getStage(player, this_quest) == 2 then
                player:startEvent(2554) -- Progresses Quest: 'A Certain Substitute Patrolman'
                return true
            end
        end
```
Since these aren't the NPC files themselves, we don't strictly _need_ to define the empty functions, so long as we have `dsp.quests.check` do a sanity check to make sure there's a function for it to actually invoke before it attempts to do so.

2) This branch could really use a rebase. I volunteer after this PR / other requested changes meet satisfaction and get merged.

3) The quest requirements were copy-pasted as a whole require of the quest table. But as I learned when trying to load the server to test, without commenting them out...

("life_on_the_frontier not found" because not scripted, a proper boilerplate for that mission table would have another require for the mission before that, then another before that, etc)

...we may just want these to be the quest/mission enums and then have some way of knowing/pulling the `stage` var for the required quest without needing the quest table itself. Probably through a `dsp.quests` function that only checks just one table without creating a require chain.

takhlaq:
for the quest stages, perhaps we should define them in each quest script instead of magic numbers and reference them by their enum (though that wouldn't work well with the requirements table)
we would need to handle quests with bit flags appropriately too here, and may need helper functions as well

the other functions in the quest script were just stubbed as an example, maybe we could group the functions by stage also, though that would be another level of indentation 

i'll merge this for now as the quests are functional for the scope of this pr and we can discuss this further on an issue or the next pr

PR #5459, opened 2018-12-11
Graoully

wrenffxi:
added to Graoully: lottery pop conditions, level, family, model, aggro, flags, and skill list


TeoTwawki:
No script for the NM?

```
Additional effect Plague
Runs at normal speed.
Resistant to Bind and Gravity.
Immune to Dark-based Sleep.
```
all possible right now, the add effect requires a mob script.

PR #5460, opened 2018-12-11
2 more lottery NMs. added 2hr to many NMs.

wrenffxi:
Added lottery pop conditions to NMs:
* Novv the Whitehearted
* Zuug the Shoreleaper

Added job special mixin to many NMs.


TeoTwawki:
question: as you add scripted 2hrs, you are remembering to check for and remove any of the old busted hard coded 2hrs that may be laying around in sql yes?

wrenffxi:
yeah, i convert 'em when i see 'em

PR #5467, opened 2018-12-12
Fix heal/unheal stutter bug.

katsith:
this fixes the bug where you re-rest after trying to stand up
it happens because darkstar is picks the action to take based on your current state instead of the field in the packet which says the requested action
the client will sometimes send multiple stop healing commands if you hold a movement key down while resting


zynjec:
While it may fix one issue, you're creating another.

What you've called request (0x04) in the packet is only `0x02` if the healing is interrupted by character movement, and will now break toggling healing (/heal, or toggling it with H or *)

katsith:
> will now break toggling healing (toggling it with /heal, H or *)

fixed!

Issue #5471, opened 2018-12-13
Automation WS can crash server since new build


KnowOne134:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 
newest

**_Source Branch_** (master/stable) **:** 
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
says NULL pointer on auto weaponskill skillschain
file dump:
https://www.dropbox.com/s/qqlffjd6qbowrur/auto_ws_crash.rar?dl=0

nfaletra:
I think I have a fix for this on Homepoint. I'll compare the code around puppet skillchains and submit a PR.

PR #5475, opened 2018-12-14
Fix PUP Skillchains occasionally crashing

nfaletra:
Core was not handling skillchain properties correctly at all. I fixed this on HomepointXI awhile back and have been meaning to port it back to DSP.

The skillchain properties are bitpacked into the Skillchain Effect's Power variable. This code properly unpacks them and also removes some superfluous code that appeared to have been copy+pasted from somewhere else.


nfaletra:
Fixed https://github.com/DarkstarProject/darkstar/issues/5471

PR #5476, opened 2018-12-14
Lamie TP moves and behavior

KnowOne134:
added mixin to toss weapon based on critical hits subanimation 0 -> 1 added some missing tp moves and can only use the right moves with correct job and animationsub
main weapon out will use that move, break it and will be reduced to hysteric barrage and tail slap.
tested against ranger mobs and couldnt get weapon to break so figure arrow deludge is always avail to use


KnowOne134:
accident made pull from master and it combined some more nyzul mob pos and group changes which is ok to merge still

scripts/globals/mobskills/dustvoid.lua
```diff
@@ -20,7 +20,9 @@ function onMobWeaponSkill(target,mob,skill)
     local dmgmod = 1
     local info = MobMagicalMove(mob,target,skill,mob:getWeaponDmg()*3,dsp.magic.ele.WIND,dmgmod,TP_NO_EFFECT)
     local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,MOBSKILL_MAGICAL,MOBPARAM_WIND,MOBPARAM_IGNORE_SHADOWS)
-
+    for i = 0,15 do
```
jmcmorris:
```suggestion
    for i = dsp.slot.MAIN, dsp.MAX_SLOTID do 
```

scripts/globals/mobskills/torrent.lua
```diff
@@ -0,0 +1,20 @@
...
+---------------------------------------------
+require("scripts/globals/status")
+require("scripts/globals/monstertpmoves")
+---------------------------------------------
+function onMobSkillCheck(target,mob,skill)
+    return 0
+end
+
+function onMobWeaponSkill(target, mob, skill)
+    for i = 0,15 do
```
jmcmorris:
```suggestion
    for i = dsp.slot.MAIN, dsp.MAX_SLOTID do 
```

scripts/mixins/families/lamie.lua
```diff
@@ -0,0 +1,35 @@
+require("scripts/globals/mixins")
+
+g_mixins = g_mixins or {}
+g_mixins.families = g_mixins.families or {}
+
+-- AnimationSub for Lamie
+-- 0 = main weapon out, sub weapon sheathed
+-- 1 = main weapon broken, sub weapon sheathed
+
+g_mixins.families.lamie = function(mob)
```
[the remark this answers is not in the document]
TeoTwawki: we could (maybe should)
do this more generally and not family specific, though still a mixin. I'm pretty sure there are several mobs with weapon changes like this using the same animation values.

KnowOne134:
Yes me a @jmcmorris were just talking about that also. All toau beast do this

KnowOne134:
I can change the 1 we currently have for qutrub to be for all

PR #5480, opened 2018-12-15
Take Damage Improvements

jmcmorris:
This patch attempts to solve a few things:

1. Resolved issue with job/pet/mob abilities not waking up the target
2. Resolved issue with all job/pet/mob abilities not setting the last attacker
3. Resolved issue with a few job abilities not factoring in stoneskin
4. Needed the attack and damage type for when damage is taken so that some mobs can react accordingly such as Flan changing mode.

The attack and damage types were mostly taken from what was already being provided but there are a few places that I updated according to ffxiclopedia, bg-wiki and ffo.jp.

I tested this with various player attacks (JA, WS, normal attacks, magic). I did not test every individual ability though (obviously) although it was mostly done with a regex find and replace. I tested "TAKE_DAMAGE" listener with a new flan family mixin which I will be submitting in a future PR.


src/map/lua/lua_baseentity.h
```diff
@@ -350,7 +350,7 @@ class CLuaBaseEntity
     int32 addHP(lua_State*);                // Modify hp of Entity +/-
     int32 setHP(lua_State*);                // Set hp of Entity to value
     int32 restoreHP(lua_State*);            // Modify hp of Entity, but check if alive first
-    int32 delHP(lua_State*);                // Subtract hp of Entity
```
TeoTwawki: woah woah, there may still be times we need the old binding in a script in the future - don't deleted it even if everything current is using `takeDamage()` after this pr.

src/map/entities/battleentity.h
```diff
@@ -234,13 +234,32 @@ enum SLOTTYPE
...
+    DAMAGE_HTH = 4,
+    DAMAGE_ELEMENTAL = 5,
+    DAMAGE_FIRE = 6,
+    DAMAGE_EARTH = 7,
+    DAMAGE_WATER = 8,
+    DAMAGE_WIND = 9,
+    DAMAGE_ICE = 10,
+    DAMAGE_LIGHTNING = 11,
+    DAMAGE_LIGHT = 12,
+    DAMAGE_DARK = 13,
```
TeoTwawki: yet _another_ list of elements to keep track of where people can get the value of the element wrong because of other lists of elements with diff values..

I don't have a change request or suggestion here. just complaining to myself about the inevitable having to explain it to ppl later. we already have multiple and ppl mix them up.

jmcmorris: Yeah, I agree this is not preferable. Unfortunately I think the previous `DAMAGETYPE` values are in the database. I suppose something we could do is offset all of the previous ones so they start at 9 after the elements. That would allow us to use the existing element values as a `DAMAGETYPE`.

If this were to happen I'm not sure if changing the database or just using an offset when reading the database would be best.

src/map/entities/battleentity.h
```diff
@@ -234,13 +234,32 @@ enum SLOTTYPE
...
+
 enum DAMAGETYPE
 {
     DAMAGE_NONE = 0,
     DAMAGE_PIERCING = 1,
     DAMAGE_SLASHING = 2,
     DAMAGE_IMPACT = 3,
-    DAMAGE_HTH = 4
+    DAMAGE_HTH = 4,
+    DAMAGE_ELEMENTAL = 5,
```
teschnei: why would elemental be separate from the elements? how could this ever be set?

src/map/entities/battleentity.h
teschnei: it should be fine the way it is - nobody should ever be interacting with this table with the raw numbers, always either this enum or the table on the lua side, so it shouldn't matter what the values actually are.  As long as we don't start using this enum in the database.

src/map/entities/battleentity.h
teschnei: maybe this was supposed to be NON elemental (such as requiescat)?

jmcmorris: The idea behind `DAMAGE_ELEMENTAL` is that it can be used as the offset for elements. So you can do `DAMAGE_ELEMENTAL + element` to get the correct damage element. I think right now I only use that in the lua scripts with `dsp.damageTypes` but I was trying to keep parity between the two. We could use it for NON elemental too but I was thinking we could just use `DAMAGE_NONE` for that?

src/map/lua/lua_baseentity.h
jmcmorris: Added it back in

src/map/entities/battleentity.h
TeoTwawki: >  As long as we don't start using this enum in the database.

at one time in the past we were. guess who nuked that.

src/map/entities/battleentity.h
teschnei: i think it's clearer to have a specific non elemental type.  we currently use DAMAGE_NONE to figure out if a character has no weapon equipped, so I wouldn't really want to combine non-elemental into that.  Plus, ELEMENT_NONE would be equivalent to this value anyways, so your use case stays consistent and doesn't need a special case for it

scripts/globals/abilities/devotion.lua
```diff
@@ -40,7 +40,7 @@ function onUseAbility(player,target,ability)
     healMP = utils.clamp(healMP, 0,target:getMaxMP() - target:getMP())
     -- printf("Devotion MP Healed: %d", healMP)
 
-    player:delHP(damageHP)
+    player:takeDamage(damageHP)
```
nfaletra: Needs to take stoneskin into account.
`Stoneskin can mitigate the amount of HP sacrificed. However, the ability does not lose potency this way.`

scripts/globals/abilities/martyr.lua
```diff
@@ -43,7 +43,7 @@ function onUseAbility(player,target,ability)
     -- Log HP Headed for Debug
     -- printf("Martyr Healed HP: %d", healHP)
 
-    player:delHP(damageHP)
+    player:takeDamage(damageHP)
```
nfaletra: Stoneskin here, also.
`Stoneskin can mitigate the amount of HP sacrificed. However, the ability does not lose potency this way.`

scripts/globals/abilities/ice_shot.lua
```diff
@@ -31,7 +31,7 @@ function onUseAbility(player, target, ability, action)
     dmg = adjustForTarget(target, dmg, dsp.magic.ele.ICE)
 
     params.targetTPMult = 0 -- Quick Draw does not feed TP
-    dmg = takeAbilityDamage(target, player, params, true, dmg, dsp.slot.RANGED, 1, 0, 0, 0, action, nil)
+    dmg = takeAbilityDamage(target, player, params, true, dmgdsp.attackType.MAGICAL, dsp.damageType.ICE, dsp.slot.RANGED, 1, 0, 0, 0, action, nil)
```
nfaletra: need a comma after dmg

nfaletra:
After looking through some of the scripts and realizing why these changes are being made, why not just add lua event triggers where OnMagicHit and OnWeaponskillHit are being called?

The weaponskill and spell are available in both places and have the damage types available to you for passing along to the event.

scripts/globals/abilities/pets/moonlit_charge.lua
```diff
@@ -18,9 +18,9 @@ function onPetAbility(target, pet, skill)
 
     local totaldamage = 0
     local damage = AvatarPhysicalMove(pet,target,skill,numhits,accmod,dmgmod,0,TP_NO_EFFECT,1,2,3)
-    totaldamage = AvatarFinalAdjustments(damage.dmg,pet,skill,target,MOBSKILL_PHYSICAL,MOBPARAM_BLUNT,numhits)
+    totaldamage = AvatarFinalAdjustments(damage.dmg,pet,skill,target,dsp.attackType.PHYSICAL,dsp.damageType.BLUNT,numhits)
     target:addStatusEffect(dsp.effect.BLINDNESS, 20, 0, 30)
-    target:delHP(totaldamage)
+    target:takeDamage(totaldamage, petdsp.attackType.PHYSICAL, dsp.damageType.BLUNT)
```
nfaletra: Another missing comma

scripts/globals/abilities/violent_flourish.lua
```diff
@@ -87,8 +87,7 @@ function onUseAbility(player,target,ability,action)
         end
 
         dmg = utils.stoneskin(target, dmg)
-
-        target:delHP(dmg)
+        target:takeDamage(dmg, player, dsp.attackType.PHYSICAL, player:getWeaponDamageType(0))
```
nfaletra: `player:getWeaponDamageType(dsp.slot.MAIN)`

scripts/globals/automatonweaponskills.lua
```diff
@@ -20,6 +20,7 @@ function doAutoPhysicalWeaponskill(attacker, target, wsID, tp, primary, action,
     -- apply WSC
     local weaponDamage = params.weaponDamage or attacker:getWeaponDmg();
     local weaponType = params.weaponType or attacker:getWeaponSkillType(0);
+    local damageType = attacker:getWeaponDamageType(0)
```
nfaletra: `attacker:getWeaponSkillType(dsp.slot.MAIN)`
`attacker:getWeaponDamageType(dsp.slot.MAIN)`

scripts/globals/items/excalibur.lua
```diff
@@ -14,16 +14,17 @@ function onAdditionalEffect(player, target, damage)
     if math.random(100) <= chance then
         local finalDMG = math.floor(player.getHP(player) / 4)
         if finalDMG > 0 then
+            local damageType = player:getWeaponDamageType(0)
```
nfaletra: `player:getWeaponDamageType(dsp.slot.MAIN)`

src/map/entities/battleentity.h
jmcmorris: I went through and changed all of the non-elemental magic attacks to use `dsp.damageType.ELEMENTAL`. There are still a number of `dsp.damageType.NONE` where I couldn't find any conclusive evidence on the damage type and it was previously set to none.

jmcmorris:
I'm still hoping to get this merged so I can make some more PRs. Any update?

PR #5495, opened 2018-12-17
Cleaned up emote packet

zynjec:


src/map/packets/char_emotion.h
```diff
@@ -28,19 +28,68 @@
 
 #include "basic.h"
 
-/************************************************************************
-*																		*
-*  																		*
-*																		*
-************************************************************************/
+enum EMOTES : uint8
```
nfaletra: `enum class`

src/map/packets/char_emotion.h
```diff
@@ -28,19 +28,68 @@
...
+    EMOTE_HURRAY = 43,
+    EMOTE_TOSS = 44,
+    EMOTE_DANCE1 = 65,
+    EMOTE_DANCE2 = 66,
+    EMOTE_DANCE3 = 67,
+    EMOTE_DANCE4 = 68,
+    EMOTE_JOB = 74
+};
+
+enum EMOTE_MODE : uint8
```
nfaletra: `enum class`

PR #5499, opened 2018-12-19
Check if battleground areas are full

cocosolos:
fixed storms of fate

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -8723,6 +8723,27 @@ inline int32 CLuaBaseEntity::registerBattlefield(lua_State* L)
     return 1;
 }
 
+inline int32 CLuaBaseEntity::battlefieldAtCapacity(lua_State *L)
+{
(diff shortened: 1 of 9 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -8723,6 +8723,27 @@ inline int32 CLuaBaseEntity::registerBattlefield(lua_State* L)
...
+    DSP_DEBUG_BREAK_IF(PZone->m_BattlefieldHandler == nullptr);
+
+    uint8 full = 0;
+
+    if (PZone != nullptr && PZone->m_BattlefieldHandler != nullptr &&
+        PZone->m_BattlefieldHandler->ReachedMaxCapacity((int)lua_tointeger(L, 1)))
+    {
+        full = 1;
+    }
+    lua_pushinteger(L, full);
```
TeoTwawki: I'd just like to point out `lua_toboolean()` is a thing. carry on. :wave:

cocosolos: @TeoTwawki thanks, I've never used lua before so I just repurposed a removed block. Hoping to improve

PR #5504, opened 2018-12-20
fix dia having no effect. remove (effect) FLAG_ from several spells.

wrenffxi:
fix dia never calling onEffectGain
remove (effect) FLAG_ being passed as subid in several spells.
dia and bio overwrite to equal or higher


TeoTwawki:
not sure what it was whoever thought putting FLAG_ERASABLE in the params would do, but I welcome its death. if anything magically breaks, we'll know what to reimplement then (I think its nil since tabling happened anyway?)

PR #5506, opened 2018-12-21
Add dsp.quests.checkRequirements()

ibm2431:
As it says in the title, I added a function to go through a quest's requirements table and return whether or not a player meets all the requirements to be given the quest. So when talking to the starting NPC, instead of checking the quest status of the quest-to-be-started, the status of previous quests in the chain, player level, etc, you can just do:
```
if dsp.quests.checkRequirements(player, this_quest) then
    -- Start the quest!
end
```
Currently implemented checks are _completed_ quests, _completed_ missions, possessed key items, fame level, current main job, and current player level.

At the moment it can't check if a player is on/past a certain stage in another quest, because...

(Tagging potentially interested parties for input on these: @takhlaq @TeoTwawki @wrenffxi @ShelbyZ @KnowOne134 and anyone else that might be interested in quests and I'm sorry for not thinking of you) 

**Current Quest Thoughts Number 1:** I can't think of a way around eventually **needing a master table of quest tables**. Regardless of how we store/pass the previous-required-quests in a given quest's table, we will eventually need that previous-quest's full table to check the `stage` var. (Unless we remove all scripter control over naming the stage var, see Thought 2 later)

So whether we store/require the entire previous-quest table in the quest-we're-checking (which will eventually lead to long require chains consisting of most quests in the game)... 

Or if we store the previous-quest in quest-we're-checking as a simple area / quest ID, we're going to need to be able to pull the previous-quest-table from just those area / quest ID enums.

Also, I was going to implement a `!resetquest` command this PR, but the command will need a way to pull the table of the quest-to-be-reset to delete it from the log, clear the quest variables, and delete any temporary key items. So whether it's invoked by `!resetquest ADOULIN 78` or `!resetquest ADOULIN A_CERTAIN_SUBSTITUTE_PATROLMAN`, the second argument has to be _some_ key in some table to pull the quest table.

Something like:
```
dsp.quests.quest_tables =
{
    adoulin =
    {
        [dsp.quests.enums.quest_ids.adoulin.A_CERTAIN_SUBSTITUTE_PATROLMAN] = require('scripts/quests/adoulin/a_certain_substitute_patrolman')
    }
}
```
And then store previous-quest-requirements as the area / quest IDs like I currently am, fetching the full quest tables when needed by consulting the table. This prevents long required-file _chains_. Of course, since this would have to be a global stored in globals/quests, the end result is that every quest DSP has will end up loading/requiring every other quest in the game by proxy.

Also, are there any thoughts on empty quest templates/tables for quests not currently implemented?

In other words, should DSP have lua files for quests it doesn't really have scripted, and then just have the master quest-table table just require everything _now_ (at the expense of a bunch of mostly-empty boilerplate files), or only add a quest-require to the master table when work has been done on it (at the expense of someone potentially trying to pull a nil table)?

Personally, I'm leaning towards a bunch of empty boilerplate, because even though it may not have events scripted, they could still have _some_ use now: like the quest's name at the very least. Things like `!addquest` could be modified to consult the master quest-table table and print the name of the quest from the otherwise-empty-boilerplate instead of some integer ID like it does now.

Let me know thoughts and I can start something for the next PR.

**Current Quest Thoughts Number 2:** **Should non-stage vars for a quest _always_ prefix a tag** for `player:get/setVar` purposes? I don't think they are at the moment.

I really, really think they should to 100% guarantee against any accidental playerVar collisions between quests (for example, I just have a var named 'waypoints' for Wayward Waypoints, and that's probably going to be used by someone else as well). I also think that the prefix/tag should be built by `dsp.quests.set/getVar` itself, instead of trusting the quest scripter to do so.

This always-prefix should also apply to the main/stage var, so instead of them being set as "[Q][Area][Quest ID]" in the quest table by the scripter, it's just called "stage" or something, and dsp.quests will handle the prefix.

**Alternatively, completely abstract the stage var. Remove it from the quest tables altogether, and dsp.quests.set/getStage will just use the "[Q][A][ID]" automatically.** This also makes it possible to check the stage of required-previous-quests _without_ the previous-quest's table, because we'll always know which playerVar to check since scripters won't have _any_ control over the name of the stage/main var.

So in the quest table, it'd look something like:
```
this_quest.vars =
{
    ["waypoints"] = { id = 1, type = dsp.quests.enums.var_types.LOCAL_VAR, repeatable = false, preserve_on_complete = false }   
}

some_npc_event = function(player, csid)
    dsp.quests.setQuestVar(player, this_quest, "waypoints", 5)
    dsp.quests.setStage(player, this_quest, 2)
end
```

And in dsp.quests:
```
dsp.quests.handleQuestVar = function(player, quest, varname, val)
    local varPrefix = "[Q]" .. "[" .. quest.area_id .. "][" .. quest.quest_id .. "]"
    if varname then
        -- Do other stuff based on contents of quest.vars.varname
        player:setVar(varPrefix .. varname, val)
    else
        -- Stage var
        player:setVar(varPrefix .. "[S]", val)
    end
end

dsp.quests.setQuestVar = function(player, quest, varname, val)
    dsp.quests.handleQuestVar(player, quest, varname, val)
end

dsp.quests.setStage = function(player, quest, stage)
    dsp.quests.setQuestVar(player, quest, "", stage)
end
```

@takhlaq: "I hate that magic number in calls to setStage()"...

**Current Quest Thoughts Number 3:** **How should stage enums be styled?** When you first mentioned them, I thought you meant stage enums might be set per-quest:
```
this_quest.enums.stages =
{
    ["got_broken_harpoon"]= 1,
    ["talked_shippilolo"] = 2
}

npc_event = function(player, csid)
    dsp.quests.setStage(player, this_quest, "talked_shippilolo")
end

--------------------------

dsp.quests.setStage = function(player, quest, stage)
    dsp.quests.setQuestVar(player, quest, "", quest.enums.stages[stage])
end
```

Which would make stages self-documenting. But potentially easier to screw up / mistype, And maybe more difficult for reviewers to track, oddly enough ("So what weird stage names is this person using, again? Need to scroll up to check").

But you also could have meant just generic enums:
```
dsp.quests.enums.stages =
{
    ['stage1'] = 1,
    ['stage2'] = 2
}

--------------------------

npc_event = function(player, csid)
    dsp.quests.setStage(player, this_quest, dsp.quests.enums.stage1)
end
```

Which would be easier to implement and also avoid the issues of magic numbers in code (for example, if you need to shift a stage, you can Ctrl+F the enum without having to worry about messing other 1's or 2's up).

I don't know if your concern about magic numbers is just making them clearly separate from other integer values, or "stage 1" or "stage 2" not being descriptive... descriptors.

### DIFF
--- a/scripts/globals/quests.lua
+++ b/scripts/globals/quests.lua
@@ -33,6 +33,13 @@ dsp.quests.enums =
         SERVER_VAR = 3,
     },
 
+    quest_status =
+    {
+        QUEST_AVAILABLE = 0,
+        QUEST_ACCEPTED  = 1,
+        QUEST_COMPLETED = 2,
+    },
+
     quest_ids =
     {
         sandoria =
@@ -597,7 +604,7 @@ dsp.quests.check = function(player, params)
                 }
                 exitLoop = checks[params.check_type](player, params)
                 player:setLocalVar("[quests]cycle", i)
-                print("Player: "..player:getName()..(targetName and " Npc: "..targetName or " Event: "..params.csid).. " \tCycle: "..i.." Quest: "..quest.name)
+                --print("Player: "..player:getName()..(targetName and " Npc: "..targetName or " Event: "..params.csid).. " \tCycle: "..i.." Quest: "..quest.name)
                 if exitLoop then
                     return true
                 end
@@ -608,6 +615,60 @@ dsp.quests.check = function(player, params)
         end
     end
     return false
+end
+
+dsp.quests.checkRequirements = function(player, quest)
+    local quest_status_check = player:getQuestStatus(quest.area, quest.quest_id)
+
+    if quest_status_check == dsp.quests.enums.quest_status.QUEST_AVAILABLE
+    or (quest_status_check == dsp.quests.enums.quest_status.QUEST_COMPLETED and quest.repeatable) then
+        -- Check all required quests
+        if quest.requirements.quests then
+            for i, required_quest in ipairs(quest.requirements.quests) do
+                local required_quest_status = player:getQuestStatus(required_quest.area, required_quest.quest_id)
+                if required_quest_status ~= dsp.quests.enums.quest_status.QUEST_COMPLETED then
+                -- or (required_quest.stage and (dsp.quests.getStage(player, required_quest) < required_quest.stage)) then -- getStage requires a full quest table
+                    return false
+                end
+            end
+        end
+        -- Check all required missions
+        if quest.requirements.missions then
+            for i, required_mission in ipairs(quest.requirements.missions) do
+                if player:getCurrentMission(required_mission.mission_log) < required_mission.mission_id then
+                -- or (required_mission.stage and (dsp.quests.getStage(player, required_mission) < required_mission.stage)) then -- getStage requires a full quest table
+                    return false
+                end
+            end
+        end
+        -- Check if player possesses all required key items to start quest
+        if quest.requirements.key_items then
+            for i, required_key_item in ipairs(quest.requirements.key_items) do
+                if not player:hasKeyItem(required_key_item) then
+                    return false
+                end
+            end
+        end
+        -- Check fame requirement
+        if quest.requirements.fame then
+            if player:getFameLevel(quest.requirements.fame.area) < quest.requirements.fame.level then
+                return false
+            end
+        end
+        -- Make sure player is on the right job, if applicable
+        if quest.requirements.main_job and (player:getMainJob() ~= quest.requirements.main_job) then
+            return false
+        end
+        -- Finally, make sure the player is high enough level
+        if quest.requirements.level and (player:getMainLvl() < quest.requirements.level) then
+            return false 
+        end
+
+        -- We haven't failed any of the requirement checks, so we must meet the quest's requirements
+        return true
+    else
+        return false
+    end
 end
 
 dsp.quests.onTrade = function(player, npc, trade, quest_table)
--- a/scripts/quests/adoulin/a_certain_substitute_patrolman.lua
+++ b/scripts/quests/adoulin/a_certain_substitute_patrolman.lua
@@ -29,16 +29,17 @@ this_quest.vars =
 
 this_quest.requirements =
 {
-    quests_missions =
-    { 
-        quests = {},
+    missions =
+    {
         {
-            -- [1] = { ['mission'] = require("scripts/globals/missions/adoulin/life_on_the_frontier") }
+            ['mission_log'] = ADOULIN,
+            ['mission_id'] = LIFE_ON_THE_FRONTIER
         }
     },
     fame =
     {
-        {this_quest.area, 1}
+        ['area'] = this_quest.area,
+        ['level'] = 1
     }
 }
 
@@ -68,17 +69,14 @@ this_quest.npcs =
         ["Rising_Solstice"] =
         {
             onTrigger = function(player, npc)
-                local ACSP = player:getQuestStatus(ADOULIN, A_CERTAIN_SUBSTITUTE_PATROLMAN)
-                if ACSP == QUEST_ACCEPTED then
-                    if dsp.quests.getStage(player, this_quest) >= 1 then
-                        if dsp.quests.getStage(player, this_quest) == 8 then
-                            player:startEvent(2552) -- Finishes Quest: 'A Certain Substitute Patrolman'
-                        else
-                            player:startEvent(2551) -- Dialogue during Quest: 'A Certain Substitute Patrolman'
-                        end
-                        return true
+                if dsp.quests.getStage(player, this_quest) >= 1 then
+                    if dsp.quests.getStage(player, this_quest) == 8 then
+                        player:startEvent(2552) -- Finishes Quest: 'A Certain Substitute Patrolman'
+                    else
+                        player:startEvent(2551) -- Dialogue during Quest: 'A Certain Substitute Patrolman'
                     end
-                elseif ACSP == QUEST_AVAILABLE then
+                    return true
+                elseif dsp.quests.checkRequirements(player, this_quest) then
                     player:startEvent(2550) -- Starts Quest: 'A Certain Substitute Patrolman'
                     return true
                 end
--- a/scripts/quests/adoulin/fertile_ground.lua
+++ b/scripts/quests/adoulin/fertile_ground.lua
@@ -23,17 +23,18 @@ this_quest.vars =
 
 this_quest.requirements =
 {
-    quests_missions =
-    { 
-        quests =
+    quests =
+    {
         {
-            -- [1] = { ['quest'] = require("scripts/globals/quests/adoulin/the_old_man_and_the_harpoon") }
-        },
-        missions = {}
+            ['area'] = ADOULIN,
+            ['quest_id'] = dsp.quests.enums.quest_ids.adoulin.THE_OLD_MAN_AND_THE_HARPOON
+        }
+        -- [1] = { ['quest'] = require("scripts/globals/quests/adoulin/the_old_man_and_the_harpoon") }
     },
     fame =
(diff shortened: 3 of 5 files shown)


takhlaq:
>Which would be easier to implement and also avoid the issues of magic numbers in code (for example, if you need to shift a stage, you can Ctrl+F the enum without having to worry about messing other 1's or 2's up).
I don't know if your concern about magic numbers is just making them clearly separate from other integer values, or "stage 1" or "stage 2" not being descriptive... descriptors.

hmmm
i was thinking the latter, with -1 being available, 0 being started, then the quest implements its own with 0xFFFFFFFF being finished or whatever
however, we'd run into the same issue of having each quest loading in all the others unless we did the requirements table in each quest's script

.. on second thought option3 would probably be a nicer alternative, though maybe we can get around typos by defining the vars like so

```lua
quest.vars =
{
  main = {id = 1, type = dsp.quests.var_type.PLAYER_VAR},
  npc_talked_to = {id = 2, type = dsp.quests.var_type.PLAYER_VAR},
  mobs_killed = {id = 3, type = dsp.quests.var_type.LOCAL_VAR},
}
...
dsp.quests.setVar(player, quest, quest.vars.main, val)
```
so we can enforce passing tabled vars (though defining the id again seems pointless since its just its index)
this would reduce quest var readability in database but that is not something i really care about since they should never be modified outside the script

scripts/globals/quests.lua
```diff
@@ -33,6 +33,13 @@ dsp.quests.enums =
         SERVER_VAR = 3,
     },
 
+    quest_status =
+    {
+        QUEST_AVAILABLE = 0;
```
takhlaq: comma instead of semicolon for new element in table

scripts/quests/adoulin/a_certain_substitute_patrolman.lua
```diff
@@ -68,17 +71,14 @@ this_quest.npcs =
...
-                local ACSP = player:getQuestStatus(ADOULIN, A_CERTAIN_SUBSTITUTE_PATROLMAN)
-                if ACSP == QUEST_ACCEPTED then
-                    if dsp.quests.getStage(player, this_quest) >= 1 then
-                        if dsp.quests.getStage(player, this_quest) == 8 then
-                            player:startEvent(2552) -- Finishes Quest: 'A Certain Substitute Patrolman'
-                        else
-                            player:startEvent(2551) -- Dialogue during Quest: 'A Certain Substitute Patrolman'
-                        end
-                        return true
+                if dsp.quests.getStage(player, this_quest) >= 1 then
```
takhlaq: for these 'magic numbers' as i called them previously, we can probably get away with a local define in the script for the stage, though we may want a callback on `setStage` to clear any vars from the previous one

scripts/globals/quests.lua
ibm2431: Whoops, don't know how I didn't catch that. Surprised Lua didn't yell at me.

scripts/quests/adoulin/a_certain_substitute_patrolman.lua
ibm2431: Are we doing non-stage vars per stage? I was going to have completeQuest nuke all left-over vars (that aren't set to be preserved) just to make sure all the vars have been cleaned up.

ibm2431:
I like vars as tables. Then each var for a quest could be stored in the DB as [Q][A][I][V]. It also lets us do some shenanigans with some `!checkQuestVars` (loop through all the "vars" and print their key-value).

We might run into issues with -1 and 0, as [setting a playerVar to zero deletes it from the db](https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_baseentity.cpp#L568).

Do we want the final stage to be 0xFFFF to signify completion? We'd end up with a lot of leftover "quest done" vars in playerVars as players complete quests. I thought the stage var would be nuked upon completion, as even without it, we could still check if a quest was completed the usual way.

scripts/globals/quests.lua
TeoTwawki: depending on the lua interpret used it sometimes allows, still we prefer the comma for consistency :+1:

TeoTwawki:
@takhlaq @ibm2431 pls def do not start doing -1 / 0 / 0xfff.. to the sql vars. leftover vars for eternity no thx. in the lua vars, sure.

Also whatever format you guys wind up with, please make sure its easy to follow when server owners wind up with ppl who got their vars fckd need to fix things. Sometimes when people multibox one char gets the KIs and somehow a completely diff character gets the variable update, leaving one or both in a broken state. I'd also advise against heavy use of masks like we do for lure of the wildcat for that same reason. I've noted that f's up so often when ppl multibox. ugh. it shouldn't but it does. Mainly chinese players..I swear its like half the country plays from the same internet cafe and all the connections get some sort of crosstalk.

ibm2431:
Git rebase fixup and force-pushed the commas and removing the needless extra fame inner-tables.

As I was lying in bed this morning thinking about stages, I pondered on stages each being their own tables with onTriggers and onAdvance functions, with all the NPC triggers split up amongst the stages in chronological order.

A stage's onAdvance would do all the messy work of deleting/adding key items, variables, etc (ex: Exchanging the Broken Harpoon KI for the Extravagant Harpoon KI). Then instead of a `dsp.quests.setStage(magic_numbers)` call after messing with KIs in the CS event finish, just do a call to `dsp.quests.advanceStage()`:

```
[dsp.quests.enums.stages[1]] {
    npcTriggers {
        ["Western Adoulin"] {
            ["Jorin"] = function(player)
                player:startEvent('Demand player hurry up')
            end
            ["Shippilolo"] = function(player)
                if (has broken harpoon) then player:startEvent(666) end
            end
        }
    }
    eventFinishes {
        ["Western Adoulin"] {
            [666] = function(player, csid, params)
                dsp.quests.advanceStage(player, this_quest)
            end
        }
    }
    onAdvance = function(player)
        delete broken harpoon KI
        give extravagant harpoon KI
    end
}

------------------

dsp.quests.advanceStage = function(player, quest)
    local on_stage = dsp.quests.getStage(player, quest)
    local on_advance = quest.stages[on_stage].onAdvance
    if on_advance then
        on_advance(player)
        dsp.quests.setStage(player, quest, on_stage + 1)
    else
        dsp.quests.completeQuest(player, quest) -- Gives rewards and nukes vars
    end
end
```

We could add yet another GM debug command: `!nextstage`, which just invokes the current stage's onAdvance. Then when a server operator, like Teo, is getting complaints about a quest being stuck for some reason, they can just advance the player to the next stage of it. Or testers could use it to quickly get to a certain phase of the quest for testing something.

takhlaq:
>I pondered on stages each being their own tables with onTriggers and onAdvance functions, with all the NPC triggers split up amongst the stages in chronological order.

i think we discussed that in a previous pr but decided not to
though i cant think of any other way to handle this

your proposed setup sounds good, though we might want to handle different progression routes like so
```lua
this_quest.stages =
{
  [1] = 
  {
    ...
    onAdvance = function (player, quest, params)
      -- if params is supplied, check whether this is being advanced from trade/talk/mob kill etc. if the next stage differs due to the method of progression for whatever reason
      -- this would include choosing the reward based on any options in cutscenes (maybe through a rewards variable or helper function)
    end,
  }
}

...
dsp.quests.advanceStage = function(player, quest, params)
    local on_stage = dsp.quests.getStage(player, quest)
    local on_advance = quest.stages[on_stage].onAdvance
    if on_advance then
        on_advance(player, quest, params)
        dsp.quests.setStage(player, quest, on_stage + 1)
    else
        dsp.quests.completeQuest(player, quest) -- Gives rewards and nukes vars
    end
end
```

ibm2431:
@takhlaq I've been using this PR still being open as an excuse not to work on DSP, but I suppose I've had long enough to play vidya and should nudge you.

takhlaq:
i was doing the exact same thing

PR #5510, opened 2018-12-23
Add Assault Excavation Duty

KnowOne134:
basic of the assault are in just missing the mine mechanics. working on that still


KnowOne134:
So i added trying to use the bomb but something is fishy, my captures gave the bomb an id but it doesnt actually exist as an npc or a mob in polutilis. gives my a null value anytime i reference it so this is as far as i can go

scripts/globals/items/qiqirn_mine.lua
```diff
@@ -0,0 +1,27 @@
...
+
+function onItemCheck(target)
+    local result = 0
+    if target:getZoneID() == LEBROS_CAVERN then
+        result = 308
+    end
+    return result
+end
+
+function onItemUse(target,npc,mob)
```
TeoTwawki: you can't just make up parameters. look at the definition of onItemUse in the core.

TeoTwawki:
> What was the ID? It could be a "generated" mob/ID like summons and trusts.
He posted it in discord. it is "generated" type.

KnowOne134:
I reverted the bomb stuff til its figured out. The rest of the assault is solid though

KnowOne134:
the bomb now shows up when item is used 1 per area. i cannot get them to agro the doors or assist me. if they did agro the countdown would start and then use mine blast on target and despawn. I'm at a loss on that part but maybe someone else can get that. feel its a good 90% + done

scripts/zones/Lebros_Cavern/mobs/Qiqirn_Mine.lua
```diff
@@ -0,0 +1,52 @@
...
+function onMobRoam(mob)
+    dsp.ally.startAssist(mob, dsp.ally.ASSIST_PLAYER);
+end;
+
+function onMobFight(mob,target)
+    local instance = mob:getInstance()
+    local players = instance:getChars();
+
+    for i,v in pairs(players) do
+        if mob:getLocalVar("Time") == 0 then
```
ibm2431: I'll be the first to admit I don't do mobs, but you might be able to get away with setting the timers all at once, without needing to increment/check your time var every `onMobFight()`. If you can't do so in `onMobSpawn()`, another option in `onMobFight()` is to run a local var like `bomb_set`, and just check for `if not bomb_set then -- set the timers to explode the bomb end`

scripts/zones/Lebros_Cavern/mobs/Qiqirn_Mine.lua
```diff
@@ -0,0 +1,52 @@
...
+            mob:timer(8000, function(mob) v:messageSpecial(ID.text.MINE_COUNTDOWN,2) end)
+            mob:setLocalVar("Time",5)
+        elseif mob:getLocalVar("Time") == 5 then
+            mob:timer(9000, function(mob) v:messageSpecial(ID.text.MINE_COUNTDOWN,1) end)
+            mob:setLocalVar("Time",6)
+        elseif mob:getLocalVar("Time") == 6 then
+            mob:timer(10000, function(mob) mob:useMobAbility(1838) end)
+            mob:setLocalVar("Time",7)
+        elseif mob:getLocalVar("Time") == 7 then
+            mob:timer(12000, function(mob) mob:setHP(0) end)
```
ibm2431: `Time` doesn't extend past 7; will this be setting up a lot of `setHP()` calls after `Time` has reached 7, but `onMobFight()` keeps getting called?

ibm2431:
> i cannot get them to agro the doors or assist me. if they did agro the countdown would start and then use mine blast on target and despawn. I'm at a loss on that part but maybe someone else can get that

Standard disclaimer I have zero mob experience, but [mob:engage(target) might be an option](https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_baseentity.cpp#L9519). Determine where the bomb is (perhaps set a var of "which rock am I blowing up" on spawn by checking coordinates), and on mob spawn/fight/whatever, engage on the appropriate rock.

scripts/zones/Lebros_Cavern/mobs/Qiqirn_Mine.lua
KnowOne134: Once hit hits hp 0 it dies and despawns. Voiding the onmobfight. Next bomb set will start counter over again. Wash repeat

scripts/zones/Lebros_Cavern/mobs/Qiqirn_Mine.lua
KnowOne134: Onmobfight only checks every 3 seconds. This way i have set will do a count every 1 second as i saw it in retail

scripts/zones/Lebros_Cavern/mobs/Qiqirn_Mine.lua
ibm2431: From what I understand, `onMobFight` gets called every combat tick, and lua timers are non-blocking~! So until the mob actually dies (12 seconds after spawn), a lot of non-blocking timers to invoke `setHP` will be set up (since Time never gets set to 8 or anything).

KnowOne134: Yea i can easily add another var set

wrenffxi:
Sorry I stole your groupid. ><

wrenffxi:
I love that you're filling the mob_groups gap from the bottom up, since I keep bein' a jerk and adding mob_groups to the top of it.

Another mob_pools conflict incoming. ><

KnowOne134:
Ill just close till your done. Or you can skip a number for me

whasf:
And this? Except for conflict, good to go?

KnowOne134:
I made my own conflict ><

KnowOne134:
Bump

TeoTwawki:
blind merging I give zero f\*\*\*'s about anyones standards now

PR #5530, opened 2018-12-31
Add Additional Effect: "Stun" for Stun Claws/+1

nfaletra:
Noticed these were missing. Copied proc rates from Mamushito.


TeoTwawki:
missing because I kept asking ppl to wait for the refactor I was working on. the one I tried to talk you into taking over

PR #5535, opened 2019-01-01
Added Vs. Ecosystem Latents

nfaletra:
Left out some Additional Effects because they require additional research into the correct sub effects and proc rates.

There are other "vs. [blank]" text that are more specific than ecosystem. E.G. Orcs, Quadavs, slimes, tigers, etc. Left those ones out, as well. Should probably be a separate latent ID.

### DIFF
--- a/scripts/globals/items/a_loutrance.lua
+++ b/scripts/globals/items/a_loutrance.lua
@@ -1,0 +1,25 @@
+-----------------------------------------
+-- ID: 18041
+-- Item: A l'Outrance
+-- Additional Effect vs. beasts: Darkness Damage
+-----------------------------------------
+require("scripts/globals/magic")
+require("scripts/globals/msg")
+require("scripts/globals/status")
+-----------------------------------------
+
+function onAdditionalEffect(player, target, damage)
+    local dmg = player:getStat(dsp.mod.INT) - target:getStat(dsp.mod.INT)
+    if dmg > 25 then
+        dmg = 25 + (dmg - 25) / 2
+    end
+    
+    local params = {}
+    params.bonusmab = 0
+    params.includemab = false
+    dmg = addBonusesAbility(player, dsp.magic.ele.DARK, target, dmg, params)
+    dmg = dmg * applyResistanceAddEffect(player, target, dsp.magic.ele.DARK, 0)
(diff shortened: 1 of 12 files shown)


scripts/globals/items/a_loutrance.lua
```diff
@@ -0,0 +1,25 @@
...
+-- Item: A l'Outrance
+-- Additional Effect vs. beasts: Darkness Damage
+-----------------------------------------
+require("scripts/globals/magic")
+require("scripts/globals/msg")
+require("scripts/globals/status")
+-----------------------------------------
+
+function onAdditionalEffect(player, target, damage)
+    local dmg = player:getStat(dsp.mod.INT) - target:getStat(dsp.mod.INT)
```
TeoTwawki: Add effect melee items don't calculate this way. ranged items do.

TeoTwawki: also still making more work for me for the global thing..

nfaletra: Ascalon specifically said it uses dMND for damage, but who knows. I'll just remove all the additional effect ones for now.

TeoTwawki: could be special case, could be shitWikiSays™

I did not see any diff from stacking INT or MND on:
Hydro Baghnakhs, Inferno Axe, Burning Cesti, Burning Cesti, Spark Spear, Spark Spear +1, Estramacon, Fire Sword, Pixie Mace, Holy Mace, Holy maul, Holy Maul +1, Cougar Baghnakhs, Shiva's Claws, Sonic Knuckles, or Fuscina

I also was looking for player lv vs mob lv to modify damage. I saw no difference.

So I thought either it doesn't effect these or the range is narrow and capped to either a maximum or minimum. I tested for level difference to. Neither my level nor the mobs level had any visible impact. Certain weapons hit notably harder and in a wider random range. Very possible there are diff rules for certain weapons and also very possible that someone got a big hit and didn't really parse a diff (I haven't seen any data corroborating Ascalon).

TeoTwawki: p.s.
in the event we can show ascalon DOES work that way, I'm inclined to think all the "vs <family>" will be that way as well. All the ones I tested were not vs any mob family,

PR #5536, opened 2019-01-01
BaseSell higher than shop sell price

ShelbyZ:
- This would allow players to buy low and sell high adding gil to the
economy
- prices are adjusted to 75% of sell price


hilts-vaughan:
There are some others that are wrong as well that I know of -- such as snapping mole for example: https://ffxiclopedia.fandom.com/wiki/Snapping_Mole

What do we do to handle these?

ShelbyZ:
The python script I created should have caught all known shop and guild shops where you could buy at a lower price than the sell.  I can see if Snapping Mole is an outlier and adjust otherwise it should have been taken care of in this or another (already merged PR).

hilts-vaughan:
I should clarify. You can't buy them. They are just drops that sell for more than they should.

PR #5542, opened 2019-01-03
separate light sleep immunity mask

godmode:
### Description
Some mobs should be immune to one form of sleep (light or dark) but not others. For example Giant Birds should be immune to Sleep/Sleep2 but not Lullaby. Since this needs significant time to research and update the database with changes over time it is backward compatible with the current mask `IMMUNITY_SLEEP`.

To improve on this in the future someone could separate repose and light shot into their own effect and add those effects to the switch in `status_effect_container.cpp`.

### Testing
Try Sleep/Sleep II/Lullaby on:
- Normal mobs with no resist
- Existing mobs with `IMMUNITY_SLEEP` such as Bomb Queen
- Perytons (should be resistant to all forms of Sleep except Lullaby in this PR)


godmode:
It doesn’t but it could if someone makes a light sleep effect and groups in all of those things. Making it the same as lullaby would give the message “receives the effect lullaby” so someone needs to make it give the proper message while having its own unique effect ID.

godmode:
@hozu I think that could work but seems a bit of a hacky way to do it

TeoTwawki:
Note 1:
Immunity and reistance are different things.

Players often say "immune" when resitance is set so high as to be impossible to land a status even at max siill with every consievable gear buff.

Note 2:
There ARE multiple intances of "sleep" status effects in the client. A retail capture can confirm which soells deliver which effect.

Note 3:
We ahoukd be checking element resist AND status effect resistance, for resist cases.

godmode:
This is specifically for immunity. It’s a fairly common thing that mobs would be immune to either light or dark sleep. Ex: skeletons. There is already a sleep and sleep 2 and lullaby effect. 

This is not a PR to write all of these effects into this. It Is simply To use the current effects as well as to lay the groundwork to be able to use the new sleep effects captured and implemented.

TeoTwawki:
I didn't think it necessary to quote people to show why I was explaining the difference between immunity and resistance - that was not about your code

TeoTwawki:
I didn't think it necessary to quote people to show why I was explaining the difference between immunity and resistance - that was not about your code

TeoTwawki:
Thanks @nfaletra I guess EFFECT_SLEEP_II must be for something really special.

nfaletra:
The "Tier" of the sleep effect is already passed as the power. Just need to add the element as SubPower and check that in core.

TeoTwawki:
You already said that.

godmode:
First thing are the new options added good? 2048 and 4096 and the constant names. The changes you’re requesting make me think of using a helper function when sleep is the effect. Is that how you’d like to see the code? If so I will close this PR and just start new.

nfaletra:
The new immunity flags are fine. You can collapse the switch cases so lullaby is included with the sleeps. I was thinking something like:

```c
        case EFFECT_SLEEP:
        case EFFECT_SLEEP_II:
        case EFFECT_LULLABY:
            if (m_POwner->hasImmunity(IMMUNITY_SLEEP))
                return false;

            uint16 subPower = PStatusEffect->GetSubPower();
            if (subPower == ELEMENT_DARK && m_POwner->hasImmunity(IMMUNITY_DARKSLEEP))
                return false;

            if (subPower == ELEMENT_LIGHT && m_POwner->hasImmunity(IMMUNITY_LIGHTSLEEP))
                return false;

            break;
```

wrenffxi:
I think effect SLEEP_II is used for [Nightmare](https://ffxiclopedia.fandom.com/wiki/Nightmare).  It's a weird case where DoTs don't wake you (or mobs, if cast by avatar), and getting punched doesn't wake you up (not true of avatar version).  Note the bit at the bottom.  It's overwritten by Sleep II and Sleepga II, but not by "other Sleep II-class sleeps"

Another example of "Sleep II-class" would be Nightmare Mandragoras using [Dream Flower](https://ffxiclopedia.fandom.com/wiki/Dream_Flower) in Dynamis - Buburimu.  But not the player version of Dream Flower.

godmode:
@wrenffxi interesting. We got around that on Eden by making a DEEPSLEEP status effect. Would be neat to see if this is how retail handles it for sure.

nfaletra:
@wrenffxi 
Nightmare does not use effect id 19 (SLEEP_II in DSP) it still uses id 2 (SLEEP in DSP). I actually had to level SMN to 29 on retail to check. XD

nfaletra:
I did not get a chance to test any BLU spells, but after talking with other OPs about it, SLEEP_II isn't used anywhere because there are some client side things going on with those status effects and SLEEP_II didn't actually do anything. Could probably use some more testing to see if anything has changed recently. My suspicion is it's a stale development id that was never cleaned up.

ValentineXI:
Why was this closed? Needs reopening pl0x :P If GM doesn't want to continue, would you consider taking this on @nfaletra ?

PR #5565, opened 2019-01-08
ZNM Velionis

KnowOne134:
Add missing TP move Maledictation. has no auto attacks only tp moves and cast spells. Fastcast goes up as HP goes down. Perma blazespikes gave it its own spell_list as it only casts tier 3 and aga 2 not all blm spells. Pops with npcutilis now


scripts/zones/Arrapago_Reef/mobs/Velionis.lua
```diff
@@ -2,12 +2,16 @@
...
 -----------------------------------
 -- Todo: blaze spikes effect only activates while not in casting animation
+function onMobInitialize(mob)
+    mob:setMobMod(dsp.mobMod.IDLE_DESPAWN, 300)
+end
 
 function onMobSpawn(mob)
+    mob:setLocalVar("[rage]timer", 3600) -- 60 minutes
     mob:SetAutoAttackEnabled(false)
     mob:addMod(dsp.mod.FASTCAST,15)
```
TeoTwawki: `setMod()`

scripts/zones/Arrapago_Reef/mobs/Velionis.lua
```diff
@@ -0,0 +1,35 @@
...
+    mob:setLocalVar("[rage]timer", 3600) -- 60 minutes
+    mob:SetAutoAttackEnabled(false)
+    mob:setMod(dsp.mod.FASTCAST,15)
+    mob:setLocalVar("HPP", 90)
+    mob:setMobMod(dsp.mobMod.MAGIC_COOL,5)
+end
+
+function onMobFight(mob,target)
+    local FastCast = mob:getLocalVar("HPP")
+    if mob:hasStatusEffect(dsp.effect.BLAZE_SPIKES) == false then
```
TeoTwawki: um, what?

if its not removable we have a flag to prevent that, no need to re-apply the effect every tic.

PR #5573, opened 2019-01-10
misc TP moves

KnowOne134:
just some misc captures i took


scripts/globals/mobskills/perdition.lua
```diff
@@ -0,0 +1,20 @@
...
+require("scripts/globals/msg")
+---------------------------------------------
+
+function onMobSkillCheck(target,mob,skill)
+    return 0
+end
+
+function onMobWeaponSkill(target, mob, skill)
+
+    skill:setMsg(dsp.msg.basic.FALL_TO_GROUND)
```
TeoTwawki: should this check for  `dsp.mod.DEATHRES`?

scripts/globals/mobskills/eternal_damnation.lua
```diff
@@ -0,0 +1,19 @@
...
+require("scripts/globals/monstertpmoves")
+require("scripts/globals/status")
+---------------------------------------------
+
+function onMobSkillCheck(target,mob,skill)
+    return 0
+end
+
+function onMobWeaponSkill(target, mob, skill)
+    local typeEffect = dsp.effect.DOOM
```
TeoTwawki: no need to var this, its unchanging and doesn't involve a function call. it just evals to a number and its already loaded.

scripts/globals/mobskills/proboscis_shower.lua
```diff
@@ -0,0 +1,26 @@
...
+require("scripts/globals/msg")
+---------------------------------------------
+
+function onMobSkillCheck(target,mob,skill)
+    return 0
+end
+
+function onMobWeaponSkill(target, mob, skill)
+    local potency = skill:getParam()
+
```
TeoTwawki: looks like pure copy pasta. at least comment where its from and that we don't have a proper formula.

scripts/globals/mobskills/perdition.lua
KnowOne134: Ive been hit with it 2 times in retail. Both times i was lvl 99 with full 119 gear and the mob was 68 and 74 and it instantly dropped me both times

TeoTwawki: 2 times is a) a pretty small sample and b) if you aren't wearing any death resist gear it can't proc and save you (there is no native source of it)

TeoTwawki: honestly I didn't know if it should be checked here but your reasoning for not doesn't really answer that.

TeoTwawki: I think we should assume yes until data says no:
https://github.com/DarkstarProject/darkstar/blob/8b920c7a9ad7db8a5f8e62dac24c32590728ff54/scripts/globals/spells/death.lua#L15

KnowOne134: Ill add it tonight. My retail account has expired now to do anymore testing

scripts/globals/mobskills/perdition.lua
```diff
@@ -12,6 +12,10 @@ function onMobSkillCheck(target,mob,skill)
 end
 
 function onMobWeaponSkill(target, mob, skill)
+    if (target:hasStatusEffect(dsp.effect.MAGIC_SHIELD) or math.random(0,99) < target:getMod(dsp.mod.DEATHRES)) then
+        spell:setMsg(dsp.msg.basic.MAGIC_NO_EFFECT)
```
TeoTwawki: mobskill is not a spell. less copy pasta son.

PR #5574, opened 2019-01-10
Correct spawn behavior of Lumber Jack - he does not spawn with enmity

TeoTwawki:
Move far enough away before the Weeping Willows dead body fades and Lumber Jack doesn't even notice you are there, will wander around till you engage, catch agro from walking to close in his line of sight, or his 10 min despawn timer is up. so no `updateEnmity()` needed here.

### DIFF
--- a/scripts/zones/Batallia_Downs/mobs/Weeping_Willow.lua
+++ b/scripts/zones/Batallia_Downs/mobs/Weeping_Willow.lua
@@ -18,5 +18,7 @@ function onMobDeath(mob, player, isKiller)
 end
 
 function onMobDespawn(mob)
+    -- Retail behavior is for it to walk back to where willow died if unclaimed *unless* willow was pulled down the cliff
+    -- In that case, it will walk back near where Willow was spawned at.
     GetMobByID(mob:getID() + 6):setSpawn(mob:getXPos(), mob:getYPos(), mob:getZPos())
 end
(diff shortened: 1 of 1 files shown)


scripts/zones/Batallia_Downs/mobs/Weeping_Willow.lua
```diff
@@ -5,17 +5,20 @@
...
 
 function onMobDeath(mob, player, isKiller)
-    local JACK = mob:getID()+6;
-    SpawnMob(JACK):updateEnmity(player);
+end
+
+function onMobDespawn(mob)
+    local JACK = mob:getID()+6
+    SpawnMob(JACK)
     GetMobByID(JACK):setPos( mob:getXPos(), mob:getYPos(), mob:getZPos(), 0);
```
zynjec: setSpawn, use it or lose it sister.

TeoTwawki: breaks behavior of him walking back up the cliff if pulled down so I wasn't sure I should make that change (I actually did then reverted it 2nd guessing myself). to get him 100% retail he should anchor to to where willow died *except* when willow dies down the cliff side at which point he goes back to the area where willow spawned at.

Still want me to switch it?

takhlaq: leave a comment in the script for now with retail behaviour vs current

TeoTwawki: done

PR #5584, opened 2019-01-11
stop using getShortID for the value of a mobmod where that value never matters

TeoTwawki:
This began as just being copy pasted from an existing script from an author that isn't around, we have no idea what his original intent was but whatever it was, it never happened. the value is never used the code just checks if this is zero or non zero.


TeoTwawki:
gee if only Wren had accepted a dev position, I could merge now..

:cry:

KnowOne134:
You ungrateful he-bitch

TeoTwawki:
well, that escalated quickly!

TeoTwawki:
AH WUZ DRUNK! SO WUZ DA BABY MAWMA!

PR #5598, opened 2019-01-13
Woodland zone znms

KnowOne134:
adding idle despawn and 1 hour rage. Barrier tusk is 90 seconds and only when tusks not broken sub(0). Flailing trunk is a 3 hit move can wipe 0 to 3 shadows. Add id rushing drub can be used by any mamool that wields a staff and sub(0). Add in stave toss for Gotoh which is what triggers its phase change to whm mode. the skill itself now forces animation sub (1). Tinin just has style edit and idle despawn and rage


scripts/zones/Wajaom_Woodlands/mobs/Gotoh_Zha_the_Redolent.lua
```diff
@@ -2,7 +2,11 @@
 -- Area: Wajaom Woodlands
 --  ZNM: Gotoh Zha the Redolent
 -----------------------------------
-mixins = {require("scripts/mixins/job_special")}
+mixins =
```
TeoTwawki:
```lua
mixins = 
{
    require("scripts/mixins/job_special"),
    require("scripts/mixins/rage")
}
```

scripts/zones/Wajaom_Woodlands/mobs/Gotoh_Zha_the_Redolent.lua
```diff
@@ -69,14 +75,14 @@ end
 function onCriticalHit(mob)
     local RND = math.random(1, 100)
     if mob:AnimationSub() == 0 and RND <= 10 then
-        mob:AnimationSub(1)
```
TeoTwawki: wrong

scripts/zones/Wajaom_Woodlands/mobs/Gotoh_Zha_the_Redolent.lua
```diff
@@ -69,14 +75,14 @@ end
...
+        mob:useMobAbility(2361)
     end
 end
 
 function onWeaponskillHit(mob, attacker, weaponskill)
     local RND = math.random(1, 100)
     if mob:AnimationSub() == 0 and RND <= 10 then
-        mob:AnimationSub(1)
```
TeoTwawki: way wrong

scripts/globals/mobskills/rushing_drub.lua
```diff
@@ -7,23 +7,24 @@
...
-
-require("scripts/globals/settings")
 require("scripts/globals/status")
 require("scripts/globals/monstertpmoves")
-
 ---------------------------------------------
 
 function onMobSkillCheck(target,mob,skill)
-    return 0
+    if (mob:getMainJob() == dsp.job.BLM or mob:getMainJob() == dsp.job.WHM) and mob:AnimationSub() == 0 then
```
TeoTwawki: what if I'm coded as a war/whm with stave?

I think there is a battlefield mob used when the AC is taken coded such a way. I could be wrong but its more flexible to just have a skill list for stave users than to tie it to main job ID, since SE has been known to do odd things like that.

Like the a certain gigas monk with the great axe, or a warrior behemoth casting meteor.

KnowOne134: can make another pull separating mamool jobs at a later date

PR #5600, opened 2019-01-14
Scintillant Lance

TeoTwawki:
No clue if resulting dmg is anywhere close to retail, numbers entirely made up. All I know is its conal light damage and wipes shadows

¯\\\_(ツ)\_/¯

**edit: which is no different from any of our other 500+ mobskills that are guessed at and eyeballed with no retail verification of the maths**


nfaletra:
Well, what’s the resulting damage?

TeoTwawki:
about the same as Wind Cutter. :wink:

PR #5606, opened 2019-01-16
ShowDebug to ShowInfo on delivery box actions, for filtering purposes

TeoTwawki:
This is about as useful as other ShowInfo type messages, which is to say almost never. I'd rather the log not be flooded by delivery box info when I need to see an actual debug message.


TeoTwawki:
I think it fits better in with the other info messages. In the map server they are all actions just like this. A new flag feels like overkill unless we have something specific like when I added the navmesh one. I've been meaning to review all our ShowWhatevers and make sure they have the appropriate category most were just tossed into debug.

TeoTwawki:
Actually nvm I'm wtong way more of this than I thought the nee one makes more sense afterall but I'll reopen later I gotta sleep first

PR #5607, opened 2019-01-16
ability(empathy): Add logic for Empathy

davismj:
Gives 200 experience points to the Wyvern and copies one status effect from the Dragoon to the Wyvern when Spirit Link is used for each merit point in Empathy.

This still needs a bit of testing, but I wanted to make sure this was on the right track first.

### DIFF
--- a/scripts/globals/abilities/spirit_link.lua
+++ b/scripts/globals/abilities/spirit_link.lua
@@ -7,6 +7,7 @@
 -----------------------------------
 require("scripts/globals/status")
 require("scripts/globals/msg")
+require("scripts/globals/wyvern")
 -----------------------------------
 
 function onAbilityCheck(player,target,ability)
@@ -69,6 +70,25 @@ function onUseAbility(player,target,ability)
     if (pet:getHP() < pet:getMaxHP()) then -- sleep is only removed if it heals the wyvern
         removeSleepEffects(pet)
     end
+
+    local empathy = player:getMerit(dsp.merit.EMPATHY)
+    if (empathy > 0) then
+        local effects = utils.shuffle(player:getStatusEffects())
+        local copied = 0
+        -- find a buff and copy it to the wyvern
+        for i, effect in ipairs(effects) do
+            if canCopyEffectToWyvern(effect) then
+                copyEffectToWyvern(pet, effect)
+                -- up to a max of empathy merits
+                copied = copied + 1
+                if (copied == empathy) then
+                    break
+                end
+            end
+        end
+        addWyvernExp(pet, empathy * 200)
+    end
+
 
     pet:addHP(healPet) --add the hp to pet
     pet:addStatusEffect(dsp.effect.REGEN,regenAmount,3,90,0,0,0) -- 90 seconds of regen
--- a/scripts/globals/abilities/spirit_link.lua
+++ b/scripts/globals/abilities/spirit_link.lua
@@ -7,7 +7,7 @@
 -----------------------------------
 require("scripts/globals/status")
 require("scripts/globals/msg")
-require("scripts/globals/wyvern")
+require("scripts/globals/pets/wyvern")
 -----------------------------------
 
 function onAbilityCheck(player,target,ability)
--- a/scripts/globals/pets/wyvern.lua
+++ b/scripts/globals/pets/wyvern.lua
@@ -287,14 +287,14 @@ function addWyvernExp(wyvern, exp)
 end
 
 function canCopyEffectToWyvern(effect)
-    return empathyEffects[effect.getType()] ~= nil
+    return empathyEffects[effect:getType()] ~= nil
 end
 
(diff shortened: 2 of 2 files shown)


scripts/globals/pets/wyvern.lua
```diff
@@ -47,6 +261,43 @@ function doHealingBreath(player, threshold, breath)
...
+            master:addMod(dsp.mod.DEFP, 4 * diff)
+            master:addMod(dsp.mod.HASTE_ABILITY, 200 * diff)
+        end
+        pet:setLocalVar("wyvern_exp", prev_exp + exp)
+        pet:setLocalVar("level_Ups", pet:getLocalVar("level_Ups") + diff)
+    end
+end
+
+function canCopyEffectToWyvern(effect)
+    return empathyEffects[effects.getType()] ~= nil
```
wrenffxi: effect.getType()

scripts/globals/abilities/spirit_link.lua
```diff
@@ -7,6 +7,7 @@
 -----------------------------------
 require("scripts/globals/status")
 require("scripts/globals/msg")
+require("scripts/globals/wyvern")
```
wrenffxi: scripts/globals/pets/wyvern

scripts/globals/pets/wyvern.lua
takhlaq: effect:getType()

davismj:
thanks for taking the time to read my crappy code

takhlaq:
im hesitant to merge this until it's been tested as that error would've been picked up straight away

davismj:
yep, but overall is the strategy and style good?

scripts/globals/abilities/spirit_link.lua
```diff
@@ -70,8 +71,27 @@ function onUseAbility(player,target,ability)
         removeSleepEffects(pet)
     end
 
+    local empathy = player:getMerit(dsp.merit.EMPATHY)
+    if (empathy > 0) then
+        local effects = utils.shuffle(player:getStatusEffects())
```
takhlaq: the shuffle might not be needed if we iterate `in pairs` www.luafaq.org/#T1.10

scripts/globals/pets/wyvern.lua
```diff
@@ -33,6 +33,220 @@ local wyvernTypes = {
...
+        -- i.e. Probably should exclude Barrage
+-- SPELLS
+    -- Includes battle spell buffs, e.g. stoneskin, spikes, regen, utsusemi, samba, minuet, rolls, sambas
+        -- i.e. Probably should include Migawari, Avatar's Favor
+    -- Excludes non-battle spell buffs, e.g. Monomi, Invisible
+        -- i.e. Probably should exclude Mazurka, Jigs
+-- CONSUMABLES
+    -- Includes attribute boosting medicines, e.g. Icarus Wing, Agility Potion, Giant's Drink
+    -- Excludes food, i.e. Instant Noodles, Meat Mithkabob
+local empathyEffects = {
```
takhlaq: this table could be moved to a flag in src/map/status_effect.h along with the flag set in status_effect db table

scripts/globals/abilities/spirit_link.lua
davismj: I'm only marginally familiar with Lua. Does "in no particular order" imply random? Will multiple consistently yield different results?

scripts/globals/pets/wyvern.lua
davismj: Yes, I was thinking about that. I didn't know if that was the best approach as it seems that new status effects that have been added might have been mis-flagged by contributors who weren't familiar with the flags, e.g. Celerity is flagged `EFFECTFLAG_DEATH` but may also need to be flagged `EFFECTFLAG_MAGIC_END` (but my Russian is rusty and I'm not sure I fully understand the flag).

Do you think it is worth it to make it a flag?

takhlaq: update status_effects set flag = flag | <val> where effectid in(list, of, status, ids)

scripts/globals/abilities/spirit_link.lua
wrenffxi: Using pairs to iterate over a non-contiguous-integer indexed table is not reliable for producing the as-written-in-code order, but it's also not exactly random, so I'd stick with Utils.shuffle.

Interesting side note: the reason this happens is because string-indexed LUA tables use hashes of the strings to order the table, and LUA now randomizes the seed on every new program run to avoid [hash vulnerability](http://lua-users.org/lists/lua-l/2012-01/msg00497.html) that used to exist.  Side effect: your string-indexed tables get mixed around every run.  But within a run, they'll always be ordered the same.

zynjec:
Closing for now, feel free to open another PR once you're able to make these changes.

Issue #5531, opened 2019-01-01
Mob level range is [min,max) when it should be [min,max]


DiscipleOfEris:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30181205_0

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Mobs with a level range of [min, max] are only spawning with a level range of [min, max-1]. To reproduce, use some method of revealing the level of mobs (e.g. widescan or /check packets), such as my MobLevel addon, and just wander around inspecting mobs. Then compare the levels against those found in `mob_groups.sql`. You will never find a mob at the maximum end of its level range unless max==min. I've seen this on the ffera.com server.

https://github.com/DarkstarProject/darkstar/blob/e3a29080064d709d2e3218312acc994566ccd512/src/map/entities/mobentity.cpp#L518-L524

Lines 518-524 should be changed to:

    // Generate a random level between min and max level
    uint8 level = dsprand::GetRandomNumber(m_minLevel, m_maxLevel+1);

The `GetRandomNumber(min, max)` function uses a half closed interval. That is, [min, max). For integers, this means the maximum value the number generator will yield is max-1, unless max==min.

https://github.com/DarkstarProject/darkstar/blob/e3a29080064d709d2e3218312acc994566ccd512/src/common/dsprand.h#L18

Sorry this isn't a pull request. I haven't set up DSP for development yet, so I can't actually test to confirm with certainty that my suggested fix will work.

wrenffxi:
this can close, was fixed by #5610

PR #5610, opened 2019-01-16
fix some dsprand::getRandomNumber() range assumptions

wrenffxi:
For reviewers, the logic changed in battleutils.cpp is on lines: 753, 797, 805.  Everything else is tab->space.

For #5531


TeoTwawki:
@teschnei 
```
[11:47 AM] Wren: the lua dependency on the core adds +1 to max when requesting dsprand:getRandomNumber()
```
you figure out wtf we should be doing kthx bai _runs away_

src/map/ai/helpers/pathfind.cpp
```diff
@@ -349,7 +349,7 @@ bool CPathFind::FindPath(const position_t& start, const position_t& end)
 
 bool CPathFind::FindRandomPath(const position_t& start, float maxRadius, uint8 maxTurns, uint16 roamFlags)
 {
-    auto m_turnLength = dsprand::GetRandomNumber(1, (int)maxTurns);
+    auto m_turnLength = dsprand::GetRandomNumber(1, (int)maxTurns + 1);
```
nfaletra: Common practice in C++ is to add 1 to the range if you want [1, max]

So you can do:
`dsp.GetRandomNumber(maxTurns) + 1`

nfaletra:
To answer the question about lua adding +1 to the max mostly has to do with where indexes start between the two languages. C++ starts at 0 (like a boss) and LUA starts at 1 (like a jerk). Doing clean rand ranges based on the size of an array becomes much simpler for both. The disconnect is confusing, but as someone from a C/C++ background, I'm used to adding 1 to my ranges where needed.

teschnei:
I would suggest changing the min_level/max_level function to just use
GetRandomNumber(min_lev,max_level+1) instead of adding the offset, it looks
better

wrenffxi:
It does this already, unless I'm misunderstanding how/where LUA is calling the function in the core.
https://github.com/DarkstarProject/darkstar/blob/ac1917100ffe811cd0de5f2c6ad4cbdc7dfe318c/src/map/lua/luautils.cpp#L319

wrenffxi:
That's what I did first, but there was some additional typecasting required that I was too lazy to futz with.  Didn't have the overload required for level = dsprand::GetRandomNumber(m_minLevel, m_maxLevel + 1)

src/map/ai/helpers/pathfind.cpp
wrenffxi: updated PR.

teschnei:
GetRandomNumber<uint8> maybe, but if that doesn't work it's not really that
important

src/map/entities/mobentity.cpp
```diff
@@ -518,9 +518,9 @@ void CMobEntity::Spawn()
     uint8 level = m_minLevel;
 
     // Generate a random level between min and max level
-    if (m_maxLevel != m_minLevel)
+    if (m_maxLevel > m_minLevel)
     {
-        level += dsprand::GetRandomNumber(0, m_maxLevel - m_minLevel);
+        level += dsprand::GetRandomNumber(0, m_maxLevel - m_minLevel + 1);
```
nfaletra: Like KJ said, we can make this just roll a range between min/max level. It's a templated function so you probably just need to declare the type explicitly for uint8.

Like this:
`level = dsprand::GetRandomNumber<uint8>(m_minLevel, m_maxLevel + 1);`

wrenffxi: didn't work.  it works if i cast both inputs from uint8 to int.
```uint8 level = dsprand::GetRandomNumber((int)m_minLevel, (int)m_maxLevel + 1);```

is that ok?

teschnei:
Just leave it how it is for now, we'll figure it out later

PR #5614, opened 2019-01-17
Enmity 2.0

nfaletra:
Used a combination of these enmity tables and forum posts from SE:
* https://kanican.livejournal.com/tag/enmity%20table!
* http://forum.square-enix.com/ffxi/threads/30629-Enmity-System-Explanation-and-Planned-Adjustments?p=420832&viewfull=1#post420832

* All enmity reductions were listed as being reduced by _approximately half_. So I reduced the listed abilities/spells by 45%.
* Cleaned up code regarding CE and VE calculations. Also made typing consistent across the board to avoid loss of data.
* Gutted m_enmityRange and let IsWithingEnmityRange handle NM ranges. Also switched the check to distanceSquared to remove a sqrt.

### Testing Done
* Tested pretty thoroughly with 2 characters using a good spread of abilities, weaponskills, and spells.
* Tested CE decay (taking damage from mob) while performing actions (heals, etc) on the target with hate.
* Did a lot of combinations of attacking while curing and debuffing, not attacking while curing and debuffing, etc.
* I tested a "PL" healing a low level player as well

In all testing cases, mob's hate behaved as I would have expected. It didn't appear to bounce around while actions against the mob or on other player were controlled. As soon I started spamming things and going wild, hate would move away from my "tank" or bounce around a lot.


src/map/enmity_container.cpp
```diff
@@ -309,12 +308,12 @@ void CEnmityContainer::SetCE(CBattleEntity* PEntity, uint16 amount)
     }
 }
 
-void CEnmityContainer::SetVE(CBattleEntity* PEntity, uint16 amount)
+void CEnmityContainer::SetVE(CBattleEntity* PEntity, const int32 amount)
```
teschnei: const value?

nfaletra: It's const correct for mutators to take in constant values. These functions shouldn't be modifying the value that goes in, it's not intuitive behavior.

src/map/enmity_container.cpp
```diff
@@ -329,18 +328,12 @@ void CEnmityContainer::SetVE(CBattleEntity* PEntity, uint16 amount)
...
 
-    uint16 mod = battleutils::GetEnmityModDamage(PEntity->GetMLevel()); //default fallback
-
-    if (m_EnmityHolder != nullptr) {//use the correct mod value
-        mod = battleutils::GetEnmityModDamage(m_EnmityHolder->GetMLevel());
-    }
-
-    auto CE = (int16)((80.0f / mod) * Damage);
-    auto VE = (int16)((240.0f / mod) * Damage);
+    int32 CE = 80 / battleutils::GetEnmityModDamage(m_EnmityHolder->GetMLevel()) * Damage;
```
teschnei: the truncation in these changes are different, is that intended? previously, truncation only happened at the end - now, truncation is happening before being multiplied by damage, which means that CE can only be a whole multiple of damage, and then VE can only be a whole multiple of (damage * 3)

src/map/enmity_container.cpp
```diff
@@ -354,21 +347,16 @@ void CEnmityContainer::UpdateEnmityFromDamage(CBattleEntity* PEntity, uint16 Dam
 *                                                                       *
 ************************************************************************/
 
-void CEnmityContainer::UpdateEnmityFromAttack(CBattleEntity* PEntity, uint16 Damage)
+void CEnmityContainer::UpdateEnmityFromAttack(CBattleEntity* PEntity, int32 Damage)
 {
-    if (m_EnmityList.find(PEntity->id) == m_EnmityList.end())
-    {
+    auto enmity_obj = m_EnmityList.find(PEntity->id);
```
teschnei: even better would be to do
```
if (auto enmity_obj = m_EnmityList.find(Pentity->id); enmity_obj != m_EnmityList.end())
{
    <do stuff with enmity_obj>
}
```
and drop the inverse condition with early return

src/map/enmity_container.cpp
teschnei: they can't modify the value that goes in because it's passed by copy

src/map/enmity_container.cpp
nfaletra: you right. I fix.

nfaletra: I'll fix other spots, too.

src/map/enmity_container.cpp
nfaletra: Not what I meant. If you call `SetThisThing(30)`, I expect ThisThing to equal 30 when it's done.

csundman:
We've been playing with this enmity update on our server for a couple days now and everything seems pretty weird now.  It just seems like the last person to do anything, be it cast regen... or use "meditate" just pulls hate.

TeoTwawki:
you guys are saying the same things I saw people saying _before_ this change. for like 3 years straight.

csundman:
Well, I can't say I know much about what people have said previously.  From my perspective aggro has been relatively predictable until I updated about 3 days ago.  Just thought I'd share the experience in case others had the same.  Maybe somebody will find something.  /shrug

src/map/enmity_container.cpp
```diff
@@ -139,16 +139,16 @@ void CEnmityContainer::UpdateEnmity(CBattleEntity* PEntity, int16 CE, int16 VE,
         if (enmity_obj->second.PEnmityOwner == nullptr) enmity_obj->second.PEnmityOwner = PEntity;
         float bonus = CalculateEnmityBonus(PEntity);
 
-        int16 newCE = (int16)(enmity_obj->second.CE + (CE > 0 ? CE * bonus : CE));
-        int16 newVE = (int16)(enmity_obj->second.VE + (VE > 0 ? VE * bonus : VE));
+        int32 newCE = (int32)(enmity_obj->second.CE + CE > 0 ? CE * bonus : CE);
```
csundman: I think this removing these parentheses busted order of operations. here

csundman:
It looks like some parentheses got removed on lines 142 and 143 of enmity_container.cpp and it may have busted order of operations.

UpdateEnmity is overwriting previous values instead of being additive.

src/map/enmity_container.cpp
nfaletra: This is a good catch. Fixing it now.

Issue #5621, opened 2019-01-19
Fighting Bomb Queen crashes server


KnowOne134:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 

**_Source Branch_** (master/stable) **:** 
Master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Spawn this big bitch and soon as she tries to bring in a goon of hers in to whomp on your ass some bam server crash.

PR #5622, opened 2019-01-19
fix bomb queen

wrenffxi:
For #5621 .
Fix crashiness.
Prevent queen from attempting to spawn multiple pets per cycle.
Prevent queen from attempting to spawn an already-spawned pet.
Clean up pets when queen dies.

### DIFF
--- a/scripts/zones/Ifrits_Cauldron/mobs/Bomb_Queen.lua
+++ b/scripts/zones/Ifrits_Cauldron/mobs/Bomb_Queen.lua
@@ -1,35 +1,53 @@
 -----------------------------------
 -- Area: Ifrit's Cauldron
 --   NM: Bomb Queen
+--  Vid: https://www.youtube.com/watch?v=AVsEbYjSAHM
 -----------------------------------
 local ID = require("scripts/zones/Ifrits_Cauldron/IDs")
 require("scripts/globals/status")
 -----------------------------------
 
 function onMobInitialize(mob)
+    mob:setMobMod(dsp.mobMod.IDLE_DESPAWN, 180)
     mob:setMobMod(dsp.mobMod.HP_STANDBACK, -1)
     mob:setMobMod(dsp.mobMod.DRAW_IN, 1)
-    mob:setMobMod(dsp.mobMod.IDLE_DESPAWN, 180)
+    mob:setMod(dsp.mod.STUNRES, 50)
 end
 
 function onMobSpawn(mob)
-    mob:addMod(dsp.mod.STUNRES, 50)
+    mob:setLocalVar("petCooldown", os.time() + 5) -- five seconds for first pet
 end
 
-function onMobFight(mob,target)
-    local battleTime = mob:getBattleTime()
+function onMobFight(mob, target)
+    -- every ~20 seconds after first pet, a new pet will spawn around queen's position.
+    -- 50% chance for Prince/Princess. 1% chance for Bastard.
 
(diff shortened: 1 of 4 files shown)


scripts/zones/Ifrits_Cauldron/mobs/Bomb_Queen.lua
```diff
@@ -1,35 +1,53 @@
...
+    -- every ~20 seconds after first pet, a new pet will spawn around queen's position.
+    -- 50% chance for Prince/Princess. 1% chance for Bastard.
 
-    if (battleTime - 5) % 20 == 0 then -- every 20 seconds after 5 seconds, source: https://www.youtube.com/watch?v=AVsEbYjSAHM
-        local random = math.random(1,100);
-        local petId = mob:getID() + 1 + math.floor(random / 25); -- approximately 1% chance to spawn bomb bastard
+    local mobId = mob:getID()
+
+    if os.time() > mob:getLocalVar("petCooldown") then
+        local petId = mob:getID() + 1 + math.floor(math.random(0, 100) / 25)
```
davismj: 0  to 100 technically makes it a 50/101 chance for the prince

davismj: That would make the minimum chance 4%, which is too high for bomb bastard. The best implementation would probably be `math.floor(math.random(0, 199) / 49)`, which would give a 24.5% chance for each prince, each princess, and 1% for bastard.

scripts/zones/Ifrits_Cauldron/mobs/Bomb_Queen.lua
```diff
@@ -1,35 +1,53 @@
...
         local pet = GetMobByID(petId)
-        local pos = mob:getPos()
-        pet:setSpawn(unpack(pos))
-        pet:spawn()
-        pet:updateEnmity(target)
-    end
 
+        if pet and not pet:isSpawned() then
```
davismj: if pet is spawned there is a problem, since the pet should have blown up by now

scripts/zones/Ifrits_Cauldron/mobs/Bomb_Queen.lua
```diff
@@ -1,35 +1,53 @@
...
+            local pos = mob:getPos()
+            pet:setSpawn(pos.x + math.random(-2, 2), pos.y, pos.z + math.random(-2, 2), pos.rot)
```
davismj: from the videos i saw, they spawn right on top of bomb queen

scripts/zones/Ifrits_Cauldron/mobs/Bomb_Queen.lua
wrenffxi: 10 second wait until using ability + the time it takes to do the ability + the time it takes to despawn the mob, can be greater than 20 seconds.  Even in the videos, you'll see a pet spawned while the previous pet is still despawning.  Also, pets could be slept or otherwise prevented from using their move at exactly 10 seconds.

scripts/zones/Ifrits_Cauldron/mobs/Bomb_Queen.lua
wrenffxi: In videos, the spawn position seems to be pretty consistent, but it's offset from the queen; you can tab to target them immediately at spawn.  If you spawn pets at queen's exact coordinates, they won't be targetable until they move.

davismj: Should an offset be forced, then? Instead of allowing a small chance for the same location?

PR #5624, opened 2019-01-20
Stability Issues

DivByXI:
People could effectively run around while healing.
People could teleport to their saved position at will (forced save via SP:0F2).

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -772,6 +772,12 @@ void SmallPacket0x01A(map_session_data_t* session, CCharEntity* PChar, CBasicPac
     break;
     case 0x12: // dismount
     {
+        if (PChar->StatusEffectContainer->HasPreventActionEffect())
+            return;
+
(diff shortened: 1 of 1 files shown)


src/map/packet_system.cpp
```diff
@@ -772,14 +772,17 @@ void SmallPacket0x01A(map_session_data_t* session, CCharEntity* PChar, CBasicPac
     break;
     case 0x12: // dismount
     {
+        if (PChar->StatusEffectContainer->HasPreventActionEffect() || !PChar->isMounted())
```
zynjec: Just curious, was there any reasoning behind the `HasPreventActionEffect()` check?

DivByXI: I just threw it there for extra measure, it's not necessarily required to stop the exploit of people running around while healing. I noticed other parts of the code that resets the player animation seem to check for preventative effects so i figured i'd follow suit to prevent any other possible abuse.

PR #5632, opened 2019-01-22
Fixed issue where blue and yellow rocks were dropped on the oppisite …

jmf9808:
…color day.


wrenffxi:
HAHA YES!  Someone did it!  I refused to switch those during the HELM rewrite because of the comment that said specifically "do not reorder this array of rocks" lol

PR #5641, opened 2019-01-23
Cleanup Footwork

nfaletra:
Reference: https://www.bg-wiki.com/bg/Footwork
* Fixed kick attack base damage calculations
* Fixed item mods for increased attack percent
* Hooked up Footwork kick attack percent
* Changed GetDamageRatio to take in a float for Attack Bonus Percent
  * Having it as a uint16 and dividing by 100 within the function wasn't intuitive
  * Cleaned up other places that called it to do the division before passing to GetDamageRatio


nfaletra:
Forgot to adjust how footwork affects "kick" weaponskills now that the kick damage is being modified directly.

PR #5642, opened 2019-01-24
add mob additional effects to -- and table functions in -- the mobs global

wrenffxi:
I notice that we have mob additional melee effects coded all over the place, often with slightly different implementations.  This is my attempt to consolidate all the logic and move it into one place, so that adding additional effects to mobs will be a one-line add from now on (okay, two if you count the mobMod).

While I was in there I tabled the phOnDespawn function too.  In separate commit for reviewer's sanity.

### DIFF
--- a/scripts/globals/mobs.lua
+++ b/scripts/globals/mobs.lua
@@ -7,6 +7,9 @@ require("scripts/globals/status")
 require("scripts/globals/zone")
 -----------------------------------
 
+dsp = dsp or {}
+dsp.mob = dsp.mob or {}
+
 -- onMobDeathEx is called from the core
 function onMobDeathEx(mob, player, isKiller, isWeaponSkillKill)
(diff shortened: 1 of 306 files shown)


TeoTwawki:
I'm fine with it, my other branch only touches the items so far.

TeoTwawki:
if we decide to reuse what I am doing for items on mobs later, this actually makes that a simpler conversion

scripts/globals/mobs.lua
```diff
@@ -1,85 +1,455 @@
...
+    return false
 end
+
+-------------------------------------------------
+-- mob additional melee effects
+-------------------------------------------------
+
+dsp.mob.additionalEffect =
+{
+    BLIND      = 0,
```
TeoTwawki: We'll definitely be revisiting this enumeration in the future. Just FYI.

PR #5644, opened 2019-01-24
Fixing spell list

zynjec:
blind merges 4 life


KnowOne134:
Order of merges conflictions

PR #5651, opened 2019-01-26
Fail Travis-CI early

ShelbyZ:
Fail before system build if:

- price_checker fails
- importing sql to db fails
- luacheck fails

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,11 +30,11 @@ install:
 before_script:
   - g++ --version
   - python3 ./tools/price_checker.py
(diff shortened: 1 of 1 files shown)


TeoTwawki:
Actually shouldn't price checker be _after_ the sql ?

ShelbyZ:
If we (price) check the sql files first and find an issue then we don't actually import into the db.

TeoTwawki:
oh your script is reading files not the actual db. Ok then.

Issue #2752, opened 2016-02-11
Mobs with 2hrs don't use them.


Hozu:
During the testing of my latest pull, I noticed that they weren't using the ability as I was smacking the master (to death) with a Kraken Club. I ended up having to force a magical move to ensure my code didn't cause a problem.

bendangelo:
This is a big problem, parties need to wipe. Will fix soon.

Hozu:
Renaming because no mobs with 2hrs use them.

abriasffxi:
Hopeful of a fix here.  Any data or testing or anything that could help?

Kushdr:
Not sure if this was intentional as part of the AI rewrite or not, but I'm pretty sure it's just this line in Mob_controller.cpp (line 298)
if (!PMobSkill->isTwoHour() && luautils::OnMobSkillCheck(PActionTarget, PMob, PMobSkill) == 0) //A script says that the move in question is valid

I would think if we remove the isTwoHour check it would stop caring and allow it but I haven't had a chance to test it

nyczducky:
I know scripted 2hrs are working.

ffxisf:
fix SMN Astral Flow

TeoTwawki:
All the ones using mobmod 2hr need converted to use the new mixin method

Issue #2063, opened 2015-09-13
Mobs use *only* 2hr with MOBMOD_2HOUR_MULTI set.


Hozu:
The problem lies in this function: https://github.com/DarkstarProject/darkstar/blob/1cb73213612af86788851ade285689ac69b32fcc/src/map/ai/ai_mob_dummy.cpp#L730

The way mob 2hrs are done is if the 2hr main mobmod is set, the mob has enough TP, and the move is eligible to be used (see the onMobSkillCheck function in each script), the mob will use it, If not, then a regular mobskill would be used. Upon use, the 2hr mobmod gets set to 0, unless 2hr multi is set. Because of this, the mob will always use the 2hr if they meet the HP requirement if they have the 2hr multi mobmod set, and never any other skill.

I don't know the best way to handle this so I'm making it an issue.

bendangelo:
I see what you're saying. The mobmod `2hour_multi` will cause mobs to only 2 hour. I think the 2 hour logic should be decoupled from mobskill into its own method. Also what mob's 2 hour multiple times? Maybe we can just script those mobs manually and remove this mod.

teschnei:
I don't think there are many. Off the top of my head, I can only think of
neo ark angels, which are all timed or have special conditions

Hozu:
The wiki has Gurfurlur and Jailer of Hope with the ability to 2hr multiple times, and I'm sure there are more. I noticed it when fighting Jailer of Hope.

TeoTwawki:
This started happening only recently, because the 2hr trigger ceased being semi random and was set by another rmobmod. before that this worked as intended (but was never really needed).

TeoTwawki:
..I know I have said this before but..
I really think all 2hr should just be scripted. Aside from dyna pretty much everything that does it gets gets a script anyway and wouldn't be that much script work to do them that way. The core bits doing 2hrs feels like ancient "still here for compatibility with old crap" code. >.>

bendangelo:
We still need this 2 hour code for dynamis. I think this mob mod should be removed and have a new `useTwoHourAbility` method implemented for scripts to use.

TeoTwawki:
No, we don't need a redundant function. We already have one that works just fine for making it go...

And dyna could easily have its mobs go without being done in core if we wanted to change that. Its not even that many scripts to add.

TeoTwawki:
Hell dyna mobs don't even work as the in core comments describe as it is.

[...]

PR #5652, opened 2019-01-26
new job_special mixin

wrenffxi:
Here we go.

I tore out the old 2hr mixin and mobMods in commit one.

In commit two, I wrote a new job_special mixin that supports a lot more customized options, and converted all the stuff that was using the old one to the new one.

If you just throw the mixin on a mob, it'll get its main job's 2hr, which it'll use once every 2 hours, at about 40-60 HPP.  This covers most typical cases.

If you want to get crazy, you can use the config function.

You can set a custom HPP trigger.  Or make a mob sometimes explode, but sometimes not.  Or you can make a mob use its 2hr multiple times.  Or give it like nine different 2hrs that all have different HPP triggers and timers.  Or have it change into a mandragora every time it uses Perfect Dodge.  It's up to you.

See the enormous comment block atop the mixin for examples.

I'm not joking:
```
function onMobSpawn(mob)
    dsp.mix.jobSpecial.config(mob, {
        specials =
        {
            {
                id = dsp.jsa.PERFECT_DODGE,
                endCode = function(mob)
                    mob:setModelId(300) -- omg i'm a mandragora
                end,
            },
        },
    })
end
```


scripts/globals/mobskills/mijin_gakure.lua
```diff
@@ -1,22 +1,17 @@
...
 require("scripts/globals/status")
-require("scripts/globals/monstertpmoves")
-
 ---------------------------------------------------
 
 function onMobSkillCheck(target,mob,skill)
-    if (mob:getMobMod(dsp.mobMod.SCRIPTED_2HOUR) == 1) then
-        return 0
-    elseif (skill:getParam() == 2 and math.random() <= 0.5) then -- not always used
+    if skill:getParam() == 2 and math.random() <= 0.5 then -- not always used
```
TeoTwawki: isn't this exact thing handled by a feature of the new mixin?

wrenffxi: yeah, the chance param determines whether mob will use a 2 hr or not. i didn't know what all these skill:getParam() things were in the 2hr code, so I left 'em in.  Can I strip them all out?  (I haven't written the mixin yet when I edited these files.)

TeoTwawki: make'em dead :ghost: 
they are all old leftovers form when 2hrs had to be called in core as a mobskill. flags to prevent mob from using it via tp, parans that said when they were allowed..all that can move to the mixin

wrenffxi:
#2063 and #2752 can be closed if/when this merges. obviously there are a good many mobs that still need 2hrs added, but prolly don't need an issue for that.

scripts/globals/mobskills/mijin_gakure.lua
wrenffxi: done

TeoTwawki:
Are the now unused mobmobs fully removed from status.lua and mobmodifier.h as well?

scripts/mixins/job_special.lua
```diff
@@ -139,6 +148,19 @@ local familyEES =
...
+{
+    [dsp.jsa.MIGHTY_STRIKES] = dsp.effect.MIGHTY_STRIKES,
+    [dsp.jsa.HUNDRED_FISTS]  = dsp.effect.HUNDRED_FISTS,
+    [dsp.jsa.MANAFONT]       = dsp.effect.MANAFONT,
+    [dsp.jsa.CHAINSPELL]     = dsp.effect.CHAINSPELL,
+    [dsp.jsa.PERFECT_DODGE]  = dsp.effect.PERFECT_DODGE,
+    [dsp.jsa.INVINCIBLE]     = dsp.effect.INVINCIBLE,
+    [dsp.jsa.BLOOD_WEAPON]   = dsp.effect.BLOOD_WEAPON,
+    [dsp.jsa.SOUL_VOICE]     = dsp.effect.SOUL_VOICE,
+    [dsp.jsa.AZURE_LORE]     = dsp.effect.AZURE_LORE,
```
TeoTwawki: Also other like Elemental Sforzo etc just in case we wind up with non default durations (eh, not the best example since its not used by any mob yet, but it exists).

PR #5664, opened 2019-01-28
Fix enmity being overwritten

nfaletra:
I done fucked up. Fixed meow. Thanks, @csundman.


csundman:
Thank you for some awesome work on enmity!

personaone:
just tested, works like a charm now :D

PR #5666, opened 2019-01-28
add blacklist table to char delete trigger

KnowOne134:


sql/triggers.sql
```diff
@@ -45,6 +45,7 @@ CREATE TRIGGER char_delete
 	BEFORE DELETE ON chars
 	FOR EACH ROW
 BEGIN
+    DELETE FROM `char_blacklist` WHERE `charid_owner` = OLD.charid;
```
zynjec: `charid_owner`?

And just for consistency, either use tabs or change the whole file to spaces.

KnowOne134: Ah yeah my edit does tabs to spaces. And yes in the blacklist table the charid is charid_owner

zynjec: Right, forgot it was named something different in blacklist.

cOnSIStEncY

KnowOne134: If you want we can do a overhaul on it and change it all to charid?

zynjec: It's fine, just do tabs to spaces and we're golden.

PR #5676, opened 2019-01-29
ability(feint): Change Feint to wear on successful attack

davismj:
Was wearing on any attack, EFFECTFLAG_ATTACK. Changed it to wear on successful attack, EFFECTFLAG_DAMAGE.


davismj:
Too naive. 

(1) If the hit connects but is absorbed by Stoneskin, Feint would apply on the next attack as well.
(2) It seems like Feint is only being applied on regular attack rounds, not WS.

Issue #4956, opened 2018-06-03
Pets should disengage when master is in CS Event


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30180427_2

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
1. Send Pet to Fight a Mob
2. Start a CS Event (Video Link features entering a displacement in Riverne)
https://youtu.be/r8SM4nQhWC0?t=1178
3. Pet should disengage from Mob and run to masters side per retail vid. It continues to fight in DSP.

hilts-vaughan:
This is basically the equivalent of asking the pet to retreat, right?

PR #5681, opened 2019-01-29
Disengage pet when entering a CS Event

DivByXI:
[Issue #4956](https://github.com/DarkstarProject/darkstar/issues/4956)


ShelbyZ:
When a player enters a CS Event are they (player/pet) also removed from the hate-list of the monster?

DivByXI:
I'm not sure if that's the case on retail but with this fix it's simply disengaging. No hate is modified (just as though the player himself issued a retreat command)

ShelbyZ:
I get that, I'm trying to dig a little deeper and see if we need to extend it to removing both the player/pet from the monster's hate-list when entering a CS Event to avoid the player dying during the event and finding out after the event completes (assuming the player didn't get whisked to another zone).

DivByXI:
In the video provided in the issue thread the pet still got attacked when it disengaged which is the current behaviour in this PR.

KnowOne134:
Correct a pet will disengage but does not remove hate. It will just keep getting attacked. Notcied last month in salvage opening chests,, count as a cs

src/map/lua/lua_baseentity.cpp
```diff
@@ -976,6 +976,13 @@ inline int32 CLuaBaseEntity::startEvent(lua_State *L)
         ((CCharEntity*)m_PBaseEntity)->StatusEffectContainer->DelStatusEffect(EFFECT_HEALING);
     }
 
+    CBattleEntity* PPet = ((CBattleEntity*)m_PBaseEntity)->PPet;
```
nfaletra: Going forward, we're trying to remove C-Style casting for complex types.

Here's what something like this should look like and I'll explain what it's doing after (for those that care)

At the very top of `startEvent` we need something like this:
```
auto PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity);
if (!PChar)
{
    ShowError("CLuaBaseEntity::startEvent: Could not start event, Base Entity is not a Character Entity.\n");
    return 0;
}
```
This needs to happen before `int32 n = lua_gettop(L);`

We have a Debug Break at the top, making sure that this function is being invoked on a TYPE_PC entity. The dynamic cast will return a nullptr if m_PBaseEntity is anything other than a CCharEntity (e.g. CMobEntity) and will protect from people trying to call startEvent on mobs or NPCs. Instead we'll get an error and go about our day.

Furthermore, this new PChar pointer can be used throughout the rest of the function instead of continuously doing casts. Casts can be expensive.

Now your new code can become:
```
if (PChar->PPet)
{
    PChar->PPet->PAI->Disengage();
}
```
Likewise, throughout the rest of this function, anywhere you see `((CBattleEntity*)m_PBaseEntity)` can be replaced with our new PChar pointer.

Also note that this fix needs to be added to `startEventString`.

Not trying to pile up on you, just trying to get some of this code cleaned up while we can and provide good examples for the future.

Thanks!

TeoTwawki:
> player/pet from the monster's hate-list when entering a CS Event to avoid the player dying during the ev

player getting hit will trigger "event skipped" unless it goes into a full blown "movie" before the mob catches up. pet will just stand there and die. Not sure if the player part is a timed duration thing or what, need some retail experimentation there.

nfaletra:
Nah, you can definitely die while in a cutscene.

TeoTwawki:
Na, you can both die and not die.



Zombie taru's, zombie taru's everywhere.

PR #5685, opened 2019-01-29
More OOB crashes

DivByXI:


DivByXI:
The LotItem/PassItem methods have the correct bounds checks but these don't and they're called beforehand. I was going to add the check in the packet handler but i don't like using macros that far out of scope :S

TeoTwawki:
You can use rebase to remove the unwanted commit(s) from the branch completely, and the a force push makes the github side and pull request match it. 

Be careful not to do a pull, or you'll bring those commits back in - most people doing this for the first time mistakenly use a regular push and get an error message, then do a pull thinking thats what it was telling them to do, and then give up in a rage.

DivByXI:
Thanks for the advice. I'm still trying to learn git so all the help is appreciated.

TeoTwawki:
No prob, feel free to hit us up in discord anytime you need. Several of us can offer command line help, and tortoisegit too.

TeoTwawki:
I'm going to manually squash them down here, coz github introduced a feature letting me do that.

PR #5691, opened 2019-01-31
ws: Make the attack multiplier vary by TP

davismj:
Iron Tempest and Sturmwind both receive a varying attack modifier based on TP. This change renames params.atkmulti to params.atk100, params.atk200, params.atk300, consistent with other TP-varying parameters.

Iron Tempest and Sturmwind are given a 2.0 and 3.5 multipler at 2 and 3 tp respectively according to jp wiki figures.

Sturmwind: http://wiki.ffo.jp/html/860.html
Iron Tempest: http://wiki.ffo.jp/html/859.html

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -15,7 +15,7 @@ require("scripts/globals/magic")
 require("scripts/globals/utils")
 require("scripts/globals/msg")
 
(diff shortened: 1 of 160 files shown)


zynjec:
`scripts/globals/weaponskills/requiescat.lua:25:68: (E011) expected statement near '+'
`

PR #5695, opened 2019-02-02
correct modifiers for dark shot and remove random buff

KnowOne134:
A feeler before i do all the quickdraw shots. cant find any info that supports that brd songs play a part so i removed. Also removed the random which buff its boosting and now it effects all buffs that are on it


KnowOne134:
Any recommendations?

KnowOne134:
Stale,,,

PR #5699, opened 2019-02-03
Nuke mobmod...

TeoTwawki:
I sinned against git/nature and used the website to edit.

PR #5700, opened 2019-02-03
2hr mixin cleanup

wrenffxi:
Follow up to #5699.

Convert some missed NMs to new 2hr mixin:
* Qn'Aern (assists Ix'Drk and Ix'Mnk)
* Steelfleech Baldarich
* Bloodtear Baldurf
* Ix'Aern (DRK)

Remove broken SCRIPTED_2HOUR mobMod from Ashu Talif Captain.  That's the last of the obsoleted mobMods removed from luas.

Question: I see that Ix'Drk was scripted to use a non-standard blood weapon.  Is this mobskill ID unique to Ix'Drk, or used by all Aern?


TeoTwawki:
> Question: I see that Ix'Drk was scripted to use a non-standard blood weapon.  Is this mobskill ID unique to Ix'Drk, or used by all Aern?

Unknown not even sure its correct animation. Its like the diff between generic mob woosh and player type animation, lets get a capture and return to that later

PR #5710, opened 2019-02-04
NMs

wrenffxi:
* Abyss Sahagin (respawn timer)
* Death from Above (additional effect, attack speed, hp, immunity, name prefix)
* Goaftrap
* Novv the Whitehearted (drops, immunity)
* Ocean Sahagin (drops)
* Sacrificial Goblet (job)
* Sozu Rogberry (skill list)
* Wurr the Sandcomber (aggro, hp, immunity, regen)
* Yarr the Pearleyed (drops, hp, immunity, name prefix, special trigger)

note: mob skill lists 243 and 244 for Tonberries were identical, so I repurposed 244 to be the "doesn't use Everyone's Rancor" skill list.  The only mob that was assigned 244 was Golden-Tongued_Culberry, so I switched it to 243.


sql/mob_groups.sql
```diff
@@ -8475,7 +8475,7 @@ INSERT INTO `mob_groups` VALUES (8411,785,159,0,128,0,0,0,75,75,0);
 INSERT INTO `mob_groups` VALUES (8412,786,159,0,128,0,0,0,75,75,0);
 INSERT INTO `mob_groups` VALUES (8413,787,159,0,128,0,0,0,75,75,0);
 INSERT INTO `mob_groups` VALUES (8414,836,159,0,128,599,10000,0,70,70,0);
-INSERT INTO `mob_groups` VALUES (8415,939,159,0,128,659,0,0,62,62,0);
+INSERT INTO `mob_groups` VALUES (8415,939,159,0,128,659,3500,0,62,62,0); -- Death from Above
```
TeoTwawki: image

PR #5726, opened 2019-02-11
[QUEST REWRITE] Reworking quest files to be stage-based + some GM commands

ibm2431:
Mainly @takhlaq , but I'd be happy for a few thumbs-up-or-down from others on this quest format, as _I_ personally don't have major changes to the format planned after this, and am concerned about people feeling it was suddenly sprung on them one day upon being merged into master.

Even if you don't have any _words_ to say, just a :+1: would show this branch is on the right track!

Anyway, I **reworked quest tables to have stage enums as their top-level** for NPC checks, resulting in being able to read the quest file from top-to-bottom and **see the steps/stages in chronological order**. ([Link to non-diff Old Man and The Harpoon for example](https://github.com/ibm2431/darkstar/blob/c3d5f4d3b8082d08167b4bdcb9c319a58cb69cfa/scripts/quests/adoulin/the_old_man_and_the_harpoon.lua#L62))

All previous occurrences of **magic numbers in scripts changed to straight enums. Added `dsp.quests.advanceStage(player, quest)`**, so a scripter will probably never need to get or set a player's stage variable.

I've been thinking about how to fetch a quest table from just an ID for a while, and still have concerns about colossal require tables being needed/loaded at the top of the global quest script (mainly rewrite/migration). Instead, I wrote a **`dsp.quests.getQuestTable(area_id, quest_id)` method that can fetch/require a quest table from just an ID** when it's called, instead of loading all quests, all the time. This still requires a table of script name strings, but at least the table isn't actively requiring _every_ quest file in the game at _all_ times.

I'm open to ideas on any better ways of solving the questID -> quest file problem.

With the ability to fetch quest tables from just an ID, I **added a GM command, `resetquest`**, which will (try to) **reset the player's progress of a quest back to zero**. It only resets quest log status, the player's stage var for the quest, any additional quest vars (as listed in the quest file), and any temporary key items used only for that quest (again as listed in the quest file). This doesn't guarantee completely resetting a quest, as things like physical items aren't touched (intentionally), and it doesn't reset any rewards for a quest either. But as it's meant primarily for debugging purposes, it should be sufficient for testing/debugging _most_ things.

I also **made an addition to the GM command `checkquest`**, in that if a quest table exists for the quest ID (as fetched by `dsp.quests.getQuestTable`), it'll **print the player's current stage and any additional quest vars**. So if a player is complaining about something not working as expected, you can check their relevant vars without having to look in the DB!

quest_stuff

But _more_ importantly, the message now displays the quest's _name_ instead of just an ID! Truly vital stuff.

And I apparently made a few changes to `dsp.quests.completeQuest` to better handle giving rewards.

### DIFF
--- a/scripts/commands/checkquest.lua
+++ b/scripts/commands/checkquest.lua
@@ -60,7 +60,30 @@ function onTrigger(player,logId,questId,target)
         [2] = function (x) status = "COMPLETED"; end,
     }
 
-    -- show quest status
-    player:PrintToPlayer( string.format( "%s's status for %s quest ID %i is: %s", targ:getName(), logName, questId, status ) );
+    -- fetch a quest table if there is one
+    local quest_table = dsp.quests.getQuestTable(logId, questId)
 
+    local quest_status_string = ''
+
+    -- show quest status
+    if quest_table then
+        quest_status_string = quest_status_string .. string.format("%s's status for %s quest '%s' is: %s", targ:getName(), logName, quest_table.name, status)
+    else
+        quest_status_string = quest_status_string .. string.format("%s's status for %s quest ID %i is: %s", targ:getName(), logName, questId, status )
+    end
+
+    -- print any known quest vars
+    if quest_table then
+        quest_status_string = quest_status_string .. string.format("\n%s is currently on stage: %i", targ:getName(), dsp.quests.getStage(targ, quest_table))
+        local quest_vars_string = ''
+        for name, var in pairs(quest_table.vars.additional) do
+            quest_vars_string = string.format(quest_vars_string.. ", %s: %i", name, dsp.quests.getVar(targ, quest_table, name))
+        end
+        if string.len(quest_vars_string) > 1 then
+            quest_vars_string = string.format("\n%s's additional quest vars are".. quest_vars_string, targ:getName())
+            quest_status_string = quest_status_string .. quest_vars_string
+        end
+    end
+
+    player:PrintToPlayer(quest_status_string)
 end;
--- a/scripts/commands/resetquest.lua
+++ b/scripts/commands/resetquest.lua
@@ -1,0 +1,72 @@
+---------------------------------------------------------------------------------------------------
+-- func: !resetquest <logID> <questID> {player}
+-- desc: Resets the target's progress in a quest (log status, quest vars, and temporary key items)
+---------------------------------------------------------------------------------------------------
+
+require("scripts/globals/quests");
+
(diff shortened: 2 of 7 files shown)


takhlaq:
>I wrote a **`dsp.quests.getQuestTable(area_id, quest_id)` method that can fetch/require a quest table from just an ID**

hmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm
can prolly use a string builder function instead of tabling each file name manually and keep a naming convention for quest script names for consistency

how do we define the lifetime of 'temporary' [key]items? within the stage or until quest completion?
there's also quests where bit flags would be needed for out of order trades (wotg chocobo quest iirc?) i cant think of many off the top of my head so maybe we can just think about that later

ibm2431:
I don't know how we'd make a ID->string connection without involving a table of strings at _some_ point. We currently don't have any strings associated with quests, as the globals used (OLD_MAN_AND_THE_HARPOON) are (currently) just static globals that resolve to an integer. We could look through the table keys and resolve the string that way, but if we're going into it with just an ID, we'd have to step through the table searching the values until we find a match (on value), then grab the key.

The first thought that comes to mind is to have quest filenames be just `[area_id]_[quest_id].lua`, but that'd make maintenance more difficult.

What could be done though, is on initializing our primary quest ID enum table (String Key->Log ID), have a function that steps through it and builds/stores a _reverse_ version of it (Log ID->String Key) and store that as a global. We'd still have a table of strings, but at least we don't have to build/maintain it manually.

On temp KIs, I was originally thinking until quest completion. Obviously the scripter should be removing the KI after the relevant stage, when it _should_ be removed. The list is mainly a backup to guarantee they _do_ get deleted by the time the quest is done (because `dsp.quests.complete` nukes them again, along with perishable vars), and for ease-of-testing with things like `resetquest`.

But I guess it's possible to add some "given_on" and "expires_after" values, and `advanceStage` can handle giving and deleting them, instead of them being handled in the `eventFinish`. It might make the quest harder to follow though, as the key items would be taken out of context of when they're given, and the scripter would have to scroll between the temp KI definitions at the top, and the stages themselves further down.

Bit flags can be used in a quest's additional vars field, and a quest's "stage" isn't passed until all the "any order" steps are satisfied (like in Wayward Waypoints, the quest doesn't progress until all the Adoulin waypoints are calibrated, but they can be calibrated in any order). What counts as a complete stage will have to be a judgment call by the scripter.

PR #5733, opened 2019-02-12
Remove WOTG tag from RDM MAB trait

isotor:
MAB1 and MAB2 traits not tied to WOTG, so simply updating the field to NULL to correct.


JulianoW:
These are "Null" strings rather than actual null. Not sure what the field is for and what it does, but I assume it would mess up any isnull()'s and such.

TeoTwawki:
Woops, I didn't catch that so I guess I'll hafta fix it. It technically works but causes the core to look for a settings.lua variable matching that string.

Issue #5442, opened 2018-12-06
Citipati is not lottery spawning


davismj:
- [x] I have searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] Provided detailed steps to diagnose and resolve the issue.

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

1. Citipati is a lottery spawn off of Corses. Corses despawn at 04:00, triggering `phOnDespawn` for each Corse, allowing Citipati to spawn.
2. Additionally, Citipati will not despawn at 04:00 since he also has the `SPAWNTYPE_LOTTERYFLAG` which would skip [this if statement](https://github.com/DarkstarProject/darkstar/blob/master/src/map/zone_entities.cpp#L693) .
3. Finally, if the two above issues are fixed, then Citipati will despawn at 04:00, which will fire the [DESPAWN listener in `phOnDespawn`](https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/mobs.lua#L69), causing Citipati's cooldown to restart.

I'm not sure if (3) is desired.

PR #5736, opened 2019-02-12
Updated Citipati respawn timers

isotor:
In advance, please forgive me for this "well actually herpderp" correction.

Citi repops every 3-6 game nights, rather than a fixed timer on Earth hours. This means he can be killed as late at 04:00 on Night 1, and respawn as early as 20:00 on Night 4 (assuming no PHs for this math). One Vana day is exactly 57m 36s so the minimum respawn is ((57*60)+36)*(2+(16/24)), or 9,216 seconds. Add three full Vana days for the minimum time on the longest window.

### DIFF
--- a/scripts/zones/Attohwa_Chasm/mobs/Corse.lua
+++ b/scripts/zones/Attohwa_Chasm/mobs/Corse.lua
@@ -10,5 +10,5 @@ function onMobDeath(mob, player, isKiller)
 end;
 
 function onMobDespawn(mob)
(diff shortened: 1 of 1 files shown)


davismj:
That's not the problem, this is the problem: https://github.com/DarkstarProject/darkstar/issues/5442

isotor:
Aside from being related to the same mob, these two things are completely unrelated

> 
> 
> That's not the problem, this is the problem: #5442

Aside from being issues with the same NM, these two things are not related. This PR is to correct the window, that's all.

PR #5749, opened 2019-02-16
Cmake

ShelbyZ:
2nd pass at moving to CMake:

- Using glob selectively for files that could change and manually requiring parts of common to avoid rebuild of entire common
- Using pkg-config as finding *.cmake files for our main dependencies feels bad (mysql/maria, zmq, luajit)
- Using a fallback method to search for mariadb to satisfy mysqlclient and if not found require finding mysqlclient
- probably still need a few compiler tweaks in compilers.cmake (add llvm, enable -Werror)


CMakeSettings.json
```diff
@@ -0,0 +1,55 @@
+{
+  // See https://go.microsoft.com//fwlink//?linkid=834763 for more information about this file.
```
TeoTwawki: Good grief, MS violates the spec with that comment. :man_facepalming:

PR #5753, opened 2019-02-18
Quest Implementation of A Timely Visit

bkesecker:
Picked up where beauxq left off and finished it up.  Logic for this quest is separated from the rest, which was untouched, except for the normal code clean-up.  Halver edits are almost all clean-up.  Wiggo provided the entire 2 hour capture, so this was easy.  Tested the battle with average geared 49 PLD, 49 RDM, and 49 WAR.  PLD did well, RDM had to use a lot of bind and rest, WAR failed badly.  One of the better story-heavy quests in my opinion.


scripts/zones/Jugner_Forest/mobs/Giollemitte_B_Feroun.lua
```diff
@@ -0,0 +1,51 @@
...
+    DespawnMob(mob:getID(), 300)
+    
+end
+
+function onAdditionalEffect(mob, target, damage)
+end
+
+function onMobFight(mob, target)
+
+    if mob:getHPP() <= 25 and mob:getLocalVar("invincible") == 0 then
```
TeoTwawki: this is handled by a mixin now
https://github.com/DarkstarProject/darkstar/blob/master/scripts/mixins/job_special.lua#L29

bkesecker: Changed to use mixins.

scripts/zones/Jugner_Forest/mobs/Giollemitte_B_Feroun.lua
```diff
@@ -0,0 +1,59 @@
...
+
+function onMobInitialize(mob)
+
+    mob:setMobMod(dsp.mobMod.IDLE_DESPAWN, 300) -- 5 minutes
+    mob:setMobMod(dsp.mobMod.ADD_EFFECT, 1)
+    
+end
+
+function onMobSpawn(mob)
+
```
TeoTwawki: Couple of things to work out here:

- you need to rebase, there are conflicts preventing merge. Sorry this happens sometimes

- the extra newlines make for weird formatting.

- the old version of the 2hr scripting is still there commented out. remove it entirely please.

- you have some `addMod()` in the `onMobSpawn()` that need explained or removed. Also they should be in `onMobInitialize()` instead, if and only if they are absolutely necessary. The mobs Job will normally give it traits that will add some mods. From what I see on wiki, this mob is a paladin. paladins already come with resist sleep baked in (and it applies to multiple forms of sleep). They don't have magic defense bonus, but if you know this guy does on retail for a fact that indicates to me either a rdm or whm sub granting that trait is likely. Wiki says "seems to".  Wiki being correct with no data is like a coin toss. I'm willing to go with it till someone proves it right or wrong, but it's probably from a job trait rather than non standard stacks tacked on. (note: those DO happen, they just aren't very common)

ibm2431:
I have submitted [a PR to your branch that might resolve those conflicts, and also tables the quest IDs](https://github.com/bkesecker/darkstar/pull/1).

zynjec:
Closing for now, feel free to open another PR once you're able to rebase and make these changes.

PR #5758, opened 2019-02-20
Table Quest IDs

ibm2431:
Heads up, @wrenffxi!

Sorry for making lines longer, @TeoTwawki. Blame @takhlaq!

Targeting master to make quest_rewrite more sane later. I'll do my best to keep the commits under control.

GM command changes are also planned; I don't intend on breaking them!

(I intend to have this done within the next few days; I just wanted an excuse to try out the Draft PR feature.)


ibm2431:
And I intend push an update to account for open PR #5753 (either to this, or PR @bkesecker's branch) depending on which of our PRs gets merged first.

TeoTwawki:
as long as the global is formattered good

wrenffxi:
could we do dsp.quest.id instead of dsp.quests.ids?  keeps everything consistent (dsp.keyitem, dsp.title, dsp.mob, etc.) -- and also two characters shorter.

ibm2431:
And so it was done! Thanks for noticing the inconsistency.

I just went with plural quests due to the usage of `dsp.quests` over in `quest_rewrite`. But those names in the branch should probably change as well, because the project has been using `dsp.` for enums, while that branch uses `dsp.quests.` for functions. Should probably change them to something else, like how we have `npcUtil.function()` instead of `dsp.npcUtil.function()`.

ibm2431:
And with the GM commands, I think that covers everything, unless anyone has concerns / requests.

I'm going to give the diff a good scroll-through before I mark this PR as ready for review, just in case the regex replacements hit some weird cases like it did with the AF hand quests and mythic WS unlock quests.

scripts/zones/Southern_San_dOria/npcs/Balasiel.lua
```diff
@@ -31,13 +31,13 @@ end
 function onTrigger(player,npc)
     local wsQuestEvent = dsp.wsquest.getTriggerEvent(wsQuest,player)
     local lvl = player:getMainLvl()
-    local aSquiresTest = player:getQuestStatus(SANDORIA, A_SQUIRE_S_TEST)
-    local aSquiresTestII = player:getQuestStatus(SANDORIA,A_SQUIRE_S_TEST_II)
-    local aKnightsTest = player:getQuestStatus(SANDORIA, A_KNIGHT_S_TEST)
+    local aSquiresTest = player:getQuestStatus(SANDORIA, dsp.quest.id.sandoria.A_SQUIRE_S_TEST)
+    local aSquiresTestII = player:getQuestStatus(SANDORIA,dsp.quest.id.sandoria.dsp.quest.id.sandoria.A_SQUIRE_S_TEST_II)
```
wrenffxi: double match

scripts/zones/Southern_San_dOria/npcs/Diary.lua
```diff
@@ -13,9 +13,9 @@ end
 
 function onTrigger(player,npc)
 
-    local aSquiresTestII = player:getQuestStatus(SANDORIA, A_SQUIRE_S_TEST_II)
-    local medicineWoman = player:getQuestStatus(SANDORIA, THE_MEDICINE_WOMAN)
-    local toCureaCough = player:getQuestStatus(SANDORIA, TO_CURE_A_COUGH)
+    local aSquiresTestII = player:getQuestStatus(SANDORIA, dsp.quest.id.sandoria.dsp.quest.id.sandoria.A_SQUIRE_S_TEST_II)
```
wrenffxi: here too

scripts/zones/Southern_San_dOria/npcs/Balasiel.lua
ibm2431: Indeed. I was an idiot at the start for the San d'Oria replace and was matching for the var names instead of the quest functions themselves (like in all commits afterward). Thought I got all these, but apparently not. Thanks wren; is fixed for this next commit I'm going through.

ibm2431:
I've noticed cases where my regex missed some replaces if there were two quest functions on a line, so I'll include those with the next commit as well.

ibm2431:
Fixed the mishaps, and did moar regex to find any quest function that didn't have a `dsp.quest.id` prefix where it should. Fixed some Shantotto Ascension checks I found by accident in the process.

Hitting the 'Ready for review' button now. Blind merge for 899 files go...!

TeoTwawki:
> Hitting the 'Ready for review' button now. Blind merge for 899 files go...!

**RUBBASTAMP**

j/k. thats Demo's thing.

takhlaq:
>Should probably change them to something else, like how we have `npcUtil.function()` instead of `dsp.npcUtil.function()`.

in that case maybe something like
```lua
dsp.helpers.table_name.func()
dsp.enums.table_name.ENUM
```
as the reason for the dsp table was to neatly keep globals together and avoid issues where scripts would reassign them due to name conflicts, and not having helpers/enums under the global dsp table defeats that purpose (though im not sure if lua has a limit on table size)

TeoTwawki:
> as the reason for the dsp table was to neatly keep globals together and avoid issues where scripts would reassign them due to name conflicts,

being in diff tables already stops the name collisions:

tableA has a thing with the same name but diff value as tableB, because we call tablename.thingy instead of just thingy, we dodge a collision 

its why we were tabling them in the first place.

so lets get careful about how deeply we nest these tables. all I am saying.

ibm2431:
Unceremonious bump!

takhlaq:
im not going through all those files
from the handful i have seen, it looks good

i wouldnt be waiting on me for anything as i wont be touching dsp for a few months at least yet

TeoTwawki:
I tell you a (not) secret demo: I'm not looking at it either

teschnei:
i will merge this as soon as conflicts are resolved

ibm2431:
Conflicts should be resolved now!

PR #5761, opened 2019-02-21
NMs

wrenffxi:
* Euryale
* Frogamander
* Killer Jonny
* Marquis Naberius
* Sabotender Corrido


scripts/zones/Eastern_Altepa_Desert/mobs/Sabotender_Corrido.lua
```diff
@@ -0,0 +1,29 @@
...
+    mob:setMobMod(dsp.mobMod.ADD_EFFECT, 1)
+    mob:setMod(dsp.mod.DOUBLE_ATTACK, 25)
+    mob:setMod(dsp.mod.MOVE, 25)
+end
+
+function onMobSpawn(mob)
+    dsp.mix.jobSpecial.config(mob, {
+        specials =
+        {
+            {id = dsp.jsa.HUNDRED_FISTS, hpp = 25},
```
TeoTwawki: quick question on usage of our new mixing. you need to declare the jsa in order to declare the hp% or would it use his mainjob of monk to get hundred fists? merging as is anyway because I like clarity but just wondered "what would happen" and I am too lazy to check when I know you wrote it

wrenffxi: you need to declare the JSA on every item in specials table, if you provide specials table.  If you leave the table off, or you don't call config at all, the mob will get its main 2hr at 40-60%, just by requiring the mixin.

TeoTwawki: good to know, thanks!

PR #5767, opened 2019-02-24
remove drawin from King Behemoth

KnowOne134:
can not find any info saying KB has a drawin effect


godmode:
It is regional. Not a standard draw in but he will draw in if you have hate and are outside of the region.

Delkathus:
He does have a draw in, i remember it vividly and it's why pld's had to kite with w.legs on retail.

TeoTwawki:
We're aware his draw in Is currently not the right type but we're leaving it until are draw in is in.

TeoTwawki:
Ahem let me amend what i just said

TeoTwawki:
We are aware _multiple_ NM ate using the wrong type of draw in, but we've been allowing it until the correct type is implemented.

KnowOne134:
Gotta. I couldnt find any info on it so thought was a mistake. Closing

PR #5769, opened 2019-02-25
Lamebus entry counted by hours instead of days.

TeoTwawki:
Also for unknown reasons the default for CoP Dyna was to let people in without meeting the requirements. I've changed that default.


zynjec:
Current retail cool down is 20 hours, if you want to change that while you're at it.

TeoTwawki:
ya know, i actually do.

image

..brb

TeoTwawki:
It is done.

PR #5773, opened 2019-02-28
NMs

wrenffxi:
* Antican Praetor (move mods from mob to pool: two Praetors share pool)
* Barbaric Weapon
* Coquecigrue
* Herensugue (triple attack, regain, drops, skills, hp)
* Shikigami Weapon (immune to silence, double attack, regen, drop rate)


scripts/zones/RoMaeve/mobs/Shikigami_Weapon.lua
```diff
@@ -41,6 +42,11 @@ local path =
     -35, -4, -36,
 }
 
+function onMobInitialize(mob)
+    mob:setMod(dsp.mod.DOUBLE_ATTACK, 10) -- "Double Attacks frequently"
```
TeoTwawki: Shiki's DA being missing are likely because we have it's job/sub wrong, its blm/blm and one of those should be war. If a mob stands back and casts until a player weakens it or comes closer, then its likely blm/war, it a mob immediately closes to melee but casts when its timer is up, its war/blm.

TeoTwawki: And of course if it DA's at a rate way higher than its main or sub would provide, then we still need a mod here.

scripts/zones/Sauromugue_Champaign_[S]/mobs/Herensugue.lua
```diff
@@ -3,6 +3,11 @@
 --   NM: Herensugue
 -----------------------------------
 
+function onMobInitialize(mob)
+    mob:setMod(dsp.mod.TRIPLE_ATTACK, 90) -- "Triple Attacks almost every round"
```
Comment (TeoTwawki) [inline scripts/zones/Sauromugue_Champaign_[S]/mobs/Herensugue.lua]: keep the mod because of the rate, but I would bet we have the same job issue on this mob as well

Comment (TeoTwawki) [reply scripts/zones/Sauromugue_Champaign_[S]/mobs/Herensugue.lua]: nvm he's thf/thf xD

scripts/zones/Xarcabard/mobs/Barbaric_Weapon.lua
```diff
@@ -0,0 +1,15 @@
+-----------------------------------
+-- Area: Xarcabard
+--   NM: Barbaric Weapon
+-- TODO: "Gains a Dread Spikes effect whenever Whirl of Rage is used."
+-----------------------------------
+require("scripts/globals/status")
+-----------------------------------
+
+function onMobInitialize(mob)
+    mob:setMod(dsp.mod.DOUBLE_ATTACK, 10)
```
TeoTwawki: He's war/war in his pool entry and should already have the base double attack from that. if he needs additional double attack on top of that you'll need to addMod

scripts/zones/RoMaeve/mobs/Shikigami_Weapon.lua
wrenffxi: It's in DB as RDM/rdm and listed as just "RDM" on wiki.  Assuming if we change it to WAR/rdm then its spells would be very weak.  Maybe RDM/war?  In the videos I watched, it opens with a debuff spell, then immediately closes toward player.  it double attacks roughly 1 in 10 attacks.

TeoTwawki: rdm/rdm eh, I must have been off by one when I looked at it. Lets give rdm/wara  try and see if its right. I think DSP's rdm's close for melee unlike blm.


in regard to melee/mage having weaksauce nukes: 

we had that problem with King Arthro as well. 2 ways to go about it, typically the melee behavior tells us which the "main" job really is. we can buff the mage side manually with magic attack but its hard to get that balance right. the easier way is we can force the engage behavior while still leaving mage main using thee mobmod to make it close for melee. 

Wiki job tends to be reliable only very loosely. if they see a spell or ability they think fits that job some random editor will add that as its job. the real fun ones have 3 or 4 jobs listed. Good luck getting those ones right first try on DSP. mob has astral flow but otherwise acts like a war in every way? wiki gonna say its a smn :rofl:

PR #5777, opened 2019-03-02
[WIP]Adding Trust AI Logic Logic

Tummie:
Adding in AI logic for Trusts using listeners in the lua to activate abilities, spells, and weapon skills.
Tried to make this as close to retail as possible.  Started off with the most basic 3 trusts from the main cities.

**Features:**
- Naji: Uses Provoke only when the player takes hate.  Uses weapon skills Red Lotus Blade, Fast Blade, and Vorpal Blade
- Excenmille: Cures party members and casts flash.  Uses weapon skills Double Thrust, Leg Sweep, Penta Thrust
- Kupipi: Casts -na spells on player only, casts protect(ra)/shell(ra), cures on party members, debuffs on mobs.  Uses weapon skills Moonlight and Starlight

**Currently there are three issues:**
- Job Abilities aren't mapped to scripts
- Game crashes when trust dies
- Trust enmity seems small or non existent.

This PR does not include the quests to unlock these three trusts.


.gitignore
```diff
@@ -163,3 +163,7 @@ log/DSP.leaks
 
 #VS Code
 .vscode
+conf/search_server.conf
```
TeoTwawki: um...undo this please

scripts/globals/pets/Kupipi.lua
```diff
@@ -0,0 +1,283 @@
...
+	finalWS = newWsList[math.random(1,#newWsList)]
+	return finalWS
+end
+
+function doBuff(mob, player)
+    local proRaList = {{63,65,128}, {47,46,127}, {27,28,126}, {7,9,125}}
+    local proList = {{63,65,46}, {47,46,45}, {27,28,44}, {7,9,43}}
+    local shellRaList = {{68,75,133}, {57,56,132}, {37,37,131}, {17,18,130}}	
+    local shellList = {{68,75,51}, {57,56,50}, {37,37,49}, {17,18,48}}
+	local battletime = os.time()
```
TeoTwawki: tabs threw off yer indents, you may want to get the editorconfig plugin for your code editor to automate this

sql/mob_skills.sql
```diff
@@ -55,8 +55,12 @@ INSERT INTO `mob_skills` VALUES (35,6,'flat_blade',0,7.0,2000,0,4,0,0,0,8,0,0);
 INSERT INTO `mob_skills` VALUES (36,4,'shining_blade',0,7.0,2000,0,4,0,0,0,4,0,0);
 INSERT INTO `mob_skills` VALUES (37,5,'seraph_blade',0,7.0,2000,0,4,0,0,0,4,0,0);
 INSERT INTO `mob_skills` VALUES (38,7,'circle_blade',1,5.0,2000,0,4,0,0,0,5,8,0);
-INSERT INTO `mob_skills` VALUES (40,9,'vorpal_blade',0,7.0,2000,0,4,0,0,0,5,8,0);
```
TeoTwawki: don't delete vorpal

src/map/ai/helpers/targetfind.cpp
```diff
@@ -243,12 +243,27 @@ void CTargetFind::addAllInAlliance(CBattleEntity* PTarget, bool withPet)
 
 void CTargetFind::addAllInParty(CBattleEntity* PTarget, bool withPet)
 {
-
     PTarget->ForParty([this, withPet](CBattleEntity* PMember)
     {
         addEntity(PMember, withPet);
+        if (PMember->objtype == TYPE_PC) {
```
TeoTwawki: we be brace on newline people in this here project

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -95,14 +102,49 @@ void CTrustController::DoCombatTick(time_point tick)
                 }
             }
         }
+		
+		POwner->PAI->EventHandler.triggerListener("TRUST_COMBAT_TICK", POwner, POwner->PMaster, PTarget);
```
teschnei: This should probably just be COMBAT_TICK - there's not really any reason to have two differently named ones

scripts/globals/pets/Kupipi.lua
Tummie: Thanks I thought I went thru the lua to check for spacing but I must have overwritten

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -118,4 +160,48 @@ void CTrustController::DoRoamTick(time_point tick)
             POwner->PAI->PathFind->WarpTo(POwner->PMaster->loc.p, RoamDistance);
         }
     }
+
+	POwner->PAI->EventHandler.triggerListener("TRUST_ROAM_TICK", POwner, POwner->PMaster, PTarget);	
+}
+
+bool CTrustController::Cast(uint16 targid, SpellID spellid)
```
teschnei: if you're overriding these to do exactly what the parent class is doing, then just don't override them and it will call the parent classes one (in other words, remove this and MobSkill)

src/map/ai/helpers/targetfind.cpp
```diff
@@ -243,12 +243,27 @@ void CTargetFind::addAllInAlliance(CBattleEntity* PTarget, bool withPet)
...
+            CCharEntity* PChar = (CCharEntity*)m_PBattleEntity;
```
teschnei: this doesn't seem right - first of all, there's no guarantee that m_PBattleEntity is a CharEntity (you checked PMember in the previous line, not m_PBattleEntity).  Then in the next line, check if PChar (which was from m_PBattleEntity) is a TRUST.. which is it? a player or a trust?
This whole block before `else if (PChar->PTrusts.size() != 0) {` I think should just be removed.  AFAIK, the only thing you want to do is add the trusts if the target is a player that has trusts.  This block is trying to add trusts if it found a trust in the loop? But if it found a trust in the loop, it wouldn't have to find the master and add the trusts because it already found one (evidenced by the fact that it is currently iterating on a trust)

src/map/entities/mobentity.cpp
```diff
@@ -141,6 +141,18 @@ CMobEntity::~CMobEntity()
     delete SpellContainer;
 }
 
+std::string CMobEntity::GetScriptName()
```
teschnei: remove this and replace where it was called with GetName

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -118,4 +160,48 @@ void CTrustController::DoRoamTick(time_point tick)
             POwner->PAI->PathFind->WarpTo(POwner->PMaster->loc.p, RoamDistance);
         }
     }
+
+	POwner->PAI->EventHandler.triggerListener("TRUST_ROAM_TICK", POwner, POwner->PMaster, PTarget);
```
teschnei: just ROAM_TICK

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -71,6 +73,11 @@ void CTrustController::Tick(time_point tick)
 
 void CTrustController::DoCombatTick(time_point tick)
 {
+    if (POwner->PMaster == nullptr || POwner->PMaster->isDead())
```
teschnei: Can you test replacing the movement stuff in `DoCombatTick` with `CMobController::Move()`?  Since this is now derived from CMobController (which it always should have been), there's a bunch of stuff that may be unnecessarily duplicated in this file

scripts/globals/pets/Excenmille.lua
```diff
@@ -0,0 +1,128 @@
...
+	local sentinelCooldown = 300
+	local master = mob:getMaster()
+	local excenmille = mob:getID()
+	mob:setLocalVar("cureTime",0)
+	mob:setLocalVar("debuffTime",0)
+	mob:setLocalVar("buffTime",0)
+	mob:setLocalVar("sentinelTime",0)
+	mob:setLocalVar("flashTime",0)
+
+	mob:addListener("TRUST_COMBAT_TICK", "TRUST_PROVOKE_TICK" .. excenmille, function(mob, player, target)
```
teschnei: do the event names need to have the id tacked on the end? you can't summon two excenmilles at the same time, right?  they also probably don't need their own name in it, calling it "PROVOKE_TICK" (without the id concatenated to the end) is good enough

teschnei:
many tabs must become 4 spaces

teschnei:
it could be a good idea to make a global trusts file to make some of the more generic things in their scripts less repetitive - like adding abilities on timers (provoke), etc.  you could just call a library function from trusts.lua or whatever to set it up for you

scripts/globals/pets/Excenmille.lua
Tummie: When I was testing, it required the ID because it caused trusts to take on weapon skills of the most recently spawned trust.  I tested again this morning after removing it and changing TRUST_COMBAT_TICK to COMBAT_TICK and it is no longer doing this.  Not sure what caused it.

teschnei: you might have been adding the listener to the player instead of the trust before, that would cause the issue you're describing

src/map/ai/helpers/targetfind.cpp
Tummie: addAllInParty doesn't see trusts as members.  I'll be resubmiting the new code for this shortly.  It looks to see if the entity is a trust, and then to get its master and iterate thru the masters Trust list to add the effect to the other trusts.  I also added a modelID check for now so it won't cast the effect on the casting trust twice.

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -95,14 +102,49 @@ void CTrustController::DoCombatTick(time_point tick)
...
 
 void CTrustController::DoRoamTick(time_point tick)
 {
     if (POwner->PMaster->PAI->IsEngaged())
     {
-        POwner->PAI->Internal_Engage(POwner->PMaster->GetBattleTargetID());
+        CCharEntity* PChar = (CCharEntity*)POwner->PMaster;
+
+        // Only engage the mob if the player has hate
+        for (SpawnIDList_t::iterator it = PChar->SpawnMOBList.begin(); it != PChar->SpawnMOBList.end(); ++it)
```
Wiggo32: Trusts do a little more than just check for hate for engagement. You have to actually perform an attack on the mob to get it to engage, doesn't matter if you hit or miss. After it has successfully engaged, then you can target lock off the mob and the trust will continue the attack up until the point where: 1 Trust dies, 2 Mob dies, or 3 Player Disengages. Here is some Video of different Targeting Mechanics for the Trusts and how they work:
https://youtu.be/0HIBwXu_BvI

scripts/globals/spells/bluemagic/magic_fruit.lua
```diff
@@ -40,7 +40,7 @@ function onSpellCast(caster,target,spell)
 
     final = final + (final * (target:getMod(dsp.mod.CURE_POTENCY_RCVD)/100))
 
-    if (target:getAllegiance() == caster:getAllegiance() and (target:getObjType() == dsp.objType.PC or target:getObjType() == dsp.objType.MOB)) then
+    if (target:getAllegiance() == caster:getAllegiance() and (target:getObjType() == dsp.objType.PC or target:getObjType() == dsp.objType.MOB or target:getObjType() == dsp.objType.TRUST)) then
```
ibm2431: These checks for trust object type might be problematic later for the likes of Sakura, Brygid, etc (sphere trusts), but those are a long ways off, so this is probably fine for now. Just noting as something to keep in mind for the future, in case we want to consider some "trust is targetable" flags / functions.

Teodor, being undead, will also require some special handling to make the cures hit for 0.

scripts/globals/spells/trust/Excenmille.lua
```diff
@@ -0,0 +1,20 @@
...
+function onMagicCastingCheck(caster,target,spell)
+    if (caster:hasHate()) then
+	    caster:messageBasic(700,0,0)
+    else
+        return 0
+    end
+end
+
+function onSpellCast(caster,target,spell)
+    -- Todo: put Trusts's spawn in dialog here or in entity script?
```
ibm2431: Probably in the entity script, because the spawn-in dialogue isn't the _only_ dialogue they have, and it'd be nice to keep it all in the same place.

src/map/utils/petutils.cpp
```diff
@@ -818,6 +826,8 @@ namespace petutils
             ref<uint16>(&PTrust->stats, counter) = (uint16)((raceStat + jobStat + sJobStat));
             counter += 2;
         }
+        PTrust->m_Weapons[SLOT_MAIN]->setDamage(GetTrustWeaponDamage(PTrust));
+        PTrust->m_Weapons[SLOT_MAIN]->setDelay((uint16)(floor(1000* (petStats->cmbDelay / 60))));
```
ibm2431: Is there a weapon type somewhere?

scripts/globals/mobskills/double_thrust.lua
```diff
@@ -0,0 +1,28 @@
...
+end
+
+function onMobWeaponSkill(target, mob, skill)
+    local numhits = 2
+    local accmod = 1
+    local dmgmod = 1.0
+    local info = MobPhysicalMove(mob,target,skill,numhits,accmod,dmgmod,TP_DMG_VARIES,1.0,1.5,2.0)
+    local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,dsp.attackType.PHYSICAL,dsp.damageType.PIERCE,info.hitslanded)
+
+    target:takeDamage(dmg, mob, dsp.attackType.PHYSICAL, dsp.damageType.PIERCE)
```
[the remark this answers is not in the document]
TeoTwawki: ws and mobskill and mostly the same thing, and either can skillchain if we flag'em too. However we have likely been doing past scripts wrong (generic mobskill with guessed at values) so at some point in the future we should audit all the mobskill versions of ws's and get them more accurate as needed. Only a select few actually do anything like player ws in their script.

src/map/utils/petutils.cpp
TeoTwawki: These should just be getting their damage and delay from their mob_pools fields

scripts/globals/pets/Kupipi.lua
```diff
@@ -0,0 +1,283 @@
...
+	    spell = 18
+    end
+
+    return spell
+end
+
+function doKupipiWeaponskill(mob)
+    local wsList = {{43,164}, {1,163}}
+    local newWsList = {}
+	local maxws = 3 -- Maximum number of weaponskills to choose from randomly
```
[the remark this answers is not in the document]
Tummie: This was checked, apparently there are special cases where they will use specific weapon skills early if they only have access to 2 like Kupipi does.  There is a screenshot of a level 5 player having Kupipi weapon skill in the dev Discord

TeoTwawki:
@Tummie _nudge nudge_

.gitignore
lion85xx: any new about this project ?

scripts/globals/pets/Kupipi.lua
```diff
@@ -0,0 +1,283 @@
...
+
+	if (player:hasStatusEffect(dsp.effect.POISON) and lvl >= 6 and mp >= 8) then
+	    spell = 14
+	elseif (player:hasStatusEffect(dsp.effect.PARALYSIS) and lvl >= 9 and mp >= 12) then
+	    spell = 15
+	elseif (player:hasStatusEffect(dsp.effect.BLINDNESS) and lvl >= 14 and mp >= 16) then
+	    spell = 16
+	elseif (player:hasStatusEffect(dsp.effect.SILENCE) and lvl >= 19 and mp >= 24) then
+	    spell = 17
+	elseif (player:hasStatusEffect(dsp.effect.CURSE_I) and lvl >= 29 and mp >= 30) then
```
NiQ1: Kupipi also casts cursna to attempt to remove doom and if I remember correctly (not 100% sure though) she prioritizes it over other status effects.

.gitignore
NiQ1: > 
> 
> um...undo this please

Maybe we should rename the config files to "map_darkstar.conf.default" and have server owners copy them to ".conf", that way there's no risk of getting your current configuration overwritten when pulling / having to stash before pushing.

zynjec:
Going to close for now, feel free to open it again if you get back to it @Tummie

PR #5778, opened 2019-03-03
add mobmod IS_NM, fix removing Temp items, ALWAYS_AGGRO always agros

KnowOne134:
is nm and always aggro needed for nyzul WIP

### DIFF
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -2180,7 +2180,7 @@ dsp.mobMod =
     AUTO_SPIKES         = 46, -- enables additional effect script to process when mob is attacked
     SPAWN_LEASH         = 47, -- forces a mob to not move farther from its spawn than its leash distance
     SHARE_TARGET        = 48, -- mob always targets same target as ID in this var
-    IS_NM               = 49, -- If set, a mob will check as a NM.
+    CHECK_AS_NM         = 49, -- If set, a mob will check as a NM.
     -- 50 Available for use
     ROAM_TURNS          = 51, -- Maximum amount of turns during a roam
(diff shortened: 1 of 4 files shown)


scripts/globals/status.lua
```diff
@@ -2180,7 +2180,7 @@ dsp.mobMod =
     AUTO_SPIKES         = 46, -- enables additional effect script to process when mob is attacked
     SPAWN_LEASH         = 47, -- forces a mob to not move farther from its spawn than its leash distance
     SHARE_TARGET        = 48, -- mob always targets same target as ID in this var
-    -- 49 Available for use
+    IS_NM               = 49, -- If set, a mob will check as a NM.
```
TeoTwawki: we should make a binding to set MOBTYPE_NOTORIOUS rather than a mobmod.

TeoTwawki: type lives in mobentity.cpp btw

KnowOne134: Dont think it changes to a true nm stat wise. Its just a check showing as impossible

TeoTwawki: hmm that thing we do about the stats I'm pretty sure is wrong we should stop that and just get their stats right instead of multiplying based on wiki conjecture form over a decade ago.

and these should still inherit other NM like things like not being able to be charmed, rather not add a ton of mobmod checks in the same spots we already have nm checks.

@teschnei https://github.com/DarkstarProject/darkstar/blob/2924080116c08539ab245f2a8947f1d448237bdf/src/map/entities/mobentity.cpp#L529

whadawedo boss?

teschnei: For this PR, I think we should just rename this variable to make it obvious its only for /check, the stats we can deal with later

PR #5785, opened 2019-03-11
Fix an agro issue

TeoTwawki:
stuff using MOBMOD_ALWAYS_AGRO were agro'ing from up to 50 yalms away instead of normal range.

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -961,13 +961,8 @@ bool CMobController::Engage(uint16 targid)
 
 bool CMobController::CanAggroTarget(CBattleEntity* PTarget)
 {
-    if (PMob->getMobMod(MOBMOD_ALWAYS_AGGRO) > 0)
-    {
-        return true;
-    }
-    
     // Don't aggro I'm neutral
-    if (!PMob->m_Aggro || PMob->m_neutral || PMob->isDead())
+    if ((PMob->getMobMod(MOBMOD_ALWAYS_AGGRO) == 0 && !PMob->m_Aggro) || PMob->m_neutral || PMob->isDead())
     {
         return false;
     }


TeoTwawki:
no luv for bug, you merge nao?

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -961,34 +961,35 @@ bool CMobController::Engage(uint16 targid)
...
 
-    if (PTarget->isDead() || PTarget->isMounted())
+    // Don't aggro I'm special
+    if (PMob->getMobMod(MOBMOD_NO_AGGRO) > 0)
     {
         return false;
     }
 
+    // Always aggro, I'm special too!
+    if (PMob->getMobMod(MOBMOD_ALWAYS_AGGRO) > 0 && CanDetectTarget(PTarget))
```
teschnei: I don't think this is right either - the intended use case for this var are monsters who normally don't aggro (ie. m_Aggro is false) and allow it to be toggled on.  Problem is, this check is after the check for m_Aggro, so the function will just return false in the first check.  I think the check for `!PMob->m_Aggro` should be `(!PMob->m_Aggro && PMob->getMobMod(MOBMOD_ALWAYS_AGGRO) == 0)`, basically "if m_Aggro is false, and also the mobmod is not set, then don't aggro"

TeoTwawki: still has a distance prob if I do that, I either have to check the range here or use CanDetectTarget() which doesn't have any m_Aggro check in it just checks if the mob can see/hear you. Always shouldn't mean from across the zone and using the existing detection makes sense. What we do not want is checking CanAggroTarget(), which would be like you were thinking.

TeoTwawki: I did fk up the order at the last second though

TeoTwawki: I fail at reading comprehension forever, that is all.

> 0 vs == 0, derp

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -961,13 +961,8 @@ bool CMobController::Engage(uint16 targid)
...
 bool CMobController::CanAggroTarget(CBattleEntity* PTarget)
 {
-    if (PMob->getMobMod(MOBMOD_ALWAYS_AGGRO) > 0)
-    {
-        return true;
-    }
-    
     // Don't aggro I'm neutral
-    if (!PMob->m_Aggro || PMob->m_neutral || PMob->isDead())
+    if ((PMob->getMobMod(MOBMOD_ALWAYS_AGGRO) == 0 && !PMob->m_Aggro) || PMob->m_neutral || PMob->isDead())
```
TeoTwawki: @teschnei better?

PR #5790, opened 2019-03-13
remove debug message from arkangel galka

wrenffxi:
Nothing to see here ...


wrenffxi:
Me, when I go to help someone do Divine Might, and I get a system message when they enter the arena: "... Well, shit."

PR #5792, opened 2019-03-14
Added misc flag to zones for auction house usage

DivByXI:
People could abuse it to buy rare elixirs etc whilst fighting


Avatarati:
The current check for disabling AH usage in jail is adequate and many servers want their players to be able to access the AH outside of cities. Besides, the only existing means of accessing the AH outside of cities is through deliberate modification of the permissions in the command script. This script would force many server owners to modify core code to get around your fix.

ibm2431:
> Besides, the only existing means of accessing the AH outside of cities is through deliberate modification of the permissions in the command script.

Or through client modification, which can be done even on retail.

@DivByXI Throwing a var into `.conf` to override the in-city check should be sufficient to appease server operators who desire the access-anywhere behavior.

Avatarati:
> > Besides, the only existing means of accessing the AH outside of cities is through deliberate modification of the permissions in the command script.
> 
> Or through client modification, which can be done even on retail.
> 
> @DivByXI Throwing a var into `.conf` to override the in-city check should be sufficient to appease server operators who desire the access-anywhere behavior.

Yes that would be best. I didn't think about client modification, either. I think some kind of solution like this would be ideal as well, considering that aspect.

zynjec:
We're trying to avoid configuration bloat for non-retail functionality, it will be up to server administrators to customize as they see fit.

DivByXI:
Just ready your message as i pushed a workaround sorry

TeoTwawki:
suggestion: zone misc flag to handle the same way we do moogles. or even reuse the moogle flag, how many places have a mog for job change that do not have an AH? a couple? selbina and mhaura.

server owners can easily set/unset that flag with 1 sql query

ibm2431:
> how many places have a mog for job change that do not have an AH?

Past cities...!

Which makes me realize that checking just for city zone type isn't sufficient. Would have to be a flag of some kind.

TeoTwawki:
> Past cities...!

but would we really care :thinking: 

eh well, we can just make a new flag anyway instead of using moogle flag, won't kill us.

zynjec:
Could you update `status.lua` to include MISC_AH as well, so there is parity between the two.

PR #5794, opened 2019-03-16
Fix an exploit with limbus and a memory corruption vuln

DivByXI:
The limbus exploit can be done by spoofing the bitmask via packet injection or modding the client. They can effectively bypass the timer allowing them to buy cosmo-cleanse KI's as much as they want.

The memory corruption vuln isn't too bad because it will only null the memory values rather than write the custom itemID supplied due to the conditions. It's relative distance is short due to it being an 8 bit value for the equipSlotId too so people can't go giving themselves a high gmLevel :)

### DIFF
--- a/scripts/zones/Port_Jeuno/npcs/Sagheera.lua
+++ b/scripts/zones/Port_Jeuno/npcs/Sagheera.lua
@@ -253,6 +253,21 @@ local abcShop =
 
 -----------------------------------
 
+function getCosmoCleanseTime(player)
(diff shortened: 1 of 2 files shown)


scripts/zones/Port_Jeuno/npcs/Sagheera.lua
```diff
@@ -253,6 +253,21 @@ local abcShop =
 
 -----------------------------------
 
+function getCosmoCleanseTime(player)
```
TeoTwawki: this can be `local function` since it is only used here

PR #5796, opened 2019-03-18
Mob family cleanup

isotor:
Freeing up unused familyIds in `mob_family_system` and tagged as available for use. Confirmed all not present in `mob_pools`. Additionally, deleted unused familyIds present inside `mob_family_mods`


TeoTwawki:
I kind of wonder wtf the originator of the list used to decide how many they thought they needed, they were obviously trying to keep some kind of logical order which is an impossibility due to later additions to the game and undefined naming conventions.

PR #5804, opened 2019-03-23
Fix Eagle Eye Shot / Violent Flourish / mob's Freezebite

TeoTwawki:
I _think_ pet abilities **don't** need the same change but I could be wrong. I didn't try them I only looked as far as seeing this: https://github.com/DarkstarProject/darkstar/blob/58539ad5a2c612b0041135b2349d99824a51e1b6/scripts/globals/automatonweaponskills.lua#L239


ibm2431:
Yeah, automation being fine was the conclusion I came to as well.

(At least in the context that they still work. Whether they're _accurate_ might be another matter. But that's how it always is. :wink:)

PR #5817, opened 2019-04-03
Fixed some item models, added shadow lord shirt script

JulianoW:
First pull request, be nice and constructive! (*'-')/

I've created a mapping of all model ID/race/slot combinations to their respective dat files, so I'll be trying to think of some ways to automate this process. (At the very least, may be able to group by item icon to check for consistency? I don't know if there will be an actual way to map them, as it's probably completely server side. Either way, I was just messing around and decided to commit some easy fixes.)

I also noticed there's a good chunk of items missing mods, so I'll try to think of a way to parse them to identify mods to help clean that up.

### DIFF
--- a/scripts/globals/items/shadow_lord_shirt.lua
+++ b/scripts/globals/items/shadow_lord_shirt.lua
@@ -4,6 +4,10 @@
 -- Teleports user to Castle Zvahl Keep
 -----------------------------------------
 
+function onItemCheck(target)
+    return 0
+end
+
 function onItemUse(target)
 	-- Teleports user to Castle Zvahl Keep
 	-- May need a retail capture to verify
--- a/scripts/globals/items/shadow_lord_shirt.lua
+++ b/scripts/globals/items/shadow_lord_shirt.lua
@@ -3,15 +3,21 @@
 -- Item: Shadow Lord Shirt
 -- Teleports user to Castle Zvahl Keep
 -----------------------------------------
+require("scripts/globals/teleports")
+require("scripts/globals/status")
+-----------------------------------------
 
 function onItemCheck(target)
-    return 0
+    local result = 0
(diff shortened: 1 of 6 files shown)


scripts/globals/items/shadow_lord_shirt.lua
```diff
@@ -0,0 +1,17 @@
...
+
+function onItemCheck(target)
+    return 0
+end
+
+function onItemUse(target)
+	-- Teleports user to Castle Zvahl Keep
+	-- May need a retail capture to verify
+	-- https://ffxiclopedia.fandom.com/wiki/Shadow_Lord_Shirt states teleports next to HP#1
+	-- so I copied the homepoint teleport position
```
TeoTwawki: I think we should use the teleport effect and global instead like other items of this nature. Thats the only thing I see off in this PR :+1:

scripts/globals/items/shadow_lord_shirt.lua
```diff
@@ -0,0 +1,17 @@
+-----------------------------------------
+-- ID: 26517
+-- Item: Shadow Lord Shirt
+-- Teleports user to Castle Zvahl Keep
+-----------------------------------------
+
+function onItemCheck(target)
```
ibm2431: Does the shirt work without having visited the zone before? If so, should make sure the area has been visited before; see [Kingdom Stable Collar](https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/items/kingdom_stables_collar.lua) for example.

JulianoW: I don't know, but I think that's a safe assumption so I made those changes. Thanks!

scripts/globals/items/shadow_lord_shirt.lua
JulianoW: Done - thanks!

PR #5824, opened 2019-04-06
String tables for quests + simplify stage constants

ibm2431:
**Adds the previously mentioned quest string key tables**, which are built from our defined quest ID tables. This presents a unified string reference for quests, which can (and now are) be used to build a filepath to a quest's file. **The short-lived `quest.filenames` has been removed**. `quests.getQuest()` can now fully pull a quest table from just an ID, provided that the file exists. Of course, this means that the filename for a quest must be the same as its string key representation.

Also **simplified stage constants** a little.

**Like with last PR, I'll try to lay out (my) future plans:**
1. **Sneak more enum tables into master, like `dsp.fame.JEUNO`, and phase the would-be-duplicate definitions out of the current quest_rewrite enums** table.
2. Revisit requirements, using quest tables being pulled from just an ID, to **be able to check if a player is on/past a given stage** for a required quest/mission
3. Tweak how quest rewards are defined, and what the default behavior is for a call to `quests.complete()` when a reward set isn't passed to it. It'd be nice to be able to **simplify the process of defining rewards** for the vast majority of quests that only have a single reward set.
4. **Reconsider how quest vars should be _stored_ in the DB**, in that maybe instead of using prefix+scripter_var_name, if they should be stored as simple prefix+integer_indexes, with the quest functions abstracting away the conversion from scripter_varname to DB_definition. This is also a concern for being able to **simplify DB migration when the entire current quest player var system is radically upended** later.
5. I'm **waffling on the best order for subtables inside a stage**. [At the moment they're stage->zone->action->NPC or event](https://github.com/ibm2431/darkstar/blob/bf5b78527879e2652da5468fa3c2ecc63182c877/scripts/quests/adoulin/a_certain_substitute_patrolman.lua#L67). I can't decide which order is the most logical for scripting, compared against ease/speed for the quest functions to access them.
If we were being as true as possible to our _current_ quest implementation (in master), it'd be stage->zone->NPC->action, and have event CS IDs be "owned" by the NPC and checked in an `onFinish` subtable inside that NPC, as opposed to the way the stage table currently considers events to be a _separate entity_ from the NPC. I want to make it as easy as possible for people to adjust to the new quest layout, but also don't want to keep a convention  _just_ for the sake of keeping the convention (events as _part_ of an NPC), if another might be better (events are simply events, not NPCs).
**How DSP handles events currently (from an NPC script), versus how it _could_ in the _future_**, might play a factor in this decision.

### DIFF
--- a/scripts/globals/quests.lua
+++ b/scripts/globals/quests.lua
@@ -1207,6 +1207,38 @@ dsp.quest.id =
     }
 }
 
+-----------------------------------
+--  QUEST STRING KEYS
+-----------------------------------
+
+-- These are the string key counterparts for a quest, and can be used to find the string
+-- key/path that DSP uses for a given quest when all you have is the quest's ID and its area.
+-- Filepaths can be built using "(AREA_FOLDER)/".. dsp.quest.string.(AREA)[(QUEST_ID)],
+-- Where (AREA) is dsp.quest.area[AREAS_LOG_ID] (a string), _not_ the default integer log ID
+local function buildQuestStringTable(dspQuestIDtable)
+    local questStringTable = {}
+    for areaKey, areaQuestTable in pairs(dspQuestIDtable) do
+        local areaQuestStringTable = {}
+        for questStringKey, questID in pairs(areaQuestTable) do
+            areaQuestStringTable[questID] = questStringKey
+        end
+        questStringTable[areaKey] = areaQuestStringTable
+    end
+    return questStringTable
+end
+
+dsp.quest.string = buildQuestStringTable(dsp.quest.id)
+
+-----------------------------------
+--  DSP QUEST SYSTEM CONSTANTS
+-----------------------------------
+
+dsp.quest.stage =
+{
+    STAGE0   =  0, STAGE1  =  1, STAGE2  =  2, STAGE3   =  3, STAGE4   =  4,
+    STAGE5   =  5, STAGE6  =  6, STAGE7  =  7, STAGE8   =  8, STAGE9   =  9,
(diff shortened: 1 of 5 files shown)


ibm2431:
Shower thought, but `quests.getQuest()` can now also be used in place of the [manually defined requires in the NPC's table of quests to check](https://github.com/DarkstarProject/darkstar/blob/quest_rewrite/scripts/zones/Western_Adoulin/npcs/Shipilolo.lua#L15)

And then the file path / location of the quests is abstracted away by either getQuest:
```
involvedQuests =
{
    quests.getQuest(dsp.quest.area.ADOULIN, dsp.quest.id.adoulin.FERTILE_GROUND)
    quests.getQuest(dsp.quest.area.ADOULIN, dsp.quest.id.adoulin.WAYWARD_WAYPOINTS)
}
```

Or a table of area, quest ID pairs:
```
involvedQuests =
{
    {dsp.quest.area.ADOULIN, dsp.quest.id.adoulin.A_CERTAIN_SUBSTITUTE_PATROLMAN},
    {dsp.quest.area.ADOULIN, dsp.quest.id.adoulin.THE_OLD_MAN_AND_THE_HARPOON},
    {dsp.quest.area.ADOULIN, dsp.quest.id.adoulin.FERTILE_GROUND},
    {dsp.quest.area.ADOULIN, dsp.quest.id.adoulin.WAYWARD_WAYPOINTS}
}
```
With a "preloader" in the NPC script like `quests.readyQuests(involvedQuests)`. (Which just calls getQuest on all of them and returns the results, but spares the scripter of having to copy-paste the function name like in the first example.)

I won't touch them for now in case someone has a strong opinion about the best approach.

Issue #5807, opened 2019-03-25
Curaga Earring 


xerodok:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **n/a** 

**_Source Branch_** (master/stable) **Stable** 

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Curaga Earring does not work. It fails 1 second after using saying you cannot use this at this time. Solo, Level Sync and standard party all affected.

ValentineXI:
No associated script in scripts/globals/items. Anytime you get an instant *Unable to use* or *Unable to cast* it usually means a missing script.

PR #5827, opened 2019-04-07
[#5807] Add Curaga Earring

zach2good:
Original Issue: https://github.com/DarkstarProject/darkstar/issues/5807

**Heal amount (60-90) taken from:**
https://www.bg-wiki.com/bg/Curaga
https://www.bg-wiki.com/bg/Cure_Formula

I couldn't find any examples of items that affect the whole party, so this is my stopgap attempt.

**Tested:**
Solo
Standard Party

**Not Tested:**
Level Sync
Alliance

**Notes:**
- Took the mods from `curaga.lua`
- The visual effect of the item only appears on the user, not on the affected members
- The chat log only shows the personal healing done, not the healing for the whole party
- Changing the `aoe` flag in the SQL entry in `items_usable` doesn't appear to have done anything

I don't know or remember what the intended behavior for these is in retail 🤷‍♂️

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -8487,24 +8487,23 @@ inline int32 CLuaBaseEntity::getPartyLeader(lua_State* L)
 inline int32 CLuaBaseEntity::forMembersInRange(lua_State* L)
 {
     DSP_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
-    DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isuserdata(L, 1));
-    DSP_DEBUG_BREAK_IF(lua_isnil(L, 2) || !lua_isnumber(L, 2));
-    DSP_DEBUG_BREAK_IF(lua_isnil(L, 3) || !lua_isfunction(L, 3));
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, 2) || !lua_isfunction(L, 2));
 
-    auto PEntity = Lunar<CLuaBaseEntity>::check(L, 1)->m_PBaseEntity;
-    auto range = (uint8)lua_tointeger(L, 2);
-    auto function = luautils::register_fp(3);
+    auto target = (CBattleEntity*)m_PBaseEntity;
+    auto range = (uint8)lua_tointeger(L, 1);
+    auto function = luautils::register_fp(2);
 
-    ((CBattleEntity*)m_PBaseEntity)->ForParty([&PEntity, &range, &function](CBattleEntity* member)
-                                              {
-                                                  if (PEntity->loc.zone == member->loc.zone &&
-                                                      distanceSquared(member->loc.p, PEntity->loc.p) < (range * range))
-                                                  {
-                                                      luautils::pushFunc(function);
-                                                      luautils::pushArg<CBattleEntity*>(member);
-                                                      luautils::callFunc(1);
-                                                  }
-                                              });
+    target->ForParty([&target, &range, &function](CBattleEntity* member)
+    {
+        if (target->loc.zone == member->loc.zone &&
+            distanceSquared(target->loc.p, member->loc.p) < (range * range))
+        {
+            luautils::pushFunc(function);
+            luautils::pushArg<CBattleEntity*>(member);
+            luautils::callFunc(1);
+        }
+    });
 
     luautils::unregister_fp(function);
 
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -8325,7 +8325,7 @@ inline int32 CLuaBaseEntity::getParty(lua_State* L)
 
     lua_createtable(L, size, 0);
     int i = 1;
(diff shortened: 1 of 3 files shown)


TeoTwawki:
This would try to act on out of range or even out-of-zone party members (the latter can trigger certain crashes)

@teschnei I think we should have a method to this to make it easy on our contributors rather than making them do a bunch of checks to for distance and zone or whatever else might need checked. Agree?

ibm2431:
`forMembersInRange(target, distance, function)` would be nice. There are other items that have AoE party effects as well, like group foods.

TeoTwawki:
I was thinking more along the lines of item target behavior like the actual spell target behavior: cast on self (or party member if item lets you do that) and core knows what to do about the aoe already. just set a flag.

TeoTwawki:
`forMembersInRange(target, distance, function)` would be a nice util function to have anyway though.

zach2good:
I was interested in figuring out how c++ and lua interacted, so I had a go at `forMembersInRange`. My notes about the visual effect and the chat logs are still the same though.

ibm2431:
I like the can-do attitude!

Does the target need to be passed in as an arg if the function is always invoked on the target?

zach2good:
I guess selecting the target as an arg is a little redundant, seeing as the function could be invoked on something else and have the same effect

src/map/lua/lua_baseentity.cpp
```diff
@@ -8477,6 +8477,39 @@ inline int32 CLuaBaseEntity::getPartyLeader(lua_State* L)
...
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isuserdata(L, 1));
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, 2) || !lua_isnumber(L, 2));
+    DSP_DEBUG_BREAK_IF(lua_isnil(L, 3) || !lua_isfunction(L, 3));
+
+    auto PEntity = Lunar<CLuaBaseEntity>::check(L, 1)->m_PBaseEntity;
+    auto range = (uint8)lua_tointeger(L, 2);
+    auto function = luautils::register_fp(3);
+
+    ((CBattleEntity*)m_PBaseEntity)->ForParty([&PEntity, &range, &function](CBattleEntity* member)
+                                              {
```
teschnei: just do a single indent level instead of lining it up with the parenthesis

src/map/lua/lua_baseentity.cpp
```diff
@@ -8477,6 +8477,39 @@ inline int32 CLuaBaseEntity::getPartyLeader(lua_State* L)
...
+                                                  if (PEntity->loc.zone == member->loc.zone &&
+                                                      distanceSquared(member->loc.p, PEntity->loc.p) < (range * range))
+                                                  {
+                                                      luautils::pushFunc(function);
+                                                      luautils::pushArg<CBattleEntity*>(member);
+                                                      luautils::callFunc(1);
+                                                      luautils::unregister_fp(function);
```
teschnei: unregister_fp should be outside the loop right?

teschnei:
ignore those 2 comments, i wasn't looking at the full diff

TeoTwawki:
its not clear to me how retails curry buns work exactly. only on page mentioning group effects and its to state that the NQs effect increases with group size.
Alternate Food Effects: (30 minutes, In party of 2~3 members)
Dexterity +3
Vitality +5
Intelligence +2
Accuracy +7
Ranged Accuracy +7
Evasion +7
Defense +20% (Max. 190)
"Resist Sleep" +4
HP recovered while healing +4
MP recovered while healing +2
In a party of 4 or more members, eating this food has the same effects as Black Curry Bun +1


are all the curry buns aoe food on retail? I never ate one in a party

zach2good:
I'm not sure how any of the group buff foods work in retail, I just added one for another place to use `forMembersInRange`. All of the group buff foods have TODOs on, so I thought it would be a good idea to start on them.

ibm2431:
I'm not seeing a group effect listed for [Black Curry Buns](https://www.bg-wiki.com/bg/Black_Curry_Bun), and old wiki doesn't seem to be listing it under [Group Foods](https://ffxiclopedia.fandom.com/wiki/Group_Food_Effects). But from my memory of eating [Lebkuchken House](https://www.bg-wiki.com/bg/Lebkuchen_House) and [Gat. aux Fraises](https://www.bg-wiki.com/bg/Gat._aux_Fraises), everyone in the party got the same effect (although, it was a party of 3 people).

TeoTwawki:
I think Turkey and candy ring / candy cane too. And of course it'll only hit members who have no active food.

zach2good:
https://ffxiclopedia.fandom.com/wiki/Black_Curry_Bun
So based on this, the right sort of logic would look like:
```
function onItemUse(target)
    target:forMembersInRange(30, function(member)
        member:addStatusEffect(dsp.effect.FOOD,0,0,3600,5764)
    end)
end

  local partySize = table.getn(target:getParty())
  if partySize == 1 then
  -- basic effects
  elseif partySize < 4
  -- middle effects
  else
  -- full effects
  endif 
end

function onEffectLose(target, effect)
  -- remove, as above
end
```

At this point, should I remove my changes for the food and open them as their own PR?

zach2good:
I realise now that this logic I suggested will break if the party size changes between effect start and end, group food effects will need a little more thought. I'll take out that part of the PR and start work on it after this PR goes in.

PR #5833, opened 2019-04-09
OSX Build Support

zach2good:
Nobody asked, but here is build support for OSX. Obviously reliant on having your `include` and `lib` paths set up correctly.

**Setup (To be added to the wiki)**
`brew install git pkg-config autoconf make gcc@7 openssl mysql@5.7 mysql-client lua@5.1 luajit zeromq zmqpp`

**Build Instructions**
Otherwise the same as the Linux instructions:
https://wiki.dspt.info/index.php/Building_the_Server


TeoTwawki:
Neat, now we can run of fruity PCs as well a potato PCs

PR #5835, opened 2019-04-10
add tools for DB updating and 1st time setup

KnowOne134:


ibm2431:
Would it be possible to pull mysql paths and connection information from .conf? If so, two less files to have to worry about stashing when pulling in changes.

Also, option to pull from stored password would be nice as well.

"Well why don't you autofill the stored db as well?" - I suppose you could do that too. But, if everything is autosaved and autofilled, be sure to still display what it's about to operate on and ask the user to confirm before firing the queries. "You are about to update database 'dspdb', using username 'root' and password 'pass'. Continue?" Then if the user says no (instead of terminate), kick them back to something where they can manually change the input if they want.

I feel that the import folder should be automatically set to the relative path for the sql folder. These are (presumably) run after a `git pull`, so the user should have the most up-to-date copy of the SQL files in `../sql/`.

As a personal preference, I wouldn't do `cls`es. I like seeing the full execution in the log!

You've got some empty `cd`s after changing to your import path.

Finally, `initial`, not `initail` :wink:

KnowOne134:
I wouldnt want all the static info. I have about 4 different clones on my
pc in differemt locations and different db names

ibm2431:
By pulling from conf, the information will be the same as for the current install/clone. You're not running four different server instances from the same folder - that's why you have the clones in different locations in the first place~!

KnowOne134:
Sometimes i do!  "You dont know my life"

Another note. Not sure how to link it anyway lol. Just something i use.
Thought i would share it with the public. Little more involved than the dsp
one on setup guide

JulianoW:
Not too familiar with batch scripting, but wouldn't it be better to exclude the tables you don't want to drop, rather than trying to include the rest?

KnowOne134:
Yea could do that way. Happen to know what the exclude comment would be? Same as lua and c ~?

JulianoW:
No idea, don't do batch scripting. Sorry.

PR #5837, opened 2019-04-11
Table Mission IDs

ibm2431:
Assault and Campaign missions deliberately left out because their nature isn't quite the same as other mission logs, [we're currently treating assaults differently from other missions](https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_baseentity.cpp#L6328), and to give the content freedom to develop its own systems that are more tailored it (ie: `dsp.assault`) without boxing it in with systems for the other mission logs.

I recognize that tabling has made references to a mission long. But, the idea is that when quests and missions are rewritten, those long references will no longer exist, as they will not only be centralized in the quest/mission's file, but replaced with a simple `thisMission.id`. (Which is ultimately shorter than before!)


wrenffxi:
I PRed some mission references that were missed to ibm's branch.

ibm2431:
Didn't realize my regexes were apparently crap. Thanks @wrenffxi .

TeoTwawki:
HAAAAAALLLLLLLVVVVVVEEEEEEEEERRRRRRRRRrrrrrrrrrr

###### and screw you Kupipi! ..erase spamming gimp mage..

ibm2431:
I'm just trying to tell myself that it's more evidence of how badly they need tabling. If I spent multiple nights trying to find all references to mission IDs and _still_ bungled the task, all the more reason for a prefix we can search for to quickly find them!

I don't suck! It was.. uhm.. all just a demonstration! Yeah, that's it! I swear!

PR #5844, opened 2019-04-15
[WIP] Refactor Cmake

zach2good:
Don't know if this is overkill or not, but this is my 2 cents. Mad props to @ShelbyZ for submitting these files to start with!

**Project Files**
If properly defined, Cmake can replace the checked in vsproj file, maintaining the folder structure and developer workflow using `source_group`'s or the original folder structure. It also removes the need for in`#includes` to use relative paths. This will add a step to Windows users in the form of: `cmake .. -G "Visual Studio 15 2017 Win64" -DDEBUG etc.`

**Library Definition**
Define each source folder as either a `STATIC` library or an `INTERFACE` library, depending on if compilation is required

**Use of GLOBs**
Remove use of GLOB, as it isn't recommended by Cmake itself:
https://cmake.org/cmake/help/v3.3/command/file.html?highlight=glob#file
```
Note We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.
```
More discussion:
https://stackoverflow.com/questions/32411963/why-is-cmake-file-glob-evil
https://stackoverflow.com/questions/1027247/is-it-better-to-specify-source-files-with-glob-or-each-file-individually-in-cmak
https://reviews.llvm.org/D31363

**Dependency Resolution**
Defining the following files:
FindMySQL.cmake
FindZMQ.cmake
FindLua.cmake
FindLuaJIT.cmake
and making them _not-awful_ makes dependency resolution pretty simple, robust and cross-platform.  They can also be customized to have verbose and understandable logging.


ShelbyZ:
I originally thought to make the entire common as a static lib that we would link into the executables, but did not make much traction as a whole.  GLOB while evil was better accepted than the laundry list of files added to the add_executable line (or setting a variable).

I initially did not have luck with any of the Findxxx.cmake and moved to using PkgConfig (at least for unix platforms).  I would be interested in finding a solution for all platforms (even windows) and removing any local dll/lib files from the project.  I'm not sure if that direction means choosing vcpkg (for windows)...

zach2good:
I'm gonna take this in a different direction, so closing until there is substantial progress 🤷‍♂️

Issue #5843, opened 2019-04-15
Experience chain multipler triggers regardless of enemy difficulty


isotor:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30190328_2

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

__Steps to reproduce:__
- Kill a bee at level 2 > get 90 exp > kill another > get 108 exp _(+20% / Chain 1 bonus)_
- You will continue receiving 108 exp for subsequent kills until you take a break _(Chain multiplier doesn't increase further)_
- If you're idle for a few minutes (break the chain timer) and kill another bee, it will award 90 exp again
- This occurs at all levels

isotor:
Just for clarification/ stating the obvious, you shouldn't receive a chain bonus on kills that check below EM: https://ffxiclopedia.fandom.com/wiki/Experience_Chain

> The bonus is obtained by defeating monsters that check as Even Match or tougher (EM+) within a time limit that starts when the last EM+ mob is killed.

zach2good:
Ok, I've tested this a bit:
- Killed a few EP's and DC's back to back, no chain.
- A few EM+'s back to back gives the appropriate chain. Killing less than an EP in the middle of a chain doesn't reset the chain, but also doesn't increase it or reset the timer, as stated on ffxiclopedia.

The only thing I wanted to clarify was if you kill your firstEM+, you get a notification that you've started a chain: Exp Chain #1!

TeoTwawki:
clarification needed:
you are chaining things weaker than even or the chain bonus is failing to higher when it is actually supposed to?

coz I know I can't chain easy prey's on stock darkstar master

zach2good:
Now that @TeoTwawki has prompted me, I've tried killing (DC) bees at Lv2 for 90xp, but a chain never forms. I'm not sure what sort of madness I was under before when I thought I had a fix for this...

TeoTwawki:
> Unsure what happens on retail if you interrupt a chain with a mob below EM but kill another EM within the original chaintime window (i.e. I'm not sure if that EP/DC in between resets the chain or is ignored)

[...]

PR #5845, opened 2019-04-15
[#5843] Add checking for level difference in exp chain calcs

zach2good:
Potential fix for:https://github.com/DarkstarProject/darkstar/issues/5843

Used this for EM calculation:
https://ffxiclopedia.fandom.com/wiki/Check
```
Even Match
Abbreviation: EM

An Even Match enemy is the same level you are.
```

Not had time to test, but will do in the next few days 🙏 

EDIT: EM logic can also be worked out with:
`charutils::GetRealExp(PPartyMember->GetMLevel(), PMob->GetMLevel()) >= 100; // EM or Higher`
But its less clear?

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -3205,7 +3205,7 @@ namespace charutils
             auto PMember = static_cast<CCharEntity*>(PPartyMember);
             uint32 baseexp = 0;
             auto exp = 0.f;
-            int8 levelDifference = PPartyMember->GetMLevel() - PMob->GetMLevel();
+            int8 levelDifference = PMob->GetMLevel() - PPartyMember->GetMLevel();
             bool evenMatchOrHigher = levelDifference >= 0;
             float permonstercap, monsterbonus = 1.0f;
             bool chainactive = false;
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -3205,6 +3205,8 @@ namespace charutils
             auto PMember = static_cast<CCharEntity*>(PPartyMember);
             uint32 baseexp = 0;
             auto exp = 0.f;
+            int8 levelDifference = PPartyMember->GetMLevel() - PMob->GetMLevel();
+            bool evenMatchOrHigher = levelDifference >= 0;
             float permonstercap, monsterbonus = 1.0f;
             bool chainactive = false;
             if (PMob->m_HiPCLvl > maxlevel) maxlevel = PMob->m_HiPCLvl;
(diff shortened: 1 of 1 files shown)


zach2good:
Ok, I've tested this a bit:

Killed a few EP's and DC's back to back, no chain.
A few EM+'s back to back gives the appropriate chain. Killing less than an EP in the middle of a chain doesn't reset the chain, but also doesn't increase it or reset the timer, as stated on ffxiclopedia.
The only thing I wanted to clarify was if you kill your first EM+, you get a notification that you've started a chain: Exp Chain #1!, is this expected behavior? I don't remember from retail

isotor:
The first EM+ shouldn’t give a chain bonus, instead the first EM+ should “set up” the chain. You can think of this as Chain 0 - though it has no message and provides no bonus itself :)

zach2good:
I've re-checked this to be absolutely sure the logic is right.
On first kill of an EM+ mob, there is no notification.
On the next kill within time of an EM+ mob, you get the notification "Exp Chain 1"

isotor:
I've just tested this and it seems to be awarding a chain in the following scenario:

1. Killed a Decent Challenge > awarded 90 exp
2. Killed an Even Match > awarded 125 Exp and Chain #1

This is not correct, the Chain #1 should only trigger on the second in a series of EM+ mobs *(i.e. EM+ > EM+ = Chain #1 on the second)*

But aside from that, it's spot on 👍

TeoTwawki:
> Also tested, using numbers at level 2 for consistency:
> 
> 1. Killed an EM+ > awarded 100 Exp
> 2. Killed a DC (level 1) > awarded 108 exp (+20%, chain 1 modifier, no notification)
> 
> You can repeat steps 1 multiple times and step 2 always gives +20% exp **without notification.** Further, you can repeat step 2 more than once and repeatedly get +20% exp from a DC-or-less mob, until the `chainTime` expires.

"without notification"?

helixhamin:
I think he means it will not show the "Chain 1" message in the box, but will still give the extra exp.

TeoTwawki:
I mean it the issue report should have explained this specifically - this is a prompt to elaborate so we know wtf we should be looking at

isotor:
@helixhamin is correct. 

@TeoTwawki @zach2good updated the original issue with a clarification of the whole issue, and hopefully a pointer towards the line that remains an issue in this PR.

TeoTwawki:
@zach2good if there is still something here you want in dsp, let me know and I'll reopen this. it needs rebased and we'll probably need to chat about it due to previous confusion - hit me up on discord

PR #5853, opened 2019-04-16
Changes as outlined by: https://wiki.dspt.info/index.php/Adjusting_th…

Bizzle-Dapp:
…e_subjob_level_to_be_different_than_the_main_job_level


Bizzle-Dapp:
Don't approve this, boys! Was only meant for my solution

Issue #5850, opened 2019-04-16
Guard proc without native hand-to-hand skill


isotor:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30190328_2

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Guard can proc on jobs without native hand-to-hand skill (any job attacking without a weapon). It should only be possible to guard with MNK or PUP set as main or sub job. Unsure how this interacts with skill gained from guard merits when otherwise the skill wouldn't exist.

zach2good:
Relevant area (maybe):
https://github.com/DarkstarProject/darkstar/blob/master/src/map/utils/attackutils.cpp#L120

PR #5856, opened 2019-04-16
[#5850] Change guard skill proc rate to check jobs, rather than equipped weapon

zach2good:
Issue: https://github.com/DarkstarProject/darkstar/issues/5850

```
Guard can proc on jobs without native hand-to-hand skill (any job attacking without a weapon). It should only be possible to guard with MNK or PUP set as main or sub job. Unsure how this interacts with skill gained from guard merits when otherwise the skill wouldn't exist.
```

As far as I can see, there is a hidden "weapon" that is inserted behind the scenes when you don't have a weapon equipped. This would allow `PWeapon->getSkillType() == SKILL_HAND_TO_HAND` to pass and allow any class without a weapon to guard. So checking instead for MNK or PUP being main/sub job will fix this.

**Not yet tested, will test all my things at the weekend :)**

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1726,6 +1726,10 @@ namespace battleutils
         // Defender must have no weapon equipped, or a hand to hand weapon equipped to guard
         bool validWeapon = (PWeapon == nullptr || PWeapon->getSkillType() == SKILL_HAND_TO_HAND);
 
+        if (PDefender->objtype == TYPE_MOB || PDefender->objtype == TYPE_PET) {
+            validWeapon = PDefender->GetMJob() == JOB_MNK || PDefender->GetMJob() == JOB_PUP;
+        }
+
         auto hasH2HSkill = PDefender->GetSkill(SKILL_HAND_TO_HAND);
 
         if (validWeapon && hasH2HSkill && PDefender->PAI->IsEngaged())
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1730,7 +1730,7 @@ namespace battleutils
             validWeapon = PDefender->GetMJob() == JOB_MNK || PDefender->GetMJob() == JOB_PUP;
         }
 
-        auto hasH2HSkill = PDefender->GetSkill(SKILL_HAND_TO_HAND);
+        bool hasH2HSkill = PDefender->GetSkill(SKILL_HAND_TO_HAND) > 0;
 
         if (validWeapon && hasH2HSkill && PDefender->PAI->IsEngaged())
         {
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1730,7 +1730,7 @@ namespace battleutils
             validWeapon = PDefender->GetMJob() == JOB_MNK || PDefender->GetMJob() == JOB_PUP;
         }
 
-        bool hasGuardSkillRank = GetSkillRank(SKILL_GUARD, PDefender->GetMJob() > 0 || GetSkillRank(SKILL_GUARD, PDefender->GetSJob() > 0;
+        bool hasGuardSkillRank = (GetSkillRank(SKILL_GUARD, PDefender->GetMJob()) > 0 || GetSkillRank(SKILL_GUARD, PDefender->GetSJob()) > 0);
 
         if (validWeapon && hasGuardSkillRank && PDefender->PAI->IsEngaged())
         {
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1730,9 +1730,10 @@ namespace battleutils
             validWeapon = PDefender->GetMJob() == JOB_MNK || PDefender->GetMJob() == JOB_PUP;
         }
 
-        bool hasH2HSkill = PDefender->GetSkill(SKILL_HAND_TO_HAND) > 0;
+        auto combinedGuardSkillCap = battleutils::GetMaxSkill(SKILL_GUARD, PDefender->GetMJob(), PDefender->GetMLevel()) +
+                                     battleutils::GetMaxSkill(SKILL_GUARD, PDefender->GetSJob(), PDefender->GetSLevel());
 
-        if (validWeapon && hasH2HSkill && PDefender->PAI->IsEngaged())
+        if (validWeapon && (combinedGuardSkillCap > 0) && PDefender->PAI->IsEngaged())
         {
             // assuming this is like parry
             float skill = (float)PDefender->GetSkill(SKILL_GUARD) + PDefender->getMod(Mod::GUARD);
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1730,10 +1730,9 @@ namespace battleutils
             validWeapon = PDefender->GetMJob() == JOB_MNK || PDefender->GetMJob() == JOB_PUP;
         }
 
-        auto combinedGuardSkillCap = battleutils::GetMaxSkill(SKILL_GUARD, PDefender->GetMJob(), PDefender->GetMLevel()) +
-                                     battleutils::GetMaxSkill(SKILL_GUARD, PDefender->GetSJob(), PDefender->GetSLevel());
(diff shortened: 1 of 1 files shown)


TeoTwawki:
I don't think a static job check is the right solution. Check for native skill + h2h would be better I think, unless we provably know for a fact that you can add guard skill to a h2h war or something and get a tiny useless proc rate.

/2cents

TeoTwawki:
*can't

zach2good:
Thats a very good point, I hadn't considered that 💡

src/map/utils/battleutils.cpp
```diff
@@ -1723,14 +1723,10 @@ namespace battleutils
     {
         CItemWeapon* PWeapon = GetEntityWeapon(PDefender, SLOT_MAIN);
 
-        // Defender must have no weapon equipped, or a hand to hand weapon equipped to guard
-        bool validWeapon = (PWeapon == nullptr || PWeapon->getSkillType() == SKILL_HAND_TO_HAND);
+        bool hasGuardSkill = (PDefender->GetMJob() == JOB_MNK || PDefender->GetSJob() == JOB_MNK ||
+                              PDefender->GetMJob() == JOB_PUP || PDefender->GetSJob() == JOB_PUP);
```
ibm2431: You don't get skillups from skills that come only from subjobs.

TeoTwawki: He's trying to effect proc rate anyway, not skill gain

ibm2431: Sorry, saw "skill rate" and thought this was about rate of gaining skill. Carry on.

zach2good:
Changed this an did some testing:

As an unarmed WHM getting beaten up: weapon was valid, but no H2H skill, so it would not go into the guard proc rate calculation, and return 0 guard skill.

As an unarmed (or armed) MNK getting beaten up: weapon was valid and H2H skill was populated, so it would calculate proc rate and return that.

zynjec:
I don't quite think that check is accurate still, only MNK and PUP get guard, but MNK, PUP, DNC, WAR, NIN, and THF all have native hand-to-hand skill (and weapons). 

Unless I'm missing something, this would allow anyone with hand-to-hand skill and hand-to-hand weapons equipped a chance to guard (albeit a small percentage), when really only MNK and PUP should be able to.

zach2good:
Even adding something like `bool canGuard = PDefender->GetSkill(SKILL_GUARD) > 0;` allows a warrior to enter the guard calculations. I guess its up to mods to decide which is appropriate, it would be very hard to confirm this in retail because of the tiny margins

zynjec:
I think its a fairly safe assumption that an entity with guard skill would be able to guard (either through their main job, or support job).

To continue your example with a warrior, they'd only satisfy the condition of `PDefender->GetSkill(SKILL_GUARD) > 0;` with a support job of MNK or PUP as they don't have guard itself. In which case, they'd be granted a guard rate based on the support jobs current guard skill (eg: 114 skill for a MNK37 subjob).

TeoTwawki:
and **if** worried about if a job that gets skill but is currently at 0 can proc, just check for the cap instead of current to tell if its a valid "guarding entity"

war/mnk will still have a cap > 0 even if current == 0. so we can and should be looking at guard skill not h2h skill.

teschnei:
that's less an "if" and more a "you do actually need to check the rank or cap or else new players will never be able to guard until they skill up once" (which is also impossible with the old style defensive skillups)

TeoTwawki:
meh, I didn't look to see if he was affecting the skillup attempts along with actual proc'ing of the skill (what order that happens in) retail lets you skill up on "failed procs" now but I guess that'd still be _after_ this check then.

zach2good:
Testing snippet for reference:
```
if (PDefender->objtype == TYPE_PC)
        {
            ShowDebug("MJob: %d, SJob: %d,validWeapon: %d, canGuard: %d, SKILL_GUARD: %d\n",
                    PDefender->GetMJob(), PDefender->GetSJob(), validWeapon, canGuard, PDefender->GetSkill(SKILL_GUARD));
        }
```

**Results:**
**MNK75/NONE0, no weapon:**
MJob: 2, SJob: 0,validWeapon: 1, canGuard: 1, SKILL_GUARD: 1

**WAR75/NONE0, no weapon:**
MJob: 1, SJob: 0,validWeapon: 1, canGuard: 0, SKILL_GUARD: 0

**WAR75/NONE0, no weapon, wearing some guard+ gear:**
MJob: 1, SJob: 0,validWeapon: 1, canGuard: 1, SKILL_GUARD: 17

**MNK75/NONE0, no weapon, wearing some guard+ gear:**
MJob: 2, SJob: 0,validWeapon: 1, canGuard: 1, SKILL_GUARD: 17

**WHM75/BLM, no weapon:**
MJob: 3, SJob: 4,validWeapon: 1, canGuard: 0, SKILL_GUARD: 0

**WHM75/MNK37, no weapon:**
MJob: 3, SJob: 2,validWeapon: 1, canGuard: 1, SKILL_GUARD: 1

teschnei:
two more tests: WAR/NON with guard+ gear (should not be able to guard) and WAR/NON with guard merits (also should not be able to guard)

TeoTwawki:
@teschnei 

```
WAR/NON with guard+ gear (should not be able to guard) and WAR/NON with guard merits (also should not be able to guard)
````
thats actually the exact thing I wasn't sure of at the begging of this.

**Comment by zach2good** [comment]
I've since changed the code in my PR, so here is some logging using the following snippet that shows all the discussed options:

```
bool validWeapon = (PWeapon == nullptr || PWeapon->getSkillType() == SKILL_HAND_TO_HAND);

        bool isJobThatHasGuardSkill = (PDefender->GetMJob() == JOB_MNK || PDefender->GetSJob() == JOB_MNK ||
                                       PDefender->GetMJob() == JOB_PUP || PDefender->GetSJob() == JOB_PUP);

        auto combinedGuardSkillCap = battleutils::GetMaxSkill(SKILL_GUARD, PDefender->GetMJob(), PDefender->GetMLevel()) +
                                     battleutils::GetMaxSkill(SKILL_GUARD, PDefender->GetSJob(), PDefender->GetSLevel());

        auto guardSkill = PDefender->GetSkill(SKILL_GUARD);

        if (PDefender->objtype == TYPE_PC)
        {
            ShowDebug("MJob: %d, SJob: %d, validWeapon: %d, isJobThatHasGuardSkill: %d, combinedGuardSkillCap: %d, guardSkill: %d\n",
                      PDefender->GetMJob(), PDefender->GetSJob(), validWeapon, isJobThatHasGuardSkill, combinedGuardSkillCap, guardSkill);
        }
```

Unless otherwise stated, the PC doesn't have any gear, merits or (a relevant) subjob.

**MNK75**
MJob: 2, SJob: 1, validWeapon: 1, isJobThatHasGuardSkill: 1, combinedGuardSkillCap: 269, guardSkill: 3

**WAR75/MNK37**
MJob: 1, SJob: 2, validWeapon: 1, isJobThatHasGuardSkill: 1, combinedGuardSkillCap: 114, guardSkill: 3

**WAR75**
MJob: 1, SJob: 3, validWeapon: 1, isJobThatHasGuardSkill: 0, combinedGuardSkillCap: 0, guardSkill: 0

**WAR75, with 5x guard merits:**
MJob: 1, SJob: 3, validWeapon: 1, isJobThatHasGuardSkill: 0, combinedGuardSkillCap: 0, guardSkill: 10

**WAR75, with guarding torque:**
MJob: 1, SJob: 3, validWeapon: 1, isJobThatHasGuardSkill: 0, combinedGuardSkillCap: 0, guardSkill: 7



To me, `isJobThatHasGuardSkill` or `combinedGuardSkillCap` seem like valid choices, and they end up being the same check

teschnei:
isJobThatHasGuardSkill is not really that great because if we ever add guard skill to another job, we'd have to edit that in order to guard too.  You guard if you have the guard skill, so it should check the skill.  Anyways, I think the proper check is to just check GetSkillRank for the main and sub job, since it doesn't depend on level or anything

zynjec:
```src/map/utils/battleutils.cpp: In function ‘uint8 battleutils::GetGuardRate(CBattleEntity*, CBattleEntity*)’:
src/map/utils/battleutils.cpp:1733:138: error: expected ‘)’ before ‘;’ token
         bool hasGuardSkillRank = GetSkillRank(SKILL_GUARD, PDefender->GetMJob() > 0 || GetSkillRank(SKILL_GUARD, PDefender->GetSJob() > 0;```

PR #5858, opened 2019-04-17
Added missing item_basic up to item 21514.

helixhamin:
Still need to verify missing items after that.
(I didn't want to flood with too many added items xD)


TeoTwawki:
> I didn't want to flood with too many added items xD

Thank you. 

What were the flag changes on existing items?

helixhamin:
The original flags were set to `28736`. This is for `no auction, no sale, no delivery, ex`

According to bg this is not the correct flags:
https://www.bg-wiki.com/bg/Handshard%3A_WAR

Flags: Obtainable from Goblin Box, Not vendorable

That should be flag set 4098, 
2=obtainable from goblin box
4096=no sale

All of these items, 9544 to 9763 share the same properties, so I changed the existing ones as I added the missing ones.

TeoTwawki:
thanks

Issue #5813, opened 2019-03-31
Aura Sculpture and Aura Statue sharing SQL info they should not


TeoTwawki:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 
n/a

**_Source Branch_** (master/stable) **:** 
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Similar to #5808 Aura Sculpture and Aura Statue are sharing drop data (I didn't look beyond that).

http://ffxidb.com/zones/178/aura-statue
http://ffxidb.com/zones/178/aura-sculpture

Reminders: 
1. use TH0 column, not the average. 
2. don't include any 0% drops
3. global drops such as crystals/seals/geodes/avatarites are not to be included.

TeoTwawki:
p.s.
If are looking for things to do to contribute, just look around our SQL files you can find a lot of minor and easily fixable things like this.

PR #5860, opened 2019-04-17
[#5813] Give aura sculpture own drops, separate from aura statue

zach2good:
Original issue:
https://github.com/DarkstarProject/darkstar/issues/5813

References:
http://ffxidb.com/zones/178/aura-statue
http://ffxidb.com/zones/178/aura-sculpture


zach2good:
Tested that killing Aura Sculptures will eventually yield a Seasoning Stone, which is not present for Aura Statues

PR #5861, opened 2019-04-17
Rework WS Damage Calculations

ibm2431:
Since I needed something to do while waiting for #5824 to be merged, I started a different project, and ended up in `weaponskills.lua`. One thing lead to another and I ended up reworking the functions used to calculate WS damage, with some fixes along the way:

- Mainly, I put together `calculateRawWSDamage`, which can be used for both physical and ranged WS - since both types calculate damage in roughly the same way, but using different values - with behavior changing based on the `calcParams` table passed into it. Broke out the single-hit calculation into its own function to reduce the amount of copy-paste code.

- Moved the positions of `doPhysicalWeaponskill, doRangedWeaponskill, doMagicalWeaponskill`, and `takeWeaponskillDamage` to be together near the top, instead of interspersed below all the utility functions

- Reordered all calls to doPhysical, doRanged, and doMagical to take their arguments all in the same order

- Removed the break that'd halt calculating further hits if the currently-tabulated damage was more than the target's HP. This was being done _before_ factoring in target resistances, resulting in situations where the WS damage calculation results in lower numbers than it should due to quitting early.

- Fixed Ranged WS being affected by Mighty Strikes

- Nullify alpha from damage calculations if the server is set to use Adoulin WS changes

- Added the hybrid hit for hybrid WSes, made hybrid WS previously marked as magical to be physical + hybrid hit. They can now also crit (when forced), and also get their offhand bonus hit. (These WSes also give Ruby light in Abyssea, furthering the case for doPhysical as a base over doMagical.)
["Base damage for these weaponskills is calculated exactly the same as for Physical Weapon Skills. However, the overall damage is calculated as if the first hit of the weapon skill is duplicated, and the duplicate hit is magical and thus is affected by modifiers of magic damage."](https://www.bg-wiki.com/bg/Weapon_Skill_Damage#Calculating_Hybrid_Weapon_Skill_Damage)

- Made the offhand hit during a WS use the damage for the offhand weapon instead of the mainhand weapon
["Each hit of the weapon skill will be calculated this way, using the appropriate base weapon damage (mainhand vs. offhand)"](https://www.bg-wiki.com/bg/Weapon_Skill_Damage)

- Added the 100% accuracy bonus for the offhand hit when Trick Attack is active.
["This ability guarantees 100% accuracy when the activation conditions are met. Accuracy bonus also applies to the offhand hit when dual-wielding."](https://www.bg-wiki.com/bg/Trick_Attack)

- Corrected the bonus damage from a THF's SA/TA being given the ftp multiplier when it shouldn't
["When added to Weapon Skills, the base damage is outside of the fTP multiplier."](https://www.bg-wiki.com/bg/Trick_Attack)

- Gave Requiescat a formless attribute in the context of damage calculation. It still counts as "physical damage" type (for now, anyway), but at least it'll cut through defense and PDT like it's supposed to
["Deals property-less damage (not Magic or Physical), but uses regular physical damage equations"](https://www.bg-wiki.com/bg/Requiescat)
["Requiescat deals non-elemental and unresistable damage, similar to Spirits Within, but in Abyssea it is still considered a physical weaponskill"](https://ffxiclopedia.fandom.com/wiki/Requiescat)

- Renamed `dsp.effect.ALL_MISS`, which was only used by wyrms when airborne, to `dsp.effect.TOO_HIGH` to more readily convey what the status means ("the wyrm is too high for your regular melee attacks to hit", not "everything misses!"). Gave Jump and High Jump params of `hitsHigh`, so DRGs can now properly jump on airborne dragons! (Ranged and Magic WS still hit as before.)
["While Ouryu is in the air, all melee attacks will miss. Dragoon Jump attacks will still connect."](https://ffxiclopedia.fandom.com/wiki/The_Savage)

- Although `doMagicalWeaponskill` doesn't use the same logic as doPhysical and doRanged (and therefore isn't using the new functions I put together), I tidied it up a little bit so it'll at least look / flow in a similar manner to the other two

One thing that I'm not certain on is _when_ Souleater should be applied during the calculations:

Is the damage added on for every hit of the WS, meaning it gets WSC and fTP?
Is it just a static value added on at the very end?
Should it stack with WS DMG+?

I looked for definitive answers, but all the internet was returning were, "Souleater increases weaponskill damage" and people a decade in the past complaining how much DRK sucks because it takes damage to deal damage. So I just left it where was, before WS DMG+ is applied.

### DIFF
--- a/scripts/globals/abilities/eagle_eye_shot.lua
+++ b/scripts/globals/abilities/eagle_eye_shot.lua
@@ -42,7 +42,7 @@ function onUseAbility(player,target,ability,action)
     params.atk100 = 1 params.atk200 = 1 params.atk300 = 1
     params.enmityMult = 0.5
 
-    local damage, criticalHit, tpHits, extraHits = doRangedWeaponskill(player, target, 0, params, 0, true, action)
+    local damage, criticalHit, tpHits, extraHits = doRangedWeaponskill(player, target, 0, params, 0, action, true)
 
     -- Set the message id ourselves
     if (tpHits + extraHits > 0) then
--- a/scripts/globals/abilities/high_jump.lua
+++ b/scripts/globals/abilities/high_jump.lua
@@ -30,6 +30,7 @@ function onUseAbility(player,target,ability,action)
     params.atk100 = 1 params.atk200 = 1 params.atk300 = 1
     params.bonusTP = player:getMod(dsp.mod.JUMP_TP_BONUS)
     params.targetTPMult = 0
+    params.hitsHigh = true
 
     if (target:isMob()) then
         local enmityShed = 50
@@ -41,7 +42,7 @@ function onUseAbility(player,target,ability,action)
 
     local taChar = player:getTrickAttackChar(target)
 
-    local damage, criticalHit, tpHits, extraHits = doPhysicalWeaponskill(player, target, 0, 0, true, action, taChar, params)
+    local damage, criticalHit, tpHits, extraHits = doPhysicalWeaponskill(player, target, 0, params, 0, action, true, taChar)
 
     if (tpHits + extraHits > 0) then
         -- Under Spirit Surge, High Jump reduces TP of target
--- a/scripts/globals/abilities/jump.lua
+++ b/scripts/globals/abilities/jump.lua
@@ -30,9 +30,10 @@ function onUseAbility(player,target,ability,action)
     local atkmulti = (player:getMod(dsp.mod.JUMP_ATT_BONUS) + 100) / 100
     params.atk100 = atkmulti params.atk200 = atkmulti params.atk300 = atkmulti
     params.bonusTP = player:getMod(dsp.mod.JUMP_TP_BONUS)
+    params.hitsHigh = true
 
     local taChar = player:getTrickAttackChar(target)
-    local damage, criticalHit, tpHits, extraHits = doPhysicalWeaponskill(player, target, 0, 0, true, action, taChar, params)
+    local damage, criticalHit, tpHits, extraHits = doPhysicalWeaponskill(player, target, 0, params, 0, action, true, taChar)
 
     if (tpHits + extraHits > 0) then
         -- Under Spirit Surge, Jump also decreases target defense by 20% for 60 seconds
--- a/scripts/globals/ability.lua
+++ b/scripts/globals/ability.lua
@@ -526,7 +526,7 @@ function AbilityFinalAdjustments(dmg,mob,skill,target,skilltype,skillparam,shado
     end
 
     --handle pd
-    if ((target:hasStatusEffect(dsp.effect.PERFECT_DODGE) or target:hasStatusEffect(dsp.effect.ALL_MISS) )
+    if ((target:hasStatusEffect(dsp.effect.PERFECT_DODGE) or target:hasStatusEffect(dsp.effect.TOO_HIGH) )
             and skilltype == dsp.attackType.PHYSICAL) then
         skill:setMsg(dsp.msg.basic.JA_MISS_2);
         return 0;
--- a/scripts/globals/automatonweaponskills.lua
+++ b/scripts/globals/automatonweaponskills.lua
@@ -108,7 +108,7 @@ function doAutoPhysicalWeaponskill(attacker, target, wsID, tp, primary, action, 
     local tpHitsLanded = 0;
     local shadowsAbsorbed = 0;
     if ((missChance <= hitrate or isSneakValid or isAssassinValid or math.random() < attacker:getMod(dsp.mod.ZANSHIN)/100) and
-            not target:hasStatusEffect(dsp.effect.PERFECT_DODGE) and not target:hasStatusEffect(dsp.effect.ALL_MISS) ) then
+            not target:hasStatusEffect(dsp.effect.PERFECT_DODGE) and not target:hasStatusEffect(dsp.effect.TOO_HIGH) ) then
         if not shadowAbsorb(target) then
             if (params.canCrit or isSneakValid or isAssassinValid) then
                 local critchance = math.random();
@@ -154,7 +154,7 @@ function doAutoPhysicalWeaponskill(attacker, target, wsID, tp, primary, action, 
             local chance = math.random();
             local targetHP = target:getHP();
             if ((chance<=hitrate or math.random() < attacker:getMod(dsp.mod.ZANSHIN)/100) and
-                    not target:hasStatusEffect(dsp.effect.PERFECT_DODGE) and not target:hasStatusEffect(dsp.effect.ALL_MISS) ) then  -- it hit
+                    not target:hasStatusEffect(dsp.effect.PERFECT_DODGE) and not target:hasStatusEffect(dsp.effect.TOO_HIGH) ) then  -- it hit
                 if not shadowAbsorb(target) then
                     pdif = generatePdif(cratio[1], cratio[2], true);
                     if (params.canCrit) then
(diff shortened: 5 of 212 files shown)


teschnei:
I'd guess Souleater is after WS DMG+ is applied, but I don't think there's enough of that stat anyways for it to matter that much.

Anyways, regarding the alpha change, there's a few things that I find odd.  bg-wiki lists the merit WS as having 85% modifiers on their stat.  If you check the history, you'll find that they used to say 100%.  The alpha at 99 is 0.85, which made these equivalent if you think about pre-alpha and post-alpha.  However, if alpha is now removed, why are they 85% still?  Did alpha really get removed or does everything just have it applied at 0.85 in advance?  Or did certain WS get nerfed when alpha was removed (as removing alpha would buff everything)?

ibm2431:
My understanding of it is that alpha was removed as part of the [June 2014 WS overhaul](http://forum.square-enix.com/ffxi/threads/42615). But as usual the official version notes aren't entirely helpful as it lists the Merit WS for Post-Adjustment as just "Varies depending on merit points allocated".

I did find the [original testing thread](https://www.bluegartr.com/threads/121610-Rehauled-Weapon-Skills-tier-lists) over at bgwiki.

On page 2 there's a poster saying:
["Apparently they removed alpha correction from the WSC values this update in addition to changing the attribute modifiers themselves (Motenten traced the source back to a JP dev post stating they were removing the 0.85 alpha mod)."](https://www.bluegartr.com/threads/121610-Rehauled-Weapon-Skills-tier-lists/page2)

(I haven't been able to find this JP dev post yet)

But on page 5, Mote himself says:
["Have heard someone mention that alpha still seems to apply to merited weaponskills. Based on the results I'm seeing in Stardiver, that seems quite likely."](https://www.bluegartr.com/threads/121610-Rehauled-Weapon-Skills-tier-lists/page5)

There's some talk on later pages about alpha still being applied to the merit WS as an SE oversight or glitch, but on Page 13, Mote muses:
["Remember that the merit weaponskills have always had what the players considered a strange way of describing the WSC mod. They weren't described as going up to 100%, but up to 85%. We treated it as 100% with the standard 0.85 multiplier, but maybe SE really didn't use the alpha multiplier on them to begin with."](https://www.bluegartr.com/threads/121610-Rehauled-Weapon-Skills-tier-lists/page13)

I'll do some more digging (I'd like to find this JP dev post that supposedly exists).

ibm2431:
Try as I might, I have been unable to find the original JP dev post that was mentioned (although I did find the actual person who said so; it wasn't Mote):

["Two new JP posts today. First one seems to be confirming that merit ws still have the 85% alpha applied to them"](https://www.bluegartr.com/threads/118737-Dev-Tracker-Discussion?p=6149136&viewfull=1#post6149136)

However, BG seems pretty convinced that alpha is gone for everything _except_ Merit WS:

[Byrth: "No. It doesn't look like SE is including Alpha in these calculations."](https://www.bluegartr.com/threads/118737-Dev-Tracker-Discussion?p=6135657&viewfull=1#post6135657)

[Prothescar: "alpha correction removal is extremely unexpected. that's another 15% boost to WSC."](https://www.bluegartr.com/threads/118737-Dev-Tracker-Discussion?p=6136880&viewfull=1#post6136880)

["With the news that alpha correction is being removed"](https://www.bluegartr.com/threads/118737-Dev-Tracker-Discussion?p=6136939&highlight=alpha#post6136939)

[Mote: "Known changes are new WSC values and the removal of the alpha multiplier."](https://www.bluegartr.com/threads/110460-DPS-Spreadsheets?p=6138982&highlight=alpha#post6138982)

[Sechs: "They also all got the indirect buff of alpha-removal."](https://www.bluegartr.com/threads/121608-DAT-Mining-6-17-2014-Update-Findings-and-Discussion?p=6140303&highlight=alpha#post6140303)

[Sechs: "Removed Alpha and done specific adjustments (the full extent of which is still unknown) to almost every WS."](https://www.bluegartr.com/threads/121608-DAT-Mining-6-17-2014-Update-Findings-and-Discussion?p=6140442&highlight=alpha#post6140442)

[Someone months later unhappy about the removal: "Removing alpha was unnecessary for weapon skills and only added to the experience of everything ending up the way it has"](https://www.bluegartr.com/threads/118737-Dev-Tracker-Discussion?p=6261769&viewfull=1#post6261769)

[Byrth in 2015: "(no Alpha anymore)"](https://www.bluegartr.com/threads/104123-Weapon-Skills?p=6338656&highlight=alpha#post6338656)

Now, **in regards to the change history on bgwiki for the merit WS**:

[Sechs: "The thing about alpha still being present for Merit WS is quite something."](https://www.bluegartr.com/threads/121610-Rehauled-Weapon-Skills-tier-lists?p=6142217&highlight=alpha#post6142217)

["but that might just be because Requiescat still has alpha pox"](https://www.bluegartr.com/threads/121610-Rehauled-Weapon-Skills-tier-lists?p=6144269&highlight=alpha#post6144269)

I'm inclined to agree with [Mote's theory that **Merit WS were never using alpha to begin with**](https://www.bluegartr.com/threads/121610-Rehauled-Weapon-Skills-tier-lists?p=6144562&viewfull=1#post6144562). My addition to that is maybe it'd have to do with the way their WSC increased with merits, so SE used a different calculation system for them. I also agree with [Mote in that **players were _assuming_ that alpha was always applied to them, so the Merit WS "were 100% then apply alpha"**](https://www.bluegartr.com/threads/121610-Rehauled-Weapon-Skills-tier-lists?p=6144588&highlight=alpha#post6144588). But after alpha was removed, either realized the mistaken assumption and lowered it to 85% (no alpha ever, we just assumed), or they lowered it to 85% because 100% mod * still-existing alpha is 85% anyway ("alpha still there for merit WS only, so if you're using our WS equations use 85% for the mod"). But you have the same result regardless how you want to represent it.

tldr conclusion: I am 99% certain that alpha is gone, and that merit WS simply never used it to begin with.

teschnei:
so when alpha was removed, did that effectively buff every single WS by 1/0.85? or did they modify the wsc to compensate?

ibm2431:
It depends on the WS. Most of them got an outright buff (the start of WS damage creep, according to some players). Others got the alpha-removal buff with some WSC changes (still in favor of increasing damage).

[Compare Herculean Slash](https://www.bg-wiki.com/index.php?title=Herculean_Slash&action=historysubmit&diff=380367&oldid=270941): 60% VIT to 80% VIT. It benefits from the alpha-removal, _and_ got a WSC buff.

[Compare Jishnu's Radiance](https://www.bg-wiki.com/index.php?title=Jishnu%27s_Radiance&action=historysubmit&diff=380450&oldid=270608): 60% DEX to 80% DEX. Like Herculean, it got double-buffed. 

[Compare Blade: Ten](https://www.bg-wiki.com/index.php?title=Blade%3A_Ten&action=historysubmit&diff=389406&oldid=271491): 30% STR / 30% INT to 30% STR / 30% DEX. Other than the WSC mod change (which is still a buff in-of-itself for NIN), the percentages weren't, but it still sees increased damage from the alpha-removal.

[Compare Refulgent Arrow](https://www.bg-wiki.com/index.php?title=Refulgent_Arrow&action=historysubmit&diff=380730&oldid=271764). 60% STR to 60% STR. It did get a fTP increase at 3000, but otherwise unchanged. Still gets the boost all other WS  (except Merit) got from removing alpha. 

[Compare Shark Bite](https://www.bg-wiki.com/index.php?title=Shark_Bite&action=historysubmit&diff=551406&oldid=268297): 50% DEX to 40% DEX / 40% AGI. The DEX was dropped down 10% (which compensates for removing alpha, but is still a 5% net increase), and AGI was added. So a net buff plus extra damage.

So, "everything except merit got a buff, and some got more of a buff than others".

It seems that [Fr0zt took care of the mod changes](https://github.com/DarkstarProject/darkstar/commit/70ee9fad8956d3f7d16b3b943e6048b53ee920c5#diff-a37784db6ad28f22bdee63f83d526638) back during #865

PR #5862, opened 2019-04-18
Add ilvl 119 Mnk/Pup items

helixhamin:
Added lots of level 99, ilvl 199, items, and corrected a few other things.


helixhamin:
I would add in hep.baghnakhs, but I am not sure how to do the Adds "Dragon Kick" effect

I know it means it allows other jobs that can get the 225 skill level natively to use dragon kick, but I have no idea how to implement that. I will let someone else add in that actual item_weapon xD

TeoTwawki:
We have a modifier that adds ws. I think it bypasses the skill requirement however. But it could be copied and altered if so. Or I might be wrong and there is a skill requirement - I can't check at the moment.

helixhamin:
Okay, I added a few more items. Stopping here for now before it becomes too much.
I can do more on another pull.

@TeoTwawki

Issue #5864, opened 2019-04-18
Learning Blue Magic


hargrave81:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 
N/A

**_Source Branch_** (master/stable) **:** 
master and stable

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
ffera.com

So unless there is some dark code in some corner of the source, or I am just very unlucky with learning blue spells, it appears the learning rate is very random and the AF hands do not affect the learning rate.

https://github.com/DarkstarProject/darkstar/blob/5b252b982291bbb50e597e6cb12f181a20ff4728/src/map/utils/blueutils.cpp

Function on Line 79: **void TryLearningSpells**

                // TODO: check for blue learning bonus and adjust base percent
                if (dsprand::GetRandomNumber(100) < 33) {

that code is in that function, it only checks a random number to see if its a 33% chance to learn.  The gloves do not alter the 33% in any way (say give it a 50% change to learn)  Rendering the gloves useless.

helixhamin:
Yep, pretty much xD

Delkathus:
rng is a bitch

PR #5866, opened 2019-04-18
[#5864] Learning Blue Magic 

zach2good:
Original Issue:
https://github.com/DarkstarProject/darkstar/issues/5864

Using this list:
http://ffxi.allakhazam.com/db/skillmods.html?fskillmod=1321

And this:
https://ffxiclopedia.fandom.com/wiki/Category:Blue_Magic#How_to_Learn_Blue_Magic
`The Magus Bazubands/Magus Bazubands +1/Assimilator's Bazubands/Assimilator's Bazubands +1 are the only known way to directly increase the chance to learn Blue Magic.`

And all entries (real or not) of Assimilator's Bazubands in the items db

Tested by logging out `chanceToLearn ` while getting hit by 1000 needles and bomb toss as a blu with and without the AF gloves


Hozu:
You'll have to create a modifier for the bonus, and assign it to the appropriate equips. Not only does this mean that the items won't be hardcoded in the core (ie. no need to modify that part if new equipment comes out with those effects), but modifiers are affected by level sync (ie. this wouldn't work below equip level), while equip ID checks are not.

src/map/utils/blueutils.cpp
```diff
@@ -136,8 +136,18 @@ void TryLearningSpells(CCharEntity* PChar, CMobEntity* PMob) {
             // make sure the difference between spell skill and player is at most 31 points
             if (playerSkillLvl >= skillLvlForSpell - 31)
             {
-                // TODO: check for blue learning bonus and adjust base percent
-                if (dsprand::GetRandomNumber(100) < 33) {
+                uint chanceToLearn = 33;
```
nfaletra: This needs to be a mod. Do not hardcode item ids like this.

zach2good:
OK, I'll look into it 👍

zach2good:
Thanks for all the help @nfaletra, all the points should be covered now

nfaletra:
No problem. As a side note, I personally think `auto` for simple types is bad. Not everyone on the project agrees with that so it's fine here. I tend to only use it for complex types (like iterators and lambdas). I feel it promotes laziness and hinders readability for someone else looking at the code when types are explicitly stated.

Issue #5865, opened 2019-04-18
Earthbreaker shouldn't wipe shadows like AOE magic


isotor:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30190328_2

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Weird one since it's a magical AOE move, right?

I vaguely tried to take a crack at submitting this directly as a PR but I'm not quite sure how to update. Earthbreaker is currently defined as a [MobMagicalMove](https://github.com/DarkstarProject/darkstar/blob/230ab12088d288122c3414757ff74e0f25a3d61b/scripts/globals/mobskills/earthbreaker.lua#L24) with param `WIPE_SHADOWS` when it should be work like a multihit AOE using something like `numhits`.

I found a video showing Earthbreaker stripping shadows twice *(2 then 1, respectively)*, timestamped here:
https://youtu.be/hy3GPJc3kEw?t=128
https://youtu.be/hy3GPJc3kEw?t=408

Not sure how many the max numhits should be, though if I had to bet I'd go with 3. Any help on how to slide this parameter into a MagicalMove would be appreciated.

Oh, and you can chalk up the inaccuracy to [#wikifails](https://ffxiclopedia.fandom.com/wiki/Earthbreaker) @TeoTwawki

isotor:
**Related:** unsure if this same issue should apply to it's little brother Earth Pounder but I rather strongly suspect it does.

PR #5867, opened 2019-04-18
[#5865] Earthbreaker shouldn't wipe shadows like AOE magic 

zach2good:
Original Issue:
https://github.com/DarkstarProject/darkstar/issues/5865

Seems simple enough

### DIFF
--- a/scripts/globals/mobskills/earthbreaker.lua
+++ b/scripts/globals/mobskills/earthbreaker.lua
@@ -22,7 +22,7 @@ function onMobWeaponSkill(target, mob, skill)
 
     local dmgmod = 1
     local info = MobMagicalMove(mob,target,skill,mob:getWeaponDmg()*6,dsp.magic.ele.EARTH,dmgmod,TP_NO_EFFECT)
-    local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,dsp.attackType.MAGICAL,dsp.damageType.EARTH,MOBPARAM_WIPE_SHADOWS)
+    local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,dsp.attackType.MAGICAL,dsp.damageType.EARTH,MOBPARAM_3_SHADOW)
 
     target:takeDamage(dmg, mob, dsp.attackType.MAGICAL, dsp.damageType.EARTH)
     return dmg


Delkathus:
According to bgwiki and ffxiclopedia it supposed to wipe shadows.

https://ffxiclopedia.fandom.com/wiki/Earthbreaker
https://www.bg-wiki.com/bg/Category:Scorpion#Abilities

TeoTwawki:
@Delkathus we have video footage showing wiki incorrect, see the reported issue for a link.

@zach2good if the param has it wipe "up to" 3 shadows but the skill does a single hit, its gonna take just one still. Confirm the behavior. 

We also need proper multihit anyway to do the skill correctly if it really is multihit, but if this change does what you were expecting that can be a separate PR.

zach2good:
Just did some testing with Serket

_MOBPARAM_WIPE_SHADOWS:_
`Test's Copy Image (3) effect wears off.`

_MOBPARAM_3_SHADOW:_
`1/2 of Test's shadows absorb the damage and disappear`

using _MOBPARAM_3_SHADOW_ recreates what we see in the videos :)

isotor:
I've tried building this and unfortunately it doesn't work as it's lacking any sort of multihit parameter. I'm guessing what this adjustment does is cause shadows to be removed using some sort of math.random when it should be based on the hitrate of the ability (see my comment about `numhits` in the report).

Without `numhits` or something similar there is never any chance of damage overspill. What I mean here is if you have *any* number of shadows Earthbreaker won't do any damageᵗ, which is incorrect.

ᵗ *I've eaten around 50 earthbreakers testing this locally with between 1-3 shadows and never taken damage*

isotor:
Addendum: not my call but this turns the skill from a truck into a kitten, personally I'd err on the side of overtuned *(current)* instead of very weak *(this PR)* until the right answer is found.

zach2good:
Now that I've slept on it, I agree with @isotor. Its better for now for the skill to be far too powerful than far too weak. I'll take another crack at this in a couple of days 👍

isotor:
A few comments for your next crack at this @zach2good:

1. I tried simply changing it to a `MobPhysicalMove` in order to use the existing `numhits` class and went to punch Serket a bunch on 75 MNK *(B+ eva)* and was avoiding all hits at a much higher rate than I should have *(compared with the video in the issue and others I reviewed yesterday)* so assuming a similar implementation I'd suggest an `accmod` greater than 1 as a baseline.
2. The issue with Earthbreaker almost certainly extends to the normal mob equivalent move Earth Pounder, so may as well kill two birds with one PR 👍

TeoTwawki:
```
I'm pretty sure from watching YT for sources on this yesterday I encountered a few instance of "[Name] evades.".
```
I'm pretty sure its physical based on that, unless a status is being evaded and not the damage thats a strong indication its a physical move.

Issue #5637, opened 2019-01-23
Capricious Cassie is weak


davismj:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30181222_1

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

https://ffxiclopedia.fandom.com/wiki/Capricious_Cassie

- She doesn't draw in _correctly_ (Needs a new feature to do correctly)
- ~~Bad Breath and Sweet Breath are conal~~
- ~~Impale doesn't reset hate~~
- ~~Uses Vampiric Lash instead of Drain Whip~~
- Doesn't drop 15,312 ~ 32,767 gil
- ~~Probably doesn't drop Amity Cape wtf~~

PR #5868, opened 2019-04-19
Make Cassie Great Again (skills update)

isotor:
Addresses Issue #5637 

1. created `drain_whip` lua using [wiki](https://ffxiclopedia.fandom.com/wiki/Drain_Whip) source
2. cleaned up formatting in `impale` lua and added target enmity reset to `mob:isMobType(MOBTYPE_NOTORIOUS)` check ([source](https://ffxiclopedia.fandom.com/wiki/Capricious_Cassie))
3. cleaned up formatting in `bad_breath` and `sweet_breath` lua files (no logic changes)
4. uncommented the Cassie/Vivian mobskill group 725-728 in `mob_skills` sql. Made both breath moves AOE and increased range to ~~50~~ 40 yalms [source](https://ffxiclopedia.fandom.com/wiki/Capricious_Cassie) & see Wiggo's comment below
5. added new skill_list for Cassie and Vivian
6. Added missing alliance Draw In effect to Cassie

**Queries:**
- Unsure if line 31 of the new `drain_whip` lua needs to be there, please advise I'm unfamiliar with lua so left it in to be safe. Further, the `math.random` check is just my first attempt, if there is a better way happy to hear it.
- Impale is duplicated in `mob_skills` sql as 316 and 725 are identical, I used the 725 ID for Cassie and Viv to keep the list tight, and it already existed just comment out. Please advise if better to stick with 319 and keep 725 commented out.


Wiggo32:
AoE for bad breath is 40 yalms, better source: https://youtu.be/i_kQsWccHb4?t=2306
You should use the animation and skill ids that are associated with the mob not random other mobskills that may have the same name. watching the whole vid will provide all of that info right on screen as the mob does its mobskills, and also a more understanding of the AoE testing.

scripts/globals/mobskills/drain_whip.lua
```diff
@@ -0,0 +1,37 @@
...
+end
+
+function onMobWeaponSkill(target, mob, skill)
+    local drainEffect = MOBDRAIN_HP
+    local dmgmod = 1
+    local info = MobMagicalMove(mob, target, skill, mob:getWeaponDmg()*3, dsp.magic.ele.DARK, dmgmod, TP_NO_EFFECT)
+    local dmg = MobFinalAdjustments(info.dmg, mob, skill, target, dsp.attackType.MAGICAL, dsp.damageType.DARK, MOBPARAM_IGNORE_SHADOWS)
+
+    local rnd = math.random()
+        if rnd < 0.33 then
```
TeoTwawki: woops, I merged to fast - bad indent. fixy pls?

scripts/globals/mobskills/drain_whip.lua
```diff
@@ -0,0 +1,37 @@
...
+            drainEffect = MOBDRAIN_TP
+        elseif rnd < 0.67 then
+            drainEffect = MOBDRAIN_MP
+        else
+            drainEffect = MOBDRAIN_HP -- does this need to be here?
```
TeoTwawki:
```lua
local rnd = math.random(3)
if rnd == 1 then
    drainEffect = MOBDRAIN_TP
elseif rnd == 2 then
    drainEffect = MOBDRAIN_MP
end
```

PR #5869, opened 2019-04-19
Move cactrot rapido pool mods to mob pools and update cmbDelay

isotor:
- `MOBMOD_SKILL_LIST` 703 updated in pools and deleted.
- cmbdelay obtained from watching [this](https://www.youtube.com/watch?v=jm40Czsj_xo) long solo vid by avesta with a stopwatch for a while and `DELAY` mod deleted.
- `MOBMOD_IMMUNITY ` 15 already present in pools so deleted.


sql/mob_pool_mods.sql
```diff
@@ -107,10 +107,7 @@ INSERT INTO `mob_pool_mods` VALUES (2675,4,25,1);
 INSERT INTO `mob_pool_mods` VALUES (1456,368,20,0);
 INSERT INTO `mob_pool_mods` VALUES (592,302,45,0);
 INSERT INTO `mob_pool_mods` VALUES (592,68,15,0);
-INSERT INTO `mob_pool_mods` VALUES (595,14,703,1);
 INSERT INTO `mob_pool_mods` VALUES (595,169,250,0);
-INSERT INTO `mob_pool_mods` VALUES (595,171,40,0);
```
TeoTwawki: this looks like a positive increase not a negative. the line should be removed, but I'd leave the pool delay as it is until someone reconfirms what the correct value should actually be.

isotor: Okay, tho anyone with 5-10 mins an a stopwatch with a lap function (https://vclock.com/stopwatch/#) can confirm this from the above link and various other available sources.

TeoTwawki: didn't even notice the link. but theres a diff problem than you are thinking: how the F we calculate h2h delay in dsp. what you put here is not the ms that the mob will have between swings if we're applying trains that adjust it (monk martial arts)

did you try this in game on dsp and see if it matched that video after your changes?

TeoTwawki: I've decided to merge as-is, because what it fixes is more valuable than what _may possibly be_ off and its better than wiki's "he's fast, he attacks fast, btw he's fast" super accurate description we can go off, when we know 360+40 for 400 delay was definitely not fast.

TeoTwawki: Btw thanks for catching and fixing the jobs. that was likely copy pasta at work.

PR #5876, opened 2019-04-19
Fix for the optional dialog chain not being optional.

Bizzle-Dapp:
Sources suggest no action other than mission active is required for Ruillont to request you to go get his Bronze Sword.
https://www.bg-wiki.com/bg/San_d%27Oria_Mission_2-1

[My first commit - be gentle...]

### DIFF
--- a/scripts/zones/Ordelles_Caves/npcs/Ruillont.lua
+++ b/scripts/zones/Ordelles_Caves/npcs/Ruillont.lua
@@ -23,7 +23,7 @@ function onTrigger(player, npc)
     if player:getCurrentMission(SANDORIA) == THE_RESCUE_DRILL then
         local missionStatus = player:getVar("MissionStatus")
 
-        if (missionStatus >= 2 && missionStatus <= 7) then
+        if (missionStatus >= 2 and missionStatus <= 7) then
             player:startEvent(1)
         elseif missionStatus >= 10 or player:hasCompletedMission(SANDORIA, THE_RESCUE_DRILL) then
             player:showText(npc, ID.text.RUILLONT_INITIAL_DIALOG + 9)
--- a/scripts/zones/Ordelles_Caves/npcs/Ruillont.lua
+++ b/scripts/zones/Ordelles_Caves/npcs/Ruillont.lua
@@ -23,7 +23,7 @@ function onTrigger(player, npc)
     if player:getCurrentMission(SANDORIA) == THE_RESCUE_DRILL then
         local missionStatus = player:getVar("MissionStatus")
 
-        if missionStatus == 7 then
+        if missionStatus <= 7 then
             player:startEvent(1)
         elseif missionStatus >= 10 or player:hasCompletedMission(SANDORIA, THE_RESCUE_DRILL) then
             player:showText(npc, ID.text.RUILLONT_INITIAL_DIALOG + 9)
--- a/scripts/zones/Ordelles_Caves/npcs/Ruillont.lua
+++ b/scripts/zones/Ordelles_Caves/npcs/Ruillont.lua
@@ -23,7 +23,7 @@ function onTrigger(player, npc)
     if player:getCurrentMission(SANDORIA) == THE_RESCUE_DRILL then
         local missionStatus = player:getVar("MissionStatus")
 
-        if (missionStatus >= 2 and missionStatus <= 7) then
+        if missionStatus >= 2 and missionStatus <= 7 then
             player:startEvent(1)
         elseif missionStatus >= 10 or player:hasCompletedMission(SANDORIA, THE_RESCUE_DRILL) then
             player:showText(npc, ID.text.RUILLONT_INITIAL_DIALOG + 9)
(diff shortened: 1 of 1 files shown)


ibm2431:
This will cause skipping steps 1 through 7 if the player talks to this NPC after the mission has been started, as the [eventFinish for cs 1](https://github.com/DarkstarProject/darkstar/blob/945ce35e64e518d3a2fbb75787d658ef0e005c5f/scripts/zones/Ordelles_Caves/npcs/Ruillont.lua#L44) advances the mission status to 8.

scripts/zones/Ordelles_Caves/npcs/Ruillont.lua
```diff
@@ -23,7 +23,7 @@ function onTrigger(player, npc)
     if player:getCurrentMission(SANDORIA) == THE_RESCUE_DRILL then
         local missionStatus = player:getVar("MissionStatus")
 
-        if missionStatus == 7 then
+        if missionStatus <= 7 then
```
ibm2431: This will cause skipping steps 1 through 7 if the player talks to this NPC after the mission has been started, as the [eventFinish for cs 1](https://github.com/DarkstarProject/darkstar/blob/945ce35e64e518d3a2fbb75787d658ef0e005c5f/scripts/zones/Ordelles_Caves/npcs/Ruillont.lua#L44) advances the mission status to 8.

ibm2431: If talking to this NPC is strictly optional and shouldn't influence the quest, the check here is fine, but you'll have to remove the mission status update in event 1, and either:
1) Find whatever NPC comes before this one and have it skip over 7 straight to 8; or
2) Find all NPCs that come after this one and decrement the required mission statuses to trigger them (and what they set the mission status too in the event finish) _after_ the player reaches mission status 7 by 1.

TeoTwawki:
last free week I did this on retail, you do at a minimum have to speak with one NPC outside but I forget which one - I tried to go straight to the cave to him and got nothing. clicking all the other NPCS in the area can be skipped. Its the guy outside the ravine, up above, that sends you down there.

Bizzle-Dapp:
Righto! I'll head back to La Theine and confirm shortly, but i believe it is Equesobillot that states "You're here to join the rescue drill, head down if so" or something of the like. It'll be this chap that's the required trigger. 
@ibm2431, regarding what you said, would leaving the current event triggers in not mean that the optional dialog still moves in the same flow if needed? But speaking with Ruillont will simply advance to the point where everyone but the required NPC states "Ruillont's sword? I don't have it." ?

Cheers for the swift review, guys. Really appreciate it - this is my first time in the realms of lua so i may be overlooking some aspects of how it works functionally.

TeoTwawki:
I had to go look him up in game, this dude, if IDs didn't shift and mix up the names: https://ffxiclopedia.fandom.com/wiki/Galaihaurat
other wiki at link there seems to back up my recolectiona nd name him as the one
```
First talk to Galaihaurat at the southern part of (E-6).
```

Bizzle-Dapp:
Yup, you're entirely right:
`
if (MissionStatus == 0) then
            player:startEvent(110);
----->
if (csid == 110) then
        player:setVar("MissionStatus",2);
`
Should i still alter the other NPC's dialog to trigger between the values as opposed to on the specific?

Bizzle-Dapp:
So regarding Ruillont - since there is an inherent sequence to the events that unfold between the NPC's, is it not easier to have Ruillont activate the request to go get the Bronze Sword on this condition:
`
if (missionStatus >= 2 && missionStatus <= 7) then
            player:startEvent(1)
`
This way, the same sequence of events can unfold in order, and only in order, yet Ruillont will advance the quest no matter what stage of the optional NPC dialog the player is at? 

Or should each of the optional NPC's trigger providing the player has spoken with the initial NPC but regardless of the sequence they are spoken to?

ibm2431:
Does the optional NPC dialog go in a specific order?

If so, you can probably leave _them_ as-is to have that in-order behavior.

If they can be spoken to in _any_ order, the onFinish for those optional events should probably be cut out of the `missionStatus` chain altogether, and their triggers be if `missionStatus == whatever_status_is_before_ruillont`.

Now that I've looked at the mission, Ruillont is required to be spoken to, so if CS 1 is him asking for his sword, then the check should be a range (>= 2 and <= 7), and the onAdvance can set it to 8 as per normal. At the moment you've just got a less than, so the player can bypass Galaihaurat.

scripts/zones/Ordelles_Caves/npcs/Ruillont.lua
```diff
@@ -23,7 +23,7 @@ function onTrigger(player, npc)
     if player:getCurrentMission(SANDORIA) == THE_RESCUE_DRILL then
         local missionStatus = player:getVar("MissionStatus")
 
-        if missionStatus == 7 then
+        if (missionStatus >= 2 and missionStatus <= 7) then
```
TeoTwawki: don't need to enclose it in (parenthesis)

Issue #5847, opened 2019-04-16
Crystal drops are determined incorrectly


isotor:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30190328_2

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Issue discovered by Eden player roclm [here](https://github.com/EdenServer/community/issues/1299). The relevant extract:

*"As it stands right now, you cannot obtain crystals in regions in which the Beastmen are the Dominant influence. This is incorrect behavior, crystals should still be obtainable in regions in which the Beastmen are the Dominant influence as long as one of the three Nations controls the region."*

tl;dr: crystlal drops should be determined by region control as assigned during the weekly conquest tally, and not by current level of influence.

isotor:
Relevant section: https://github.com/DarkstarProject/darkstar/blob/2924080116c08539ab245f2a8947f1d448237bdf/src/map/entities/mobentity.cpp#L840

Would replacing `GetInfluenceGraphics` with `GetRegionOwner` do the trick?

PR #5887, opened 2019-04-23
Fix crystal drop check in ROZ areas

isotor:
Should drop based on regional owner *(either San d'Oria, Bastok, of Windurst)* not current level of influence.

Addresses issue #5847.

### DIFF
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -837,7 +837,7 @@ void CMobEntity::DropItems(CCharEntity* PChar)
 
     if (validZone && charutils::GetRealExp(PChar->GetMLevel(), GetMLevel()) > 0)
     {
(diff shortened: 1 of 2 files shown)


isotor:
I've not been able to get this working checking locally, but the local is so simple at this point I'm willing to accept I'm probably just doing something wrong in my test build lol. Anyone else welcome to try this - feedback appreciated.

isotor:
Reference: https://ffxiclopedia.fandom.com/wiki/Category:Conquest

> You only get crystals in regions under the control of San d'Oria, Windurst or Bastok. Regions controlled by beastmen will prevent crystals from dropping from monsters killed within that region. 

This doesn't specifically address address when `GetRegionOwner() = 4` (no owner) but I think it's better to assume you don't get crystals in this case.

PR #5893, opened 2019-04-25
Added a couple missing items.

helixhamin:
Submitting now in case I did this whole "Make a weaponskill" thing wrong, so I don't do too many other changes before realizing I have no idea what I am doing xD
Please feel free to let me know if I messed this up completely...


sql/weapon_skills.sql
```diff
@@ -253,6 +253,7 @@ INSERT INTO `weapon_skills` VALUES (217,'trueflight',0x0000000000000000000001000
 INSERT INTO `weapon_skills` VALUES (218,'leaden_salute',0x00000000000000000000000000000000010000000000,26,0,255,228,2000,16,1,9,1,0,1,34);
 INSERT INTO `weapon_skills` VALUES (220,'wildfire',0x00000000000000000000020000000000020000000000,26,0,2,235,2000,16,1,14,9,0,1,48);
 INSERT INTO `weapon_skills` VALUES (221,'last_stand',0x00000000000100000000010000000000010000000000,26,357,0,241,2000,16,1,11,5,0,1,0);
+INSERT INTO `weapon_skills` VALUES (241,'final_paradise',0x01010101010101010101010101010101010101010101,1,0,0,25,2000,5,1,13,11,0,0,0);
```
zynjec: Shouldn't its ID be 228?

scripts/globals/weaponskillids.lua
```diff
@@ -29,6 +29,7 @@ dsp.weaponskill =
     TORNADO_KICK        = 13,
     VICTORY_SMITE       = 14,
     SHIJIN_SPIRAL       = 15,
+    FINAL_PARADISE      = 241,
```
zynjec: Should also be 228.

sql/item_mods.sql
```diff
@@ -26336,6 +26336,7 @@ INSERT INTO `item_mods` VALUES (21507,8,15);      -- jolt_counter_+1 STR+15
 INSERT INTO `item_mods` VALUES (21507,10,15);     -- VIT+15
 INSERT INTO `item_mods` VALUES (21507,25,30);     -- Accuracy+30
 INSERT INTO `item_mods` VALUES (21507,291,18);    -- Counter+18
+INSERT INTO `item_mods` VALUES (21509,811,25);    -- Final Paradise WS
```
zynjec: Don't think this is right, modifier 811 doesn't exist?

Think you want modID 355 (ADDS_WEAPONSKILL) with a value of the weaponskill it adds (in this case, 228)

helixhamin: Oh, is that how this works? I was honestly confused on that.
I will fix that, once I have the WS number figured out xD

sql/weapon_skills.sql
helixhamin: I thought it would just be increment. The highest number is 240, but yes, there is no WSes between 226-238. Sorry to ask a stupid question, but how did you figure out the ID number?

helixhamin: nvm, I'm an idiot.
Check windower. Find weaponskill IDs:
`    [228] = {id=228,en="Final Paradise",ja="ファイナルパラダイス",element=6,icon_id=590,prefix="/weaponskill",range=2,targets=32},
`

Yeah, lol, at least now I know xD

helixhamin:
@zynjec Thank you for the feedback.
Those items you highlighted should be fixed now.

sql/item_mods.sql
helixhamin: All fixed now

scripts/globals/weaponskillids.lua
helixhamin: All fixed now

sql/weapon_skills.sql
helixhamin: All fixed now.

helixhamin:
So, any love for this pull from someone with approve/merge authority?

helixhamin:
I need to fix some flags. I will fix tomorrow. Please don't merge until then (4 flags should be 4100, not 4098).

helixhamin:
Okay, manually fixing the conflict items didn't work...
No clue anymore...
Unless someone has a better idea, I will just close this, make a new branch, then do this again. I can just copy everything based on changes I made here.

helixhamin:
Seriously, didn't I resolve?

helixhamin:
I give up...
I will just resubmit in a new PR...

ibm2431:
Manually changing files won't do anything - git doesn't care about the actual contents of a "conflict"; git is concerned about your local branch not having _some_ commit in master that also touched the same lines that your current commits do.

You need to do a rebase off of Darkstar master, at which point your git software will yell about a merge conflict. I personally use Visual Studio to resolve these. You need to compare the changes made by that "some" commit in master, and the commit that you have, and either:
A) Pick which one the final file output should use, or
B) Manually create a meld between the two.

In Visual Studio, it'll show your source and your target side by side with checkboxes next to changed lines that you can click to choose which change from which commit to use in the final product. When you click on one of them, it'll add the changes in a third window below your split source/target - this window shows what your final output will be.

If you need to make manual changes because line changes in the source and target contradict, you can manually edit the final output in the third lower window. When you've fully melded the changes made by the two commits together into the file, you just save the file - this is going to get staged and become your "merge" commit.

This new commit will be the offspring of both that conflicting commit in master, and the commit with the changes you're trying to make. Git now considers the changes in DSP master "accounted for" in the history of your local branch, and when pushed, should no longer present a merge conflict going into master.

PR #5894, opened 2019-04-25
[WIP] Add item scripts for anima

isotor:
Current issues:

1. __Terror__ causes the target to freeze. The correct behavior is the target runs around erratically for the duration and gains a flee (+100% movement) effect.
2. __Amnesia__ also works as a hack: currently it prevents the target from attempting TP moves. The correct behavior is the target is able to attempt TP moves but they fail and TP is lost *(same as attempting to use weaponskills when unable to see target)*.
3. There shouldn't be a message when the effect wears on the target *(confirmed from watching a few retail Promyvion videos on YT)*.
4. Unsure if the family system check is strictly necessary as all mobs in these zones are Empty anyway, but I suppose it's technically correct to include.
5. Using these on a passive target should result in aggro/claim.


Wiggo32:
https://youtu.be/EU7Rbubxoe4

isotor:
Thanks Wiggo 😄 from your link I can determine that there is a flee effect applied to terror, the duration is better input as `math.random(20, 30)`, and the correct animation is 55. These are now all corrected locally, but won't push until I get some additional feedback. 

Afraid I'm pretty new here so not yet figured out how to apply a claim/aggro on usage. Tinkering now.

scripts/globals/items/bottle_of_hysteroanima.lua
```diff
@@ -0,0 +1,32 @@
...
+require("scripts/globals/zone")
+-----------------------------------------
+
+function onItemCheck(target)
+    local result = dsp.msg.basic.ITEM_UNABLE_TO_USE -- default result is fail
+    local family = target:getSystem()
+    local zone = target:getZoneID()
+    local validZoneList =
+            {
+                16,17,18,19,20,21,22,23 -- only usable in Promyvion areas
```
TeoTwawki: does it work on just any mob within these zones? honestly I never tried it against non bosses

TeoTwawki: also your indent is wrong

KnowOne134: Thats what i was thinking. Be funny watching the memory recepticle dance around

isotor: Wiki says on any Empty enemies in Promyvion zones which is exactly what the check does, and Wiggo's link shows it working on a Wanderer. Unsure of behavior against MR, but in that case the fix might be changing its systemid to something so it's not categorized as an Empty.

Hozu:
re: the anima that causes their TP moves to do nothing, could you instead have it set a timer on the mob, and modify all of the empty mobskills to simply do nothing until the timer elapses?

Wiggo32:
> 
> 
> re: the anima that causes their TP moves to do nothing, could you instead have it set a timer on the mob, and modify all of the empty mobskills to simply do nothing until the timer elapses?

pretty sure that's called "amnesia" for a set amount of time and he already has that set correctly except for not removing the chat log message with `addStatusEffectEx`

TeoTwawki:
@Wiggo32 no he's saying let the skills "go off" but their script does noting so the TP is lost. but I'd prefer a solution that doesn't have us editing all of the skill scripts for a conditional "just in case that anima is used"

isotor:
So that's what `addStatusEffectEx` means! I was wondering, easy enough to fix.

ibm2431:
To be clear, the target freezing in place is what "Terror" _should_ be (`dsp.status.TERROR`) and shouldn't be changed. I'd call the behavior for using Terroanima on Empty something like "Panic" instead, to clearly differentiate between Terror ("frozen in") and Panic ("running around in").

For Psychoanima, `dsp.effect.DOUBT` is for Bully and is [what allows Sneak Attacks from any direction](https://github.com/DarkstarProject/darkstar/blob/e4c9bc9fd9a17e05f85bb2eb3f43601204b543e7/scripts/globals/weaponskills.lua#L71). You probably want `dsp.effect.INTIMIDATE` instead. Of course, we're [apparently not checking for any base intimidation effect in isIntimidated](https://github.com/DarkstarProject/darkstar/blob/311b593d1f292784dd65d82eb93474e5aa4a6fc2/src/map/utils/battleutils.cpp#L2692), so that's probably something that should be added (then you could add `dsp.effect.INTIMIDATE` with a power of 100 and be done with the item).

scripts/globals/items/bottle_of_hysteroanima.lua
```diff
@@ -0,0 +1,32 @@
...
+require("scripts/globals/status")
+require("scripts/globals/zone")
+-----------------------------------------
+
+function onItemCheck(target)
+    local result = dsp.msg.basic.ITEM_UNABLE_TO_USE -- default result is fail
+    local family = target:getSystem()
+    local zone = target:getZoneID()
+    local validZoneList =
+        {
```
TeoTwawki: indent still wrong

Wiggo32:
> For Psychoanima, `dsp.effect.DOUBT` is for Bully and is [what allows Sneak Attacks from any direction](https://github.com/DarkstarProject/darkstar/blob/e4c9bc9fd9a17e05f85bb2eb3f43601204b543e7/scripts/globals/weaponskills.lua#L71). You probably want `dsp.effect.INTIMIDATE` instead. Of course, we're [apparently not checking for any base intimidation effect in isIntimidated](https://github.com/DarkstarProject/darkstar/blob/311b593d1f292784dd65d82eb93474e5aa4a6fc2/src/map/utils/battleutils.cpp#L2692), so that's probably something that should be added (then you could add `dsp.effect.INTIMIDATE` with a power of 100 and be done with the item).

if you use EFFECT_DOUBT, value of 100 would not be correct:
https://github.com/DarkstarProject/darkstar/blob/311b593d1f292784dd65d82eb93474e5aa4a6fc2/src/map/utils/battleutils.cpp#L2725
if you use EFFECT_INTIMIDATE, there's nothing...

Wiggo32:
> Edit 2: Other than the obvious mobs of these Empty and Qiqirn bombers, I distinctly remember herding Bluffalo during some event, so there are multiple mobs who can be considered cowardly. I don't know if it might also be useful for Trusts who constantly try to move out of range of mobs.

Qiqirn and Buffalo do not run in an erratic pattern when pathing away from a player. And trusts are even more deliberate with their pathing. TERROANIMA is the only thing in those examples that produces a random and erratic pathing for 30 sec.

Wiggo32:
> @Wiggo32 no he's saying let the skills "go off" but their script does noting so the TP is lost. but I'd prefer a solution that doesn't have us editing all of the skill scripts for a conditional "just in case that anima is used"

and i have no response to this because i've been toooo busy/lazy to go on retail and see if they actually 'try' to do mobskills which would make them lose their tp or just recieve an 'amnesia' type effect allowing them to keep their tp.

ibm2431:
> if you use EFFECT_DOUBT, value of 100 would not be correct:

[`dsprand::GetRandomNumber(integer)`, is a half-open interval. It doesn't generate the supplied max number.](https://github.com/DarkstarProject/darkstar/blob/master/src/common/dsprand.h#L34).

> if you use EFFECT_INTIMIDATE, there's nothing...

Hence why I said `we're apparently not checking for any base intimidation effect in isIntimidated, so that's probably something that should be added`. :wink:

Wiggo32:
> `dsprand::GetRandomNumber(integer)` is a half-open interval. It doesn't generate the supplied max number.

dsprand generates 100, the problem is the '<' (less than) sign not including the assigned value leaving a 1% chance of not being intimidated after putting in a value of 100.  should be `<=`

ibm2431:
`dsprand` generates 100 only when given a max higher than 100. This is [sometimes done automatically by functions](https://github.com/DarkstarProject/darkstar/blob/master/src/map/entities/mobentity.cpp#L523) whose input expect the result to be a closed interval, or _always_ in the case of our [overriding of lua's random](https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/luautils.cpp#L311). 

A call inside core to `dsprand::GetRandomNumber(100)` [produces numbers in the interval \[0, 100)](https://github.com/DarkstarProject/darkstar/blob/master/src/common/dsprand.h#L50), or, all integers from 0 to 99, including both, but not 100.

Wiggo32:
ok, well 2% chance of not being intimidated... even worse...

scripts/globals/items/bottle_of_hysteroanima.lua
godmode: You could also probably just use the family system Empty for it and not have to iterate.

zynjec:
Feel free to PR again once you've completed the changes posted here.

PR #5896, opened 2019-04-26
[Quest_Rewrite] Hopefully bring quest table system to near-completion

ibm2431:
**I _intend_ for this to be my last PR to this branch, in hopes that my next PR will be merging quest_rewrite into master.** (After which this branch can be closed.) If you think there's a course correction to be made, or the brakes need to be tapped, now is the time to say so! (Now is also the time to give thumbs up in approval too!)

At the moment, this new system is serviceable enough to start actively using and benefiting from. I foresee no major changes to the _underlying system_, and any deficiencies or new features can be added later as needed (ex: BCNM-related functions and hooks).

**"Why master after this PR? Quests aren't actually rewritten yet!"**
Trying to rewrite all our quests - and keep quests rewritten - while people are actively adding to and modifying master would just be a conflict nightmare, slow down progress, and make would-be contributors wary/unsure about adding new stuff (in the current style) due to the "threat" of some imposing new system that'll arrive in an unspecified future. The sooner the new system is in place and can be used from master, the quicker people can start using it when implementing new quest stuff! (Also the sooner I can begin making those long references I introduced in #5758 start vanishing so people don't complain about them.)

To assuage fears: **this new system does not break any quests currently in master. It has been explicitly designed such that quests can be rewritten in a new style, and exist alongside quests still in the older style.** (And it also shouldn't break any custom content, since you have to proactively set out to use these new functions for them to be applicable.) I'm not like, deleting `player:completeQuest()` or anything.

**"Give me the tl;dr on changes to quest_rewrite in this PR."**
One thing led to another and I ended up going full object on quests. Quests have methods now. I included a bunch of wrappers for quest-related functions so people don't have to worry about what function is in what "namespace". Other changes resulted in reducing the number of typed characters required to do stuff. It's real neato.

**"You [said something before about player var SQL changes](https://github.com/DarkstarProject/darkstar/pull/5824)?"**
The DB doesn't understand or care about what we tag values as. It doesn't care if a SQL player var is called "WaywardWaypoints", "WW_progress" or "waypoints_calibrated". But once _one_ person decides to refer to something by a certain player var name, we're stuck with it, as we can't go willy-nilly modifying player var names without potentially screwing with player progress on live servers.

So with the new system, player vars for quests are intended to be _unique_ and mostly _anonymous_. In that the scripter can use _whatever_ name they want for referencing a variable _in the script_, but the new quest var wrappers will handle storing/fetching the value _in the DB_. Stored player vars are guaranteed to be unique to the quest (via the quest's prefix, example: "[Q][2][26]"), and capable of being "renamed" later (a key such as "[1]" to be added after the quest prefix). This results in DB player vars of "[Q][2][26][1]", "[Q][2][26][2]", etc.

This is done through the `db_name` field for a quest var, defined at the top of the quest file. So long as these db_names follow a mandated style for their creation (`thisQuest.varPrefix .. "[1]"`), we can freely rename our references to that variable inside a script later, as no matter what we call it _in code_, it'll still resolve to the _same DB var_.

This also gives us the added benefits of letting us be as verbose as we want in what references we use for variables (as player var names in the DB _are_ character limited, while they are not in scripts), but more importantly, _eliminates any concern over collisions_. One quest can refer to a variable about waypoints as, well, "waypoints", and another quest also dealing with waypoints can _also_ call a variable "waypoints" inside its script. The scripter doesn't need to fret over prepending an acronym for the quest, as the new quest system abstracts all that away for you.

**"But our current player vars _don't_ function that way. Do you have a migration plan?"**
By using the quest var's db_name field, of course!

It'd suck to have to run SQL queries to migrate player vars every single PR that rewrites a quest. So while quests are slowly being rewritten and converted to the new system, simply have the db_name for the appropriate var be what is _currently_ the SQL var name (`"WW_waypoints_calibrated"`). Then, once all existing quests are rewritten, compile a list of mappings of old player var -> new player var, put them in a migration script, and _then_ push the single commit that changes all the db_names in quest files to reflect more anonymous names (`"[Q][9][25][1]"`).

Server operators only need to run the migration once, and then not have to worry about people wanting to change (quest-related) SQL player var names again, as they're now mapped to generic vars that can be "renamed" (inside the script) later.

**"Hold up, I don't know what [Q][9][25][1] refers to. What if I want to look up a player's quest vars to fix something?"**
Hey, _I_ don't know what it refers to either, but I can run `!checkquest AREA QUEST_NAME` and it'll tell me what all the player's vars associated with that quest are, with the variable's "human friendly" scripter-defined names! Then if I need to change it for a player, I can do a `!setquestvar AREA QUEST_NAME FRIENDLY_NAME VALUE` and the command will make sure the right SQL entry gets updated.

So now in this scenario, you no longer have to find the right NPC file and then search it for the correct `player:setVar()` definition. New GM commands will tell you upfront with no files opened!

(Also, if you ever _actually_ need to do this, someone really screwed up, as with this new system, quests should be harder to break.)

**"Okay, but how do I write quests with this new system?"**
I will have documentation and How-Tos ready before attempting a merge of quest_rewrite into master. It'll take me a bit, because as you all have no doubt seen, I can tend to be... verbose.

But for now, the eager can check out `quests/adoulin/the_old_man_and_the_harpoon` and `zones/western_adoulin/npcs/jorin`. The Old Man and the Harpoon is probably one of the most basic of quests, and can make for a good study.

**"Why do some of your quest methods just call methods we already have?"**
Consistency! All quest-related actions a scripter would want to perform are in the form of `thisQuest.doFoo`. The scripter doesn't need to remind themselves that `npcUtil.giveKeyItem` is the one that includes the message, while `player:giveKeyItem` doesn't, or that calls to lua-exposed core functions use `:` while our lua helper library functions use `.` The first argument is also always `player` for more consistency. The coding instinct becomes "I want to do bar, I will type `thisQuest.bar(player,`"

And by always going through a `thisQuest` wrapper, we can:
1) Enforce certain conventions, like not allowing a scripter to delete a key item from the player if it's not in the quest's temporary key items table (which avoids mistakes, _and_ makes people clearly define what they intend to delete later in the file)
2) Help the scripter out a little with returns, to shorten the number of lines required. Since the new quest system requires a return of `true` to indicate to the calling NPC that control should be diverted away from the rest of its script, we can write `return thisQuest.startEvent(player, csid)` which gives us a free `true`, instead of having to:
`player:startEvent(csid)`
`return true`
The fewer lines we have, the closer we are to Altana!
3) If we ever need/want to change an underlying function we're using later (like changes to how trades work), we just need to update the code in the method for the base quest object, and not have to worry about mass-find-and-replaces later!

**"Who is getting tagged?"**
- @takhlaq so he can say if this branch is everything he envisioned
- @TeoTwawki because I'd _like_ for this to meet his definition of easy-enough-to-use-while-adding-needed-functionality
- @teschnei because while I don't need to since he already got a notification, I didn't want him to feel left out, and I'm totally not doing a little brown-nosing because he's the most likely candidate to hit the merge button
- @zynjec to hedge my bets
- @whasf because he was the one to hit the merge button on my last PR, so apparently he's a merge-candidate as well, and I gotta cover as many bases as I can
- @wrenffxi because I care about his input in particular (even if just a silent emoji), because he's one of the more likely candidates to immediately start using this system other than me
- @KnowOne134 because I know he's already looked this system over and may/may not have an opinion
- @Bizzle-Dapp , @helixhamin , and @isotor because they're easy "thumbs up emoji" candidates, and seeing the emoji makes me puff out my chest in pride

### DIFF
--- a/scripts/globals/quests.lua
+++ b/scripts/globals/quests.lua
@@ -1355,7 +1355,7 @@ quests.newQuest = function()
         if entity:isPC() then
             entity:PrintToPlayer(message)
         end
-        print("[Quest Error] "..entity:getName()..": "..message)
+        print("[Quest Error]".. this.string_key .." "..entity:getName()..": "..message)
     end
 
     ---------------------------------------------------------------
@@ -1418,8 +1418,16 @@ quests.newQuest = function()
             if this.requirements.quests then
                 for i, requiredQuest in ipairs(this.requirements.quests) do
                     local requiredQuestStatus = player:getQuestStatus(requiredQuest.log_id, requiredQuest.quest_id)
-                    if requiredQuestStatus ~= dsp.quest.status.COMPLETED then
-                    -- or (required_quest.stage and (dsp.quests.getStage(player, required_quest) < required_quest.stage)) then -- getStage requires a full quest table
+                    if requiredQuest.stage then
+                        local quest = quests.getQuest(requiredQuest.log_id, requiredQuest.quest_id)
+                        if quest then
+                            if quest.getStage(player) < requiredQuest.stage then
+                                return false
+                            end
+                        else
+                            return false
+                        end
+                    elseif requiredQuestStatus ~= dsp.quest.status.COMPLETED then
                         return false
                     end
                 end
@@ -1531,6 +1539,25 @@ quests.newQuest = function()
         return true
     end
 
+    -- Give KI to player, while going through this quest's wrapper
+    ---------------------------------------------------------------
+    this.giveKeyItem = function(player, key_item)
+        return npcUtil.giveKeyItem(player, key_item)
+    end
+
+    -- Give KI to player, while going through this quest's wrapper
+    ---------------------------------------------------------------
+    this.delKeyItem = function(player, key_item)
+        for _, ki in pairs(this.temporary.key_items) do
+            if ki == key_item then
+                player:delKeyItem(key_item)
+                return true
+            end
+        end
+        error(player, "Key Item " .. key_item .." not listed as temporary in quest file! No action taken.")
+        return false
+    end
+
     -- Completes the quest in the player's log, gives out rewards, and cleans up leftover vars and key items
     ---------------------------------------------------------------
     this.complete = function(player, reward_set)
@@ -1547,7 +1574,7 @@ quests.newQuest = function()
             end
             if reward_set then
                 -- todo: check inventory (including stack space), award items, return false if cant complete
-                rewards_given = npcUtil.completeQuest(player, this.area, this.quest_id, reward_set)
+                rewards_given = npcUtil.completeQuest(player, this.log_id, this.quest_id, reward_set)
                 if not rewards_given then
                     error(player, message.. "Unable to give quest rewards.")
                 end
@@ -1559,7 +1586,7 @@ quests.newQuest = function()
         end
 
         if rewards_given then
-            this.setVar(player, this.vars.stage, 0)
+            this.setVar(player, this.vars.stage, 0) -- Stage should ALWAYS be erased
             for name, var in pairs(this.vars.additional) do
                 if not var.preserve then
                     this.setVar(player, name, 0, message)
--- a/scripts/globals/quests.lua
+++ b/scripts/globals/quests.lua
@@ -4,9 +4,12 @@ require("scripts/globals/zone")
 
 dsp = dsp or {}
 dsp.quest = dsp.quest or {} -- "Solid" enum definitions which are currently in use by master
+quests = {}
+dsp.quests = {}
 
--- These should be tabled enums at some point in the future,
--- but that'll require mass-replacing in master first
+-- These should be tabled enums for rewritten quests, but
+-- these globals are kept so old-style quests will work
+-- until they're all converted (and then these can be removed)
 QUEST_AVAILABLE = 0;
 QUEST_ACCEPTED  = 1;
 QUEST_COMPLETED = 2;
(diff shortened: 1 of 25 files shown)


helixhamin:
lol, is the quest vars related to this? (I just finished updating this page)
https://wiki.dspt.info/index.php/Log_IDs

For example, DESTINYS_DEVICE
So,
[Q] = quest flag (If yes, will we do one for missions too?)
[9] = SoA quests (Adoulin)
[25] = DESTINYS_DEVICE (Found in the quests.lua under area 9, SoA)
[1] = the var stage (So, first step?)

Just curious, because that could be very nice to help follow the actual flow of the quests.lua, rather than just pick an arbitrary var name.

ibm2431:
Yes for everything except that last tag! (I definitely want to rewrite missions as well!)

The 'stage' for a quest (ie: an unchanging order) should be under:
`thisQuest.vars.stage`, and stored in the DB as just the varPrefix (so `[Q}[9][25]`).
That will cover a lot of quests.

The additional [1] tag is for an "additional" var, that would be under:
`thisQuest.vars.additional.some_var_name_here`, and stored as `[Q][9][25][1]`

There are a couple quests that need to track things that are more than just "step 1" to "step 2". I'll use [Wayward Waypoints](https://github.com/DarkstarProject/darkstar/blob/0b24461bf639e7475f8079887ca852c8b99e5f4e/scripts/quests/adoulin/wayward_waypoints.lua) as an example. For the first steps of the quest you just talk to people (so stage, stored in DB as just "[Q][9][29]"). However, later on you need to talk to seven separate waypoints (key note: in any order), and can't advance the quest until you've done so.

So you'll still be on [stage 1](https://github.com/DarkstarProject/darkstar/blob/0b24461bf639e7475f8079887ca852c8b99e5f4e/scripts/quests/adoulin/wayward_waypoints.lua#L88), but the waypoints you've calibrated will be stored in [an "additional var"](https://github.com/DarkstarProject/darkstar/blob/0b24461bf639e7475f8079887ca852c8b99e5f4e/scripts/quests/adoulin/wayward_waypoints.lua#L16). Inside the script file you can call it whatever you want ("waypoints_calibrated", or just "waypoints") via calls like `thisQuest.setVar(player, 'waypoints', 5)`, but when the quest method goes to store that value in the DB, it'll resolve to "[Q][9][29][1]" (or, "our first additional var").

If the first scripter decides to call it "waypoints_calibrated", it can be changed to "waypoints" by someone on a var-shortening-spree (or vice-versa for someone on a "make var names more clear" spree), without any effect on it always being "[Q][9][29][1]" in the DB.

A call to `!checkquest ADOULIN WAYWARD_WAYPOINTS` would produce:
```
PLAYER's status for 'Wayward Waypoints' is: ACCEPTED
PLAYER is on stage: 1
PLAYER's additional quest vars are - waypoints: 5
```
(Pretend that five represents a bitmask indicating that you calibrated the first and third waypoints.)

When you're done calibrating all the waypoints and advance to the next stage of the quest, the scripter _should_ have deleted the additional var from the player (so it's no longer stored in the DB).

**But if they messed up and didn't delete it, all calls to `thisQuest.complete(player)` will nuke both stage and additional vars**, so there's a guarantee there won't be phantom vars in the SQL after a player has completed a quest.

Once a player triggers whatever tells the script that they're done with the calibration step, our main 'stage' var will be incremented as normal via `thisQuest.advanceStage(player)`, which tracks/stores via our usual `[Q][9][25]` DB var (but is still displayed as "stage"). There need be no further action regarding the additional 'waypoints' var, as its no longer needed, and never interfered with your main 'stage' var.

(I suppose we could store stages as `[Q}[9][25][0]` for added clarity.)

helixhamin:
Love it! But, lol, you are in for a lot of work redoing the quests and finding a way to update already existing vars in databases.
Let me know if I can help any. Not sure I can do much, but who knows.

Bizzle-Dapp:
> and make would-be contributors wary/unsure about adding new stuff (in the current style) due to the "threat" of some imposing new system that'll arrive in an unspecified future.

I was told to wait - and thus wait i am doing 🗡 

Looking forward to your verbose How_To!

ibm2431:
> I was told to wait - and thus wait i am doing :dagger:

You were definitely the person I had in mind when I wrote that sentence,

KnowOne134:
Approved

TeoTwawki:
> Hold up, I don't know what [Q][9][25][1] refers to.

quest log 9 id 25 step 1 ?

###### I am strange

TeoTwawki:
```
I intend for this to be my last PR to this branch, in hopes that my next PR will be merging quest_rewrite into master. (After which this branch can be closed.) If you think there's a course correction to be made, or the brakes need to be tapped, now is the time to say so! (Now is also the time to give thumbs up in approval too!)
```
can we include one example quest before it goes to master? Just one, so the scripters have a guide.

KnowOne134:
There are 2 examples done in the adoulin areas with the re write. Havent
built myself to test but looks like it to me. Scripts globals quest
[questname]

TeoTwawki:
I saw them, Please see my edit: I'm looking for something ppl were already familiar with. most servers don't even have ToAU enabled.

TeoTwawki:
P.s. they don't need to be in THIS pull request, I'm only talking about when it all arrives at master.

KnowOne134:
The good news is this doesnt break any current quest. But can be a gradual
tradition to new style code. Headache for new server owners will be the
remembering of changing player vars as it continues to progress

TeoTwawki:
I know, the original PR comment explained that

megberry:
I've just started a new job, and haven't been active on DSP, so I've only scratched the surface of this quest rewrite, but I love what I've seen so far.  Huge props for the effort you've put into this.  When this new quest system's ready to go, and I'm settled enough to start contributing again, I'll absolutely help convert existing quests over.

wrenffxi:
I've just started a new job, and haven't been active on DSP, so I've only scratched the surface of this quest rewrite, but I love what I've seen so far. Huge props for the effort you've put into this. When this new quest system's ready to go, and I'm settled enough to start contributing again, I'll absolutely help convert existing quests over.

isotor:
This is worth a 👍 for the new commands alone.

scripts/zones/Western_Adoulin/npcs/Clemmar.lua
```diff
@@ -1,24 +1,24 @@
...
 -- !pos -12 0 12 256
 -----------------------------------
 require("scripts/globals/missions");
 require("scripts/globals/quests");
 
-local quest_table =
+local quests =
 {
-    require("scripts/quests/adoulin/a_certain_substitute_patrolman")
+    {dsp.quest.log_id.ADOULIN, dsp.quest.id.adoulin.A_CERTAIN_SUBSTITUTE_PATROLMAN}
```
teschnei: i think i prefer the `require` method here - it lets you know exactly what the filename containing the quest data is (ie. i can yank inside the quotes and then open the file directly with that), and it already handles the whole "don't load it if it's already loaded" scenario (assuming it assigns it to a global in the same way mixins do)

scripts/globals/quests.lua
```diff
@@ -1416,223 +1632,171 @@ dsp.quests.complete = function(player, quest, reward_set)
...
-        end
-        if cycle > 0 and not needsCycling then
-            player:setLocalVar("[quests]dialogCycle", 0)
-        end
-        local i = cycle + 1
-        while i <= #quest_table do
-            local quest = quest_table[i]
+    -- Helper function to check all of our involvedQuests for a given quest event
+    local check = function(player, params)
+        local i = 1 -- Doing i loop instead of pairs so quests checked in order
```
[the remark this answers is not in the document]
ibm2431: It is. It was late. I am ashamed.

ibm2431:
@teschnei I have a natural habit of using `['string']` when I want the element to stand out from others due to string syntax highlighting (as opposed to blending in with other variables). I'm not firmly attached to it (and have already changed them locally), but wanted to verify if you want _all_ such occurrences changed (specifically, references to NPC names in quest files).

scripts/zones/Western_Adoulin/npcs/Clemmar.lua
ibm2431: At the moment, `getQuest()` is what fetches a table from a quest file (which just returns its table without global assignment) - if it wasn't already in the global (`getQuest()` then stores the quest in the global if it had to fetch it from file). I'll make changes to have the quest files assign themselves to the global the same way mixins do.

After a quest file assigns itself to the global, should it return the table it just assigned (like mixins return their function), or just the Area/Quest pairs for the `involvedQuests` object to stash? Just the Area/Quest pair would be easiest (and at the point a quest is in the global, I don't know of additional value in returning the full table; all relevant functions just pull from the global based on Area/Quest index).

teschnei: to be honest, I think it should a) return the quest object, and b) the functions should use the quest object instead of the area/quest indices.  These tables are all passed by reference, so there shouldn't be any harm in passing the tables around.  I think it makes it easier to understand when the function is requesting an actual quest object instead of logid/questid

ibm2431:
Made requested changes to how quests are loaded and checked.

Also, at @TeoTwawki's request, rewrote The Brugaire Consortium. Which was actually more complicated than one might first assume due to the fact that you can drop quest items, pay to get them back, and the NPC will apparently hold parcels if you're full (with a custom event no less).

But it was an excuse to add in functions that will hold items for the player if the player is full. These work both during the quest, and for the final rewards (in a behavior [like Halver's San d'Orian flag](https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Chateau_dOraguille/npcs/Halver.lua#L171) - everything else will be rewarded, but the flag will be held until the next time the player talks to him with sufficient inventory).

(If anyone is interested, I wouldn't mind a capture of this quest, for verifying how the "hold parcel if player is full" behavior works. I don't fully "trust" the event -> event chain currently in the logic for Fontoumant, as the behavior seems a little weird.)

TeoTwawki:
"easy fix"

*runs away*

ibm2431:
I've been working on documentation, and as such have been mulling over (not-major) changes. I'm going to go ahead and say I plan another PR to quest_rewrite _after_ this one before attempting a merge on master:

- I'm not fully satisfied with how held items are handled (note to future self: "`stage.COMPLETE`", also be careful not to dupe). Don't worry Teo! The stage var will still be nuked from the DB upon quest completion, "`stage.COMPLETE`" will just be a value to grab a final "stage" with if quest status is `dsp.quest.status.COMPLETED`.
- I want to rename quest var types from `dsp.quest.var.LOCAL` and `dsp.quest.var.CHAR` to something like `dsp.quest.var.TEMP` and `dsp.quest.var.PERM`, to better convey the nature of these variables, and calling something a local var is just begging to have it confused with a local var.
- I saw [the monstrosity of Regine](https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Port_San_dOria/npcs/Regine.lua#L63), and now _really_ want some built-in bitwise var functions _before_ "launch"
- I've realized that migration from old vars to new vars won't be as clean as I would have liked, because previous scripters did some weird things with vars that will not be replicated when a quest is rewritten (ex: above point). So we can't just switch from using the old var names until the last moment to using new names.
Debating between pushing quest rewrites in large "chunks" instead of "whenever rewritten" (not preferable), or setting up some "rolling" rewrite/migration system that uses a server var (in the form of say, a date), which gets checked (maybe by `quests.lua`) and alerts the server operator that quest vars have been shifted since they last ran the migration script (with what's being compared against set by the PR that rewrote a quest), so they may want to run it again (which will update the server var). Of course, on final rewrite PR, nuke the server var and kill the check logic.
- Once again stuck in the eternal "should it be `dsp.quest.log_id`, `dsp.quest.quest_log`, or just `dsp.quest.log`?" debate. At the moment I'm heavily leaning towards the third option and wondering why I didn't pick it in the first place.

However, to avoid PR creep, **I'm going to go ahead and leave this PR alone unless more changes are requested**.

helixhamin:
While I have no authority for anything here, I approve anyways. At this time, I don't see anything broken, and hopefully it will make it better going forward 😄

scripts/zones/Port_San_dOria/npcs/Fontoumant.lua
```diff
@@ -1,117 +1,53 @@
...
 require("scripts/globals/settings");
 require("scripts/globals/keyitems");
-require("scripts/globals/titles");
 require("scripts/globals/quests");
+
+local quests =
+{
+    require("scripts/quests/sandoria/the_brugaire_consortium")
+}
+quests = dsp.quest.involvedQuests(quests)
```
teschnei: instead of assigning quests and then reassigning it immediately, do you think it looks nicer or not to do 
```
local quests = dsp.quest.involvedQuests(
{
    require("scripts/quests/sandoria/the_brugaire_consortium")
})
```
?

teschnei: it probably looks more confusing at first but this is gonna be copy/pasted in a lot of places

ibm2431: Hey, if project style were up to me, that opening table bracket would be on the same line as the opening paren~! :wink:

```
local quests = dsp.quest.involvedQuests({
    require("scripts/quests/sandoria/the_brugaire_consortium")
})
```

For your example, I guess there is a small concern that newer contributors might have a little difficulty making sure all parens/brackets for the construction call are there. But that's really of minor consideration in regards to template style. If it's just going to be copy-pasted, it probably won't pose a problem.

I have no preference either way~! Up to you!

TeoTwawki: *eyes you like a monster*

teschnei: yeah, I like your style better, lets do that

then I think we're done with this and I'll merge it

ibm2431:
Regarding The Brugaire Consortium, I just did a capture on retail, and as suspected, the quest as implemented is not correct (and the way the quest used to be implemented gave me wrong impressions).

Summary:
- He doesn't hold any items, you must replay the entire event again (even after paying gil)
- The "custom you're too full" event he is associated with is not used _at all_; one event does not chain into another, and he doesn't check your inventory space at any time to say it (either accepting quest, getting parcel, or completing quest)
- Paying gil outright skips the stage; you're not given another parcel
- Receiving NPCs only have the "I didn't ask for this" on the stage you're supposed to deliver to them on (they respond to all items during this step, not just parcels)
- You can not give the NPC the correct parcel after skipping the stage by paying gil; they'll just ignore you like usual
- You can pay more 100 gil to skip, but only 100 will be taken
- I unfortunately spaced trying to give him _less_ than 100, but I assume he'll just ignore the trade

I'll post the capture to Captures channel, and fix the quest script, at a later date.

PR #5899, opened 2019-04-28
Add Shield Bash MOD to Fenian Ring

neuromancerxi:
Augments Shield Bash is Damage +50 (https://www.bg-wiki.com/bg/Fenian%20Ring).


TeoTwawki:
```sql
Importing item_mods.sql into the database...ERROR 1064 (42000) at line 18584: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '--Fenian Ring Shield Bash +50
INSERT INTO `item_mods` VALUES (15832,85,3)' at line 1
```
you need a space after the 2 dashes

Issue #5881, opened 2019-04-22
Feature request: Mog Safe expansion mini quest line


isotor:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30190328_2

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Just a request for the following mini-series if anyone is looking to implement some quests:

1. Give a Moogle a Break
2. The Moogle's Picnic!
3. Moogles in the Wild 

Link: https://ffxiclopedia.fandom.com/wiki/Mog_Safe

PR #5903, opened 2019-04-30
Mog Safe Quests

zach2good:
Original issue: https://github.com/DarkstarProject/darkstar/issues/5881

Started looking at these quests in the last few days. This is my first attempt at quest scripting so all feedback is appreciated!


src/map/packet_system.cpp
```diff
@@ -5349,6 +5349,8 @@ void SmallPacket0x0FA(map_session_data_t* session, CCharEntity* PChar, CBasicPac
             PChar->getStorage(LOC_STORAGE)->AddBuff(PItem->getStorage());
 
             PChar->pushPacket(new CInventorySizePacket(PChar));
```
Wiggo32: if you're going to go the route of char_vars, you can eliminate two steps and unnecessary bindings by just having the core do that work here:
```
            const char* itemQuery = "INSERT INTO char_vars(charid, varname, value) VALUES (%u,'[MH]%s',%u);";
            Sql_Query(SqlHandle, itemQuery, PChar->id, PItem->getName(), CVanaTime::getInstance()->getVanaTime() + 1009810800);
```

src/map/packet_system.cpp
```diff
@@ -5423,6 +5425,8 @@ void SmallPacket0x0FB(map_session_data_t* session, CCharEntity* PChar, CBasicPac
                 PChar->getStorage(LOC_STORAGE)->AddBuff(-(int8)PItem->getStorage());
 
                 PChar->pushPacket(new CInventorySizePacket(PChar));
```
Wiggo32: of course same thing here but opposite:
```
                const char* itemQuery = "DELETE FROM char_vars WHERE charid = %u AND varname = '[MH]%s' LIMIT 1";
                Sql_Query(SqlHandle, itemQuery, PChar->id, PItem->getName());
```

src/map/packet_system.cpp
TeoTwawki: We may have other future things we want to trigger via furniture changes without recompiling. This was discussed on discord already.

TeoTwawki:
@teschnei I was looking for where we store the layout after placing items in the house, I was thinking it was a table and became confused when I didn't see it. if it were table I'd say store the timestamp there instead of char var. coz char_vars is kinda bloaty as it is. I just saw where that is done and frew up in mouth just a lil... looks like it _has_ to be a char_var unless we refactor how furniture is even done. But I am barely awake atm. Pls tell me I am wrong.

scripts/globals/moghouse.lua
```diff
@@ -42,8 +48,21 @@ function moogleTrigger(player,npc)
...
         end
 
+        local homeNationFame = player:getFameLevel(player:getNation())
+        local GiveMoogleABreak = player:getQuestStatus(OTHER_AREAS_LOG,GIVE_A_MOOGLE_A_BREAK)
+
         if player:getVar("MoghouseExplication") == 1 then
             player:startEvent(30000)
+        -- Start GiveMoogleABreak
+        elseif player:getLocalVar("QuestSeen") == 0 and GiveMoogleABreak == QUEST_AVAILABLE and homeNationFame >= 3 and
+               player:getVar("MogSafe1Time") + 604800 <= os.time() then -- 604800 One week in Unix epoch
```
ibm2431: The wait is until the next conquest tally, not just a full week. I _want_ to say there's a function somewhere that returns the next time until tally, but I'm drawing a blank on its name / where it is. How you'd frame the `safeTime` var (and when) would depend on what that function returns (time remaining until tally, or the absolute next time a tally should occur).

For example, if it returns absolute time, you'd want to set `safeTime` to that absolute time on the furniture placed event, and then here check if the current time is more than the time you stored before.

src/map/packet_system.cpp
TeoTwawki: @ibm2431 looks like github also puts a priority on "review" comments so they appear first instead of in chronological order.

ibm2431: @TeoTwawki That was an impressive bit of magic, wasn't it. I think that may be because I had it queued in the "review" and was still looking through stuff when you chimed in, and Github wanted to keep it in that order to show at the time I added the comment, I "hadn't seen" yours yet (even though I did in another tab).

scripts/globals/moghouse.lua
```diff
@@ -27,8 +29,9 @@ function moogleTrade(player,npc,trade)
             end
         end
 
-        GiveMoogleABreak = player:getQuestStatus(OTHER_AREAS_LOG,GIVE_A_MOOGLE_A_BREAK)
-        if GiveMoogleABreak == QUEST_ACCEPTED and trade:hasItemQty(17161, 1) and trade:hasItemQty(13457, 1) then
+        local GiveMoogleABreak = player:getQuestStatus(OTHER_AREAS_LOG, dsp.quest.id.otherAreas.GIVE_A_MOOGLE_A_BREAK)
+        if GiveMoogleABreak == QUEST_ACCEPTED and npcUtil.tradeHas(trade, {17161, 13457}) then
+            player:confirmTrade()
```
ibm2431: It's usually better to have t`confirmTrade`s in the `eventFinish`, to avoid situations where the event gets canceled and we took the player's items without advancing their variables.

scripts/globals/moghouse.lua
Wiggo32: is this what you're looking for:
https://github.com/DarkstarProject/darkstar/blob/5b438424754a60893716759c458147a70a17709c/scripts/globals/common.lua#L80
and you're right... it would be hard to do conquest tally without the bindings...

Wiggo32:
> I was looking for where we store the layout after placing items in the house

the layout info is stored in the item's extra data. and that would be stored in `char_inventory`

TeoTwawki:
> @TeoTwawki
> 
> > I was looking for where we store the layout after placing items in the house
> 
> the layout info is stored in the item's extra data. and that would be stored in `char_inventory`
> its a blob
> though, i suspect you might get some opposition to that idea:
> char_inventory

That why I was nauseated. because its in inventory as exdata making char var a better solution. If someone tried to add a column to inventory I'd reject. Had furniture been in its own table I'd say we should store timestamps in it instead.

scripts/globals/moghouse.lua
ibm2431: Yep, looks like it, thanks Wiggo!

scripts/globals/moghouse.lua
```diff
@@ -42,8 +57,38 @@ function moogleTrigger(player,npc)
             end
         end
 
+        local homeNationFame = player:getFame(player:getNation())
```
ibm2431: Everything looks good to me save for this should be `getFameLevel`~!

The [comment for `player:getFame()` is a bit confusing in that it says "fame level"](https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_baseentity.cpp#L5546), but it's actually returning raw fame.

You'll [want `'player:getFameLevel()'`](https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_baseentity.cpp#L5751) because it [converts the player's raw fame into the 1-9 integers](https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_baseentity.cpp#L5773) used for fame checks.

PR #5916, opened 2019-05-02
Update overwrite params of Dia/Bio/Elegy

isotor:
~~1. Berserk and Defender should overwrite each other~~
2. Dia and Bio should land initial damage, but the status should not overwrite for an equal power effect *(e.g. Goblin Leechers spamming Dia shouldn't be able to refresh effect duration on the target)*
3. An equal tier elegy should overwrite/refresh the effect duration (discussed on BG [here](https://www.bluegartr.com/threads/54661-BRD-formulas?p=1908883&viewfull=1#post1908883))

- Also update status message check for Dia/Bio, and update powers for [April 2019 Version Update](http://forum.square-enix.com/ffxi/threads/55263-April.-3-2019-%28JST%29-Version-Update).


TeoTwawki:
> Berserk and Defender should overwrite each other

No they should not.

when bg wiki says:
`When Defender is activated at the same time as Berserk, the two effects cancel out.`
it does not mean they overwrite. They don't. They have equivalent stats (normally) so having both up is the same as having neither up, that is what they mean.

ffxiclopedia makes the distinction:
`When Defender is activated at the same time as Berserk, the two abilities effectively cancel each other out, however both effects will remain active unless canceled. Should one ability be canceled, the other will resume having an effect.`

TeoTwawki:
need to adjust scripts for the bio/dia thing

https://github.com/DarkstarProject/darkstar/blob/9ecf04adef295121fc9c9c17a2f7e5bde843d5d4/scripts/globals/spells/dia.lua#L56
```lua
    if target:addStatusEffect(dsp.effect.DIA, 1 + dotBonus, 3, duration, 0, 5, 1) then
        spell:setMsg(dsp.msg.basic.MAGIC_DMG)
    else
        spell:setMsg(dsp.msg.basic.MAGIC_NO_EFFECT)
    end
```
if the status doesn't apply the message doesn't get set (which is wrong, because we need to still do initial damage)

This went unnoticed prior to you correcting the overwrite of the status.

Additional concern: I know retail will let me gain skillups on same tier's initial damage. You can spam Dia to the enfeebling cap on retail without ever waiting for it to wear off as long as the target mob is high enough level.  I do not know if the way our core is structured expects overwrite-ability for that to happen.

TeoTwawki:
actually...in retail doesn't it still do the damage message even when a higher tier or bio is on?

assuming the skillup isn't tied to the status check, we should be able to just get rid of that conditional completely if it always displays the damage in the log

isotor:
> actually...in retail doesn't it still do the damage message even when a higher tier or bio is on?

Yep, the initial damage is independent of status effect checks. I'll update the script and append to this PR when I have a min, probably tomorrow 👍

isotor:
Regards updating the Atk/Def down percentages, [source here](http://forum.square-enix.com/ffxi/threads/55263-April.-3-2019-%28JST%29-Version-Update). 

- Diaga II and III not specified but took a liberty with common sense vis-a-vis Dia/ Diaga commonality. - Unsure how to proceed for Bio IV - a buff here makes sense on paper to me but up to you guys if you want it without official confirmation.

scripts/globals/spells/bio_ii.lua
```diff
@@ -1,6 +1,6 @@
 -----------------------------------------
 -- Spell: Bio II
--- Deals dark damage that weakens an enemy's attacks and gradually reduces its HP.
+-- Deals dark damage that weakens an enemy's attacks and gradually reduces it's HP.
```
helixhamin: its was correct. It's = it is. Reduces it is HP?
its = belong to it. Reduce the hp belonging to it? Sounds better.

helixhamin:
its is correct, it's (it is) is not.

scripts/globals/spells/bio_ii.lua
TeoTwawki: whatever the helptext on retail says /shrug

scripts/globals/spells/diaga_iii.lua
```diff
@@ -15,52 +15,47 @@ end
...
-    end
+    -- do it!
+    target:addStatusEffect(dsp.effect.DIA, 3 + dotBonus, 3, duration, 0, 20, 3)
+    spell:setMsg(dsp.msg.basic.MAGIC_DMG)
 
-    -- Try to kill same tier Bio
-    if BIO_OVERWRITE == 1 and bio ~= nil then
-        if bio:getPower() <= 3 then
+    -- try to kill same tier Bio (non-default behavior)
+    if (BIO_OVERWRITE == 1 and bio ~= nil) then
```
TeoTwawki: no parens

isotor:
> 
> 
> its is correct, it's (it is) is not.

[I hate to grammar Nazi you but](https://www.dummies.com/education/language-arts/grammar/using-apostrophes-to-show-possession/)

scripts/globals/spells/diaga_iii.lua
isotor: will fix along with whatever else comes, I just picked one for consistency as some had some didn't.

scripts/globals/spells/bio_ii.lua
TeoTwawki: Bio description text:
image (lowers an enemy is attacks..ok SE)

Dia description text:
image

scripts/globals/spells/diaga_iii.lua
```diff
@@ -27,33 +27,33 @@ function onSpellCast(caster, target, spell)
...
+    -- Check for Bio
     local bio = target:getStatusEffect(dsp.effect.BIO)
 
-    -- do it!
+    -- Do it!
     target:addStatusEffect(dsp.effect.DIA, 3 + dotBonus, 3, duration, 0, 20, 3)
     spell:setMsg(dsp.msg.basic.MAGIC_DMG)
 
-    -- try to kill same tier Bio (non-default behavior)
+    -- Try to kill same tier Bio (non-default behavior)
```
TeoTwawki:
```
Capitalize comments because Teo hates me
```
:rofl:

isotor:
You can revert that stupidity yourself :P

TeoTwawki:
I already did.

helixhamin:
:sigh
@isotor https://www.grammarly.com/blog/its-vs-its/

> It’s is a contraction of “it is” or “it has.” Its is a possessive determiner we use to say that something belongs to or refers to something.

Issue #2158, opened 2015-09-30
Mobs dropping too many items


Deadwing888:
A mob (even one with more than ten possible drops) should never drop more than ten items.

It is also generally accepted that Treasure Hunter (especially TH4) is overpowered on Darkstar. I don't know the specifics of how it's currently coded but I would say it's overpowered by at least a factor of two. I have a feeling I'm about to receive a "tag you're it" to do some treasure hunter parsing and I'm willing to do take that on if nobody has the goods on treasure hunter mechanics. 

It's also possible that drops seem more common because treasure hunter is getting baked in twice. For instance wiki reports the drop rate for Herald's Gaiters at 25%, but Tiamat would almost always be killed with TH4 on... I think in this fashion it's possible that treasure hunter is already baked into the wiki drop rates for some items and then gets applied again when there is a thief present.

Also Dynamis armor drops are more common than they should be, but it's hard to say if that's OP TH talking or drop rates. Relic currency in Dynamis Jeuno is similarly too common.

teschnei:
IIRC, every level of treasure hunter adds an extra chance for every drop,
which is definitely not how it works on retail. TH13 absolutely does not
give you 13 chances for an item. I think it's been thought to just give a
flat 1% bonus chance for th3+. You can just look at the ffxidb drop rates
for data

TeoTwawki:
@teschnei I want to add a column to the drop table: drop slot. most of the loot pool overflows come from not just abnormally high drop rates but the fact both items are dropping independently when they should not be. And with permission, I'll change TH to being +chance instead of re-rolls on our imaginary dice.

TeoTwawki:
@teschnei unless you want either-or type drops done a diff way. Then I am all ears. I had an alternative worked out in case you didn't want a new column for it, but I forgot wtf it was now.

teschnei:
nah thats probably fine

towbes:
Seems like adding the drop slot column is a solution on its own, why change
the re-rolls aspect as well?  The research done here (before all the TH5+
changes) http://khoisan.livejournal.com/15302.html indicates that the
re-roll aspect is indeed correct.  Perhaps research at current live levels
would shed some light on TH5+ being more chance% increase.

TeoTwawki:
[...]

PR #5918, opened 2019-05-02
Update drop tables for "loot-of-interest" in various 75-cap areas

isotor:
__Changes:__
- Fix a bunch of NM drop rates for implemented NMs in 75-cap era zones.
- Implemented a bunch of loot groupings.
- Adjust some known overly-generous farming spots *(bird blood, sahagin scrolls, davoi gold masks, antlion jaws)*.

All changes based on FFXIDB.com, not usually blindly following the TH0 rate as the sample is a small proportion of the total kills in most cases. Looking at [other drops with huge samples](http://ffxidb.com/zones/88/drachenlizard) in a lot of cases I've taken approx ~33% of the TH3+ rate *(rounded)*, first checking this falls within the error bars for the TH0 rate.

__To do:__
- general NPC trash drops in known farming hotspots *(Castle O. and Davoi)*

__Feedback:__
- As presented, Sky Gods currently have 1 guaranteed abjuration drop same as grounds kings, not sure this is correct, happy to adjust based on feedback.

### DIFF
--- a/sql/mob_droplist.sql
+++ b/sql/mob_droplist.sql
@@ -30536,15 +30536,15 @@ INSERT INTO `mob_droplist` VALUES (4128,0,0,1000,880,210);
 INSERT INTO `mob_droplist` VALUES (4129,2,0,1000,880,0); -- (Wendigo, Inner Horutoto Ruins, WAR) bone_chip: steal
 INSERT INTO `mob_droplist` VALUES (4129,0,0,1000,880,300); -- bone_chip
 INSERT INTO `mob_droplist` VALUES (4129,0,0,1000,880,270); -- bone_chip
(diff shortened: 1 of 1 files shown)


isotor:
General point I think just about everything here is a nerf - with the exception of Bounding Boots and Lindwurm skins, primarily because a lot of the existing data if from ffxiclopedia meaning TH is being double-counted, raised in https://github.com/DarkstarProject/darkstar/issues/2158.

DanteMccloud:
The Sky grouping here seems to fix most issues, however I noticed something on a few of the sky gods. You removed the possible 2nd drop for some of their items. Kirin's items for example are missing the possible 2nd drop on Damascus_ingot, and Shining cloth, raise3, quake, Orichalcum_ingot, was that by design?

zynjec:
Needs to be rebased, reopen when you have.

PR #5919, opened 2019-05-02
Correct Slendlix Spindlethumb modelid

isotor:
sgVkyPP


wrenffxi:
Wanted to say thanks for all the commits recently.

Issue #2440, opened 2015-11-26
Issues with the Crafting System


Deadwing888:
I'm making this to summarize everything off about the crafting system. Issues #1357 #2391 would be fixed as a result of fixing the following retail inconsistencies:

1) In retail there is no such thing as fractions of a crafting skill level-- it is always a whole number. Day/Moon/Direction do influence HQ rate, but they do not directly affect skill level. Simply removing this will fix the two issues listed above automatically.

2) Darkstar is slightly too generous with HQs. I have no parses to back me up on this, but as a level 100 crafter on retail my calculations for the base rate of HQ were: 

T0= 0-10 skill above cap = 1/64 hq rate
T1= 11-30 skill above cap = 1/16 hq rate
T2= 31-50 skill above cap = 1/4 hq rate
T3= 51+ skill above cap = 1/2 hq rate

Now I do believe day direction and moon all played a slight roll in HQ rate, but the key word here is slight. A 10% bonus for correct day and 10% penalty for incorrect day sounds about right to me. Moon and direction had even less effect ~5% tops.

3) Subcrafts should matter when it comes to HQ tier. With main craft 50 levels over and subcraft 11 levels over would result in a T1 hq rate.

4) Uncapping your FPS will cause you to craft ~4x faster with a good GPU.

5) Going to the synthesis history menue and spamming enter fast enough will result in being able to craft without crystal in inventory.

6) I never saw a a skillup value higher than .1 after passing level 60. 
It was always my belief that skillups higher than .1 ceased after level 60; although I have since been told I was wrong about this. 

7) Skillups are slightly too easy to earn at the lower levels, and become slightly too hard at the highest levels. It's close enough to be workable, but it's not quite right. Again, no parses to back me up.

8) #2434 

Source: I had level 100 smithing in retail, 4 crafts 90+ on Nasomi

teschnei:
not going to do anything about day/moon/direction unless it can be 100% confirmed either way - my experience is it does not, so it will stay that way by default (the setting can be changed to affect rate instead of level, though)

for 4, that's client-side - is the server stopping subsequent synths?

teschnei:
At any rate, issue "groups" are very annoying to deal with as you can't
mark certain parts as closed unless someone edits the original post (and
the title is not as descriptive)

Deadwing888:
I can 100% confirm day/moon/direction do not affect skill in retail.

[...]

PR #5924, opened 2019-05-03
Update HQ rates for various tiers

isotor:
Updating HQ tier success chances to the values [literally referenced just a few lines below](https://github.com/DarkstarProject/darkstar/blob/7eeef20eb4f9bd01a8f7948336a59a203fe3cb77/src/map/utils/synthutils.cpp#L403). Quick search finds this adjustment was also suggested previously within #2440.

Are fractions valid here? I *think* so but not 100%.


nfaletra:
Jesus, why is this all doubles in the first place? (not your fault, isotor) We don't need that much precision. . .

PR #5926, opened 2019-05-03
Minor changes to experience table based on wiki information

isotor:
Okay okay so wiki sucks but I figure filling in a table over a period of 7 years is probably something they can handle. Checked values against the [MRV](https://ffxiclopedia.fandom.com/wiki/Experience_Points?oldid=1277542) prior to the Feb 14 2011 update that added a multiplier and it got all fucked up.

And yes, I actually did this. Send help.


TeoTwawki:
_incredibly easy prey when?_

PR #5928, opened 2019-05-03
Swap modelid of THF and RNG specters in Fei'Yin

isotor:
THF model acts like a RNG, and vice versa. See here for pic of THF model:

https://ffxiclopedia.fandom.com/wiki/Specter *(on the current build this model pings arrows at you)*

This error is probably known to anyone whose tried farming for a Vali's Bow and/or Retaliators. This fixes. 🎉


TeoTwawki:
RE: oddness
our PH ID looks wrong compared to retail. off by one.
group table comments a supposedly retail captured model ID.

PR #5934, opened 2019-05-05
Fix increasing mog safe size

ibm2431:


ibm2431:
@zach2good The function name fooled me too!

PR #5936, opened 2019-05-05
Prevent multiple completions during the same conquest tally

ibm2431:
I don't remember if multiple quests can be completed in the same conquest tally, but if not, here you go!

### DIFF
--- a/scripts/globals/items/mahogany_bed.lua
+++ b/scripts/globals/items/mahogany_bed.lua
@@ -7,8 +7,9 @@ require("scripts/globals/quests");
 -----------------------------------------
 
 function onFurniturePlaced(player)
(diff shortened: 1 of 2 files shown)


ibm2431:
What's even more concerning is that [bgwiki doesn't even mention a conquest tally at all](https://www.bg-wiki.com/bg/Mog_House)!

TeoTwawki:
well..shit
```
[2:10 PM] Ivaar: They removed wait till conquest tally
[2:10 PM] Tēo: you know why your plant die in teh mog hous ebut not at the mog garden? because the pots are perfect for makin moogle moonshine
[2:10 PM] Nyu: There we go. Thanks Ivaar.
```

PR #5940, opened 2019-05-06
Stable fix back into master

TeoTwawki:
......


TeoTwawki:
hey @lyptt ? pr to master in the future, not stable, I didn't even see it till it was too late.

lyptt:
@TeoTwawki sure, no worries! I wasn't sure what the branching strategy was for this project so I just branched off what I'm using on my own machine.

TeoTwawki:
I would rather we sent bug fixes to both branches, but we haven't got the manpower to deal with it like that, so we just throw everything to master and periodically merge to stable before rebuilding the installer.

lyptt:
Okay cool, I'll do the latter in future 🙂

PR #5941, opened 2019-05-06
Fix synth support to remove existing support from other guild

isotor:
First pass just with the Bastok Goldsmith and Smith NPCs for feedback. Includes various bits of formatting *(ordering, semicolons, spaces, parens)* but the only logic change is `player:delStatusEffectsByFlag(dsp.effectFlag.SYNTH_SUPPORT, true)` onEventFinish to strip existing support which is [how it works on retail](https://ffxiclopedia.fandom.com/wiki/Synthesis_Image_Support)

> This effect will be replaced by synthesis image support for any craftskill except the craftskill currently supported by the existing effect.

__Other issues:__

1. `if player:hasStatusEffect(dsp.effect.SMITHING_IMAGERY) == false then` is probably not the correct check for the advanced support NPCs since they blow you off if you have basic support. I need to confirm this on retail but I strongly suspect this check should only catch existing *advanced* image support.
2. Fatimah still has `0xB0001AF`

### DIFF
--- a/scripts/zones/Bastok_Markets/npcs/Fatimah.lua
+++ b/scripts/zones/Bastok_Markets/npcs/Fatimah.lua
@@ -4,43 +4,44 @@
 -- Type: Goldsmithing Adv. Synthesis Image Support
 -- !pos -193.849 -7.824 -56.372 235
 -----------------------------------
-require("scripts/globals/status");
-require("scripts/globals/crafting");
-local ID = require("scripts/zones/Bastok_Markets/IDs");
+local ID = require("scripts/zones/Bastok_Markets/IDs")
+require("scripts/globals/status")
+require("scripts/globals/crafting")
 -----------------------------------
 
 function onTrade(player,npc,trade)
-end;
+end
 
 function onTrigger(player,npc)
-    local guildMember = isGuildMember(player,6);
-    local SkillLevel = player:getSkillLevel(dsp.skill.GOLDSMITHING);
-    local Cost = getAdvImageSupportCost(player, dsp.skill.GOLDSMITHING);
+    local guildMember = isGuildMember(player, 6)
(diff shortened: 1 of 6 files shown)


ibm2431:
You're going to hate me @isotor , but I've gotta get this out:
"Would it be possible to have a function in the crafting global to help reduce this repeated code?"

Table of crafts if we don't already have one, table linking guilds to `dsp.skill.CRAFT_SKILL` and `dsp.effect.CRAFT_IMAGERY`, then functions that takes the guild, event CS, and params the specific NPC requires:
```
synthSupport = {
    [dsp.craft.GOLDSMITHING] = {dsp.skill.GOLDSMITHING, (dsp.effect.GOLDSMITHING_IMAGERY}
}

function startSupportEvent(guild, event, params)
    local guildMember = isGuildMember(player, guild)
    local skill, imagery = synthSupport[guild][1], synthSupport[guild][2]
    local SkillLevel = player:getSkillLevel(skill)
    local Cost = getAdvImageSupportCost(player, skill)
```
Then replace the logic in the NPC scripts with something like:
`startSupportEvent(guild, event_id, unique_npc_param x4)`

And a `handleSupportFinish(guild, event)` or something.

Not saying this is vital, but if you're considering going through all the synth support NPCs...! (You should wait on input from others though.)

isotor:
That's kinda why I stopped and opened a PR with just a few changed as examples. I got burned once already re the Signet message and learned my lesson :kill_me:

ibm2431:
I do recognize that I have a table fetish!

So you may want a second opinion from a more neutral party~

TeoTwawki:
RE: Fatimah 
TO DA CAPCHTA DISCORD!

TeoTwawki:
```
continuous-integration/appveyor/pr Expected — Waiting for status to be reported
```
can someone look into that pls? and decide wtf to do with this and all the others stuck on that message

PR #5942, opened 2019-05-06
Bandaid fix on purchasing HQ crystals with GP

isotor:
Okay so there's undoubtedly a better fix for this but here's my offer for a temporary/bandaid solution. Right now the client let's you order as many HQ crystals as you like but only gives you a max of 12 (one stack). This is a quick fix to limit the ~~amount of GP paid to the same maximum (12 x 200 GP each = 2400)~~ number of crystals you can request in one trade to 12 *(tho the client will still let you input however many you want)*.


ibm2431:
If there's a cap on crystals given, then cap the crystals :wink:

```
        local quantity = math.min(bit.rshift(option, 9), 12);
        local cost = quantity * i.cost;
```

PR #5944, opened 2019-05-06
lua cleanup a few job abilities

isotor:


scripts/globals/abilities/chi_blast.lua
```diff
@@ -29,4 +29,8 @@ function onUseAbility(player,target,ability)
     target:updateEnmityFromDamage(player,dmg)
     player:delStatusEffect(dsp.effect.BOOST)
     return dmg
+
+	if dmg > 0 and target(isAsleep) then
```
TeoTwawki: the return above means this will never be hit.

isotor: Woops, it's early plz forgive

ibm2431:
[target:takeDamage in core should already be accounting for this](https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_baseentity.cpp#L7451)

isotor:
Ah, fair point. Removed, but I figure may as well leave the PR open just for the free lua cleanup.

PR #5945, opened 2019-05-06
Update first hHP/hMP recovery to occur at second healing tic

isotor:
https://ffxiclopedia.fandom.com/wiki/MP_Recovered_While_Healing

> The first recovery tick requires 20 seconds of rest and each subsequent recovery tick occurs every 10 seconds. 

Rest is just formatting.

### DIFF
--- a/scripts/globals/effects/healing.lua
+++ b/scripts/globals/effects/healing.lua
@@ -1,12 +1,12 @@
 -----------------------------------
 --
 -- dsp.effect.HEALING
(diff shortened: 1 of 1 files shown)


ibm2431:
Apparently applies to hHP too:

[HP Recovered While Healing](https://ffxiclopedia.fandom.com/wiki/HP_Recovered_While_Healing):
> A /healing tick is 10 seconds, but the first tick takes 20 seconds

PR #5946, opened 2019-05-06
Change Greater/Colibri subjob to BLM to improve ele and drk magic skill

isotor:
... ranking when reflecting stuff


zynjec:
That might help elemental and dark magic, but what about ninjutsu and songs that can also be reflected?

It might make sense to look in to MAB/MACC as well as/instead of just changing the support job.

TeoTwawki:
per discord chat we either need to directly add skill via modifiers or we need to change the mimic code to just act as if we have A rank skill on cast for whatever magic type it is

zynjec:
Going to close this for now, we can discuss on Discord about potential other fixes (like Teo's idea of capped skill), but I feel this is a lateral movement for reflected spell power.

PR #5948, opened 2019-05-06
Fix a few physical mob TP moves to only proc addeffect on damage

isotor:
If misses or hits for 0 addeffect shouldn't be applied. Few bits of formatting.

### DIFF
--- a/scripts/globals/mobskills/bone_crunch.lua
+++ b/scripts/globals/mobskills/bone_crunch.lua
@@ -18,11 +18,10 @@ function onMobWeaponSkill(target, mob, skill)
     local dmgmod = 2.5
     local info = MobPhysicalMove(mob,target,skill,numhits,accmod,dmgmod,TP_NO_EFFECT)
     local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,dsp.attackType.PHYSICAL,dsp.damageType.BLUNT,info.hitslanded)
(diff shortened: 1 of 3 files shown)


scripts/globals/mobskills/bone_crunch.lua
```diff
@@ -1,29 +1,29 @@
...
-    local typeEffect = dsp.effect.PLAGUE
-
-    MobPhysicalStatusEffectMove(mob, target, skill, typeEffect, 5, 0, 30)
 
     local numhits = 1
     local accmod = 1
     local dmgmod = 2.5
     local info = MobPhysicalMove(mob,target,skill,numhits,accmod,dmgmod,TP_NO_EFFECT)
     local dmg = MobFinalAdjustments(info.dmg,mob,skill,target,dsp.attackType.PHYSICAL,dsp.damageType.BLUNT,info.hitslanded)
+    local typeEffect = dsp.effect.PLAGUE
```
TeoTwawki: no need for a variable since it is never changing and typeEffect is only called once, just use it directly

isotor: Good deal 👍

TeoTwawki: same on the other scripts, guess I should have said so in the first place

PR #5950, opened 2019-05-06
[WIP] Set cap on CP awarded for rank bar overspill on trade

isotor:
__Closing an exploit with arbitrary limit.__

Just tried this on local master build at Rank 2 - traded a stack of Earth crystals to set me very close to the rank cap, then traded 8 full stacks of Dark crystals and was awarded a pretty hilarious 48,000 CP. I don't believe this is an intended use of this feature so propose setting an arbitrary cap of 1k CP for any amount of excess crystals above filling the rank bar. 

I have absolutely no support for the 1k cap, and the correct fix is probably an hefty reduction on CP-per-crystal, but this PR is just an option for consideration/discussion.

### DIFF
--- a/scripts/globals/conquest.lua
+++ b/scripts/globals/conquest.lua
@@ -939,7 +939,7 @@ dsp.conquest.overseerOnTrade = function(player, npc, trade, guardNation, guardTy
                         break
                     else
                         trade:confirmItem(crystalId, count)
-                        addPoints = addPoints + count * math.floor(4000 / (player:getRank() * 12 - crystalWorth))
+                        addPoints = addPoints + count * math.floor(400 / (player:getRank() * 12 - crystalWorth))
                     end
                 end
             end
@@ -947,7 +947,7 @@ dsp.conquest.overseerOnTrade = function(player, npc, trade, guardNation, guardTy
             if addPoints > 0 and pRank ~= 1 and pRankPoints < 4000 then
                 if pRankPoints + addPoints >= 4000 then
                     player:setRankPoints(4000)
-                    player:addCP(math.min(pRankPoints + addPoints - 4000, 1000))
+                    player:addCP(pRankPoints + addPoints - 4000)
                     player:showText(npc, mOffset + 44) -- "Your rank points are full. We've added the excess to your conquest points."
                 else
                     player:addRankPoints(addPoints)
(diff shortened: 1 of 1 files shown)


TeoTwawki:
you can do this on retail, but you won't get that much CP for doing it. once the rank bar is full they refuse trades entirely but I did get a lot of my CP by getting as close to full as I could and then bombing the NPC with as many Dark Crystals as I could.

No way would retail give that much though, so I dunno wtf is going on here.

isotor:
It's giving 500 CP per dark crystal, I agree a better fix would be knocking a 0 off of the CP received per excess crystal - but up to you if that's how you want to proceed as I don't have specific proof for any of this it's just clearly super-borked in it's current implementation

TeoTwawki:
yeah lets make'em 50 each for now..
pretty sure on retail the value per crystal would involve what rank we are and the NPC sale value of the crystal (making light and dark more valuable here)

isotor:
@TeoTwawki `crystalWorth` values seem fine ([link](https://github.com/DarkstarProject/darkstar/blob/fa7ad5c82257a542fb2a047328a89fe8adfd8368/scripts/globals/conquest.lua#L526)), the issues is in the equation:

https://github.com/DarkstarProject/darkstar/blob/fa7ad5c82257a542fb2a047328a89fe8adfd8368/scripts/globals/conquest.lua#L942

which I suspect is just made up using a few known factors *(overspill, diminishing returns at higher ranks)*. 4000 seems totally arbitrary outside of being related to the rank bar itself, so knocking a 0 off to achieve the aim of reducing CP-per-crystal by a factor of 10 until someone decides to figure out the correct equation on retail.

isotor:
Also a few [Wikiclaims™](https://ffxiclopedia.fandom.com/wiki/Talk:Conquest_Points) taken together (Rank 6  =~ 600, and Rank 10 = 96) confirm the current/above equation is fundamentally wrong. Just a tidbit for whoever decides to fix this properly in 2039.

isotor:
Setting a [WIP] tag on this for a day or so while I enlist some help on retail to figure out the best way to proceed.

zynjec:
Going to close, feel free to PR once you're ready for it to be reviewed.

PR #5961, opened 2019-05-06
Implement tiered MB accuracy bonus

isotor:
Stole `local casterLevel` from Helix duration calc further down.

Selected tiers as 32/64/128, increasing at the level Black Mage acquires a higher tier nuke (Stone II at 26, Stone III at 51).

Not particularly married to any of it so wade in with your hot take <3

### DIFF
--- a/scripts/globals/magic.lua
+++ b/scripts/globals/magic.lua
@@ -581,7 +581,6 @@ function getSpellBonusAcc(caster, target, spell, params)
         else
             levelBonus = 128
         end
-
         magicAccBonus = magicAccBonus + levelBonus
         return magicAccBonus
     end
@@ -1039,6 +1038,8 @@ end;
 
 function getHelixDuration(caster)
     --Dark Arts will further increase Helix duration, but testing is ongoing.
+
+    local casterLevel = caster:getMainLvl();
     local duration = 30; --fallthrough
     if (casterLevel <= 39) then
         duration = 30;
--- a/scripts/globals/magic.lua
+++ b/scripts/globals/magic.lua
@@ -570,8 +570,18 @@ function getSpellBonusAcc(caster, target, spell, params)
     magicAccBonus = magicAccBonus + params.AMIIaccBonus;
 
     --add acc for skillchains
-    if (skillchainTier > 0) then
-        magicAccBonus = magicAccBonus + 25;
+    local casterLevel = caster:getMainLvl()
+    if skillchainTier > 0 then
+        local levelBonus = 0
+        if (casterLevel <= 25) then
+            levelBonus = 32;
+        elseif (casterLevel <= 50) then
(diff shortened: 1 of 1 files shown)


TeoTwawki:
logics wrong but even if it wasn't I don't think this is a good solution

isotor:
Fixed the return, regarding the nature of the solution I'll wait for a better idea as this is what was suggested yesterday in the Discord group.

scripts/globals/magic.lua
```diff
@@ -1027,8 +1039,6 @@ end;
 
 function getHelixDuration(caster)
     --Dark Arts will further increase Helix duration, but testing is ongoing.
-
-    local casterLevel = caster:getMainLvl();
```
nfaletra: You can't remove this variable. It's in a separate function scope. If you cast a helix spell with this code as it is, your server will crash.

scripts/globals/magic.lua
```diff
@@ -569,9 +569,21 @@ function getSpellBonusAcc(caster, target, spell, params)
     --add acc for BLM AMII spells
     magicAccBonus = magicAccBonus + params.AMIIaccBonus;
 
-    --add acc for skillchains
-    if (skillchainTier > 0) then
-        magicAccBonus = magicAccBonus + 25;
+    -- Add acc for skillchains
+    local casterLevel = caster:getMainLvl()
+
+    if skillchainTier > 0 then
```
nfaletra: I'm a little confused. Where is this info being pulled from?

scripts/globals/magic.lua
isotor: Woops, will correct.

scripts/globals/magic.lua
isotor: The figures? Nowhere, literally out of my behind. Just a general feeling I've heard from lots of DSP server players that Magic Burst accuracy bonus feels very underwhelming on DSP, and a static+25 all the way to 75 sort of supports that sentiment. The tiers and new figures I just made up based on what a few people suggested I try yesterday on Discord (instead of just changing the 25 to a bigger value and calling it good).

nfaletra:
No offense to you or anything, but we don't really like changes like this without at least some documented proof to back them up. There are a lot of issues with magic damage/accuracy calculations in general and that stems from lack of data from retail. Changes like this are best left to the discretion of individual server devs. Feel free to share your ideas with them directly and see how they like them, but we don't want changes like this on master.

scripts/globals/magic.lua
TeoTwawki: I don't think retail even does a macc bonus like we have here at 25. I think its the mobs resistance to the SC's elements that is changed on retail, during the SC. Which isn't as similar as one might think.

PR #5963, opened 2019-05-06
Fix typo in Evasion Down effect script

### LINKED ISSUE #5701: Evasion down effect sets evasion to 0 in some situations
**_I have:_**

**_Client Version_** (type `/ver` in game) **:** 30181205_0

Problem is this if line in scripts/globals/effects/evasion_down.lua

    if (target:getMod(dsp.mod.EVA) - effect:getPower() < 0) then
        effect:setPower(target:getStat(dsp.mod.EVA))

When you have less evasion+ from gear/skills than the effect is applying (such as wearing just Emperor Hairpin and getting hit by Infrasonics) then it sets Evasion to 0, if you have more then it just applies it as normal. It was added when a bunch of THF stuff was implemented, possibly just to check if Feint is reducing evasion below 0? Although from tests it seems like that is already handled without this check so it could probably just be removed.

isotor:
Fixes #5701.

------

Edit: I figure this just wasn't explained very well. My reading of the code suggests the following situation:

- Target natural evasion: 240
- Target evasion mod total: 10 *(say from gear)*
- Effect power: 40

Check currently calculates that 40 (**Effect**) is bigger than 10 (**Mod**), and so sets the effect power to 240 (**Stat**)

Correct check is whether the effect power exceeds the target's natural evasion - not sure if maybe this should also include gear from mods but then there's the issue of what happens if you're set to 0 then take off your +eva gear?

https://github.com/DarkstarProject/darkstar/blob/1a6eaa31d76bf449b91b972142cf815ddda97bfb/scripts/globals/status.lua#L884-L885

### DIFF
--- a/scripts/globals/effects/evasion_down.lua
+++ b/scripts/globals/effects/evasion_down.lua
@@ -7,24 +7,24 @@ require("scripts/globals/status")
 -----------------------------------
 
 function onEffectGain(target,effect)
-    if (target:getMod(dsp.mod.EVA) - effect:getPower() < 0) then
+    if target:getStat(dsp.mod.EVA) - effect:getPower() < 0 then
         effect:setPower(target:getStat(dsp.mod.EVA))
     end
-    target:addMod(dsp.mod.EVA,-effect:getPower())
+    target:addMod(dsp.mod.EVA, -effect:getPower())
 end
 
 function onEffectTick(target,effect)
     -- Only Feint uses the tick, restore 10 evasion every tick
     local evaDownAmt = effect:getPower()
-    if (evaDownAmt > 0) then
-        effect:setPower(evaDownAmt - 10)
-        target:delMod(dsp.mod.EVA, -10)
+    if evaDownAmt > 0 then
+        effect:setPower(math.max(evaDownAmt - 10, 0))
+        target:delMod(dsp.mod.EVA, 10)
     end
 end
 
 function onEffectLose(target,effect)
     local evaDownAmt = effect:getPower()
-    if (evaDownAmt > 0) then
-        target:delMod(dsp.mod.EVA,-effect:getPower())
+    if evaDownAmt > 0 then
+        target:delMod(dsp.mod.EVA, effect:getPower())
     end
 end


TeoTwawki:
@nfaletra 95af1b832ab0c7e051946748a51baaa8bb02bdd1 was yours, what was the intent there and does this do what you intended by using getStat instead?

zynjec:
`getStat` didn't support evasion before this commit. (added https://github.com/DarkstarProject/darkstar/commit/aaf1055c7fc7d9aa56dc59e49a1ff5f47053473b)

That issue is probably stale, and fixed as of @nfaletra's commit.

nfaletra:
I don't see what the issue with #5701 is? Is it supposed to set evasion below zero? Because that check is just making sure it doesn't set your evasion below zero. Changing it to getStat won't actually change any functionality and I doubt this code was tested.

nfaletra:
The "typo" is actually in onEffectTick
`target:delMod(dsp.mod.EVA, -10)`

That should not be a negative value.

teschnei:
also onEffectLose

zynjec:
Well, there are multiple things.

- Double negatives from `delMod`
- Shouldn't be using `onEffectTick` for anything but Feint, it's otherwise *not* a gradual decrease.
- That `onEffectTick` is causing `EVASION_DOWN`'s power to overflow, giving hilarious amounts of evasion if left on a long period of time.

image

isotor:
I figure this just wasn't explained very well. My layman's reading of the code suggests the following:

__Hypothetical:__
- Target base evasion total = 240
- Target evasion mod total = 10 *(say an empy pin)*
- Evasion down power = 40
- Check reads that 40 (**effect**) is bigger than 10 (**mod**), and so sets the effect power to 250 (**stat**)

This is what the issue report also suggests is happening. Is this wrong? If so then apologies for not being more clear.

isotor:
Think I've resolved the issues except:

- Shouldn't be using `onEffectTick` for anything but Feint, it's otherwise not a gradual decrease.

PR #5965, opened 2019-05-06
change table item_armor to item_armor to item_equip

KnowOne134:
do we want to also change the file names and pointers for item_armor?


src/search/data_loader.cpp
```diff
@@ -103,7 +103,7 @@ std::vector<ahItem*> CDataLoader::GetAHItemsToCategory(uint8 AHCategoryID, int8*
     const char* fmtQuery = "SELECT item_basic.itemid, item_basic.stackSize, COUNT(*)-SUM(stack), SUM(stack) "
         "FROM item_basic "
         "LEFT JOIN auction_house ON item_basic.itemId = auction_house.itemid AND auction_house.buyer_name IS NULL "
-        "LEFT JOIN item_armor ON item_basic.itemid = item_armor.itemid "
+        "LEFT JOIN item_equip ON item_basic.itemid = item_armor.itemid "
```
helixhamin: Should this have `= item_armor.idemid`, or should it be changed to `= item_equip.itemid`?

KnowOne134: good find and i found another, search doesnt search very well

helixhamin:
lol, I just want to know if this is happening, so I know when I can start adding in more missing items xD
Guess I will need to poke around to see if there is anything else I can work on :o

src/search/data_loader.cpp
zynjec: You could use something like `git grep item_armor` instead of depending on whatever search you used.

teschnei:
also yes, we should change the CItemArmor type name too

KnowOne134:
We can merge this as is and i can start working on pointers tomorrow so it doesnt stop work thats going on in the equip sql. 
Merging as is will not break anything

PR #5967, opened 2019-05-07
Switching instruments shouldn't wipe TP

isotor:
[Fix proposed by beauxq](https://github.com/bdhs/darkstar/commit/7f2901a8ae733de1e52f01e08ed2867333b78d9c)

Tested locally, does what it says on the tin.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -1620,7 +1620,9 @@ namespace charutils
                         PChar->look.ranged = 0;
                     }
                     PChar->m_Weapons[SLOT_RANGED] = nullptr;
-                    PChar->health.tp = 0;
+                    if (((CItemArmor*)PItem)->getJobs() != 512) {  // Is not instrument
+                        PChar->health.tp = 0;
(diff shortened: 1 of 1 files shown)


src/map/utils/charutils.cpp
```diff
@@ -2142,7 +2144,9 @@ namespace charutils
         }
         if (equipSlotID == SLOT_MAIN || equipSlotID == SLOT_RANGED || equipSlotID == SLOT_SUB)
         {
-            PChar->health.tp = 0;
+            if (equipSlotID != SLOT_RANGED || PItem == nullptr || PItem->getJobs() != 512) {  // Is not instrument
```
nfaletra: Any ranged item with BRD only as the job will get through this check. Add a function to CItemArmor called like ```IsInstrument``` and have that function check the skill type. I'm sure there are other places in code that would benefit from that abstraction.

zynjec:
Going to close this since there was no changes made.

Simply checking for BRD isn't an appropriate check.

PR #5968, opened 2019-05-07
Add key item message to Monberaux at PM1-1 end

isotor:
https://ffxiclopedia.fandom.com/wiki/The_Rites_of_Life


scripts/zones/Upper_Jeuno/npcs/Monberaux.lua
```diff
@@ -86,6 +86,7 @@ function onEventFinish(player,csid,option)
     elseif (csid == 10) then
         player:setVar("PromathiaStatus",0);
         player:addKeyItem(dsp.ki.MYSTERIOUS_AMULET_DRAINED);
+        player:messageSpecial(ID.text.KEYITEM_OBTAINED,dsp.ki.MYSTERIOUS_AMULET);
```
ibm2431: Does dsp.ki.MYSTERIOUS_AMULET_DRAINED not display as the same?

isotor: No idea, but not sure it matters as it's just a message. Can check if needed?

On retail: https://youtu.be/pKxQd3AbPWU?t=253

PR #5970, opened 2019-05-07
Update Skulker's Cape bonus effect

isotor:
Previously estimated as 1.5x from BGwiki. This was also the case on wiki *(with a verification needed tag)* until [27 March 2018](https://ffxiclopedia.fandom.com/wiki/Skulker%27s_Cape?oldid=1651833) when it was updated to +30 seconds and the tag removed. The entry on BG wiki [hasn't been updated since 2016](https://www.bg-wiki.com/index.php?title=Skulker%27s_Cape&action=history) so I suspect is using the old estimate from wiki.

Also: yes I saw the comment on making a mod, will look into that once the bonus is agreed on 👍

### DIFF
--- a/scripts/globals/spells/invisible.lua
+++ b/scripts/globals/spells/invisible.lua
@@ -16,9 +16,11 @@ end
 function onSpellCast(caster, target, spell)
     if not target:hasStatusEffect(dsp.effect.INVISIBLE) then
 
-        local duration = calculateDuration(math.random(42, 54) * 10, spell:getSkillType(), spell:getSpellGroup(), caster, target)
+        local duration = calculateDuration(math.random(420, 540), spell:getSkillType(), spell:getSpellGroup(), caster, target)
 
-        duration = calculateDurationForLvl(duration, 25, target:getMainLvl()) + player:getMod(dsp.mod.INVISIBLE_DURATION)
+		duration = duration + target:getMod(dsp.mod.INVISIBLE_DURATION)
+		
+        duration = calculateDurationForLvl(duration, 20, target:getMainLvl())
 
         spell:setMsg(dsp.msg.basic.MAGIC_GAIN_EFFECT)
         target:addStatusEffect(dsp.effect.INVISIBLE, 0, 10, math.floor(duration * SNEAK_INVIS_DURATION_MULTIPLIER))
--- a/scripts/globals/spells/invisible.lua
+++ b/scripts/globals/spells/invisible.lua
@@ -18,8 +18,8 @@ function onSpellCast(caster, target, spell)
 
         local duration = calculateDuration(math.random(420, 540), spell:getSkillType(), spell:getSpellGroup(), caster, target)
 
-		duration = duration + target:getMod(dsp.mod.INVISIBLE_DURATION)
-		
+        duration = duration + target:getMod(dsp.mod.INVISIBLE_DURATION)
+	
         duration = calculateDurationForLvl(duration, 20, target:getMainLvl())
 
         spell:setMsg(dsp.msg.basic.MAGIC_GAIN_EFFECT)
--- a/scripts/globals/spells/invisible.lua
+++ b/scripts/globals/spells/invisible.lua
@@ -21,7 +21,7 @@ function onSpellCast(caster, target, spell)
 
         -- FIXME: Create mod and use that instead of an itemID check
         if (target:getEquipID(dsp.slot.BACK) == 13692) then -- skulker's cape
-            duration = duration * 1.5
+            duration = duration + 30
         end
 
         spell:setMsg(dsp.msg.basic.MAGIC_GAIN_EFFECT)
--- a/scripts/globals/spells/invisible.lua
+++ b/scripts/globals/spells/invisible.lua
@@ -1,12 +1,13 @@
 -----------------------------------------
 -- Spell: Invisible
 -- Lessens chance of being detected by sight.
--- Duration is random number between 30 seconds and 5 minutes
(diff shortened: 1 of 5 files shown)


TeoTwawki:
Would you mind killing the itemID check and making this into a mod, then any future items added could use the mod as well.

scripts/globals/spells/invisible.lua
```diff
@@ -15,19 +16,14 @@ end
...
-        -- last 7-9 minutes
-        local duration = calculateDuration(math.random(420, 540), spell:getSkillType(), spell:getSpellGroup(), caster, target)
-        duration = calculateDurationForLvl(duration, 25, target:getMainLvl())
+        local duration = calculateDuration(math.random(42, 54) * 10, spell:getSkillType(), spell:getSpellGroup(), caster, target)
 
-        -- FIXME: Create mod and use that instead of an itemID check
-        if (target:getEquipID(dsp.slot.BACK) == 13692) then -- skulker's cape
-            duration = duration * 1.5
-        end
+        duration = calculateDurationForLvl(duration, 25, target:getMainLvl()) + player:getMod(dsp.mod.INVISIBLE_DURATION)
```
nfaletra: This should only increase the duration on the wearer. Also, it should be added into the base duration, not tacked on after a level check (won't matter because you can't cast it on yourself if you're level synced below the spell level)

scripts/globals/spells/sneak.lua
```diff
@@ -1,33 +1,29 @@
...
-        -- last 7-9 minutes
-        local duration = calculateDuration(math.random(420, 540), spell:getSkillType(), spell:getSpellGroup(), caster, target)
-        duration = calculateDurationForLvl(duration, 20, target:getMainLvl())
+        local duration = calculateDuration(math.random(42, 54) * 10, spell:getSkillType(), spell:getSpellGroup(), caster, target)
 
-        -- FIXME: Create mod and use that instead of an itemID check
-        if (target:getEquipID(dsp.slot.BACK) == 13692) then -- skulker's cape
-            duration = duration * 1.5
-        end
+        duration = calculateDurationForLvl(duration, 20, target:getMainLvl()) + player:getMod(dsp.mod.SNEAK_DURATION)
```
nfaletra: This should only increase the duration on the wearer. Also, it should be added into the base duration, not tacked on after a level check (won't matter because you can't cast it on yourself if you're level synced below the spell level)

scripts/globals/spells/sneak.lua
```diff
@@ -1,33 +1,29 @@
...
 
-function onSpellCast(caster,target,spell)
-    if (not target:hasStatusEffect(dsp.effect.SNEAK)) then
+function onSpellCast(caster, target, spell)
+    if not target:hasStatusEffect(dsp.effect.SNEAK) then
 
-        -- last 7-9 minutes
-        local duration = calculateDuration(math.random(420, 540), spell:getSkillType(), spell:getSpellGroup(), caster, target)
-        duration = calculateDurationForLvl(duration, 20, target:getMainLvl())
+        local duration = calculateDuration(math.random(42, 54) * 10, spell:getSkillType(), spell:getSpellGroup(), caster, target)
```
nfaletra: This is making the duration be locked to 10 second intervals. If that's intentional, I'm not sure where the proof for that behavior exists.

scripts/globals/spells/invisible.lua
```diff
@@ -15,19 +16,14 @@ end
 function onSpellCast(caster, target, spell)
     if not target:hasStatusEffect(dsp.effect.INVISIBLE) then
 
-        -- last 7-9 minutes
-        local duration = calculateDuration(math.random(420, 540), spell:getSkillType(), spell:getSpellGroup(), caster, target)
-        duration = calculateDurationForLvl(duration, 25, target:getMainLvl())
+        local duration = calculateDuration(math.random(42, 54) * 10, spell:getSkillType(), spell:getSpellGroup(), caster, target)
```
nfaletra: This is making the duration be locked to 10 second intervals. If that's intentional, I'm not sure where the proof for that behavior exists.

scripts/globals/spells/sneak.lua
TeoTwawki: he asked me about it and I ok'd it because when do you see 93 seconds or 231 second durations ever. I figure when SE beefed up the durations on retail they just multiplied it. didn't realize what I was really okaying I was thinking of the setting. my bad.

TeoTwawki: actually i still think it'd be ok. we won't know if its accurate or not though without a lot of checking retail with same items

isotor: I'll dump it from this PR and if I can find support add it separately, since my thoughts on this would extend to pots and jigs and stuff also.

scripts/globals/spells/invisible.lua
```diff
@@ -15,19 +16,16 @@ end
...
         local duration = calculateDuration(math.random(420, 540), spell:getSkillType(), spell:getSpellGroup(), caster, target)
-        duration = calculateDurationForLvl(duration, 25, target:getMainLvl())
 
-        -- FIXME: Create mod and use that instead of an itemID check
-        if (target:getEquipID(dsp.slot.BACK) == 13692) then -- skulker's cape
-            duration = duration * 1.5
-        end
+		duration = duration + target:getMod(dsp.mod.INVISIBLE_DURATION)
+		
+        duration = calculateDurationForLvl(duration, 20, target:getMainLvl())
```
nfaletra: cool extra space

nfaletra:
Fix the one whitespace thing and it looks good.

PR #5971, opened 2019-05-07
Add Sword Strap 2H delay mod

isotor:
https://ffxiclopedia.fandom.com/wiki/Sword_Strap

Technically ~~171~~ 380 is just ~~`DELAY`~~ `DELAYP` but like it's on a strap so I figure that's ok.


isotor:
Thanks @zynjec, updated per your comments. The TP thing isn't ideal but as you say it's better for now if you decide to merge.

PR #5974, opened 2019-05-07
Increase critical damage cap

isotor:
The cap for critical damage is currently the same as the cap for normal damage - for critical hits the cap for the multiplier should be 3 (see: https://ffxiclopedia.fandom.com/wiki/Level_Correction_Function_and_pDIF)

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -2329,7 +2329,7 @@ namespace battleutils
 
         if (PAttacker->objtype == TYPE_PC)
         {
-            ratioCap = 2.25f;
+            ratioCap = isCritical ? 3 : 2.25f;
         }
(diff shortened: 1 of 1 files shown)


TeoTwawki:
your source hasn't been updated since 2013 and I am pretty sure things changed since then. Verify by..Anything else. please.

nfaletra:
First of all, the title doesn't accurately describe what this PR is trying to do. This change won't affect the critical damage cap at all. It's still capped at +100%, which is correct.

There is more wrong with this function that simply setting the ratio cap to 3 if the hit is a critical. The ratio cap needs to be adjusted based on the type of weapon that is being used.

Here's some more up-to-date info on how to correct the math. (still circa 2016, but most recent I could find).

https://www.bluegartr.com/threads/127523-pDIF-Changes-%28Feb.-10th-2016%29

PR #5975, opened 2019-05-07
Updates for True Will quest (Ninja AF3)

isotor:
Don't need to be NIN to pick up quest True Will, see: https://ffxiclopedia.fandom.com/wiki/True_Will

> *It is not required to be Ninja to pick up this quest.*


isotor:
Second commit adds 2hrs to the mobs, and enables receiving the Trick Box with only one mob killed (see above link to quest walkthrough)

> *Only one Kappa needs to be defeated,*

ibm2431:
> As with other fights using this strategy, you must wait for the other two Kappas to de-spawn before defeating the one you choose to fight.

Might need to: in mob death -> do a check to see if the others are despawned -> if so then set var saying NMs "killed".
Then in the qm, just check for the var, don't need to check if dead.

isotor:
Touched pools since [wiki says Bonze can't be slept](https://ffxiclopedia.fandom.com/wiki/True_Will)

TeoTwawki:
> > As with other fights using this strategy, you must wait for the other two Kappas to de-spawn before defeating the one you choose to fight.
> 
> Might need to: in mob death -> do a check to see if the others are despawned -> if so then set var saying NMs "killed".
> Then in the qm, just check for the var, don't need to check if dead (because the "can get by with just one" behavior has already been accounted for).

I'd suggest looking at if the way Headstone Pilgrimage mobs were done is accurate here.

PR #5978, opened 2019-05-07
Add global setting for Survival Guides and set NPC content tag

isotor:


zynjec:
Trying to get away from having era settings, and just targeting current retail so going to close this.

isotor:
I don't really understand why as this isn't "era" it's just preference - no different from an FOV/GOV content tag?

ibm2431:
Would it be acceptable, where if instead of individual toggles, FOV, GOV, and Survival all shared a bitmask flag of `BOOKS_ENABLED`?

helixhamin:
I agree, we already have the following:
-- FIELDS OF VALOR/Grounds of Valor settings
ENABLE_FIELD_MANUALS  = 1; -- Enables Fields of Valor
ENABLE_GROUNDS_TOMES  = 1; -- Enables Grounds of Valor
REGIME_WAIT = 1; -- Make people wait till 00:00 game time as in retail. If it's 0, there is no wait time.
LOW_LEVEL_REGIME = 0; -- Allow people to kill regime targets even if they give no exp, allowing people to farm regime targets at 75 in low level areas.

But yeah, maybe as IBM mentions, the setting should be enable books or something, since I don't see a situation where we enable FoV/GoV and not enable SG.

TeoTwawki:
We could set the books to be the Expansion they came in with.

I don't think we even need a mask - one on off. I can modify my own server to kill the ones I don't want easy enough if I want to go non standard (its one query to change NPC status)

TeoTwawki:
p.s. I think the explorer moogle settinsg can go as well. its just 3 NPC scripts / 3 SQL lines, its bloaty to have that ina  setting.

TeoTwawki:
p.p.s
@zynjec in some cases I feel we should still have settings. Just think we need to axe all the trivial/redundant stuff we accumulated, restructure how we go about things so that its clearer to people what things will be rejected. I removed the limbus portal setting for example but that commit is in the bcnmcrap branch I think.

PR #5979, opened 2019-05-07
[WIP] Consolidate first batch of settings into OLDSCHOOL_ENABLED

isotor:
First trial batch of consolidating global settings into a single switch. Will add more but wanted to get (#engagecorporate) gauge the temperature of the room before spending more time on it.

Please Don't merge yet, there is plenty more >.>

@zynjec

### DIFF
--- a/scripts/globals/magic.lua
+++ b/scripts/globals/magic.lua
@@ -1125,7 +1125,7 @@ function doElementalNuke(caster, spell, target, spellParams)
     local V = 0;
     local M = 0;
 
-    if USE_OLD_MAGIC_DAMAGE and spellParams.V ~= nil and spellParams.M ~= nil then
+    if OLDSCHOOL_ENABLED and spellParams.V ~= nil and spellParams.M ~= nil then
         V = spellParams.V; -- Base value
         M = spellParams.M; -- Tier multiplier
         local I = spellParams.I; -- Inflection point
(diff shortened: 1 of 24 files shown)


ibm2431:
Death to `STONESKIN_CAP`!  :fire: :mountain:
And `AQUAVEIL_COUNTER`! :fire: :droplet:

(These have never been changed. And current players on retail are getting hit _harder_ these days, so 350 should be ample!)

isotor:
Can I also kill the DIA/BIO_OVERWRITE now they work properly pleeeeeease @TeoTwawki

zynjec:
This still doesn't resolve having to maintain multiple "bases" for some code, lots of if trees, and now more in-fighting of what is considered "old school" and if it should/could be included in it.

TeoTwawki:
@DarkstarProject/contributors  @DarkstarProject/core what say you to a "oldschool.lua" that the rest of us can simply ignore

maxtherabbit:
"oldschool" is too ambiguous, if we were going to consolidate settings I would suggest having specific expansions as the dividing line only

TeoTwawki:
>  specific expansions as the dividing line only

I suggested that earlier today, when my lunch break hits today I'm trying something in a branch to see how it looks (splitting into separate tables)

PR #5980, opened 2019-05-07
[WIP] Murder settings that have a known retail default

isotor:
I have one last attempt in me today >.>

### DIFF
--- a/scripts/globals/settings.lua
+++ b/scripts/globals/settings.lua
@@ -113,16 +113,10 @@ AF2_FAME = 40; -- base fame for completing an AF2 quest
 AF3_FAME = 60; -- base fame for completing an AF3 quest
 OldSchoolG1 = false; -- Set to true to require farming Exoray Mold, Bombd Coal, and Ancient Papyrus drops instead of allowing key item method.
 OldSchoolG2 = false; -- Set true to require the NMs for "Atop the Highest Mountains" be dead to get KI like before SE changed it.
-FrigiciteDuration = 30; -- When OldSChoolG2 is enabled, this is the time (in seconds) you have from killing Boreal NMs to click the "???" target.
 
 -- SPELL SPECIFIC SETTINGS
-DIA_OVERWRITE = 1; --Set to 1 to allow Bio to overwrite same tier Dia.  Default is 1.
-BIO_OVERWRITE = 0; --Set to 1 to allow Dia to overwrite same tier Bio.  Default is 0.
-STONESKIN_CAP = 350; -- soft cap for hp absorbed by stoneskin
-BLINK_SHADOWS = 2;   -- number of shadows supplied by Blink spell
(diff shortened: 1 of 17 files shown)


TeoTwawki:
yer gonna make zynjec happy while annoying me

isotor:
This must be how interns at the United Nations feel 😄

TeoTwawki:
well ya know we could try..discussing things all the way before the work gets done instead of..After.

I merge 30 pr's close ONE and have to see some bullshit about how you've made to many and its time to chase you away by some assclown on discord, doesn't make me very happy here.

isotor:
I'm just trying to do what people want <.< the goalposts keep changing depending on who is awake, and the PRs are raised as "hey would you like this" and "okay how about this" not "yo merge my stuff".

tagban:
TBH we need more configuration options not less. It allows for more unique servers without having to modify the main source code, thus easing the ability to keep up with updates to the core code.

helixhamin:
@tagban The main issue is about who will manage it. We discussed this on the discord. It would be better to find a way to easily allow others to modify the code without dealing with tons of bloated settings toggles that make it more difficult for development. Ideas on how to do that are open.

tagban:
The easiest way to do that is literally push everything to Lua if you wanted to do that. :-\ But that's just as complex. If not more?

tagban:
Plus would contribute to slower loading I'd imagine.

TeoTwawki:
@isotor the goalpost did not change, there is no agreement on what is to be done overall, and going forward before that is not a good idea. if you leave it to zynjec you can kiss all settings goodbye.

@tagban I have a pipe dream of moving "old school" settings to a table in another file, and just load that into settings.lua. Have default behavior for the things those settings would do if they aren't present - old school ppl just insert their setting into script.

tagban:
> @isotor the goalpost did not change, there is no agreement on what is to be done overall, and going forward before that is not a good idea. if you leave it to zynjec you can kiss all settings goodbye.
> 
> @tagban I have a pipe dream of moving "old school" settings to a table in another file, and just load that into settings.lua. Have default behavior for the things those settings would do if they aren't present - old school ppl just insert their setting into script.

I mean, I'd be fine if they moved to SQL since it would NEVER get overwritten and could easily be backed up. But I'm lazy?

ibm2431:
Apologies for my limited understanding of status effects, but might it be possible to handle custom Dia/Bio behavior with just the SQL negative IDs?

isotor:
Yes it would, set them to overwrite 3 (always) instead of 1 (higher only) under a common `negative_id` and boom. I need to submit a PR to harmonize the negid anyway.

TeoTwawki:
previously the settings were removed and a contributor put them back. and correct me if I misunderstand but overwrite always would mean overwriting the lower tier as well. EFFECTOVERWRITE_EQUAL_HIGHER is a thing though.

isotor:
That is the ticket yes, sorry I have too many windows open...

PR #5981, opened 2019-05-07
Harmonize Dia and Bio negative_id

isotor:
Sorry @TeoTwawki! last one on this damnable pair.

Currently they are set to [`OVERWRITEEFFECT_HIGHER`](https://github.com/DarkstarProject/darkstar/blob/361ab1f03bcb008af3dc91bb34ae755df96afc94/src/map/status_effect.h#L37) meaning they can share the same `negative_id` and by default not overwrite each other. The individual scripts then override this default position depending on the global setting. 

Side bit: as @ibm2431 said in another convo, you could achieve the customization of allowing same-tier-overwrites by setting the `overwrite` to 3 in this sql once the `negative_id` is harmonized.


TeoTwawki:
Same tier of Bio and Dia are not equal for overwriting.

Bio III > Dia III > Bio II > Dia II > Bio > Dia 

Dia 1 will not overwrite Bio 1, but bio 1 will overwrite Dia.

using `EFFECTOVERWRITE_ALWAYS = 3, // always overwrite no matter` you can also have Dia 1 overwrite Dia 3 or Bio3, since that's **always**.

`EFFECTOVERWRITE_EQUAL_HIGHER = 0, // only overwrite if equal or higher (tier, power)` should achieve the same thing the setting was doing, I think. But with the same negative ID, are they behaving in retail fashion without the setting? As in, will Dia 1 mistakenly overwrite Bio 1 now?

PR #5983, opened 2019-05-07
Add Latents for Berserk for Conqueror 75 -> 119v3

neuromancerxi:
Adds Conqueror's Latent Berserk Buffs. Stats based on data from BGWiki: https://www.bg-wiki.com/bg/Conqueror

ATTP +x% (varies by tier see SQL Comments)
DEFP -x% (varies by tier see SQL Comments)
Crit Rate +x% (varies by tier see SQL Comments)
DA +5%

### DIFF
--- a/scripts/globals/abilities/berserk.lua
+++ b/scripts/globals/abilities/berserk.lua
@@ -14,7 +14,7 @@ function onAbilityCheck(player,target,ability)
 end
 
 function onUseAbility(player,target,ability)
-    player:addStatusEffect(dsp.effect.BERSERK,1,0,180)
+    player:addStatusEffect(dsp.effect.BERSERK,player:getMod(dsp.mod.BERSERK_EFFECT),0,180)
 
     return dsp.effect.BERSERK
 end
--- a/scripts/globals/abilities/berserk.lua
+++ b/scripts/globals/abilities/berserk.lua
@@ -14,7 +14,7 @@ function onAbilityCheck(player,target,ability)
 end
 
 function onUseAbility(player,target,ability)
-    player:addStatusEffect(dsp.effect.BERSERK,player:getMod(dsp.mod.BERSERK_EFFECT),0,180)
+    player:addStatusEffect(dsp.effect.BERSERK,25 + player:getMod(dsp.mod.BERSERK_EFFECT),0,180)
 
     return dsp.effect.BERSERK
 end
--- a/scripts/globals/effects/berserk.lua
+++ b/scripts/globals/effects/berserk.lua
@@ -7,16 +7,16 @@ require("scripts/globals/status")
 -----------------------------------
 
 function onEffectGain(target,effect)
-    target:addMod(dsp.mod.ATTP, 25 + effect:getPower())
-    target:addMod(dsp.mod.RATTP, 25 + effect:getPower())
-    target:addMod(dsp.mod.DEFP, -25 - effect:getPower())
+    target:addMod(dsp.mod.ATTP, effect:getPower())
+    target:addMod(dsp.mod.RATTP, effect:getPower())
+    target:addMod(dsp.mod.DEFP, - effect:getPower())
 end
 
 function onEffectTick(target,effect)
 end
 
 function onEffectLose(target,effect)
-    target:delMod(dsp.mod.ATTP, 25 + effect:getPower())
-    target:delMod(dsp.mod.RATTP, 25 + effect:getPower())
-    target:delMod(dsp.mod.DEFP, -25 - effect:getPower())
+    target:delMod(dsp.mod.ATTP, effect:getPower())
+    target:delMod(dsp.mod.RATTP, effect:getPower())
+    target:delMod(dsp.mod.DEFP, - effect:getPower())
 end
--- a/scripts/globals/effects/berserk.lua
+++ b/scripts/globals/effects/berserk.lua
@@ -7,16 +7,18 @@ require("scripts/globals/status")
 -----------------------------------
 
 function onEffectGain(target,effect)
-    target:addMod(dsp.mod.ATTP, effect:getPower())
-    target:addMod(dsp.mod.RATTP, effect:getPower())
-    target:addMod(dsp.mod.DEFP, - effect:getPower())
+    local power = effect:getPower()
+    target:addMod(dsp.mod.ATTP, power)
+    target:addMod(dsp.mod.RATTP, power)
+    target:addMod(dsp.mod.DEFP, -power)
 end
 
 function onEffectTick(target,effect)
 end
 
 function onEffectLose(target,effect)
-    target:delMod(dsp.mod.ATTP, effect:getPower())
-    target:delMod(dsp.mod.RATTP, effect:getPower())
-    target:delMod(dsp.mod.DEFP, - effect:getPower())
+    local power = effect:getPower()
+    target:delMod(dsp.mod.ATTP, power)
+    target:delMod(dsp.mod.RATTP, power)
+    target:delMod(dsp.mod.DEFP, -power)
 end
--- a/scripts/globals/effects/berserk.lua
+++ b/scripts/globals/effects/berserk.lua
@@ -7,16 +7,16 @@ require("scripts/globals/status")
 -----------------------------------
 
 function onEffectGain(target,effect)
-    target:addMod(dsp.mod.ATTP,25)
-    target:addMod(dsp.mod.RATTP, 25)
-    target:addMod(dsp.mod.DEFP,-25)
+    target:addMod(dsp.mod.ATTP, 25 + effect:getPower())
+    target:addMod(dsp.mod.RATTP, 25 + effect:getPower())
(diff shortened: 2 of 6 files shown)


TeoTwawki:
Reading BG the effects on berserk persist even when it is removed. A latent won't do that. Double attack isn't confirmed.

```
Attack to +33.5% (±86/256), Defense -33.5% (Upgraded from the original Berserk effect of +25%/-25%)
Effect on Attack persists even after Conqueror removal.
Critical Hit Rate +14%
Effect on Critical Hit Rate does not persist when Conqueror is removed.
May also give some amount (3~5%) of Double Attack. Verification Needed
```

TeoTwawki:
Reading BG the effects on berserk attack/defense persist even when it is removed. A latent won't do that. Double attack isn't confirmed. The crit is correct as a latent though.

```
Attack to +33.5% (±86/256), Defense -33.5% (Upgraded from the original Berserk effect of +25%/-25%)
Effect on Attack persists even after Conqueror removal.
Critical Hit Rate +14%
Effect on Critical Hit Rate does not persist when Conqueror is removed.
May also give some amount (3~5%) of Double Attack. Verification Needed
```

neuromancerxi:
The ATTP/DEFP buff could check for main hand item ID in abilities/berserk instead, which would allow it persist through un-equip since it would be applied at usage. If that's an acceptable method, I'll rework my PR.

Any recommendations on the DA stat? Omit entirely due to the verification needed tag, or go middle of the road?

TeoTwawki:
> could check for main hand item ID in 

_Nope._
Anytime you think you would check an ItemID you should make a modifier instead, and use that modifier in whatever script needs to adjust the value.

TeoTwawki:
Check for another source on the double attack, look for if there is a general consensus that it exists or if its just we don't know how much. If we just don't know how much, go with the low estimate (page says 3-5 so 3)

neuromancerxi:
> make a modifier instead
Can do. 

>Check for another source
DA/TA/QA is really rough to get a good sample size, but I'll pour over the usual places and see what I can find. Have a feeling outside of BGWiki, details will be scarce.

TeoTwawki:
github formatting tip: extra newlines after the lines with >

:wink:

sql/item_latents.sql
```diff
@@ -3234,4 +3234,57 @@ INSERT INTO `item_latents` VALUES(21661, 405, 3, 10, 0);      -- Rune Algol: Dra
...
 -- INSERT INTO `item_latents` VALUES(21667, 30, 10, ??, 0); -- Futhark Claymore: (D): Magic Accuracy+10
 -- INSERT INTO `item_latents` VALUES(21668, 25, 10, ??, 0); -- Peord Claymore: (D): Accuracy+10
--- INSERT INTO `item_latents` VALUES(21668, 30, 10, ??, 0); -- Peord Claymore: (D): Magic Accuracy+10
\ No newline at end of file
+-- INSERT INTO `item_latents` VALUES(21668, 30, 10, ??, 0); -- Peord Claymore: (D): Magic Accuracy+10
+
+-- -------------------------------------------------------
+-- Conqueror Enhances Berserk I - V
+-- -------------------------------------------------------
+INSERT INTO `item_latents` VALUES(18991,62,5,13,56);   -- Conqueror 75 Attack +5% if Berserk Active
```
helixhamin: BG Wiki states the following (For the lvl 75 version):

> Augments Berserk effect gives the following:
> Attack to +30% (±77/256), Defense -30% (Upgraded from the original Berserk effect of +25%/-25%)
> Effect on Attack persists even after Conqueror removal.
> Critical Hit Rate +5%
> Effect on Critical Hit Rate does not persist when Conqueror is removed.
> May also give some amount (3~5%) of Double Attack.

Your value says +5%

Reference: https://www.bg-wiki.com/bg/Conqueror_(Level_75)

You may need to recheck your numbers.

scripts/globals/abilities/berserk.lua
```diff
@@ -14,7 +14,7 @@ function onAbilityCheck(player,target,ability)
 end
 
 function onUseAbility(player,target,ability)
-    player:addStatusEffect(dsp.effect.BERSERK,1,0,180)
+    player:addStatusEffect(dsp.effect.BERSERK,player:getMod(dsp.mod.BERSERK_EFFECT),0,180)
```
TeoTwawki: Is there ever a time berserk potency is less than 25? would make more sense to set this 25+modifier than to repeatedly say 25 and the effect power in the effect script.

scripts/globals/effects/berserk.lua
```diff
@@ -7,16 +7,16 @@ require("scripts/globals/status")
 -----------------------------------
 
 function onEffectGain(target,effect)
-    target:addMod(dsp.mod.ATTP,25)
-    target:addMod(dsp.mod.RATTP, 25)
-    target:addMod(dsp.mod.DEFP,-25)
+    target:addMod(dsp.mod.ATTP, 25 + effect:getPower())
+    target:addMod(dsp.mod.RATTP, 25 + effect:getPower())
+    target:addMod(dsp.mod.DEFP, -25 - effect:getPower())
```
TeoTwawki: yeeeahh lets put the 25 up in the ability script and then you can just put effect:getpower() to a variable in this script and +/- the variable in all these lines.

sql/item_latents.sql
```diff
@@ -3234,4 +3234,47 @@ INSERT INTO `item_latents` VALUES(21661, 405, 3, 10, 0);      -- Rune Algol: Dra
...
+INSERT INTO `item_latents` VALUES(19948,165,14,13,56); -- Crit Rate +14% if Berserk Active
+INSERT INTO `item_latents` VALUES(19948,288,3,13,56);  -- Double Attack +3% if Berserk Active
+
+-- Conqueror 119
+INSERT INTO `item_latents` VALUES(20837,165,14,13,56); -- Crit Rate +14% if Berserk Active
+INSERT INTO `item_latents` VALUES(20837,288,3,13,56);  -- Double Attack +3% if Berserk Active
+
+-- Conqueror 119 AG
+INSERT INTO `item_latents` VALUES(20838,165,14,13,56); -- Crit Rate +14% if Berserk Active
+INSERT INTO `item_latents` VALUES(20838,288,3,13,56);  -- Double Attack +3% if Berserk Active
```
TeoTwawki: I take it you found enough to convincingly say the DA is real and it was just the amount that was in need of verification?

TeoTwawki: eh, let me rephrase: you are reasonably sure. good enough for Teo.

scripts/globals/abilities/berserk.lua
neuromancerxi: Yes, potentially a separate value for the defense penalty if AF2 feet are equipped, but I think that can be handled entirely by latent since they need to be equipped for it to take effect and its lost on un-equip.

scripts/globals/effects/berserk.lua
neuromancerxi: Will do.

sql/item_latents.sql
neuromancerxi: I'd love to say that I found more than the BGWiki entries, not so much. The "search" feature on the forums is... not so useful.

TeoTwawki: ...prolly less hassle to remove/change later if we're wrong than to keep it out and have someone possibly open new PR's trying to add it,

scripts/globals/effects/berserk.lua
```diff
@@ -7,16 +7,16 @@ require("scripts/globals/status")
 -----------------------------------
 
 function onEffectGain(target,effect)
-    target:addMod(dsp.mod.ATTP, 25 + effect:getPower())
-    target:addMod(dsp.mod.RATTP, 25 + effect:getPower())
-    target:addMod(dsp.mod.DEFP, -25 - effect:getPower())
+    target:addMod(dsp.mod.ATTP, effect:getPower())
```
TeoTwawki: what I meant by variable:
```diff
-    target:addMod(dsp.mod.ATTP, 25 + effect:getPower())
-    target:addMod(dsp.mod.RATTP, 25 + effect:getPower())
-    target:addMod(dsp.mod.DEFP, -25 - effect:getPower())
+    local power = effect:getPower()
+    target:addMod(dsp.mod.ATTP, power)
+    target:addMod(dsp.mod.RATTP, power)
+    target:addMod(dsp.mod.DEFP, -power)
```

TeoTwawki: That way `getPower()` isn't called 6 times in 2 functions. and yeah I just put a diff inside a diff.  (ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

helixhamin:
@TeoTwawki Are we sure about these berserk enhance values?
BG has a different value than ffxicyclopedia. BG states the attack+ starts at 30% on level 75 version of the weapon, but I notice this still has the 5%.

helixhamin:
Scratch that, it is attack +30% (Up from 25%, so 30-25=5).
Sorry for the unneeded ping due to failed maths...

PR #5986, opened 2019-05-08
Fix some values/logic in prelude songs

isotor:
1. Updated the minimum power based on retail testing. 
2. Up[dated potency caps per [BG Wiki](https://www.bg-wiki.com/bg/Prelude).
2. Fixed logic in the power adjustment - I don't know the formulae but surely whatever the inflection is it should be consistent within the formula? Currently a combined skill between 130-200 will result in a power below the base for Archer's Prelude.

The new params cap out at combined skill 320 for Hunter's Prelude (BRD54), and 815 for Archer's Prelude (BRD99 + extra gear/merits) and is consistent with the "Skill cap is < 900" claim on [BG Wiki](https://www.bg-wiki.com/bg/Archer%27s_Prelude).

**I can extend this PR (or make a new one) to cover other "buff" songs - just wanted to get a green light on the changes/ methodology first.**

-------

*1.* Sorry, I did this over Discord screen share for all songs with a friend on his newly-minted-99 BRD (power-leveled, 0 skill) and didn't think to take screenshots. The updated `local powers` were ascertained using /checkparam.

### DIFF
--- a/scripts/globals/spells/archers_prelude.lua
+++ b/scripts/globals/spells/archers_prelude.lua
@@ -1,6 +1,6 @@
 -----------------------------------------
 -- Spell: Archer's Prelude
--- Gives party members ranged attack accuracy
+-- Enhances ranged attack for party members within area of effect.
 -----------------------------------------
 require("scripts/globals/status")
 require("scripts/globals/msg")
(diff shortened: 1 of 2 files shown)


isotor:
Update: the logic (human) in a71404 is incorrect

> This is also assumed from Archer's Prelude that you need 600 combined skill to cap out at 30 ranged accuracy. 

it's pretty well known that lower level songs can cap out well below level 75, see [Army's Paeon](https://ffxiclopedia.fandom.com/wiki/Army%27s_Paeon).

PR #5992, opened 2019-05-08
[bcnm] Retable quest/mission IDs + some needed fixes I spotted

ibm2431:
I didn't intend to pick up the CMake commits, dunno how that happened. But I guess they're coming along too.


TeoTwawki:
meh its fine

PR #5994, opened 2019-05-09
add mod for fotia gear

KnowOne134:


src/map/ai/states/weaponskill_state.cpp
```diff
@@ -90,8 +91,15 @@ void CWeaponSkillState::SpendCost()
     }
     else
     {
-        tp = m_PEntity->health.tp;
-        m_PEntity->health.tp = 0;
+        if (dsprand::GetRandomNumber(100) < m_PEntity->getMod(Mod::WS_NO_DEPLETE))
```
TeoTwawki:
```cpp
        tp = m_PEntity->health.tp;

        if (m_PEntity->getMod(Mod::WS_NO_DEPLETE) < dsprand::GetRandomNumber(100))
        {
            m_PEntity->health.tp = 0;
        }
```
maybe, to avoid saying tp = whatevs twice. 

might have meant <= above, family nagging me to death while I wrote this

Issue #3561, opened 2016-11-27
SMN spririts casting flaws


dwn134:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**

**_Server Version_** (type `@revision` in game) **:**

**_Source Branch_** (master/stable) **:**

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
if you retreat a spirit and assualt it will begin again to cast another spell over and over again. Just seems to restart the timer to 0.
also I'm not sure if it should be casting immediately after being summoned I thought it had a cool down from 1st cast also.
and last thing is the Light Spirit doesnt cure, which would probably need some type of AI or HP checks to function correctly. I have had it cast haste on me once other than that it does banish dia and flash.

my server is currently down and i can get the ver and misc when its back up in an edit

dwn134:
ver 30161103_1

revision dec03e7567fe7d37016c44a9b5c2867ced2f8ffa

daftpunk1983:
I did some testing with this, And it seems the Thunder ele, and fire ele are the main ones on this, however the thunder ele will use an aga and a single target nuke, and a stun to cycle, where the fire will straight use aga and single target and not burn or others, thus not being a cycle.

wrenffxi:
re: "I'm not sure if it should be casting immediately after being summoned I thought it had a cool down from 1st cast also."

There should be a forced delay before a spirit will start casting spells.  It is calculated by [this formula](https://ffxiclopedia.fandom.com/wiki/Thunder_Spirit):
```
Casting Time (in seconds) = 48 + (max Summoning Skill - current Summoning Skill) / 3
If Spirit is cast on corresponding day: -3 seconds
If Spirit is cast on weak day: +3 seconds
If Spirit is cast during corresponding weather: -2 seconds
If Spirit is cast during weak weather: +2 seconds
If using Summoner's Spats: -5 seconds
If using Astral Flow: -5 seconds
Light Spirit in healing mode or buffing mode: 1/2 Casting Time
```

hargrave81:
[...]

PR #5999, opened 2019-05-09
Address Summoner Elemental Casting Time - and Light Spirit Idle

hargrave81:
This supports falling back to the existing "monster 24s casting mode" if too many players do not like the retail nature.

The petcontroller was modified to handle elementals and support controlling to some extent the actions speed/and or spells via lua script. Some modification to scope for the mob controller was necessary, though further clean up could change that. This was just V1.

#3561 SMN Spirit flaws issue

What changed? Elementals were casting based on the monster's default cooldown of about 24 seconds. They also "opened up" casting almost all the time on a "sic" command. Now elementals will cast based on SMN skill (base of 45s at skill cap, faster with higher than cap, slower with lower than cap). Weather and other modifiers were added as well.

Light Spirit now casts buffs on master and party members, as well as heals them as needed.

AI for light spirit is heavily contained within lua script onRoam. The default casting AI is used during battle for light spirit, it will randomly cure v.s. attack.

TODO: I have not had much time testing, so lots of testing, make sure **other pets** are not negatively impacted. Also put in additional code in lua script to read the gear smn skill bonus gear. I don't think the default helper does that. Fix the counter from being reset everytime the monster re-engages the monster.

Other Notes: 
- If there's going to be unique AI for each spirit, then the AI needs to be broken out into individual files.  Potentially this would be a nice upgrade to allow each spirit their own AI (healing v.s. non-healing or custom spirits)
- Clean up lua scripts to be more readable and unify the elemental (dsp.elements.WIND for example) and the database having a different number than other element Enums.

Reference https://github.com/DarkstarProject/darkstar/pull/5966 for other additional notes.


hargrave81:
@ibm2431 , @teschnei - I have cleaned up the code commits.  I also cleaned up the boolean default to false noted.  I updated the notes for the pull request as well.

~Chris

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,286 @@
...
+        return sms * - 1
+    end
+    return 0
+end;
+
+function getDayMod(avatar)
+    local element = getElement(avatar)
+    local ms = -3000    
+    local summoner = avatar:getMaster()    
+    if VanadielDayElement() == dsp.day.ICEDAY and element == dsp.subEffect.ICE_DAMAGE then
```
TeoTwawki: tables my dude. tables.
```lua
local myTableNameHere =
{
    [index here] = what value we want from selecting it,
    [index here] = what value we want from selecting it,
    [index here] = what value we want from selecting it
}

if myTable[input to get index here] then
```

TeoTwawki: pretty much anytime you have to type "else" that many times, you are working to hard

TeoTwawki: nvm, didn't realize this was your other PR re sent in, I expected the new commits to replace what was on the other one (github will auto update the pr if the branch's content changes, so force pushing to overwrite it works)

teschnei:
agh, all my comments are gone

it's better to force push over an existing branch with a PR open cause it'll keep the PR open but replace the commits with the new ones

teschnei:
@ibm2431 you wanna improve the scripts in here and let me do the core?

ibm2431:
@teschnei Yeah, I can help out!

(May not be able to start working on it today though; it's getting late for me.)

ibm2431:
https://github.com/hargrave81/darkstar/pull/5

helixhamin:
@hargrave81 
Please don't change default settings like that. The ver lock is set to 1 because version mismatch can cause problems. The server owner can change it themselves if they want. If they don't understand enough to change it, then what do you think will happen when version mismatch causes strange behavior.

version.info
```diff
@@ -6,7 +6,7 @@ CLIENT_VER: 30190328_2
 #0 - disabled (every version allowed)
 #1 - enabled - strict (only exact CLIENT_VER allowed, default)
 #2 - enabled - greater than or equal  (matching or greater than CLIENT_VER allowed)
-VER_LOCK: 1
+VER_LOCK: 2
```
helixhamin: Should not be changed...

hargrave81:
I'm not at my pc, I'll revert this.  Forgot this would change the pull request on the fly.  Had a client that couldn't connect because POL update is newer.  I'll change my deploy scripts to modify that during deploy.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

teschnei:
I pushed the changes (minus the version.info commit) manually to the smn-spirits branch, so this is on our repo now. @ibm2431 can you PR to that branch instead? also, I fixed getFamily, but I have to wait for someone to merge it before I can put it on the branch

PR #6000, opened 2019-05-09
cleanup mobskills using mobs HP as power

TeoTwawki:
Also removing anecdotal unsourced bull about thf job.

***CITATION NEEDED*** if you want to put that back.


TeoTwawki:
Pending: rename that function so this doesn't happen again

TeoTwawki:
Done.

TeoTwawki:
I also killed the tabs while I was there.

PR #6004, opened 2019-05-12
Smn spirits

ibm2431:
Not sure how onMobFight is supposed to be choosing delay. I'm currently overwriting your mob cool - that function is a replacement for the:
```
    if (pet:getSystem() == 11) then         -- elemental spirit
        delay = getTimeCost(pet) -- base 45s and adjusted based on skill 
```
block that was overwriting your mob cool anyway.

I'm just throwing the script work yesterday at this, then we can get the two to play nice.


teschnei:
oh yeah, I guess just remove the getMobMod part for onMobFight since it didn't do anything anyways

teschnei:
side note: are all the spirits really called "avatar" in the db? is that why the pets/avatar.lua file is named so?

ibm2431:
https://github.com/DarkstarProject/darkstar/blob/5c520560358d42340b1af528484496a324f58422/src/map/entities/petentity.cpp#L66-L72

https://github.com/DarkstarProject/darkstar/blob/5c520560358d42340b1af528484496a324f58422/src/map/entities/petentity.h#L29-L31

teschnei:
ah, makes sense, just "avatar" is a bad name for it I guess

teschnei:
actually nevermind, we can totally do GetName like PETTYPE_TRUST does which would hopefully get us the specific avatar's name

ibm2431:
Is MOBMOD_MAGIC_COOL an internal counter we're using that prevents spellcasting? Looking at this onMobFight again, I think it's being constantly overwritten with the (base - SMN mods) time every round, preventing anything from ever being cast during combat.

TeoTwawki:
> Is MOBMOD_MAGIC_COOL an internal counter we're using to prevent spellcasting? Looking at this onMobFight again, I think it's being constantly overwritten with the (base - SMN mods) time every round, preventing anything from ever being cast during combat.

it was _supposed to be_ the base cooldown time between casts set up during init

teschnei:
When checking whether to cast a spell or not, it's the number that is added
to the time of the last cast and then the result of that is compared with
the current time. The mobmod itself doesn't contain a timestamp so it
should be okay to overwrite it constantly

A thought though: if you change your gear to modify the cooldown, will it
take effect for the current cast or only the next one?

ibm2431:
You'd think it'd function like Fast Cast in that it only calculates it once a spell goes off.

Wiki claims: ["The spirit timer is dynamic and can change if any of the factors affecting it change at any time."](https://ffxiclopedia.fandom.com/wiki/Spirit_Guide_by_Spira)

But that's wiki.

Might need some testing come free week (if SCH spells work for the weather mod).

hargrave81:
So I'm working through the testing, running into issues getting the spirits to work as expected in the previous code.  I'm going to dig deeper, but it looks like

https://github.com/ibm2431/darkstar/blob/a49f27150525b8f92624e309f52f22da096e516d/scripts/globals/pets/avatar.lua#L52

It is using the classic onMobRoam, rather than the altered PetRoam that tracks when the last cast was. (ms parameter)

hargrave81:
Another thing to note.  To get things to "kinda" work I changed the local spirits array to something like this:
[102] = dsp.pet.spirit.create(dsp.magic.element.FIRE),
and included a require to the spirit lua.
However, the delay (while calculated what appears to be accurate) does not actually affect the recast speed correctly.
**CLuaBaseEntity::setMobMod Expected type mob (17278720) but its a (8)**

The Pet Controller no longer has a roam function so the onMobRoam will never be triggered (as the pet controller does not currently execute the base functionality of that overloaded event)  So Light Spirit does nothing.
refer to void CPetController::DoRoamTick(time_point tick)

ibm2431:
So seeing this pop up in my GitHub notifications, I realized I completely spaced testing spirits during this free period to see if the cast timer functions "dynamically" or if it functions like Fast Cast.

I won't be able to squeeze it in on this last day of the free period, so if someone would like to confirm how it functions, better get the testing in while you still have retail access~!

PR #6006, opened 2019-05-12
Create Masan

AikoSachi:
Adding new page with era gil amounts.


helixhamin:
Isn't there already gil drop values in the mob drops table? Why would we need to add a specific lua for him?

KnowOne134:
You could add it to pool_mods

megberry:
BTW, we'll need a Masan.lua eventually to add its melee attack additional effects, so NBD if one's created for it.

wrenffxi:
We'll need a Masan.lua eventually, to add its melee attack additional effects, so NBD if one's created for it.

PR #6008, opened 2019-05-13
Client Update 30190430-0

wrenffxi:


helixhamin:
I bet they just think to themselves 

> Let's mess with DSP just to annoy them

PR #6010, opened 2019-05-15
missing lady bug Tp moves

KnowOne134:
the ids are already stored and open in the mobskills and skill_list

### DIFF
--- a/scripts/globals/mobskills/nepenthean_hum.lua
+++ b/scripts/globals/mobskills/nepenthean_hum.lua
@@ -18,7 +18,6 @@ function onMobSkillCheck(target,mob,skill)
 end
 
 function onMobWeaponSkill(target, mob, skill)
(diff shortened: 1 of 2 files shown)


scripts/globals/mobskills/nepenthean_hum.lua
```diff
@@ -0,0 +1,26 @@
...
+function onMobSkillCheck(target,mob,skill)
+    if VanadielHour() >= 6 and VanadielHour() <= 18 then
+        return 0
+    else
+        return 1
+    end
+end
+
+function onMobWeaponSkill(target, mob, skill)
+    local message = dsp.msg.basic.SKILL_MISS
```
TeoTwawki: This isn't needed, you are already setting the message to whatever MobStatusEffectMove() returns below

PR #6022, opened 2019-05-20
Set default for mob level capping to uncapped

ibm2431:
So by default they'll be the levels we've been defining for them in mob groups!


TeoTwawki:
these are supposed to be enabled/disabled by way of a 3rd setting, so they don't need to be default zero here

TeoTwawki:
got pulled away by work again:
at least I thought that was how that worked and the commenting plus placement in the settings lua didn't make it obvious. I'll go look for the commit it came in with to be sure.

TeoTwawki:
ok yeah you are correct and my memory is faulty, merging, and in the future I'd like to see it tied to the expansion content restrictions like I thought it was.

PR #6026, opened 2019-05-23
remove or assign some groupids that related to 0 rows in mob_spawn_points

wrenffxi:
I identified a bunch of mob groups that did not match any mobs in mob_spawn_points.  For each,
* if it should be assigned to mobs in mob_spawn_points, I did so.
* if not, I removed it.  These were mostly redundant groups, or ones that had been allotted for trusts or cutscene NPCs.

I also remove one redundant mob pool (Razorback).


sql/mob_groups.sql
```diff
@@ -1044,7 +921,6 @@ INSERT INTO `mob_groups` VALUES (996,2412,37,0,128,4414,0,0,75,80,0);  -- Lights
 INSERT INTO `mob_groups` VALUES (997,2413,37,0,128,4416,7000,1000,75,80,0);  -- Light_Elemental      OK
 INSERT INTO `mob_groups` VALUES (998,2691,37,0,128,4411,9000,500,75,80,0);  -- Moblin_Dustman          OK
 INSERT INTO `mob_groups` VALUES (999,636,37,0,128,0,15000,0,75,80,0);  -- Mystic_Avatar (carbuncle)
-INSERT INTO `mob_groups` VALUES (1000,2836,37,0,128,0,0,0,75,80,0);  -- Nguruvilu
```
TeoTwawki: was Nguruvilu a duplicate or what?

wrenffxi: Dunno, but Nguruvilu shouldn't have a group in zone 37 (Temenos).  It does already have a group in Abyssea - La Theine.
https://ffxiclopedia.fandom.com/wiki/Nguruvilu

TeoTwawki: That's "yes" then.

PR #6031, opened 2019-05-25
"Say It With a Handbag" quest line

kaincenteno:
Fixed doors in Windurst Waters [S]
Added Events for NPCs
Create missing NPC Hampu-Kampu
Made a comment in item-latents as latents for crafting now implemented on latent_effect.h (only is missing the case in latent_effect.h that matches leathercraft)
Added note in quests.lua to show that quest can be completed
Added new special message on Windurt Waters [S] IDs.lua

Would appreciate feedback as this is my first pull request quest to improve notes on the next one.
Thank you 😸


scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -6,18 +6,41 @@
...
 require("scripts/globals/crafting");
+require("scripts/globals/quests");
 local ID = require("scripts/zones/Southern_San_dOria/IDs");
 -----------------------------------
 
 function onTrade(player,npc,trade)
+    local sayItWithAHandbagCS = player:getVar("sayItWithAHandbagCS")
+
+    if (player:hasKeyItem(dsp.ki.TORN_PATCHES_OF_LEATHER) and sayItWithAHandbagCS == 2 and trade:getItemCount() == 3) then
+        if (trade:hasItemQty(2012,1) and trade:hasItemQty(850,1) and trade:hasItemQty(816,1)) then
```
ibm2431: [`npcutil.tradeHas` can shorten this](https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/npc_util.lua#L347-L360)

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -6,18 +6,41 @@
...
 require("scripts/globals/status");
 require("scripts/globals/crafting");
+require("scripts/globals/quests");
 local ID = require("scripts/zones/Southern_San_dOria/IDs");
 -----------------------------------
 
 function onTrade(player,npc,trade)
+    local sayItWithAHandbagCS = player:getVar("sayItWithAHandbagCS")
+
+    if (player:hasKeyItem(dsp.ki.TORN_PATCHES_OF_LEATHER) and sayItWithAHandbagCS == 2 and trade:getItemCount() == 3) then
```
ibm2431: Drop all outer parens per [style guide](https://github.com/DarkstarProject/darkstar/blob/master/CONTRIBUTING.md)~!

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -6,18 +6,41 @@
...
     local SkillCap = getCraftSkillCap(player, dsp.skill.LEATHERCRAFT);
     local SkillLevel = player:getSkillLevel(dsp.skill.LEATHERCRAFT);
 
-    if (guildMember == 1) then
+    if (sayItWithAHandbag == QUEST_COMPLETED and sayItWithAHandbagCS == 4 and player:hasItem(19110)) then
+        player:startEvent(914)
+        player:setVar("sayItWithAHandbagCS", 5)
+    elseif (player:hasKeyItem(dsp.ki.REPAIRED_HANDBAG) and sayItWithAHandbagCS == 4) then
+        player:startEvent(913)
+    elseif (sayItWithAHandbagCS == 3 and player:needToZone() == false) then
```
ibm2431: If something evaluates to `true` or `false`, you don't need to do a comparison, just use the evaluated value:
```
elseif sayItWithAHandbagCS == 3 and not player:needToZone()
```

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -6,18 +6,41 @@
...
+        player:startEvent(912)
+    elseif (sayItWithAHandbagCS == 3 and player:needToZone() == true) then
+        player:startEvent(911)
+    elseif (player:hasKeyItem(dsp.ki.TORN_PATCHES_OF_LEATHER) and sayItWithAHandbagCS == 2) then
```
ibm2431: Is the key item check needed if you're already using a stage variable?

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
KnowOne134: Dont need outter ()

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
KnowOne134: Can we make this work with npc_util

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -29,11 +52,22 @@ function onTrigger(player,npc)
...
 function onEventFinish(player,csid,option)
-    if (csid == 651 and option == 1) then
+    if (csid == 912) then
+        player:giveKeyItem(dsp.ki.REPAIRED_HANDBAG)
+        player:setVar("sayItWithAHandbagCS", 4)
+    elseif (csid == 910) then
+        player:delKeyItem(dsp.ki.TORN_PATCHES_OF_LEATHER)
+        player:setVar("sayItWithAHandbagCS", 3)
+        player:needToZone(true)
+        player:tradeComplete()
```
KnowOne134: tradeConfirm and then it will only take items asked for

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -6,18 +6,41 @@
...
 
 function onTrigger(player,npc)
+    local sayItWithAHandbag = player:getQuestStatus(CRYSTAL_WAR, dsp.quest.id.crystalWar.SAY_IT_WITH_A_HANDBAG)
+    local sayItWithAHandbagCS = player:getVar("sayItWithAHandbagCS")
     local guildMember = isGuildMember(player,7);
     local SkillCap = getCraftSkillCap(player, dsp.skill.LEATHERCRAFT);
     local SkillLevel = player:getSkillLevel(dsp.skill.LEATHERCRAFT);
 
-    if (guildMember == 1) then
+    if (sayItWithAHandbag == QUEST_COMPLETED and sayItWithAHandbagCS == 4 and player:hasItem(19110)) then
```
KnowOne134: All these have no reason for outter ()

scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua
```diff
@@ -0,0 +1,33 @@
...
+end
+
+function onEventUpdate(player,csid,option)
+end
+
+function onEventFinish(player,csid,option)
+    if (csid == 171 and option == 1) then
+        player:setVar("sayItWithAHandbagCS", 1)
+    end
+    if (csid == 171 and option == 0) then
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua]: `elseif`, or combine with an outer csid check:
```
if csid == 171 then
    if option == 1 then
        player:setVar("sayItWithAHandbagCS", 1)
    else
        -- bug: quest music
    end
end
```

Although is checking for option 0 even required? You're not doing anything with it; is it supposed to _do_ something and you're just noting the bug? If choosing option 0 is supposed to have zero effect on anything, you can just ignore it.

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,54 @@
+-----------------------------------
+-- Area: Windurst Waters (S)
+--  NPC: Hampu-Kampu
+-- Type: Quest NPC
+-----------------------------------
+local ID = require("scripts/zones/Windurst_Waters_[S]/IDs")
+require("scripts/globals/quests");
+require("scripts/globals/keyitems");
+
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: I predict a certain someone will tell you to add a comment of dashes here:
`-----------------------------------`

scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua
```diff
@@ -0,0 +1,33 @@
+-----------------------------------
+-- Area: Windurst Waters (S)
+--  NPC: Door Acolyte hostel
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua]: May consider adding a !pos here

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,54 @@
+-----------------------------------
+-- Area: Windurst Waters (S)
+--  NPC: Hampu-Kampu
+-- Type: Quest NPC
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: !pos here too~

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,54 @@
...
+        player:startEvent(169)
+        player:setVar("sayItWithAHandbagCS",0)
+    end
+end
+
+function onEventUpdate(player,csid,option)
+end
+
+function onEventFinish(player,csid,option)
+    if (csid == 174) then -- Option doesn't matter as NPC will take key item if yes or no
```
Comment (KnowOne134) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: Npc_util has a quest complete function. Changes everything here to 1 line

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,54 @@
...
+        if (player:getFreeSlotsCount() < 1) then
+            player:messageSpecial(ID.text.ITEM_CANNOT_BE_OBTAINED,19110);
+        else
+            player:delKeyItem(dsp.ki.REPAIRED_HANDBAG)
+            player:addItem(19110)
+            player:completeQuest(CRYSTAL_WAR, dsp.quest.id.crystalWar.SAY_IT_WITH_A_HANDBAG)
+            player:messageSpecial(ID.text.ITEM_OBTAINED, 19110)
+        end
+    elseif (csid == 172) then
+            player:giveKeyItem(dsp.ki.TORN_PATCHES_OF_LEATHER)
```
Comment (KnowOne134) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: Silent give of keyitem? If not npc_util

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,54 @@
...
+function onTrade(player,npc,trade)
+end
+
+function onTrigger(player,npc)
+    local sayItWithAHandbag = player:getQuestStatus(CRYSTAL_WAR, dsp.quest.id.crystalWar.SAY_IT_WITH_A_HANDBAG)
+    local sayItWithAHandbagCS = player:getVar("sayItWithAHandbagCS")
+
+    if (sayItWithAHandbag == QUEST_COMPLETED) then
+        player:startEvent(175)
+    elseif (player:hasKeyItem(dsp.ki.REPAIRED_HANDBAG) and sayItWithAHandbagCS == 4) then
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: Again, since you're using a stage variable, you don't need to check the key item~

I think you can cut out all your KI checks. (Make sure you still give/remove the KI from the player though, even if you're not using them. Just don't need to check for them.)

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,54 @@
...
+        player:startEvent(174)
+        print("repaired bag and 4")
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: Debug print

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,54 @@
...
+    elseif (player:hasKeyItem(dsp.ki.TORN_PATCHES_OF_LEATHER) or sayItWithAHandbagCS == 3) then
+        player:startEvent(173)
+        print(player:hasKeyItem(dsp.ki.REPAIRED_HANDBAG))
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: Debug print here too

sql/npc_list.sql
```diff
@@ -10258,7 +10258,7 @@ INSERT INTO `npc_list` VALUES (17162680,'blank','     ',192,-50.000,-5.349,212.5
 INSERT INTO `npc_list` VALUES (17162681,'Myron_CA','Myron, C.A.',223,-13.534,-4.999,119.040,14,40,40,0,0,0,2,27,0x01000F0196103920963096409650006000700000,32,'WOTG',1);
 INSERT INTO `npc_list` VALUES (17162682,'Wenonah_CA','Wenonah, C.A.',194,-2.175,-1.000,10.184,8,40,40,0,0,0,0,27,0x0100070296103920963096409650006000700000,32,'WOTG',1);
 INSERT INTO `npc_list` VALUES (17162683,'Hampu-Kampu','Hampu-Kampu',245,-115.597,-1.000,-158.703,1,40,40,0,0,0,0,27,0x01000C0577107B20693069406950006000700000,32,'WOTG',1);
-INSERT INTO `npc_list` VALUES (17162684,'Door_Acolyte_hostel','Door:Acolyte hostel',0,124.000,-3.000,222.215,32769,40,40,0,0,0,0,3,0x0000340000000000000000000000000000000000,0,'WOTG',0);
+INSERT INTO `npc_list` VALUES (17162684,'Door_Acolyte_hostel_up','Door:Acolyte hostel',0,146.619,-8.525,242.874,32769,40,40,0,0,0,0,3,0x0000340000000000000000000000000000000000,0,'WOTG',0);
```
ibm2431: I assume these doors were mixed up at some point, explaining the swap.

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: Indent one too many.

Does the quest not advance at all after 172?

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,54 @@
...
+    end
+end
+
+function onEventUpdate(player,csid,option)
+end
+
+function onEventFinish(player,csid,option)
+    if (csid == 174) then -- Option doesn't matter as NPC will take key item if yes or no
+        if (player:getFreeSlotsCount() < 1) then
+            player:messageSpecial(ID.text.ITEM_CANNOT_BE_OBTAINED,19110);
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: Semi-colon at end of line

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,54 @@
...
+        else
+            player:delKeyItem(dsp.ki.REPAIRED_HANDBAG)
+            player:addItem(19110)
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: Could you add a comment saying what item ID 19110 is?

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,54 @@
...
+            player:completeQuest(CRYSTAL_WAR, dsp.quest.id.crystalWar.SAY_IT_WITH_A_HANDBAG)
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: You can use [npcUtil.completeQuest](https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/npc_util.lua#L268-L285) to reduce this down. The first thing it tries to do is attempt to give items, and if it can't it'll return false. So you can just check against it; if the player can receive the item, well, you've completed the quest. Make sure you only delete the KI if it returned true.

Also, you forgot to clear out your stage var~!

```
if npcUtil.completeQuest(player, CRYSTAL_WAR, dsp.quest.id.crystalWar.SAY_IT_WITH_A_HANDBAG, {item=19110}) then
    player:delKeyItem(dsp.ki.REPAIRED_HANDBAG)
    player:setVar("sayItWithAHandbagCS",0)
end
```

(`npcUtil.completeQuest` can take a var in its table too to clear out, but the line might get a little long)

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,54 @@
...
+    elseif (player:hasKeyItem(dsp.ki.TORN_PATCHES_OF_LEATHER) or sayItWithAHandbagCS == 3) then
+        player:startEvent(173)
+        print(player:hasKeyItem(dsp.ki.REPAIRED_HANDBAG))
+    elseif (sayItWithAHandbag == QUEST_ACCEPTED and sayItWithAHandbagCS == 1) then
+        player:startEvent(172)
+    elseif (sayItWithAHandbag == QUEST_ACCEPTED) then
+        player:startEvent(170)
+    elseif (sayItWithAHandbag == QUEST_AVAILABLE) then
+        player:startEvent(169)
+        player:setVar("sayItWithAHandbagCS",0)
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: You don't need to _initialize_ player vars at zero. Only set them to zero to delete them when you're done.

(And make sure not to do any setting in onTriggers, only eventFinishes!)

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -6,18 +6,41 @@
...
+        player:startEvent(914)
+        player:setVar("sayItWithAHandbagCS", 5)
+    elseif (player:hasKeyItem(dsp.ki.REPAIRED_HANDBAG) and sayItWithAHandbagCS == 4) then
+        player:startEvent(913)
+    elseif (sayItWithAHandbagCS == 3 and player:needToZone() == false) then
+        player:startEvent(912)
+    elseif (sayItWithAHandbagCS == 3 and player:needToZone() == true) then
```
ibm2431: Could also:
```
if sayItWithAHandbagCS == 3 then
    if player:needToZone() then
        player:startEvent(911)
    else
        player:startEvent(912)
    end
end
```

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -6,18 +6,41 @@
...
+    local sayItWithAHandbag = player:getQuestStatus(CRYSTAL_WAR, dsp.quest.id.crystalWar.SAY_IT_WITH_A_HANDBAG)
+    local sayItWithAHandbagCS = player:getVar("sayItWithAHandbagCS")
     local guildMember = isGuildMember(player,7);
     local SkillCap = getCraftSkillCap(player, dsp.skill.LEATHERCRAFT);
     local SkillLevel = player:getSkillLevel(dsp.skill.LEATHERCRAFT);
 
-    if (guildMember == 1) then
+    if (sayItWithAHandbag == QUEST_COMPLETED and sayItWithAHandbagCS == 4 and player:hasItem(19110)) then
+        player:startEvent(914)
+        player:setVar("sayItWithAHandbagCS", 5)
```
ibm2431: What's this 5 being used for?

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
ibm2431: `if sayItWithAHandbag == QUEST_COMPLETED then` should be sufficient by itself~

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -29,11 +52,22 @@ function onTrigger(player,npc)
 end;
 
 function onEventUpdate(player,csid,option)
+
 end;
 
 function onEventFinish(player,csid,option)
-    if (csid == 651 and option == 1) then
+    if (csid == 912) then
+        player:giveKeyItem(dsp.ki.REPAIRED_HANDBAG)
```
ibm2431: `player:giveKeyItem` doesn't exist, it's either `player:addKeyItem` (if using the [raw core function](https://github.com/DarkstarProject/darkstar/blob/master/src/map/lua/lua_baseentity.cpp#L6460-L6467)) or `npcUtil.giveKeyItem(player, dsp.ki.REPAIRED_HANDBAG)` (if using [npcUtil function](https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/npc_util.lua#L235-L244)). Core's version just gives it silently. npcUtil's version gives it _and_ displays the obtained message.

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: @KnowOne134  no () in all statements that are if or elif? (sorry trying to get clarification :)

KnowOne134:
Nope dont need the (if.....) thens

scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua
Comment (kaincenteno) [reply scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua]: @ibm2431 Well the issue here is that there' 2 pop up message for choices.
The first one if you choose Yes, it will lead to the second one, which if you choose No then the bug will show up.
If you choose No from the beginning then the CS will end without having that bug with the quest music still ongoing after the CS got declined. This is why I put it on 2 different if statements

sql/npc_list.sql
kaincenteno: Yeah actually those door were named exactly the same 😹 and one of the scripts was MIA 🙀

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
Comment (kaincenteno) [reply scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: @ibm2431 i used the number per the other quests but I think it also works if i use the actual name. Should I just leave with number (as prefered by other scripts) and add a comment of item name? 😸

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: O M G🙀! there's documentation 👍 Totally awesome, fixing now 😸

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: changed to player:confirmTrade()
Thanks for the back and forth in discord, scripting is still new to me

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: removed the ```== true```, but left the ```== false``` because it needs to check if false to display the next CS.

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: @ibm2431  The used is so that CS is only played once, as Kipopo is supposed to go back to her standard dialog ```player:startEvent(651) -- Standard Dialogue``` once quest is done as ```player:startEvent(914)``` is an optional CS not part of the questline.

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: @ibm2431  this check is just like mentioned on ```What's this 5 being used for?``` Checking if CS == 4 to play this one, also to play this cutscene the item needs to be on the user's inventory, otherwise it won't play. I'm not sure how accurate that is to retail but it seems plausible as Kipopo refers to the item in the cutscene 😿

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: It's needed otherwise it will jump order.
CS == 1 is attained at the door acolyte hostel
then the TORN_PATCHES OF LEATHER are obtained from Hampu Kampu.

for CS == 2  I removed the keyitem, you are right is not needed for scenario.

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: will try to remember this one, you are right as they are evaluating the same CS 😸

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
Comment (kaincenteno) [reply scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: @ibm2431  OMG I am loving npcUtil ❤❤❤
Unfortunately I cannot set the Var to  0 cause it has an optional CS afterwards
I guess I can set it to 0 but then I would still have to change it to 5, but wouldn't that be sort of confusing for reading? If it didn't have an after cutscene I would totally do it to make it simpler. What do you think?

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
Comment (kaincenteno) [reply scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: thank you. 🙏

 OMG, can't believe I put that inside the onTrigger. Will make sure to spot check better 🙀

scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua
Comment (kaincenteno) [reply scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua]: combined with outer csid check. Thank you 😸

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
Comment (kaincenteno) [reply scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: Changed var to 0 as using a different one to track Bonus CS

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: another var handles Bonus CS. No longer using 5 anymore. Thank you for your feedback 😸

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: changed to ```sayItWithAHandbag == QUEST_COMPLETED and sayItWithAHandbagBonus CS == 1 and player:hasItem(19110)``` (See comment above for comment regarding player:hasItem)

kaincenteno:
Submitted new commit with fixes with recommendations from @ibm2431  and @KnowOne134 😸

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -2,38 +2,77 @@
...
+require("scripts/globals/quests")
+require("scripts/globals/npc_util")
+local ID = require("scripts/zones/Southern_San_dOria/IDs")
 -----------------------------------
 
 function onTrade(player,npc,trade)
+    local sayItWithAHandbagCS = player:getVar("sayItWithAHandbagCS")
+
+    if player:hasKeyItem(dsp.ki.TORN_PATCHES_OF_LEATHER) and sayItWithAHandbagCS == 2 then
+        if npcUtil.tradeHasExactly(trade, {2012, 850, 816}) then
```
KnowOne134: No need to have a separate if here can make it all function

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -2,38 +2,77 @@
...
         else
-            player:startEvent(651,SkillCap,SkillLevel,1,239,player:getGil(),7128,0,0);
+            player:startEvent(912)
+        end
+    elseif sayItWithAHandbagCS == 2 then
+        player:startEvent(909)
+    elseif player:hasKeyItem(dsp.ki.TORN_PATCHES_OF_LEATHER) and sayItWithAHandbagCS == 1 then
+        player:startEvent(908)
+    elseif guildMember == 1 then
+        if player:hasStatusEffect(dsp.effect.LEATHERCRAFT_IMAGERY) == false then
```
KnowOne134: If not player:hasstatus....
Remove the == false

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
KnowOne134: All 1 line i mean

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -24,13 +24,13 @@ end;
 function onTrigger(player,npc)
     local sayItWithAHandbag = player:getQuestStatus(CRYSTAL_WAR, dsp.quest.id.crystalWar.SAY_IT_WITH_A_HANDBAG)
     local sayItWithAHandbagCS = player:getVar("sayItWithAHandbagCS")
+    local sayItWithAHandbagBonusCS = player:getVar("sayItWithAHandbagBonusCS")
     local guildMember = isGuildMember(player,7)
     local SkillCap = getCraftSkillCap(player, dsp.skill.LEATHERCRAFT)
     local SkillLevel = player:getSkillLevel(dsp.skill.LEATHERCRAFT)
 
-    if sayItWithAHandbag == QUEST_COMPLETED and sayItWithAHandbagCS == 4 and player:hasItem(19110) then
+    if sayItWithAHandbag == QUEST_COMPLETED and player:hasItem(19110) and sayItWithAHandbagBonusCS == 1 then
```
ibm2431: Can drop the `player:hasItem(19110)`. They must have been given it for the quest to be completed, _and_ you want them to be able to get the optional dialogue (more importantly: clear out their SQL var) even if they've tossed the item!

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
```diff
@@ -0,0 +1,52 @@
...
+    if csid == 174 then -- Option doesn't matter as NPC will take key item if yes or no
+        if npcUtil.completeQuest(player, CRYSTAL_WAR, dsp.quest.id.crystalWar.SAY_IT_WITH_A_HANDBAG, {
+            item = 19110, -- Trainee's Needle
+            var = "sayItWithAHandbagCS"
+        }) then
+            player:delKeyItem(dsp.ki.REPAIRED_HANDBAG)
+            player:setVar("sayItWithAHandbagBonusCS", 1)
+        end
+    elseif csid == 172 then
+            npcUtil.giveKeyItem(player, dsp.ki.TORN_PATCHES_OF_LEATHER)
```
Comment (ibm2431) [inline scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: One indent too many

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
TeoTwawki: This is for demonstration purposes:
```lua
    if 
        npcUtil.tradeHasExactly(trade, {2012, 850, 816})
        and player:hasKeyItem(dsp.ki.TORN_PATCHES_OF_LEATHER) 
        and sayItWithAHandbagCS == 2 
    then
```
to avoid double `if` and a wide line.

This one really isn't that wide I guess, but wanted to show you that if need be you can break them up this way and it still works.

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: @KnowOne134 I dunno if that will break imagery, I haven't touched imagery  😿️ dunno why it shows that as a change.

KnowOne134: Can change old code to new style as we edit on is all

kaincenteno: Will do my best, will make changes and test 😸️

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: Thank you for noticing this. I changed it to two lines as it was going over 120 characters. 😸

scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua
Comment (kaincenteno) [reply scripts/zones/Windurst_Waters_[S]/npcs/Hampu-Kampu.lua]: omg that was so  bad of me, thank you for noticing 😸

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
kaincenteno: I didn't think about that scenario.  Thanks for pointing it out. Will be removing it so player can continue the quest even tho they have tossed the item to clear the var. this has been already added onEventFinish for csid == 914.

Thanks for your feedback 😸

kaincenteno:
Made changes. Hopefully this time is okay @ibm2431  and @KnowOne134 . Made sure to test it on dsp before committing the files 😸

scripts/zones/Southern_San_dOria/npcs/Kipopo.lua
```diff
@@ -2,38 +2,78 @@
...
-local ID = require("scripts/zones/Southern_San_dOria/IDs");
+require("scripts/globals/status")
+require("scripts/globals/crafting")
+require("scripts/globals/quests")
+require("scripts/globals/npc_util")
+local ID = require("scripts/zones/Southern_San_dOria/IDs")
 -----------------------------------
 
 function onTrade(player,npc,trade)
+    local sayItWithAHandbagCS = player:getVar("sayItWithAHandbagCS")
```
KnowOne134: Just noticing. No real reason to make a local unless you are using it more than 1 time per function. Could just make it part of the if statement

scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua
```diff
@@ -0,0 +1,36 @@
...
+require("scripts/globals/missions")
+require("scripts/globals/settings")
+require("scripts/globals/quests")
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end
+
+function onTrigger(player,npc)
+    local sayItWithAHandbag = player:getQuestStatus(CRYSTAL_WAR, dsp.quest.id.crystalWar.SAY_IT_WITH_A_HANDBAG)
```
Comment (KnowOne134) [inline scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua]: Same here

Comment (ibm2431) [reply scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua]: My guess here is because the tabled name is long, so the assignment is to make the line for the if check shorter.

Comment (kaincenteno) [reply scripts/zones/Windurst_Waters_[S]/npcs/Door_Acolyte_hostel_up.lua]: @ibm2431   that is correct, but I can remove it. I'll go through the script and make changes to the guild locals too :)

PR #6032, opened 2019-05-26
Bcnmcrap

whasf:
Merge branch into master
This will likely break current BCNMs that aren't in the branch.


whasf:
branch needs to be cleaned of debug print lines and such

whasf:
Cleaned up BCNM, lets try again

scripts/globals/battlefield.lua
```diff
@@ -0,0 +1,250 @@
...
+                                end
+                                break
+                            end
+                            players[1]:addTreasure(entry.itemid, npc)
+                        end
+                        break
+                    end
+                end
+            else
+                printf("fuckin loot groups")
```
TeoTwawki: missed one

PR #6034, opened 2019-05-27
Debug messages, Engrish, and you

TeoTwawki:
Added quantity to tossed item and reserved item prints got rid of "too much quantity" grammar.


KnowOne134:
Di we want to un // the packetsystem debug?
Approve

TeoTwawki:
the commented out one is more of a notice (which is why ShowNotice instead of ShowDebug) So I leave that decision to server owners, as its not very important and gets very spammy.

src/map/utils/charutils.cpp
```diff
@@ -1358,9 +1358,12 @@ namespace charutils
             PChar->pushPacket(new CInventoryItemPacket(nullptr, LocationID, slotID));
             return 0;
         }
+
+        uint32 ItemID = PItem->getID();
```
whasf: uint32 for an item ID? Isn't that kind of overkill

TeoTwawki: I dunno, its what we already had. I just moved it up from line 1376

KnowOne134:
Boy scouts are always prepared!

TeoTwawki:
@whasf afaik, item id's are always going to stop at 65535 with is uint16 max and where SE put gil at. So I went ahead and changed it to uint16 from the uint32 someone else already had there.

Issue #6035, opened 2019-05-28
WS damage rework causing 0-dmg automaton attacks / atonement / spirits within


wrenffxi:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30190430_0

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

#5861 changed the parameter list of function takeWeaponskillDamage from

`function takeWeaponskillDamage(defender, attacker, params, primary, finaldmg, attackType, damageType, slot, tpHitsLanded, extraHitsLanded, shadowsAbsorbed, bonusTP, action, taChar)`

to

`function takeWeaponskillDamage(defender, attacker, wsParams, primaryMsg, attack, wsResults, action)`

However, there are several places in the code base where this function is still called with the old parameters.

~~`darkstar\scripts\globals\automatonweaponskills.lua(210): finaldmg = takeWeaponskillDamage(target, attacker, params, primary, finaldmg, dsp.attackType.PHYSICAL, damageType, dsp.slot.MAIN, tpHitsLanded, extraHitsLanded, shadowsAbsorbed, bonusTP, action, taChar)`~~ Will be fixed by #6037

~~`darkstar\scripts\globals\automatonweaponskills.lua(501): finaldmg = takeWeaponskillDamage(target, attacker, params, primary, finaldmg, dsp.attackType.RANGED, attacker:getWeaponDamageType(dsp.slot.RANGED), dsp.slot.RANGED, tpHitsLanded, extraHitsLanded, shadowsAbsorbed, bonusTP, action, nil)`~~ Will be fixed by #6037

`darkstar\scripts\globals\weaponskills\atonement.lua(85): damage = takeWeaponskillDamage(target, player, wsParams, primary, damage, dsp.attackType.BREATH, dsp.damageType.ELEMENTAL, dsp.slot.MAIN, tpHits, extraHits, 0, 0, action, nil)`

`darkstar\scripts\globals\weaponskills\spirits_within.lua(58): damage = takeWeaponskillDamage(target, player, {}, primary, damage, dsp.attackType.BREATH, dsp.damageType.ELEMENTAL, dsp.slot.MAIN, tpHits, 0, 0, 0, action, nil)`

PR #6037, opened 2019-05-29
Fix automaton zero weaponskill damage

ibm2431:
This partially address #6035 

The "quick fix" would have been to touch the automaton functions as little as possible and just fiddle their output into what takeWeaponskillDamage expects. But that's boring, and why write a `calculateRawWeaponskillDmg` function if you're not going to use it?

Notes:
- This naturally fixes an issue where automaton crit rates were being capped by "natural" crit rates. I don't know if their are any non-natural critical mods for automatons, but they can go past cap now.
- Renamed `getRangedDmgForRank` to `getRangedDmgRank` per convention, as requested by the comment in core
- Removed the dmgRank multiplication by 9 in `lua_baseentity` that was only getting divided by 9 in the lua weaponskill script. Weaponskill script now simply uses the divided-by-9 value originally given by `battleEntity->Get[Melee/Ranged]WeaponRank`.
- Renamed base_dmg in fSTR to weapon_rank; the variable was probably the actual damage before (hence the division mentioned above), but wasn't renamed when it was changed to use weaponRank from lua_baseentity. (That's my guess anyway.)
- For players, corrected both melee and ranged WS using a weird combination of fSTR (melee) and fSTR2 (ranged). The approximations in the original function were for fSTR2, but it was using fSTR caps. Melee/Ranged WSes now use separate functions for their appropriate values, and `do[Physical/Ranged]Weaponskill` were changed to calculate their fSTR upfront before passing it in to `calculateRawWSDmg`. This also lets automatons precalculate their clamped dSTR value, allowing them to use `calculateRawWSDmg` as well. See [bgwiki's fSTR page](https://www.bg-wiki.com/bg/FSTR) for player fSTR/fSTR2 formulae and caps.
- Removed weaponDmgRank from calcParams because it's no longer used by calculateRawWSDmg (it's only used to calculate fSTR)
- Tweaked how calculateRawWSDmg sets/uses alpha, to let an alpha be passed in (to overwrite calculateRaw default behavior if desired), and added a `calcParams.bonusWSmods`. These are mainly for automatons right now, but could potentially be hooked into by other things in the future.
- Had to move hitRate calculations back into their original `do[Physical/Ranged]Weaponskill` functions, but are still put into `calcParams`.
- Fencer bonus calculated upfront by player `doPhysical/Ranged` , outside `calculateRawWeaponskillDmg`, because automatons don't have inventories.
- Bonus fTP and accuracy from Elemental Gorgets and Belts are now also pre-calculated before going into `calculateRaw`. The calcParams passed in for these in `calculateRaw` can also be used for generic bonus fTP, such as being used for an automaton's Flame Holder

Speaking of Flame Holder, I'm fairly certain that [the values it uses](https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/abilities/pets/attachments/flame_holder.lua#L28-L40) aren't [in line with retail](https://www.bg-wiki.com/bg/Flame_Holder). But that was outside of "get automaton WS working again", and I didn't want fixing it to delay getting a PR in.

The non-working player WSes mentioned in #6035 are still on my docket, but again, I wanted to get an automaton fix PR in relatively swiftly.

(And for once, my attempts at simplifying something actually resulted in net deletions! :tada: )

### DIFF
--- a/scripts/globals/automatonweaponskills.lua
+++ b/scripts/globals/automatonweaponskills.lua
@@ -92,6 +92,72 @@ function doAutoPhysicalWeaponskill(attacker, target, wsID, tp, primaryMsg, actio
     return finaldmg, calcParams.criticalHit, calcParams.tpHitsLanded, calcParams.extraHitsLanded, calcParams.shadowsAbsorbed
 end
 
+-- params contains: ftp100, ftp200, ftp300, str_wsc, dex_wsc, vit_wsc, int_wsc, mnd_wsc, canCrit, crit100, crit200, crit300, acc100, acc200, acc300, ignoresDef, ignore100, ignore200, ignore300, atkmulti, accBonus, weaponDamage
+function doAutoRangedWeaponskill(attacker, target, wsID, wsParams, tp, primaryMsg, skill, action)
+    -- Determine cratio and ccritratio
+    local ignoredDef = 0
+    if (wsParams.ignoresDef == not nil and wsParams.ignoresDef == true) then
+        ignoredDef = calculatedIgnoredDef(tp, target:getStat(dsp.mod.DEF), wsParams.ignored100, wsParams.ignored200, wsParams.ignored300)
+    end
+    local cratio, ccritratio = getAutocRatio(attacker, target, wsParams, ignoredDef, false)
+
+    -- Set up conditions and wsParams used for calculating weaponskill damage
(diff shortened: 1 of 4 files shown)


wrenffxi:
Can confirm this PR makes automaton ranged attacks do damage.

PR #6042, opened 2019-05-31
Fix real Padfoot loot

wrenffxi:
setting its loot back to default dropid in onMobDeath is too early, so real padfoot would end up dropping same loot as the fake ones.

### DIFF
--- a/scripts/zones/Lufaise_Meadows/mobs/Padfoot.lua
+++ b/scripts/zones/Lufaise_Meadows/mobs/Padfoot.lua
@@ -1,37 +1,38 @@
 -----------------------------------
 -- Area: Lufaise Meadows
---  MOB: Padfoot
+--   NM: Padfoot
 -- !pos -43.689 0.487 -328.028 24
 -- !pos 260.445 -1.761 -27.862 24
 -- !pos 412.447 -0.057 -200.161 24
 -- !pos -378.950 -15.742 144.215 24
 -- !pos -141.523 -15.529 91.709 24
 -----------------------------------
-local ID = require("scripts/zones/Lufaise_Meadows/IDs");
+local ID = require("scripts/zones/Lufaise_Meadows/IDs")
 -----------------------------------
 
 function onMobSpawn(mob)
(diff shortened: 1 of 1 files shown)


TeoTwawki:
There's wren beating me to stuff as usual :)

I did this on demiurge to see if it fixed and it did, but didn't get it confirmed till yesterday and didn't have time to put it to darkstar:
```lua
-- Area: Lufaise Meadows
--  MOB: Padfoot
-- !pos -43.689 0.487 -328.028 24
-- !pos 260.445 -1.761 -27.862 24
-- !pos 412.447 -0.057 -200.161 24
-- !pos -378.950 -15.742 144.215 24
-- !pos -141.523 -15.529 91.709 24
local ID = require("scripts/zones/Lufaise_Meadows/IDs")

function onMobSpawn(mob)
    if GetServerVariable("realPadfoot") == 0 then
        SetServerVariable("realPadfoot",math.random(1, 5))
    end

    if mob:getID() == ID.mob.PADFOOT[GetServerVariable("realPadfoot")] then
        mob:setDropID(4478)
    else
        mob:setDropID(2734)
    end
end

end

function onMobDespawn(mob)
    local mobId = mob:getID();
    if mobId == ID.mob.PADFOOT[GetServerVariable("realPadfoot")] then
        local respawn = math.random(75600, 86400) -- 21-24 hours
        for _, v in pairs(ID.mob.PADFOOT) do
            GetMobByID(v):setRespawnTime(respawn)
            if v ~= mobId and GetMobByID(v):isSpawned() then
                GetMobByID(v):setHP(0)
                DespawnMob(v)
            end
        end

        SetServerVariable("realPadfoot", math.random(1, 5))
    end
end
```

what I found was the server was getting confused about which one was the "real" one, and the previous spawns treasure set would be on the current spawns fake causing the one with the astral earring to be one of the despawned ones, so putting both the treasure set AND the random for which is real in onSpawn cleared it all up. I also moved soem code into onDespawn to make it more in line with the other NMs timing wise.

PR #6046, opened 2019-06-01
fix Vermilion-Eared Noberry spell list

wrenffxi:
give Vermilion-Eared Noberry (SMN) an avatar summon spell list, rather than BLM spells.


TeoTwawki:
we have a spell list for that?

kinda thought core just made them spawn spirits and astral flow made'em pop avatars.

/noobdev

PR #6053, opened 2019-06-02
[WIP] A Moogle Kupo d'Etat

zach2good:
https://ffxiclopedia.fandom.com/wiki/Category:A_Moogle_Kupo_d%27Etat_Missions

Back again with Moogle related fun. Lots still to do, but the overall structure is starting to come together.

**TODO:**
~~Fix strange positioning at end of opening cinematic (AMK1)~~
~~Restricting the HELM gathering section to pre-RotZ areas~~
Check level restrictions for each mission
Extend orb drops to all Cardians
Restrict chocobo digging section to randomly chosen area only
All of AMK9
AMK10 BCNM
AMK13 Games
AMK14 BCNM
A placeholder reward (or nothing) until augments are available
Any additional treasure coffer things
Add !pos information to modified and new scripts
~~Shantotto gil check (and packet capture to figure it out)~~

~~Zone/Expansion table~~

More use of npcUtils for keyItems, messaging and mob popping
~~Make sure there are no uses of MKD instead of AMK (bcnm.lua etc.)~~
Squintrox Dryeyes trades in Port Jeuno

**Useful information**
~~a rough droprate for the key items in AMK4, ~~
~~a rough droprate for the orb key items in AMK6,~~
~~Packet caps for trying to pay shantotto when you have less than 5000g at the end of AMK6,~~
Packet caps for the AMK9 puzzle,
Video/logs of the AMK10 fight,
Packet caps of the games in AMK13,
Video of the AMK14 fight

**Not Done**
AMK6 Confrontation Cardian fight - Confrontation doesn't work yet?


KnowOne134:
i see your using a different var for each mission leg but can we set it like missions where they all use the same var thoughout the entire thing like [Mission]AMK or something

scripts/zones/Castle_Zvahl_Baileys/npcs/Flame_of_Fate.lua
```diff
@@ -0,0 +1,18 @@
+-----------------------------------
+-- Area: Castle Zvahl Baileys 161
+-- NPC: Flame of Fate
+-- A Moogle Kupo d'Etat
```
ibm2431: Could I ask for some !pos in these headers~? :sweat_smile:

scripts/zones/Sea_Serpent_Grotto/npcs/Shady_Sconce.lua
```diff
@@ -0,0 +1,25 @@
...
+-- !pos
+-----------------------------------
+local ID = require("scripts/zones/Sea_Serpent_Grotto/IDs")
+-----------------------------------
+
+function onTrade(player, npc, trade)
+end
+
+function onTrigger(player, npc)
+    if player:getCurrentMission(AMK) == dsp.mission.id.amk.LENDER_BEWARE_READ_THE_FINE_PRINT and player:getVar("AMK8") == 0 then
```
ibm2431: I'll extrapolate in main discussion, but as an example, if you tweak your var usage to "stage-based" (here: AMK8 == 1), you could do things like drop currentMission checks

scripts/zones/Upper_Jeuno/npcs/Inconspicuous_Door.lua
```diff
@@ -0,0 +1,88 @@
...
+    local hasTreeBark = player:hasKeyItem(dsp.ki.PIECE_OF_RUGGED_TREE_BARK)
+    local hasLambRoast = player:hasKeyItem(dsp.ki.SAVORY_LAMB_ROAST)
+    local hasWormeatenChest = player:hasKeyItem(dsp.ki.MOLDY_WORMEATEN_CHEST)
+
+    if jamInJeuno then
+        player:startEvent(10178)
+        player:setVar("AMK4", 1)
+    elseif myDecrepitDomicile and hasMetalStrip and player:getVar("AMK4") == 1 then
+        player:startEvent(10179) -- Metal Strip handed in
+    elseif myDecrepitDomicile and player:getVar("AMK4") == 1 then
```
ibm2431: Does the player have to obtain/give the KI in order?

scripts/zones/Windurst_Walls/npcs/Shantotto.lua
```diff
@@ -57,6 +57,16 @@ function onTrigger(player,npc)
     elseif (player:getQuestStatus(WINDURST,dsp.quest.id.windurst.CLASS_REUNION) == QUEST_ACCEPTED and player:getVar("ClassReunionProgress") == 3) then
         player:startEvent(409) -- she mentions that Sunny-Pabonny left for San d'Oria
     -------------------------------------------------------
+
+    -- AMK
+    elseif player:getCurrentMission(AMK) == dsp.mission.id.amk.CURSES_A_HORRIFICALLY_HARROWING_HEX then
+        player:startEvent(506)
+    elseif player:getCurrentMission(AMK) == dsp.mission.id.amk.AN_ERRAND_THE_PROFESSORS_PRICE and player:hasKeyItem(dsp.ki.RIPE_STARFRUIT) then
+        player:startEvent(507)
+        -- TODO: Where to check for 5000gil being available?
```
ibm2431: Does the event take a param telling it how much gil the player currently has? If the player says yes to the CS, when the CS has started, but they have less than 5,000 gil, what happens?

I have a feeling that it does take a gil value (at least, newer events do).

scripts/zones/Windurst_Walls/npcs/Shantotto.lua
```diff
@@ -179,6 +189,15 @@ function onEventFinish(player,csid,option)
         player:setVar("MissionStatus",8)
     elseif (csid == 399) then
         player:setVar("ShantottoCS",0)
+    elseif csid == 506 then
+        player:completeMission(AMK,dsp.mission.id.amk.CURSES_A_HORRIFICALLY_HARROWING_HEX)
+        player:addMission(AMK,dsp.mission.id.amk.AN_ERRAND_THE_PROFESSORS_PRICE)
+    elseif csid == 507 and option == 0 then -- Dont Pay
+        player:needToZone(true)
+    elseif csid == 507 and option == 1 then -- Pay
+        player:delGil(5000)
```
ibm2431: Make sure you check greater than 5000 gil here before you delete it (if "gil in param" above).

ibm2431:
@KnowOne134 
`can we set it like missions where they all use the same var thoughout the entire thing like`

This may be a point of contention, but personally, I think it'd be better for us to move away from that. Separating vars eliminates risk of cross-mission contamination ("Halver."), makes it easier to tell what the variable is currently being used for ("Oh, AMK8, must be the KIs" versus "well some mission in this line is doing something"), and actually serves as a built-in mission check if the vars are used properly:

Setting/checking AMK8 to 1 or higher combines the mission check with the mission progress check - the only way that'd get set >= 1 would be if the player is currently on that mission. And if the vars are properly being deleted, it'll be back to 0 when the player is done with that mission. So if progress/stage vars are kept separate between missions (and the vars are used with this in mind), you don't need to do `player:currentMission` checks at all! In fact, I'm not positive I'm checking for a quest being accepted over in quest_rewrite - just checking if the player is on a stage of the quest via stage var.

So long as vars are being properly destroyed when we're done with them, there should only ever be one var per character in the DB. And as a small bonus, if for some reason you're looking at vars in the DB, you can see what mission a character is currently on right when you see the variable. Would also make it pretty obvious if there's a script that's not properly cleaning them up - if you ever see that a player has _two_ "AMK" vars in the DB, you know someone dun goofed.

scripts/zones/Upper_Jeuno/npcs/Inconspicuous_Door.lua
zach2good: Once `myDecrepitDomicile` is flagged, the items can all be collected in any order, but they are handed in in order, as there are the specific reminders about what mog wants to be handed in next (cs: 10186, 10187 and 10188)

TeoTwawki:
> This may be a point of contention, but personally, I think it'd be better for us to move away from that. Separating vars eliminates risk of cross-mission contamination ("Halver."),

@ibm2431 I disagree. It makes mission code generally easier to read and _Halver's disease_ type errors was caused by people fixing a thing, having players who has started down the code path in the broken state, and others "fixing" them by adjusting the NPC a 2nd time instead of fixing the broken state of their players to "make it go" combined with very poor documenting of wtf was going on and trash script formatting (indents) confusing which conditionals where in which part. 

Since you can't have 2 active missions in the same category, we should rightly have just one variable _and make our code easy to read_.

scripts/zones/Outer_Horutoto_Ruins/npcs/qm1.lua
```diff
@@ -0,0 +1,37 @@
...
+local ID = require("scripts/zones/Outer_Horutoto_Ruins/IDs")
+-----------------------------------
+
+function onTrade(player, npc, trade)
+end
+
+function onTrigger(player, npc)
+    
+    for Custom_Cardian = ID.mob.CUSTOM_CARDIAN_OFFSET + 0, ID.mob.CUSTOM_CARDIAN_OFFSET + 13, 1 do
+        -- TODO: Spawn with mods, depending on key items
```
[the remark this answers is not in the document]
zach2good: Would it make more sense to have this extra function as an extra member of `params`, or as its own argument - changing the signature?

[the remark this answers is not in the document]
ibm2431: I'd probably make it be part of params, and then you don't need to worry about argument list changing. I don't know what you'd call it though, params.onSpawn would be my first guess.

scripts/globals/helm.lua
```diff
@@ -1430,34 +1430,31 @@ dsp.helm.onTrade = function(player, npc, trade, helmType, csid)
...
-            -- zoneId == dsp.zone.ZERUHN_MINES -- TODO: Find a way of filtering original RotZ areas
-        then
-            npcUtil.giveKeyItem(player, dsp.ki.SAVORY_LAMB_ROAST)
+            if
+                helmType == dsp.helm.type.MINING and
+                not player:hasKeyItem(dsp.ki.STURDY_METAL_STRIP)
+                -- zoneId == dsp.zone.ZERUHN_MINES -- TODO: Find a way of filtering original RotZ areas
+            then
+                npcUtil.giveKeyItem(player, dsp.ki.STURDY_METAL_STRIP)
+            end
```
ibm2431: Can save _more_ lines if you combine your `end`s for your type check with your next `if` (for type) into `elseif`~!

ibm2431:
@TeoTwawki But built-in current mission checks, Teo..! One less function you need to call...! Just grab the var and go...!

(I realize I probably will not win this one.)

(And re: DB lookup, I was envisioning a `LIKE`.)

sql/mob_groups.sql
```diff
@@ -24,25 +24,23 @@ DROP TABLE IF EXISTS `mob_groups`;
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `mob_groups` (
   `groupid` int(10) unsigned NOT NULL,
-  `poolid` int(10) unsigned NOT NULL DEFAULT 0,
-  `zoneid` smallint(3) unsigned NOT NULL DEFAULT 0,
-  `name` varchar(24) DEFAULT NULL,
```
TeoTwawki: Looks like you goofed when merging the masters changes into your branch, as the diff indicates undoing the changes rather than integrating them.

TeoTwawki: nvm I'm looking at an old diff. Thanks github!

zach2good: No, you're right, I dun goofed. I've messed up `mob_groups` and `mob_spawn_points`

zach2good:
Now that large chunks of this mission line that are complete, I'm closing this PR and submitting it in a couple of separate smaller PRs so it's easier to review. 
(Also I broke my db and I don't want to go through aggressive rebasing 🙏)

sql/mob_groups.sql
TeoTwawki: explorer moogle stuff to?

TeoTwawki:
Oh thank you..I was actually worried you were going to keep going till it was one gigantic PR for the entire mini expansion :sweat_smile:

PR #6058, opened 2019-06-04
Implement Lost In Translocation quest

kaincenteno:
Added all npcs needed for the quest. Also changed name in npc_list to match file names of npcs.

I appreciate your feedback on my scripting, hopefully there's not as many errors as my first pull request.

As always, thank you 😸


scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua
```diff
@@ -0,0 +1,65 @@
...
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end
+
+function onTrigger(player,npc)
+    local lostInTranslocation = player:getQuestStatus(CRYSTAL_WAR, dsp.quest.id.crystalWar.LOST_IN_TRANSLOCATION)
+    local leftMapPiece = player:hasKeyItem(dsp.ki.LEFT_MAP_PIECE)
+    local middleMapPiece = player:hasKeyItem(dsp.ki.MIDDLE_MAP_PIECE)
+    local rightMapPiece = player:hasKeyItem(dsp.ki.RIGHT_MAP_PIECE)
```
Comment (ibm2431) [inline scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: Can pop a `hasAnyPiece` here which is your `or`s, then below check for:
`var == 1 and hasAnyPiece`
`ACCEPTED and hasAnyPiece`
`ACCEPTED and not hasAnyPiece`

scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua
```diff
@@ -0,0 +1,65 @@
...
+
+function onEventFinish(player,csid,option)
+    if csid == 107 then
+        if player:hasKeyItem(dsp.ki.MAP_OF_GRAUBERG) then
+            npcUtil.completeQuest(player, CRYSTAL_WAR, dsp.quest.id.crystalWar.LOST_IN_TRANSLOCATION, {
+                gil = 2000,
+                xp = 2000,
+                var = "lostInTranslocationCS"
+            })
+            player:delKeyItem(dsp.ki.LEFT_MAP_PIECE)
```
Comment (ibm2431) [inline scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: Can save three lines by breaking these deletions out of the Grauberg Map check (but still inside cs 107).

sql/npc_list.sql
```diff
@@ -20496,8 +20496,8 @@ INSERT INTO `npc_list` VALUES (17494713,'Layton','Layton',105,382.679,-39.999,3.
 INSERT INTO `npc_list` VALUES (17494714,'Erik','Erik',125,258.643,-33.249,99.901,9,40,40,0,0,0,0,27,0x0100060102100220023002400250486000700000,32,'WOTG',1);
 INSERT INTO `npc_list` VALUES (17494715,'Theodor','Theodor',132,258.693,-33.250,-60.032,1,40,40,0,0,0,0,27,0x01000C0102100220023002400250616000700000,32,'WOTG',1);
 INSERT INTO `npc_list` VALUES (17494716,'RedAxe','Red Axe',130,304.628,-27.500,13.955,26,40,40,0,0,0,0,27,0x01000A08021002200230024002504F6100700000,32,'WOTG',1);
-INSERT INTO `npc_list` VALUES (17494717,'Gravestone','Gravestone',0,254.428,-32.999,20.001,1,40,40,0,0,0,0,3,0x0000340000000000000000000000000000000000,0,'WOTG',1);
-INSERT INTO `npc_list` VALUES (17494718,'Sarcophagus','Sarcophagus',0,336.594,-33.500,-56.728,1,40,40,0,0,0,0,3,0x0000340000000000000000000000000000000000,0,'WOTG',1);
+INSERT INTO `npc_list` VALUES (17494717,'Gravestone_map_quest','Gravestone',0,254.428,-32.999,20.001,1,40,40,0,0,0,0,3,0x0000340000000000000000000000000000000000,0,'WOTG',1);
```
ibm2431: Surprisingly, Gravestone appears to be unique in this zone, so the suffix for it isn't necessary (to others: the second Sarcophagus is 17494754).

scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua
```diff
@@ -0,0 +1,65 @@
...
+            npcUtil.completeQuest(player, CRYSTAL_WAR, dsp.quest.id.crystalWar.LOST_IN_TRANSLOCATION, {
+                keyItem = dsp.ki.MAP_OF_GRAUBERG,
+                var = "lostInTranslocationCS"
+            })
+            player:delKeyItem(dsp.ki.LEFT_MAP_PIECE)
+            player:delKeyItem(dsp.ki.MIDDLE_MAP_PIECE)
+            player:delKeyItem(dsp.ki.RIGHT_MAP_PIECE)
+        end
+    elseif csid == 105 then
+        player:setVar("lostInTranslocationCS", 1)
```
Comment (ibm2431) [inline scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: I'm curious, what's this var being used for? What's the difference between event 105 and 106? (There apparently isn't a dump for Batalia [S]).

ibm2431:
You might be asked to rebase to squash all those commits into just one. Mostly copy-pasting my last explanation:

In git bash you can use `git rebase -i HEAD~x` (where x is however many commits back you want to start squashing commits from) to begin an Interactive Rebase.

This will bring up a window where you can mark commits as `fixup` (may need to hit Insert to be able to type into the window).

You can also `reword` the first commit to something like `Implement Lost in Translocation quest`.

When you're done, Esc should take you back to the "bottom input line", where you can type `:wq` and press Enter to save the file. Git will then attempt to follow your instructs and "meld" all your commits into just one commit.

You can then do a `git push -f` to update your remote branch (on Github), and it will auto-update your PR here.

See [About Git Rebase](https://help.github.com/en/articles/about-git-rebase). I think there's also an autosquash feature of rebase as well, but I've never used it, so I can't help you there.

scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua
Comment (kaincenteno) [reply scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: Awesome! thank you for pointing that out 😸️ just fixed it this way to make it easy to read.

scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua
Comment (kaincenteno) [reply scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: can't believe I miss this, thank you for pointing it out  😸️

sql/npc_list.sql
kaincenteno: @ibm2431  how does the map know which one to choose? I posted in development and I kinda understood that the names were unique. Would love insight to get a better understanding 😸️

scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua
Comment (kaincenteno) [reply scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: @ibm2431  the var is used so if you talk to the NPC at csid == 105 it will give you a longer CS, compared if you talk to him again, he will still refer to the same dialog that was on csid == 105, but the dialog will only be a 1- 2 sentence dialog.

sql/npc_list.sql
ibm2431: It [builds an execution path based on the current zone and the triggered NPC's name](https://github.com/DarkstarProject/darkstar/blob/6b9f3d77eb2201b923160c50d13379d9f3b1ee8a/src/map/lua/luautils.cpp#L1440-L1442).

scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua
Comment (ibm2431) [reply scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: I suppose the var is fair then; just making sure it's required!

(Although, does it only play just that one time, or one per each zone? Would a entity local var be preferable?)

Comment (kaincenteno) [reply scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: @ibm2431  it doesn't play only once as NPC doesn't have a "standard greeting", instead he is stuck on that until the keyitems attained changed to have a few, or all of them. So for instance if the PC would leave to do something else like other quests or missions, and then eventually come back to this quest it should still play csid == 106.

With my explanation, do you think entity local var is still needed? Is there an example of a entity local var in one of the files? (sorry for asking but usually i refer to other quests people have done, to guide myself) so that way if I do another quest I can think of using it if needed 😸️

Comment (ibm2431) [reply scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: If the NPC should always say that even after the player zones and comes back, then I guess it should be a DB var.

But if you're interested in entity local vars (which DSP currently refers to as "local vars"; terminology which can be confusing), I think the [most recent examples of `setLocalVar` and `getLocalVar` can be found with the Moghouse quests](https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/moghouse.lua).

You use them just like normal vars that'd otherwise go in the SQL DB, but the difference is that these vars are stored in memory (for each specific entity), and are lost upon zoning. They can also be used for mobs as well as players, which may be useful if the mob needd to do anything out of the ordinary.

scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua
```diff
@@ -0,0 +1,63 @@
...
+    local leftMapPiece = player:hasKeyItem(dsp.ki.LEFT_MAP_PIECE)
+    local middleMapPiece = player:hasKeyItem(dsp.ki.MIDDLE_MAP_PIECE)
+    local rightMapPiece = player:hasKeyItem(dsp.ki.RIGHT_MAP_PIECE)
+    local anyMapPiece = leftMapPiece or middleMapPiece or rightMapPiece
+
+    if lostInTranslocation == QUEST_COMPLETED then
+        player:startEvent(108)
+    elseif lostInTranslocation == QUEST_ACCEPTED and leftMapPiece and middleMapPiece and rightMapPiece then
+        player:startEvent(107)
+    elseif player:getVar("lostInTranslocationCS") == 1 and anyMapPiece then
```
Comment (ibm2431) [inline scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: One tiny last thing I noticed: if the var is set after 105, then we already know the player has a map piece and don't need to check again. :wink:

Comment (kaincenteno) [reply scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: good catch 😸️ made the change. Thank you

scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua
Comment (kaincenteno) [reply scripts/zones/Batallia_Downs_[S]/npcs/Thorben.lua]: Awesome info 😸️ Thank you @ibm2431  I'll keep it in mind next time if it doesnt need to be a DB var.

sql/npc_list.sql
kaincenteno: Reverted the name back to the original. Thank you 😸️

kaincenteno:
Should I make any other changes for this to be merged? 😸

PR #6073, opened 2019-06-10
create global item table and test scripts

KnowOne134:


zynjec:
Do we really need a ginormous lua table for every item? It's just another thing that will need to be maintained and more than likely become stale.

If for some reason someone pushes this through, I really dislike the naming conventions you used (eg: _HQ2), at least have it match what we use in the database, and you also have errors.

TeoTwawki:
@zynjec kind of work in progress and he's run it by myself and lotus already. We're looking at switching the DB (and thus scripts) to use short names, becaus ethe names are stupidly long and..well just plain stupid.

plate_of_sole_sushi.lua
vs
sushi.lua

and we're unable to put = and minus symbols in the variable name, so I suggested using "HQ" and "NEG" instead,

zynjec:
This is literally the worst idea this project has ever had.

TeoTwawki:
¯\\\_(ツ)_/¯

TeoTwawki:
I too would like to start using the name fields in the actual SQL tables..for items and some other things as well. Lets discuss on discord. This really wasn't ready for a PR anyway.

nfaletra:
After brief discussion, this isn't quite the direction we want to take this stuff. Let's continue to talk about possible solutions for hardcoded item ids in discord.

PR #6075, opened 2019-06-10
Consolidate/remove some settings

TeoTwawki:
### Removed:
- **LOW_LV_REGIME** comment out the lv check to get the same effect.
- **AF specific fame.** Why was this ever a thing?
- **VISITANT_BONUS** past me apologies for placeholder-ing this.

### Merged, functionality retained:
- **Explorer moogle settings** are just one setting now. Set its level to 0 to disable it, instead of a setting for on/off and another for the level.


TeoTwawki:
### PENDING IN A SEPARATE PR TO COME:
- I think bio/dia shiz can be done entirely by SQL flags, need to double check.
- I want change how we handle settings (and perhaps confs) and when I have something ready to show discuss with the team and pray a buncha negative nancy's don't complain that its still to many settings. I'm envisioning a way to package out things where we don't need a bunch of requires all over creation and those of us that don't want something can more easily ignore its existence (less cluttery crap to maintain while still enabling a lot of adjustable-ness)

TeoTwawki:
Fixed conflicted file..

scripts/zones/La_Theine_Plateau/npcs/Dimensional_Portal.lua
```diff
@@ -2,19 +2,18 @@
 -- Area: La Theine Plateau
 --  NPC: Dimensional Portal
 -----------------------------------
+local ID = require("scripts/zones/La_Theine_Plateau/IDs")
 require("scripts/globals/keyitems")
```
ibm2431: Looks like this was a c/p of the Konschtat one with a KI require instead of missions.

(Tahrongi is fine.)

PR #6076, opened 2019-06-11
Add some missing items to Windurst Waters Cooking guild: Miso, Soy Sauce, Dried Bonito

TeoTwawki:
~~Todo: make old style guild shops optional..Or just nuke them since they generally caused complaints since forever and retail has fully eliminated them and I've been getting bad info from people "from retail" because of it for the last FIVE years http://forum.square-enix.com/ffxi/threads/46481-Any-chance-of-removing-crafting-caps?p=555187&viewfull=1#post555187~~

> [9:23 PM] ibm2431: Yeah, if it's on Chomo, I would rubber stamp that
> [9:23 PM] Teo: if you check the miso and its not 2500
> [9:23 PM] Teo: then it means that it does flux
> [9:23 PM] ibm2431: Because I know Chomo is the "always in stock" one
> [9:29 PM] ibm2431: I'll have to wait until 5:00 for Kopopo, but have some screenshots while you wait
_snip_
> [9:30 PM] Teo: ok so they took the corn off her, left a lot of th other stuff
> [9:30 PM] Teo: their patch notes say nothing about the arrangement of some items being limited still causing me and several other people to think this is gone
> [9:30 PM] ibm2431: Kopopo may have some of the same items, but at "fluctuation" prices
> [9:31 PM] Teo: the thread I linked, was someone asking for guild limits to be removed and the player base telling him it already was, quoting patch notes
> [9:31 PM] ibm2431: I do remember a time I was able to get a synth mat for cheaper from the "old" guild NPC than the hard-set "new" NPC
> [9:32 PM] Teo: from what you are telling me..I don't want a setting, just fuck this whole mess and replicate modern retail.
> [9:32 PM] ibm2431: Yep, fuck the whole mess and replicate modern retail!
> [9:32 PM] Teo: the "new" shop with ranks, can be set up as normal shops with an added param for craft rank like we have 1 for nation rank
> [9:33 PM] Teo: copy pasting this into guthub if you don't mind
> [9:37 PM] ibm2431: Not at all


ibm2431:
Both new and old style coexist on retail. There are two NPCs behind the counter, one is the new style, the other is the old style. In cases where both NPCs behind the counter aren't being used for this, they'll use a NPC standing next to the counter instead (ie: Northern San d'Oria Woodworking)

TeoTwawki:
from what I've read on the official forums, and BOTH wiki,s and ffxiah forums...None of the prices or stock levels change anymore at all on ANY npcs at all.

ibm2431:
I had the Windurst Cooking Guild run out of stock on me yesterday for Millicorn... but I didn't check price and/or stock after the guild reopened the next day.

For new/old style, are we talking about the hard-set items that get unlocked as your rank goes up?

TeoTwawki:
They were claiming the items never run out anymore. Am confused now. 

old style would be what we have now. New style would be...not doing this shit (so I thought) rank crap would be implemented on new style not touching the old was my thinking (if they were to coexist with some kind of setting).


If I have things wrong I'd like to amend my commit message. But I wnat to know exactly WTF IS HAPPENING first, because their own official word is being quoted and I'm being told it ain't true now.

ibm2431:
I have some more characters that require some fame boosting; I'll record a video of the guild~! Client updating now.

TeoTwawki:
perhaps this is a "we shoulda used commas or something" situation: where just the new items in the same fiuggin item list never run out

TeoTwawki:
like the 3 I just used = infinite, 
million corn = not infinite

TeoTwawki:
yeah fuggit, amending message someone else can deal with that crap this PR doesn't touch it anyway

ibm2431:
To others: Video capture with packet logs is available on the capture Discord server, in the hobbies-crafting channel.

TeoTwawki:
Guild crap needs a full overhaul. 1 NPC needs to us eteh non guild shop function, which needs a change to its arguments to support doing craft rank similar to how we do nation rank.

***this PR will not be addressing any of that. just add the 3 items with modern retail params as closely as our current system allows***

davismj:
Can you add some Dried Bonito to my local grocery store pls & ty.

PR #6019, opened 2019-05-19
Reassign groupids, with each zone getting a batch

wrenffxi:

PR #6087, opened 2019-06-17
mob group overhaul

Wiggo32:
The goal is to get Dynamis Zones working with bcnmcrap merge. I'm not going to do that until mob groups get organized and become easier to work with, so Phase 1:
What I did was separate groups into smaller groupings per zone and assigned them to mobs sequentially in order by mobid. Each zone will have its own set of groups starting at 1 and be uniquely identified by both the groupid AND the zoneid per lotus' suggestion on discord chat. This will make adding additional groups easier for future endeavors. 
The next phase will be reorganizing the mob_groups.sql so that they appear in order of zoneid like it does in mob_spawn_points.sql. Which I will start on after this is accepted to make sure I'm not doing frivolous work.
After that, I plan to take the Dynamis Zones one at a time and correctly organize/overhaul per current retail data which I have already collected.
Questions? Comments? Concerns?


ibm2431:
Tagging @wrenffxi on this, because I know he was interested in solving this issue with #6019 (Although I think the approach was different.)

I am in favor of getting groups under control; I have been hesitant towards doing work with mobs due to how intimidating it is to keep all the groups straight with our current system.

_And_ no more "rebase because my ID was stolen"!

helixhamin:
https://github.com/DarkstarProject/darkstar/blob/1a1ef6d87252f1c27dfacac05122b33b0a14ab19/sql/mob_groups.sql#L10424

Why do we suddenly lose zone ordering here? The zone ids suddenly jump around after this point.

ibm2431:
> Why do we suddenly lose zone ordering here? The zone ids suddenly jump around after this point.

Reorganizing to be sorted by zone ID hasn't been done yet; they're still in the order of the original groups.

TeoTwawki:
I guess you missed the conversation where we said not to do this because we want to move to per-zone group IDs instead

Wiggo32:
that's exactly what this is... per zone group ids

TeoTwawki:
sorry looked like just re-ordering it isn't showing me full diff (because massive)

I'll have to pull local

Wiggo32:
and wren's approach was a bit more complicated than just renumbering groups per zone which is what lotus suggested and is much more simplier/easy to comprehend/understand.

Wiggo32:
also, there will need to eventually be another phase where we fix the broken links in mob_groups.sql that haven't had any reference to any mobs in the mob_spawn_points.sql... for now I just left them the same number that they were to avoid conflicts.

TeoTwawki:
ok yeah, my bad that is what we were looking for. thanks Wiggo.

Wiggo32:
@helixhamin I have left them in original order so people could more easily check/track changes I did to the groups themselves... for Q&A purposes

Wiggo32:
> 
> 
> ok yeah, my bad that is what we were looking for. thanks Wiggo.

it's cool, I did suggest doing it a different way when it was first discussed in discord so it's not your fault. It's just even my original idea is even more complicated than this.

PR #6094, opened 2019-06-19
fix hq synth rates

wrenffxi:
divide int literals (1/2), get an int output (0)

### DIFF
--- a/src/map/utils/synthutils.cpp
+++ b/src/map/utils/synthutils.cpp
@@ -390,10 +390,10 @@ uint8 calcSynthResult(CCharEntity* PChar)
 
                     switch(hqtier)
                     {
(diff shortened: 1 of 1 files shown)


TeoTwawki:
can we possibly recast it to enable using actual division?

wrenffxi:
I mean if you're averse to float representation and really want fractional representation you could divide float literals (1./2.) and get the same number.  Just can't divide int literals, or you'll get an int result.

TeoTwawki:
@isotor so about that testing our changes my dude :)

TeoTwawki:
@wrenffxi I'm fine with however this turns out but there was some discussion about clarity, /shrug I'll let someone either merge as is or tell you what they want here.

PR #6099, opened 2019-06-22
A Moogle Kupo d'Etat Missions 1-5

zach2good:
https://ffxiclopedia.fandom.com/wiki/Category:A_Moogle_Kupo_d%27Etat_Missions

Breaking up my previous giant/broken PR (https://github.com/DarkstarProject/darkstar/pull/6053) into reasonable chunks.

Missions 1-5, stopping when Shantotto asks you to go fight some Cardians.

Opening CS is locked behind `ENABLE_AMK `, and all other missions are locked behind their various `player:getCurrentMission(AMK)` values.


scripts/globals/helm.lua
```diff
@@ -1430,6 +1431,31 @@ dsp.helm.onTrade = function(player, npc, trade, helmType, csid)
...
 
+        local amkChance = 20
+        if 
+            player:getCurrentMission(AMK) == dsp.mission.id.amk.WELCOME_TO_MY_DECREPIT_DOMICILE and
+            broke ~= 1
+        then
+            if
+                helmType == dsp.helm.type.MINING and
+                not player:hasKeyItem(dsp.ki.STURDY_METAL_STRIP) and
+                dsp.expansionZone.ORIGINAL_ROTZ[zoneId] and math.random(0,99) < amkChance
```
TeoTwawki: I really dislike doing `0,99` in math.random calls. I would like to see this conditional altered to just used `math.random(100)` instead. That will roll 1-100 which means it won't start at zero and is more intuitive in my opinion for things that are being treated as a % chance. You'll have to alter the comparison to account for that shift by 1.

scripts/globals/zone.lua
```diff
@@ -384,6 +384,143 @@ dsp.zone =
     WALK_OF_ECHOES_P                = 298,
 }
 
+dsp.expansionZone = {}
+dsp.expansionZone.ORIGINAL_ROTZ = {
+    [dsp.zone.WEST_RONFAURE]                  = true,
+    [dsp.zone.EAST_RONFAURE]                  = true,
```
TeoTwawki: I think this could be shortened by a lot - do you really need to explicitly check the zone by expansion or can you just check if the region or that the region is NOT one form the other expansions? I think the only other helm area's would be wotg and cop. https://github.com/DarkstarProject/darkstar/blob/master/src/map/zone.h#L347

(might be able to bind contententtype and use it depending on if cop area's round as middle lands or "other", but I did not look..prolly would need to region it at a min..In the future I'd like to do something outside of lua for zone expansion info but that is outside the scope of this PR)

scripts/globals/zone.lua
```diff
@@ -384,6 +384,143 @@ dsp.zone =
     WALK_OF_ECHOES_P                = 298,
 }
 
+dsp.expansionZone = {}
+dsp.expansionZone.ORIGINAL_ROTZ = {
```
TeoTwawki: we're newline brace people. Just FYI.

scripts/globals/moghouse.lua
```diff
@@ -17,6 +19,28 @@ MOGLOCKER_ACCESS_TYPE_ALLAREAS = 1
...
 
+function isInMogHouseInHomeNation(player)
+    if not player:isInMogHouse() then
+        return false
+    end
+
+    local currentZone = player:getZoneID()
+    local nation = player:getNation()
+    if nation == dsp.nation.BASTOK then
+        if currentZone >= 234 and currentZone <= 237 then
```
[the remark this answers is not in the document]
TeoTwawki: yes actually

PR #6100, opened 2019-06-22
Fishing skill up

nuno-pires-cruz:
Implementation of Fishing skill up logic.

Chance for skill up formula, similar to craft, but with a twist. The higher the difference between user skill and fish max skill needed, lower the chance to skill up.

Uses same craft vars, from map config file, for control of skill up chance and skill up multiplier.

Added new var to map config file, _fishing_daily_limit_, to enable a daily limit, like on retail.

Any input is welcome! Feel free to comment and suggest changes!


ibm2431:
> but with a twist. The higher the difference between user skill and fish max skill needed, lower the chance to skill up.

Why?

nuno-pires-cruz:
I checked fishing_fish table, there's a total of 133 fishes. of those, about half, have min = 0 and max > 30, meaning, a player with skill 0 can catch a level 30+ fish, so he would just need to find 3-4 types of fishes, to get to skill 100. which is what this [thread](https://www.bluegartr.com/threads/109359-FISHING-What-you-need-to-know-(hidden-gems-from-the-fishing-thread)) on BG demonstrates.

the only fishes that have min > 0 , are :
  Moat Carp min = 5 , Crayfish min = 2 and legendary min = 90

all guides seem to know the cap level, but not starting level to catch them.

the formula is a proposal. so players pick gaps of 5-10 levels for higher chance of skill up. they can pick higher gaps, but at the expense of chance to skill up.

TeoTwawki:
We only care about emulating what retail actually does, officially. If retail does that, then we want to. if that's not what retail does, then we don't.

ibm2431:
[When I first started going after Mercanbaligi (cap 86) at level 54](https://user-images.githubusercontent.com/13112942/60061526-0958c300-96e5-11e9-9c79-be59ff9112a0.png) compared to [when I reached level 56](https://user-images.githubusercontent.com/13112942/60061523-0958c300-96e5-11e9-810a-6d044d85ea5d.png). According to timestamps, those two levels took roughly 16 minutes.


[When I first started going after Armored Pisces (cap 108) at level 95](https://user-images.githubusercontent.com/13112942/60061524-0958c300-96e5-11e9-9b90-772198d1c0bf.png) compared to when to [when I reached level 96](https://user-images.githubusercontent.com/13112942/60061525-0958c300-96e5-11e9-83ee-2e3c2ce2896f.png). According to those timestamps, that one level took me 52 minutes.

nuno-pires-cruz:
ibm, that was on retail? If so, it's the opposite of my formula. I can change it to get similar behaviour.

ibm2431:
Yeah, was on retail (it would be nice if DSP had RoE though!). I _might_ be able to dig through my old logs at some point for harder data, but don't hold me to that - I think that was around 2014, and Logger may or may not have started dropping timestamps from logs / doing the line duplication thing at the time. Those screenshots were just when I saw double-skills.

src/map/utils/fishingutils.cpp
```diff
@@ -65,6 +66,13 @@ void LoadFishingMessages()
     });
 }
 
+void ClearPlayersFishingCatches()
+{
+    Sql_Query(SqlHandle, "update char_vars set value = 0 WHERE varname = 'FishingCatches';");
```
whasf: Are you missing " and charid = " otherwise you will clear everyone's counters!

src/map/utils/fishingutils.cpp
```diff
@@ -76,6 +84,28 @@ uint16 GetMessageOffset(uint16 ZoneID)
...
+// get player skill with item mod bonus added
+uint16 getPlayerFishingSkill(CCharEntity* PChar)
+{
+    uint16 skill = 0;
+
+    // player skill
+    skill += PChar->RealSkills.skill[SKILL_FISHING];
+
+#ifdef _DSP_SYNTH_DEBUG_MESSAGES_
+    ShowDebug(CL_CYAN"player skill = %u\n" CL_RESET, skill);
```
whasf: "Player fishing skill" (so we know what skill you're talking about when reading the log file)

src/map/utils/fishingutils.cpp
```diff
@@ -76,6 +84,28 @@ uint16 GetMessageOffset(uint16 ZoneID)
...
+#endif
+
+    // mod bonus
+    skill += PChar->getMod(Mod::FISH)*10;
+
+#ifdef _DSP_SYNTH_DEBUG_MESSAGES_
+    ShowDebug(CL_CYAN"player skill with mod bonus = %u\n" CL_RESET, skill);
```
whasf: "Player fishing skill with mod bonus ="

src/map/utils/fishingutils.cpp
nuno-pires-cruz: It's to clear everyone's counter. It's to be called at server midnight. I just realized i didn't commit my changes to time_server.cpp, will fix that.

src/map/utils/fishingutils.cpp
nuno-pires-cruz: Done

src/map/utils/fishingutils.cpp
nuno-pires-cruz: Done

src/map/utils/fishingutils.cpp
TeoTwawki: why would you not just check and set this the next time the specific player would need to access (read or write) that variable _anyway_ instead of setting up a new global timer to hit the entire server?

nuno-pires-cruz: I didn't set a new timer, I put it next to "guildutils::UpdateGuildPointsPattern()" call. Shouldn't everything that needs to be midnight reset, be in same place?
How to implement your suggestion?

TeoTwawki: Leave it like it is for now I guess, if its just piggy backing an existing one. It'll actually be more trouble and lines of code than its worth to explain and do to in place check. Honestly, I don't even like our existing implementation but I'm not going to ask you to rewrite it that'd be a little crazy/cruel. 

(if I had my way most of it would move out to global/zone lua's like digging, if only to do away with rebuilding/reimporting when we need to add/fix something)


###### more/slashes/just/because.

src/map/map.cpp
```diff
@@ -1339,6 +1340,10 @@ int32 map_config_read(const int8* cfgName)
         {
             map_config.skillup_bloodpact = atoi(w2);
         }
+        else if (strcmp(w1, "fishing_daily_limit") == 0)
+        {
+        map_config.fishing_daily_limit = atoi(w2);
```
TeoTwawki: missing indent

src/map/utils/fishingutils.cpp
```diff
@@ -149,8 +179,13 @@ bool CheckFisherLuck(CCharEntity* PChar)
 		return false;
 	}
 
+    // if there's a daily limit and the player reached it, then he can't get more fish
+    if (map_config.fishing_daily_limit != 0 && PChar->fishingCatches >= map_config.fishing_daily_limit)
+        return false;
+
 	CItemFish* PFish = nullptr;
 	CItemWeapon* WeaponItem = nullptr;
+    uint16 playerFishingSkill = getPlayerFishingSkill(PChar);
```
TeoTwawki: indent is wrong

TeoTwawki: scratch that, I was fooled by preexisting tabs.

src/map/map.cpp
nuno-pires-cruz: fixed

scripts/globals/player.lua
```diff
@@ -134,6 +134,7 @@ local function CharCreate(player)
     player:addItem(536) -- adventurer coupon
     player:addTitle(dsp.title.NEW_ADVENTURER)
     player:setVar("MoghouseExplication", 1) -- needs Moghouse intro
+	player:setVar("FishingCatches", 0) -- counter for fishing
```
TeoTwawki: You've placed this is a section that only applies to newly created characters, at which point this variable would already be zero (any variable not found returns zero)

src/map/utils/fishingutils.cpp
```diff
@@ -76,6 +84,28 @@ uint16 GetMessageOffset(uint16 ZoneID)
...
 
+// get player skill with item mod bonus added
+uint16 getPlayerFishingSkill(CCharEntity* PChar)
+{
+    uint16 skill = 0;
+
+    // player skill
+    skill += PChar->RealSkills.skill[SKILL_FISHING];
+
+#ifdef _DSP_SYNTH_DEBUG_MESSAGES_
```
TeoTwawki: do we really need a define for this? I never liked that in the crafting utils either. Debug Messages are filterable and we could always comment/uncomment them. Its not even a synth.

scripts/globals/player.lua
nuno-pires-cruz: Done

src/map/utils/fishingutils.cpp
nuno-pires-cruz: Done

src/map/utils/fishingutils.cpp
```diff
@@ -352,6 +392,106 @@ void RodBreaks(CCharEntity* PChar)
...
+                satier = 3;
+            else if ((baseDiff >= 31) && (baseDiff < 41))
+                satier = 4;
+            else if (baseDiff >= 41)
+                satier = 5;
+
+            for (uint8 i = 0; i < 4; i++)
+            {
+                random = dsprand::GetRandomNumber(1.);
+#ifdef _DSP_SYNTH_DEBUG_MESSAGES_
```
TeoTwawki: I thought you already removed these /shrug

TeoTwawki: not really important, just noticing coz Nyu made me look. :stuck_out_tongue:

ibm2431:
Saw this thread pop up in my notifications. Sorry I forgot about it.

[My chatlogs on retail from 2015/09/12 to 2015/09/15, during which I took Fishing from level 18 to level 100.](https://github.com/DarkstarProject/darkstar/files/3568709/fishing_skill.zip)

**Note:** This was during a skill-up campaign. I was also using Fisherman Feast and Pelican Ring.

I have edited the logs with some reg-ex to tag all the fishing-related messages (by type), and also remove non-fishing messages (linkshell, tell, etc). The only lines remaining _should_ be related to fishing, but if any slipped past, I apologize.

PR #6104, opened 2019-06-26
Fixed BCNM.lua line 470, and Added SQL information from Nyzul ilse

JusticeTroll:


TeoTwawki:
you have other stuff mixed in. if you can wait a bit I already have a pr incoming anyway, or you can fix this and I'll merge

TeoTwawki:
#6106

JusticeTroll:
I mean they are all 3 things that I figured I could do as 1 PR instead of three... 
1.) Byakko Loot Table Fix
2.) The Loot Table and prep work for Nyzul
3.) the BCNM line 470 fix.

Should I do these as three different PRs?

TeoTwawki:
empty PR description had me thinking you had messed up a merge (I didn't read the entire title). In general we try to keep things separated into multiple PRs since git+github allows us to use as many branches as we want and open as many as needed at once. we do allow minor things to be combined. But in this case the only thing mentioned was the BCNM error.

PR #6106, opened 2019-06-26
Misc odds and ends

TeoTwawki:
1. Fix the invalid variable in bcnm.lua
2. correct typing in lua/lua_baseentity.cpp for showPosition()
3. Eliminate magic numbering for zone ID checks that look for the last available zoneID - MAX_ZONE is now a thing.


TeoTwawki:
Now also includes fixes to the embassy NPCs in RuLude Gardens, they were overwriting missions if the player accidentally went to the wrong embassy by mistake.

PR #6108, opened 2019-06-26
Ghelsba Outpost BCNMs

wrenffxi:
Thanks to Brierre for compiling all this data and getting it working on the old BCNM system. I'm converting her work to the new BCNM system in this PR.

I've opened BCNMs that are scripted with full or nearly-full challenge.

I've closed BCNMs that are missing key elements that make the fight challenging.  If server owners want to enable these fights, they can be toggled on in scripts/globals/bcnm.lua, and they'll be enter/win/loot-able.  I've only disabled them because there's little challenge to getting the win.

* Petrifying Pair
OPEN: fully implemented
* Toadal Recall
OPEN: missing the shroom-in-cap mechanic for using queasy/numb/shake-shroom. Currently coded as regular mobskills.
* Wings of Fury
OPEN: missing Slipstream and Turbulence mobskills.


wrenffxi:
updated this PR so that it won't conflict with my open Horlais Peak PR.

TeoTwawki:
> OPEN: missing the shroom-in-cap mechanic for using queasy/numb/shake-shroom. Currently coded as regular mobskills.

we need a mixin for that sorta thing in general for shrooms. even shrooms in west ronfaure and la theine will "lose" a tp move after losing their cap for it.

> The "-shroom" attacks are used in the following order: Queasyshroom, Numbshroom, Shakeshroom. Each attack will consume one mushroom on the funguar's cap. There are three total. A successful steal may remove 1 mushroom from the funguar's cap and possibly change the order of "-shroom" attacks. On Coppercap, successfully stealing a woozyshroom seems to remove Queasyshroom from its arsenal while stealing sleepshroom seems to have no such effect on Exoray. 

There is a skill some use that recharges their caps too.

PR #6112, opened 2019-06-27
Sandoria 5-2 BCNM fix

DanteMccloud:


scripts/globals/bcnm.lua
```diff
@@ -481,7 +481,7 @@ function checkReqs(player, npc, bfid, registrant)
         [ 102] = function() return ( mjob == dsp.job.WHM and mlvl >= 66                                                                                ) end, -- Quest: Shattering Stars (WHM LB5)
         [ 103] = function() return ( mjob == dsp.job.SMN and mlvl >= 66                                                                                ) end, -- Quest: Shattering Stars (SMN LB5)
         [ 128] = function() return ( roz == dsp.mission.id.zilart.THE_TEMPLE_OF_UGGALEPIH                                                                                 ) end, -- ZM4: The Temple of Uggalepih
-        [ 160] = function() return ( mission == 15 and natStat == 3                                                                                 ) end, -- Mission 5-2
+        [ 160] = function() return ( sandy == dsp.mission.id.sandoria.THE_SHADOW_LORD and natStat == 3                                                                                 ) end, -- Mission 5-2
```
wrenffxi: This line covers 5-2 for all three nations. Mission 15 for sandy is THE_SHADOW_LORD. For Bastok, XARCABARD_LAND_OF_TRUTHS. For Windy, THE_SHADOW_AWAITS. So checking mission number == 15 is correct. It looks like the variable we want is nat rather than mission though. So it should read
[ 160] = function() return ( nat == 15 and natStat == 3 ) end, -- Mission 5-2

DanteMccloud: Updated to reflect [ 160] = function() return ( nat == 15 and natStat == 3 ) end, -- Mission 5-2

PR #6116, opened 2019-06-27
Sky Drop fixes

DanteMccloud:
I have compared the drop rates from ffxidb.com, BG-wiki.com, ffxi.allakhazam.com, and ffxiclopedia to reproduce a much more retail accurate drop rate. Barrowed Isotor's groups and some information from #5918 and focused it to be exclusive to Sky. These have been tested on my server and results have been pretty identical to a retail experience, minus the fighting for rights to roll on the drops with make believe LS points...

I was going to leave this to Isotor's PR but noticed missing items and rates seemed a bit off from where they should be for a hand full items, and seeing as that touches a lot more than just Sky I thought this would be easier to break into several smaller PR's to isolate fixes within specific content.

This resolves all the seals not dropping issues, as well as fixes grouping and rates for the other items.

Added identification comments for easy look ups and reference points.


TeoTwawki:
> minus the fighting for rights to roll on the drops with make believe LS points...

GET'IM BOYS! Thats one less person in the w.legs competition!

PR #6122, opened 2019-07-01
Nyzul-isle SQL update "prep work"

DanteMccloud:
I explained to Teo... I super broke the crude out of the old PR... fixed this one up.  In the future I'll keep the PR open instead of closing to fix it. My bad, I tried to salvage it haha.


zynjec:
Needs to be rebased, can PR again once it has.

Issue #3544, opened 2016-11-15
No indicator as to whether or not a synth recipe is a desynthesis.


Hozu:
**_I have:_**

- [X] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [X] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
current

**_Server Version_** (type `@revision` in game) **:**
b8a258559e8178b7924e43a95805e1d4992609bf

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
When I was looking through synthutils.cpp for a recent pull request, I noticed that it's hardcoded to decrease the success rate for synths using a lightning element crystal. Not all such synths are desynths (ie. the alchemy ones to make Mercury from various fish), and not all desynths use a lightning element crystal (I can't think of one offhand). The synth_recipes table does not have a column for desynth. I did notice there is a column named Type, however it does not appear to even be read from judging by the query (if I'm reading it correctly). It looks like it has a value of either 0 or 1, though I didn't look at all of the entries closely.

teschnei:
moblin armor (leathercraft) is a desynth that uses wind

Deadwing888:
Quadav backplate desynth with a wind crystal

TeoTwawki:
Quadav Backplates (and a ton of others) have multiple desynth recipes, 1 wind 1 lightning.

If that column is unused, lets rename it to make it obvious, and then use it to tell if a recipe is a desynth.

wrenffxi:
I've QCed two craft recipe lists so far (alchemy and bonecraft) and the `type` column is consistently 1 for synths and 0 for desynths, so I believe we can reliably use this column to distinguish.

maxtherabbit:
We will use it, but I'm almost positive at least some of the recipes in the
db are marked incorrectly. Appreciate you verifying them.

ibm2431:
Should that column be renamed from `type` to something more informative?

maxtherabbit:
I mean 'type' is a pretty accurate description for what it is, I think most
of the confusion around it rose from the fact it was never referenced in
the code

ibm2431:
Just floating the idea of flipping the two values, and renaming the column from `type` to `is_desynth` ~!

TeoTwawki:
[...]

PR #6128, opened 2019-07-03
Synth/Desynth | Swap 1 to 0 and Vice Versa

Cryptexx:
Changes made:

1. In synth_recipes, renamed column 'Type' to 'Desynth'.
2. Swapped 1's and 0's so normal synths are now 0 and Desynth is a 1 under 'Desynth'.
3. Added column 'ResultName' that lists the NQ result for human readability.
4. Changed core files to reflect 0 being normal synth and 1 being desynth.
5. Updated query in synthutils to reflect new 'Desynth' column name.


helixhamin:
Not sure about this

> Added column 'ResultName' that lists the NQ result for human readability.

I understand. One reason I created my excels, to let me see these things, but this just causes more unneeded bloat to the tables. I guess the people with actual authority to merge will need to weigh in though.

TeoTwawki:
That "unneeded bloat" was to spare us from future loss of data via lost comments when redumping. It has no functional difference its just adding human readability so we don't have to go running to item_basic or ffxiah to see what the result item was or really on maintaining a comment next to the row. Just moving that comment into the row itself basically. A bunch of us on discord chimed in that we preferred to at least preserve the item name.

If I have to right scroll a mile, then I consider it bloaty. code wise mysql does extremely well with massive amounts of rows and columns. Pretty much exists _because_ a machine can do well with such things while a human cannot.

helixhamin:
Sure, I asked before about something like that, that would basically make it more of a permanent comments type of thing, but I was told it would not be accepted. Why I thought this would not really be approved. If it is, well, no concern to me.

TeoTwawki:
Told by who? Most thinsg kinda depends who you ask, which is why this isn't insta merged - give everyone a chance to object.

ibm2431:
Regarding Name as a _column_: How often are redumps done/used? Would it be useful?

Not weighing in either way; asking because I honestly don't know since I've never worked on something that'd benefit from a DB redump.

ibm2431:
And this has been linked to a few times recently on Discord, but for those that only follow DSP through GitHub, this PR is addressing #3544. Previous discussion/some ideas can be found there.

Cryptexx:
I've seen some discussion before as to whether desynth was actually being used in any calculations (looks like it's referenced in that PR too). I just want to clarify that Type (now called Desynth) **is** being read while verifying the recipe and whether it's a normal synth/desynth does have an effect on the mods being used during synthesis calculation.

TeoTwawki:
@ibm2431 this happened specifically because the person PR'ing this had a situation that would be made a ton easier by redumping, but would lose comments. But with 1 join on the query we could get the item name back so that happened. Looking at the file, it looks like the comments were manually re-added _anyways._ As a rule redumps have been avoided and should not happen often, but the reason for that (in most tables) has always been comment loss or people that don't keep it sorted making the diffs hellish to review.

@Cryptexx the part that would make desynths have a drastically lower success rate compared to synths isn't in there yet.

Cryptexx:
@TeoTwawki Yes, I agree. Desynths aren't fully coded yet, but it is accounted for and the stage is set for further development.

PR #6129, opened 2019-07-05
Nyzul Isle npc placement

KnowOne134:
setup for a dynamic pos on lamps and rune of transfer as they are changed every floor, called and set by instance. Individual script support for runic lamps, currently in progress. individual script support for the starter rune of transfer and acts differently than the rest


KnowOne134:
Npc have to be at least set to 1,1,1 to be usable in the instance because if set at 0,0,0 you get a load error

ibm2431:
For lamps, you could probably get by with a generalized `onTrigger` that does all on/order/same time based on the lamp's number calculated from the passed-in NPC's ID (`this_lamp = npc:getID() - IDs.LAMP_START`).

That way you'd only have to worry about maintaining the one `runic_lamp.lua` instead of 5 copy-paste luas~!

KnowOne134:
Problem I ran into is if people are pressing the on trigger at same time.
This way each lamp is just checking npc:getID()+1 vector animationsub 3
based on how many lamps there are. This way its individual scripts per npc
causes no crossed wires

teschnei:
could you elaborate on that? i don't think there's any difference between
using one script or multiple scripts

KnowOne134:
In my testing few times when you hit 2 lamps at exact same time only 1 was responding. If times perfectly. I was gonna make it a global function anyway. Except lamp1 will have on if stage == register then. Else trigerLamp(player,npc) end

Wiggo32:
same kinds of things happen when multiple people trigger the same npc... even those that have events ... some of them don't get to finish their events (i.e. memory stream in promys)

KnowOne134:
Made the changes back to the lamps and seems to work out ok with them all named the same

PR #6131, opened 2019-07-07
Fix logic so check will pass correctly

whasf:


TeoTwawki:
had muh ternary results backwards

src/map/lua/lua_baseentity.cpp
```diff
@@ -8734,8 +8734,8 @@ inline int32 CLuaBaseEntity::checkKillCredit(lua_State *L)
 
     bool credit = false;
     int lvlDiff = (int)(PMob->GetMLevel()) - (int)(PChar->GetMLevel());
-    int maxDiff = (lua_isnil(L, 2) || !lua_isnumber(L, 2)) ? (int)lua_tonumber(L, 2) : 15;
-    float range = (lua_isnil(L, 3) || !lua_isnumber(L, 3)) ? (float)lua_tonumber(L, 3) : 100;
+    int maxDiff = (!lua_isnil(L, 2) && lua_isnumber(L, 2)) ? (int)lua_tonumber(L, 2) : 15;
```
ibm2431: Does `lua_isnumber` actually return true if the value is nil?

whasf: Yes
**https://pgl.yoyo.org/luai/i/lua_isnumber**

ibm2431: But.. but.. _nil_ isn't a string! :cry:

PR #6144, opened 2019-07-14
strange apparatus cleanup

wrenffxi:
move strange apparatus elementals into zones table
remove numeric zone ids
combine the two apparatus data tables into one with keys

### DIFF
--- a/scripts/globals/strangeapparatus.lua
+++ b/scripts/globals/strangeapparatus.lua
@@ -2,8 +2,12 @@
 --
 --     Strange Apparatus
 --
+-- https://www.bg-wiki.com/bg/Strange_Apparatus
+-- TODO: adjust drops rates per zone
+-- TODO: add Rune Kris, Rune Algol, Rune Scythe
 -----------------------------------
 require("scripts/globals/npc_util")
+require("scripts/globals/zone")
 ------------------------------------
 
 dsp = dsp or {}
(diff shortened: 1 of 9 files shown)


scripts/globals/strangeapparatus.lua
```diff
@@ -28,62 +29,227 @@ local DARK_CLUSTER      = 4111
...
-    [200] = {17093, 0.0400, 1, 17461, 0.0800, 1, 18084, 0.1200, 1, 17158, 0.1600, 1, 16563, 0.2000, 1, 12742, 0.2400, 1, 16647, 0.2800, 1, 18206, 0.3200, 1, 16408, 0.3600, 1, 221, 0.4000, 1, 17333, 0.5200, 6, 1229, 0.7000, 2, 931, 1.0000, 8},
+    [dsp.zone.DANGRUF_WADI] =
+    {
+        suffix  = 'DW',
+        uid     = 0,
+        chip    = RED_CHIP,
+        cluster = FIRE_CLUSTER,
+        drop    =
+        {
+            17093, 0.0400, 1,
```
TeoTwawki: I can has comments saying which drop is what? to save future contributors having to run to item_basic/ffxiah

code looks good. I'm assuming you tested.

wrenffxi: Added item names in comments, and some TODOs.  The drop data that's in there now is identical per zone.  Wiki indicates that certain items are more likely to drop in certain zones.  Also it seems that three new items were added to the drop pools in 2017 (locations and rates unknown).

I tested before PRing.

Issue #6115, opened 2019-06-27
Server crash in Shattering Stars/SMN on pet release


xipies:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 
30190605_0

**_Source Branch_** (master/stable) **:** 
Old School server (updated 2019-06-26)

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Entered Shattering Stars/SMN as 75smn.
Ate rolanberry pie.
Drank melon juice.
Summoned Garuda.
Used Aerial Armor.
Used Release.
Crashed.

whasf:
Here is the crash dump: https://1drv.ms/u/s!AvyTIXTmkiRwpXz8Bt6u6F7NMqXm

xipies:
Reproduced locally.

Exception:
Unhandled exception at 0x01614D07 in DSGame-server.exe: 0xC00000FD: Stack overflow (parameters: 0x00000001, 0x00132FD8). occurred

Stack trace:
 	DSGame-server.exe!std::_Iterator_base12::_Orphan_me() Line 214	C++	Symbols loaded.
 	DSGame-server.exe!std::_Iterator_base12::_Adopt(const std::_Container_base12 * _Parent) Line 187	C++	Symbols loaded.
 	DSGame-server.exe!std::_Tree_unchecked_const_iterator<std::_Tree_val<std::_Tree_simple_types<std::pair<unsigned short const ,CBaseEntity *> > >,std::_Iterator_base12>::_Tree_unchecked_const_iterator<std::_Tree_val<std::_Tree_simple_types<std::pair<unsigned short const ,CBaseEntity *> > >,std::_Iterator_base12>(std::_Tree_node<std::pair<unsigned short const ,CBaseEntity *>,void *> * _Pnode, const std::_Tree_val<std::_Tree_simple_types<std::pair<unsigned short const ,CBaseEntity *> > > * _Plist) Line 40	C++	Symbols loaded.
[...]

PR #6145, opened 2019-07-14
Check for infinite loop that is causing stack overflow when releasing…

xipies:
… a pet in a battlefield.

Resolves issue #6115.

I did not see a particular code pattern in dsp for addressing an infinite loop in a dtor, so looked online for existing C++ patterns and went with this one.


teschnei:
the typical solution would be to figure out why it's an infinite loop and stop it from happening in the first place.  the probably solution is probably to unset PBattlefield inside RemoveEntity

Issue #6119, opened 2019-06-28
Mobs not superlinking in BCNMs


wrenffxi:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** n/a

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

There's some [code in the core](https://github.com/DarkstarProject/darkstar/blob/6b9f3d77eb2201b923160c50d13379d9f3b1ee8a/src/map/utils/mobutils.cpp#L845-L850) that looks like it's supposed to be superlinking mobs in zones with ZONETYPE_BATTLEFIELD, but neither PMob->m_bcnmID nor PMob->m_battlefieldID are ever being set, so this code does nothing.

I've been superlinking battlefields mobs in their onMobSpawns as I work my way through the battlefields.  But having it be handled in SetupBattlefieldMob instead would be a nice-to-have.

PR #6152, opened 2019-07-16
Brothers ENM

wrenffxi:
* add mob HP, NM flag, resistances, apocalyptic ray, draw-in
* add chance to drop nothing to the loot groups


scripts/zones/Bearclaw_Pinnacle/mobs/Eldertaur.lua
```diff
@@ -0,0 +1,20 @@
...
+-----------------------------------
+
+function onMobInitialize(mob)
+    mob:setMobMod(dsp.mobMod.DRAW_IN, 1)
+    mob:setMod(dsp.mod.DMGMAGIC, -10)
+    mob:setMod(dsp.mod.SLEEPRES, 75)
+end
+
+function onMobSpawn(mob)
+    mob:setMobMod(dsp.mobMod.SUPERLINK, mob:getShortID())
```
TeoTwawki: @teschnei didn't we have this being handled automatically for bcnm mobs without having to set this?

TeoTwawki: We should fix that if it broke

zynjec: Probably because of https://github.com/DarkstarProject/darkstar/issues/6119

TeoTwawki: yeah ok, we needa fix that then after we'll nuke all these manual superlinks. waiting on appveyor atm to merge this.

PR #6155, opened 2019-07-18
Bestiary adjustments: beasts and lizards

wrenffxi:
Reordered mob_pool_mods (in its own commit)

Mob | Adjustment
-- | -- 
Buds Bunny | group name, add drops
Carriage Lizard | job change, remove spell list
Coeurl | skill lists based on expansion
Fume Lizard | group name, add drops
Grand Marid | remove aggro
Labyrinth Lizard | some aggro, some dont; split into two pools
Lesser Wivre | remove aggro
Pygmytoise | job change; add aggro, immunities, resistances, and skill list
Snow Lizard | remove aggro
Various Tigers | adjust aggro, linking, and truesightedness
Virulent Peiste | add 5% XP bonus


sql/mob_groups.sql
```diff
@@ -10661,7 +10661,7 @@ INSERT INTO `mob_groups` VALUES (14,5895,193,'Donggu',0,32,4671,3500,0,42,44,0);
 INSERT INTO `mob_groups` VALUES (15,1690,193,'Goblin_Mugger',720,0,1389,0,0,22,26,0);
 INSERT INTO `mob_groups` VALUES (16,1683,193,'Goblin_Leecher',720,0,1366,0,0,22,26,0);
 INSERT INTO `mob_groups` VALUES (17,1666,193,'Goblin_Gambler',720,0,1333,0,0,22,26,0);
-INSERT INTO `mob_groups` VALUES (18,378,193,'Beach_Bunny',330,0,0,0,0,83,89,0);
+INSERT INTO `mob_groups` VALUES (18,378,193,'Buds_Bunny',330,0,360,0,0,83,89,0);
```
TeoTwawki: :carrot:

scripts/zones/Gustav_Tunnel/mobs/Pygmytoise.lua
```diff
@@ -3,8 +3,14 @@
 --  MOB: Pygmytoise
 -----------------------------------
 require("scripts/globals/regimes")
+require("scripts/globals/status")
 -----------------------------------
 
+function onMobInitialize(mob)
+    mob:setMod(dsp.mod.MDEF, 250) -- "Highly resistant to magic damage"
+    mob:setMod(dsp.mod.THUNDERRES, 75) -- "Very resistant to Lightning damage"
```
TeoTwawki: A few things:
1. I think you meant 25 MDEF not 250, or confused MDEF with MDT ([wrong scale](https://github.com/DarkstarProject/darkstar/search?q=MDEF&unscoped_q=MDEF))
2. I'm having trouble locating any info about these having double attacks, if they do not we can probably provide this via subbing them with whm, else the mod is actually what we want. I've asked in capture discord to see if we can confirm. I'm ok leaving this as a war sub with the mod for now if we get the amount corrected.
3. Retail seems to have 2 class of turtle: 1 is very resistant to earth and thunder but susceptible to ice (all else neutral), the other is strong vs water but actually weak to thunder (again neutral to the rest). Genbu falls in the later. Pygmy in teh former. so we should handle the element via mob_families.sql.

wrenffxi: I tested the 250 MDEF value before PRing.  Blizzard4 from a well-geared BLM did ~1400 damage at MDEF 0.  It did ~1322 with MDEF 25.  It did ~650 damage at MDEF 250.  That's more like it.  You might be right that the correct mod to use is -MDT, though it's impossible for me to know without retail access.  Wiki just says "resistant/highly resistant/very resistant" which doesn't help much.

Wiki also says these mobs are particularly resistant to lightning, on top of their already very high magic resistance.  But there's no source for any of this info.

I'm just trying to make this camp not a total pushover for manaburn parties, who are nuking the hell out of the worms without drawing any aggro from the tortoises.

TeoTwawki: Nyu checked for us on retail, could not get any turtles to do a double attack so its probably not a war sub, and the whm sub would give it the MDEF so thats probably what retail actually has here is pld/whm turtles.

TeoTwawki: I should have refreshed before posting

TeoTwawki: 250 MDEF is _insane_ so if thats what it take sto get them close to retail accurate something is way wrong and you probably want 25% damage reduction on everything instead. And that leaves me with no clue what its proper sub should be we just know war isn't it.. I guess go pld/pld with MDT and see if in the right ballpark.

Hozu: I think to get the magic damage done properly, Specific Damage Taken would have to be implemented.

https://www.bg-wiki.com/bg/Resist#Elemental_SDT

The basic idea is elemental resistance is tiered, and below certain tiers guarantees an increasing number of resists. And resists are calculated by rolling a number of times, 4 for nukes 2 for enfeebles, and the number of failures is how strong the resists are.

Anyway, according to the spreadsheet in [this BG post](https://www.bluegartr.com/threads/130595-Legends-of-the-Hidden-jp-testing?p=7019830&viewfull=1#post7019830) the adamantoise family is guaranteed to get a full 1/8th resist on thunder, earth, and water elements, a minimum of 1/2 resist on fire, wind, light, and dark, leaving ice as the only element that is able to land unresisted. (actually MBing fire/wind/light/dark would raise it above the 50% tier, which would allow unresisted, but the 130% tier for ice makes it do far more damage when MB'd)

However, all of that would require a rather large overhaul beyond the scope of this pull request.

wrenffxi:
rebased

scripts/zones/Gustav_Tunnel/mobs/Pygmytoise.lua
TeoTwawki: Nice find Hozu.

PR #6158, opened 2019-07-20
Adjust Adamantoise family elemental resistances to hopefully be more retail accurate

TeoTwawki:
Also fixed Genbu's crystal element, he's water instead of earth.

@wrenffxi when you have time, would you mind comparing the results of this with the what you originally had? I have marked this PR as a "draft" until I can gauge if damage numbers are in the right ballpark.

### DIFF
--- a/scripts/zones/Gustav_Tunnel/mobs/Pygmytoise.lua
+++ b/scripts/zones/Gustav_Tunnel/mobs/Pygmytoise.lua
@@ -6,6 +6,24 @@ require("scripts/globals/regimes")
 require("scripts/globals/status")
 -----------------------------------
 
+function onMobInitialize(mob)
+    --[[
+    Attempt to Aproximate retail damage ratios,
+    the current resist rates can't do the job..
+    What retail appears to do per info on BG is FORCE a minimum resist teir
+    along with a damage bonus on ice (all spell get a partial resist).
+    These are annoyingly x/256 scaled.
+    ]]
+    mob:setMod(dsp.mod.FIREDEF, 128)
+    mob:setMod(dsp.mod.ICEDEF, 52)
+    mob:setMod(dsp.mod.WINDDEF, 128)
+    mob:setMod(dsp.mod.EARTHDEF, 200)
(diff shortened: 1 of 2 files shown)


TeoTwawki:
I know its not going to be perfect, because those columns I adjusted are both damage resistance % and resist rate in one. Retail would have those as separate things and [BG indicates](https://www.bg-wiki.com/bg/Resist#Elemental_SDT)  neither is exactly the right behavior for this.

We should probably move this out of mob families into its own table and ad a "resistID" to mob pools instead someday, but I will give y'all a heart NOT IT on that.

wrenffxi:
Just tested this with a level 75BLM with good gear.  Thunder4 and Blizzard4 were both doing ~1400-1500 damage, with occasional resist for half damage.

TeoTwawki:
so still unresisted way the F to often even with resist set to 50%.

I'll modify this some more tomorrow.

wrenffxi:
Oh, Pygmytoise are familyId 275 'MiniAdam' so they were unaffected by this SQL change.

TeoTwawki:
_slaps self_ I didn't even notice it!

TeoTwawki:
Re-pushed for both covering both IDs now..

wrenffxi:
Hm.  Still doesn't seem to do very much.  Sometimes they resist for half damage, but I wouldn't expect any of these T4 spells to do over 1k damage.
image

TeoTwawki:
hmm yeah blizzard I could see but thunder should have been drastically cut, so this isn't working out. Back to plan be when I get time tomorrow.

wrenffxi:
Those mods definitely cut damage.  Blizzard does the most, thunder/earth do the least, and the other elements do damage in between.

image

TeoTwawki:
@teschnei this is kinda half assed, I will understand if we don't want to do this. I feel that at some point we should move the mobs special resistances out of family and put them to a new table and include a resists ID in mob pools like we have for skill lists and spell lists while also adding this weird "minimum resist" thing these turtles do that BG is describing on its resist page.

PR #6160, opened 2019-07-20
Bestiary adjustments: bees n' beetles

wrenffxi:
Made the following group/pool adjustments for bees and beetles:

Mob | Adjustment
-- | -- 
Digger Wasp | some aggro, some don't, so created two distinct pools.
Vespo | no longer using Wespe pool. no longer aggro.
Eschan Wasp | no longer aggro.
Armet Beetle | no longer exist in game. replaced by Bonnet Beetle.
Beady Beetle | no longer aggro, because the groups that aggroed have been removed from game. now link.
Deathwatch Beetle | no longer using Chamber Beetle pool. added drops.
Goliath Beetle | some aggro, some don't, so created two distinct pools.
Helm Beetle | some aggro, some don't, so created two distinct pools.
Starborer | no longer using Starmite pool. no longer aggro. fixed drops.
Targe Beetle | no longer using Chamber Beetle pool. no longer aggro. added drops.


sql/mob_droplist.sql
```diff
@@ -33063,6 +33063,11 @@ INSERT INTO `mob_droplist` VALUES (4803,0,0,1000,644,212); -- (Aura Sculpture) C
 INSERT INTO `mob_droplist` VALUES (4803,0,0,1000,3541,182); -- (Aura Sculpture) Seasoning stone
 INSERT INTO `mob_droplist` VALUES (4803,0,0,1000,955,30); -- (Aura Sculpture) Golem shard
 INSERT INTO `mob_droplist` VALUES (4803,0,0,1000,3543,91); -- (Aura Sculpture) Fossilized fang
+INSERT INTO `mob_droplist` VALUES (4804,0,0,1000,846,295); -- (Starborer) insect_wing
+INSERT INTO `mob_droplist` VALUES (4804,0,0,1000,889,200); -- (Starborer) beetle_shell
+INSERT INTO `mob_droplist` VALUES (4804,0,0,1000,894,92); -- (Starborer) beetle_jaw
+INSERT INTO `mob_droplist` VALUES (4804,0,0,1000,3541,77); -- (Starborer) seasoning_stone
```
TeoTwawki: It looks like you used the average value on ffidb, that will be inaccurate since the value we have in this db column is before any TH is applied. When possible, use the TH0 value, otherwise make your best guesstimate - I see the only data they have is TH2 so I'd say ~50 on the seasoning stone. The others all have TH0 data we can use.

TeoTwawki: | 

| Item | Count | Average | TH0 | TH1 | TH2 | TH3+
-- | -- | -- | -- | -- | -- | --
Insect wing | 177 out of 600 | 29.5% | 20.0% | 0% | 40.0% | 55.2%
Beetle shell | 120 out of 600 | 20.0% | 25.0% | 0% | 30.0% | 32.8%
Beetle jaw | 55 out of 600 | 9.2% | 15.0% | 0% | 0.00% | 15.5%
Seasoning stone | 46 out of 600 | 7.7% | 0.00% | 0% | 20.0% | 12.1%
Fossilized bone | 37 out of 600 | 6.2% | 15.0% | 0% | 0.00% | 8.6%

TeoTwawki: eh, I kinda see why you did that now... Its very skewed by high TH

TeoTwawki: Every rule gets an exception I suppose..merging.

PR #6168, opened 2019-07-24
clean up mission/quest battlefield scripts

wrenffxi:
couple small fixes included:
* Trial by Element battlefields now award titles (before, only Ice did).
* Trial by Lightning now enforces the one-day wait on loss, like the other Trials.
* Moved mobIds into zones table.
* Moon Reading: record arguments are now passed into event.


Hokuten85:
I believe this introduced a bug with the mini-trial fights.

in the onBattlefieldLeave function

dsp.quest.id.**otherAreas**.TRIAL_SIZE_TRIAL_BY_FIRE

should be

dsp.quest.id.**outlands**.TRIAL_SIZE_TRIAL_BY_FIRE

I believe this type of error shows up in most of the mini-trial fight files

PR #6179, opened 2019-08-09
Increase maneuver duration

Andrew6ta:
Maneuver base duration is 60 seconds and gradually increases to 5 minutes while the automaton is engaged in combat (deployed). This was changed in the 3/26/15 update.

### DIFF
--- a/scripts/globals/abilities/dark_maneuver.lua
+++ b/scripts/globals/abilities/dark_maneuver.lua
@@ -7,6 +7,7 @@
 -----------------------------------
 require("scripts/globals/status")
 require("scripts/globals/magic")
+require("scripts/globals/utils")
 -----------------------------------
 
 function onAbilityCheck(player,target,ability)
@@ -42,8 +43,9 @@ function onUseAbility(player,target,ability)
             target:removeOldestManeuver()
         end
 
-        local bonusDur = target:getPet():getDeployDuration()
-        target:addStatusEffect(dsp.effect.DARK_MANEUVER, bonus, 0, 60 + bonusDur)
+        local dur = player:getPet():getLocalVar("MANEUVER_DURATION")
+        target:addStatusEffect(dsp.effect.DARK_MANEUVER, 0, 0, utils.clamp(dur,60,300))
+
     end
 
     return dsp.effect.DARK_MANEUVER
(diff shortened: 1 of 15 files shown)


wrenffxi:
Note: these files have tabs in them.  Project standard is to use 4-spaces.  Recommend checking the settings in your editor of choice, and setting/enabling

* strip trailing spaces from lines when saving
* default tab spacing / indent size = 4
* convert new tabs to spaces
* convert existing tabs to spaces when saving

wording for these settings my vary, depending on editor.

zynjec:
Wouldn't it make more sense to just store when the entity was created, and have the binding return 
its age? Having something increment every tick seems a little over kill.

Then you could just do something like:
```lua
local bonus = math.max(entityAge, 240)
```

Andrew6ta:
Hi Zynjec,

The duration only increments when the automaton is deployed (engaged with a mob). Just Activating the pet and standing around doesn't increase the duration.

teschnei:
The feature doesn't need any cpp changes.  Please replace it with a localvar and listener that is incremented using the COMBAT_TICK or EFFECTS_TICK (depending on the granularity you actually need) events.

Andrew6ta:
I have this done using the "AUTOMATON_AI_TICK" event since "COMBAT_TICK" is never triggered in the automaton AI.  It's in a new branch on my local computer. Would it be best to close this PR and open a new one or use this same PR somehow? I'm not super familiar with git best practices.

Thanks

teschnei:
keep using this PR so we can keep the conversation history

I think you should use COMBAT_TICK and just add it to DoCombatTick though,
AUTOMATON_AI_TICK only fires when the automaton's allowed to try an action
(ie. decisions are not on cooldown)

teschnei:
also, if the power increases every 3 seconds like most status effects, then
you should use the effect tick instead

Andrew6ta:
I implemented the duration bonus using COMBAT_TICK and added a function to make sure it's only once every 3 seconds.  For EFFECT_TICK, no maneuvers have to be active in order for the duration to increment, so that one doesn't make sense to me. The duration of a maneuver effect doesn't increase dynamically once it's applied to the master, but the duration of future maneuver effects increases the longer the automaton is engaged in battle (deployed).

I added automaton.lua to the pets folder to handle adding and removing the listener.

teschnei:
EFFECTS_TICK just fires every 3 seconds, it is nothing more than that (it
fires when effects would tick)

Andrew6ta:
I can implement using EFFECTS_TICK and checking if the automaton has a target to make sure it only increments during battle, if we think that's a more elegant solution?

PR #6195, opened 2019-08-17
[Dynamis] refill statue droplists, various Dyna SQL fixes

wrenffxi:
### Drop Lists
* Created distinct loot pools for HP/MP refill statues so that they drop correct according to:
  * drop beastcoins and singles (no hundos, no proccing) in all zones
  * drop ginuva's battle theory in beauc, xarc, and dreamworld zones
  * drop sparkling stones in cities, beauc
  * drop infinity cores in cities, beauc, xarc
* Fixed Vanguard Dragontamer dropid in Bubu and Quf (had Valk's dropid)
* Removed dropid from Arch Diabolos (had regular megaboss's drops)
* Started labeling drop lists (Va, Bu, Qu, Ta), which will preserve my sanity for future commits

### Miscellany
* Turned on MOBTYPE_CALLED flag on several NM pets.
* Adjusted Arch-Diabolos level range.
* Made a Rearguard Eye time extension group in Beauc, rather than a single eye.


wrenffxi:
Rebased

PR #6198, opened 2019-08-22
Rhapsodies of Vanadiel 1-1 to 1-10

zach2good:
Making progress

**TODO:**
~~Basic confrontation mechanic~~
Ophiotaurus fight - https://www.bg-wiki.com/bg/Ophiotaurus
Siren fight
Rhapsody in White/Umber/Azure exp mod
Character/Expansion blocking, alternate text -> ROV Helper Script
Intro CSs in Windy/Bastok/Sandy - possibly moved to the ROV helper, since they're all the same logic and csid's
Additional rewards: Escha access, Gilgamesh's letter of introduction etc.
Formatting, new files headers, cleanup etc.
Lots of testing


src/map/utils/charutils.cpp
```diff
@@ -2724,6 +2724,11 @@ namespace charutils
                 }
                 MaxSkill = MaxSkill * 10;
 
+                if(hasKeyItem(PChar, 2884)) // Rhapsody in White
```
TeoTwawki: unsquish pls. `if (things)`

scripts/zones/Mhaura/npcs/Tonasav.lua
```diff
@@ -0,0 +1,27 @@
...
+-- Type: ROV NPC
+-----------------------------------
+require("scripts/globals/missions")
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end
+
+function onTrigger(player,npc)
+    if player:getCurrentMission(ROV) == dsp.mission.id.rov.THE_BEGINNING and player:getVar("ROV") == 0 then
```
TeoTwawki: If thats a mission status variable, we have an existing convention for those. ZilartStatus PromathiaStatus etc. etc.. so RhapsodiesStatus probably. RovStatus maybe? The mission category seems to be using ROV already, so whichever.

kaincenteno:
@zach2good Try using npcutil for the trade for Ekoko.lua

src/map/utils/charutils.cpp
```diff
@@ -4404,6 +4409,11 @@ namespace charutils
         else
             exp = exp + bonus;
 
+        if (hasKeyItem(PChar, 2884)) // Rhapsody in White
```
ibm2431: I'm unfamiliar with AddExpBonus, and this may be in "premature optimization" territory, but I'm guessing AddExpBonus is called every time upon a mob kill, and there will eventually be 9 ROV KIs.

I'm wondering that, since these KI effects are permanent upon being obtained (they don't wear off like Dedication would), if instead of 9 `if has KI` checks and combining them upon each exp event, if instead the ROV bonus should be calculated and stored as a `MOD_ROV_EXP` or something on zone/character load/whatever, and then in AddExpBonus just `exp *= MOD_ROV_EXP`

(Same thing for skill-ups)

TeoTwawki: Like I said in discord yesterday, probably store a total in either a mod or a variable (you will be checking those KI even if it were a mod) and then multiply by that value at the end just before exp dispersement.

TeoTwawki:
> Teo's going to hate this, but as one can see from my experimenting with these messages, there are times it's treated as a player var. 

I assume you mean I hate the extra char_vars table entries. What I don't know is what you mean by "treated as" one. Retail doesn't have these things, it has behaviors we made crap up to act like retail. The reason we (not just me) get on everyon's case about players vars is they have been misused in the past to holy hell and back. char_vars are explicitly for things that need to persist, and when we are done with them they need zero'd out after. Anything not needing to persist should be something else, and anything that will -never- be zero'd should be stored elsewhere permanently (homepoints..)

ibm2431:
When it comes to reminders by NPCs on retail (general quest reminders), they predominantly fall into different versions of "local var" - the NPC either says it once and then doesn't say it again until you zone, or the NPC toggles between two dialogues. No permanent var to worry about to track whether or not an NPC reminded a player where to go.

But for ROV, some of these "unable to continue due to event in other expansion" messages _do_ keep permanent track of them being already flagged. If you trip it, it displays once, and then never again, even across zoning and logging. (To be clear, these aren't cutscenes that actually advance anything, just messageSpecials.) So to accurately emulate this behavior for some dumb message, "moar char vars" are needed, rather than just sufficing with a simple "local".

Another issue is that the ROV "unable to proceed" message behavior isn't always _consistent_. Some times it is "once only ever" but other times it's "every time you trip the check". So care might be needed to not confuse instances of "always notify" with "notify once" and potentially leaving the reminder var behind: "oh, this ROV should be the always remind type so a var wasn't set, the var doesn't need to be cleared out".

TeoTwawki:
A helper for RoV sounds plausible given the number of instances RoV is likely to have of it, but yeah as long as they are wiped after we no longer need them, this is what char_vars are for.

zach2good:
Messed up my rebase 😰, will come back to this soon

kaincenteno:
yay @zach2good  u are back 😸

scripts/zones/Port_Windurst/Zone.lua
```diff
@@ -15,6 +15,11 @@ end;
 
 function onZoneIn(player,prevZone)
     local cs = -1;
+
+    if ENABLE_ROV and player:getCurrentMission(ROV) == dsp.mission.id.rov.RHAPSODIES_OF_VANADIEL and player:getMainLvl()>=3 then
```
zynjec: You have to explicitly check for 1 here, only `nil` and `false` are truly 'false' in Lua.

Also if/until the quest rewrite stuff makes it in to master, it still makes sense to have this after the opening cutscene, nothing trumps that.

(Both these issues exist with the starting cutscenes you added for Seekers of Adoulin as well)

scripts/zones/Windurst_Walls/Zone.lua
```diff
@@ -18,6 +18,13 @@ end;
 
 function onZoneIn(player,prevZone)
     local cs = -1;
+
+    if ENABLE_ROV and player:getCurrentMission(ROV) == dsp.mission.id.rov.RHAPSODIES_OF_VANADIEL and player:getMainLvl()>=3 then
```
nfaletra: 0 evaluates to tru in lua. ENABLE_ROV will always be true in this case. XD

nfaletra: gdi bot

zach2good: I was wondering why this wasn't working... 🤡

kaincenteno:
keep up the great work @zach2good  😸

PR #6201, opened 2019-08-23
Dynamis Mixin adjustments

wrenffxi:
### Related adjustments
* Add missing beastmen NM scripts.  These don't contain any special battle mechanics.  Just 2hr and dyna_beastmen mixins.
* Remove from mob scripts ASTRAL_FLOW_MAAT and EES_* job special declarations, because these are defaults.
* SQL: assign QuadavNM family to Dynamis Quadav NMs.

### Beastmen Mixin
* Make beastmen currency family-based rather than declaring it in every beastmen script.
* Add THlevel adjustment.  ~~I added a getDeathTH binding, because the getTHlevel() binding always reports 0 by the time mob death listener is triggered.~~
* Add a base hundo drop slot to NMs. (starts at 0.5%, reaches 5% at TH4, which [matches bgwiki](https://www.bg-wiki.com/bg/Ree_Nata_the_Melomanic))

### Dreamworld Mixin
* Add THlevel adjustment.
* Make mixin work for mobs that don't have specific currency (e.g. Nightmare Taurus)
* Add a base hundo drop slot to NMs.


TeoTwawki:
```
Add THlevel adjustment to beastmen mixin. I added a getDeathTH binding, because the getTHlevel() binding always reports 0 by the time mob death listener is triggered.
```
Ironically that binding is exactly what my original implementation was, before I had to change it because it was reporting 0 before mob death because of someone else's change moving the line that sets TH onto mob. (it used to be non zero in both cases)

Maybe we should add a conditional checking if mob is dead, so it can be one binding again?

wrenffxi:
Rebased

PR #6209, opened 2019-09-02
Fixed lua completeQuest saving quest list to DB even if no changes ar…

Cryptexx:
…e made.


Cryptexx:
Been in limbo for a while, so I'll just reiterate that the main problem with this is repeatable quests, like Only The Best, triggering a write operation for each subsequent trade. It's already completed, so no need to update the Quests column.

Issue #6166, opened 2019-07-22
Conquest not working properly


Starfret:
**_I have:_**

- [X] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [X] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_**  **:30190605_0** 

**_Source Branch_** (master/stable) **: Master** 

**_Additional Information_** (Steps to reproduce/Expected behavior) **: Even if on the "Region Info" map the zones are under San Dorian control,** , vendors do not sell items just like the zones were still under beastmen control, and the guards at the gates can't assign supply quests. There are no guards at the outposts in the zones under San Dorian Control, and mobs do not drop crystals.

zynjec:
Related to https://github.com/DarkstarProject/darkstar/issues/6047

Fails to get region owners because of VS bug.

Issue #6047, opened 2019-06-02
Conquest system sql error crashing map server


thunderdragon21:
[02/Jun] [02:31:53][1;35m[SQL][0m DB error - You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 1
SQL: SELECT region_control FROM conquest_system WHERE region_id = 
I am not sure what is causing it. Just updated to the new stuff of the site and now getting this problem Id appreciate any help.

KnowOne134:
Problem with vs19. Downgrade back to vs2017 and toolset 241 until it's fixed

thunderdragon21:
ty but where do i find toolset 241 i looked all over for it?

TeoTwawki:
toolset 141, not 241.

![toolset](https://user-images.githubusercontent.com/6871475/58769182-82e01400-8572-11e9-9f18-058d27fed970.jpg)

thunderdragon21:
Severity	Code	Description	Project	File	Line	Suppression State
Error	MSB8020	The build tools for v142 (Platform Toolset = 'v142') cannot be found. To build using the v142 build tools, please install v142 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution".	DSConnect-server	C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Microsoft.Cpp.Platform.targets	67	
Getting this error with 141.

TeoTwawki:
because you need to select v141 before it can use it..it litteraly tld you how to fix it in that message you pasted.

thunderdragon21:
ok i got it now ty sorry just dawned on me

ZatuyouTantou:
`int` ret = Sql_Query(SqlHandle, Query.c_str(), region);`

use static_cast.

`int ret = Sql_Query(SqlHandle, Query.c_str(), static_cast<uint8>(region));`

Begocent:
Hello, what i did to fix this error was to change the enum REGIONTYPE from uint8 to uint16 at src\map\zone.h and now i compile the project with vs19.

PR #6210, opened 2019-09-02
Conquest system sql bugfix

DFPercush:
Addressing the conquest system sql bug reported in issues #6047 and #6166. The call to sprintf() in template function Sql_Query() treats the enum REGIONTYPE as an untranslated binary value, resulting in invalid ascii characters when building the query string. Applied casts to a more primitive integer type in several places.

### DIFF
--- a/src/map/conquest_system.cpp
+++ b/src/map/conquest_system.cpp
@@ -104,8 +104,8 @@ namespace conquest
 
         influences[nation] += lost;
 
-        Sql_Query(SqlHandle, "UPDATE conquest_system SET sandoria_influence = %u, bastok_influence = %u, "
-            "windurst_influence = %u, beastmen_influence = %u WHERE region_id = %u;", influences[0], influences[1], influences[2], influences[3], static_cast<uint8>(region));
+        Sql_Query(SqlHandle, "UPDATE conquest_system SET sandoria_influence = %d, bastok_influence = %d, "
+            "windurst_influence = %d, beastmen_influence = %d WHERE region_id = %u;", influences[0], influences[1], influences[2], influences[3], static_cast<uint8>(region));
     }
 
     /************************************************************************
--- a/src/map/conquest_system.cpp
+++ b/src/map/conquest_system.cpp
@@ -104,8 +104,8 @@ namespace conquest
 
         influences[nation] += lost;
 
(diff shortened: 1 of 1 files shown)


src/map/conquest_system.cpp
```diff
@@ -105,7 +105,7 @@ namespace conquest
         influences[nation] += lost;
 
         Sql_Query(SqlHandle, "UPDATE conquest_system SET sandoria_influence = %d, bastok_influence = %d, "
-            "windurst_influence = %d, beastmen_influence = %d WHERE region_id = %d;", influences[0], influences[1], influences[2], influences[3], region);
+            "windurst_influence = %d, beastmen_influence = %d WHERE region_id = %d;", influences[0], influences[1], influences[2], influences[3], static_cast<uint8>(region));
```
TeoTwawki: Since its unsigned that should be `%u` instead of `%d`.

Also this only needs cast because of a compiler bug. The first toolset that shipped with VS2019 didn't have this problem nor did any past ones, this all started with updates to visual studio. Dunno if anyone reported it to MS yet.. We should still make it explicit anyway but I think thats why nobody made this change yet. ¯\\\_(ツ)_/¯

kaincenteno: will there be any issues if this is changed for ppl that compile on linux?

TeoTwawki: doubt it, prolly be just fine. travis is stuck atm or I'd point to what it says - travis builds linux for us while appveyor builds windows.

TeoTwawki:
@DFPercush could you make the %u change and push it? I wouldn’t insist on it but travis is stuck in a way where I can’t even force restart the build, which is preventing me from merging. I think it travis sees a new push it’ll make it go. For some reason travis doesn’t even see this pr at all yet. If that doesn’t work I’ll contact travis-ci devs about it.

TeoTwawki:
@teschnei  Travis CI is saying "abuse detected" on all commits it tried to build from this Pull Request. I have emailed Travis CI support about it but they are notorious for taking forever to respond to  email. They have a forum too, but there are several complaints of this exact problem that either took awhile to get a reply or..never did get a reply at all. For now the problem seems isolated to just this one pull request.

TeoTwawki:
_kicks travis_ _**Move!**_

PR #6212, opened 2019-09-04
Fix the Pet Crashing in Battlefield

Safhaven:
Fix the Pet Crashing in Battlefield. Removed a few delete calls that were causing a delete loop

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -467,6 +467,7 @@ bool CBattlefield::RemoveEntity(CBaseEntity* PEntity, uint8 leavecode)
                 {
                     m_AllyList.erase(std::remove_if(m_AllyList.begin(), m_AllyList.end(), check), m_AllyList.end());
                 }
+                PEntity->status = STATUS_DISAPPEAR;
                 return found;
             }
             else
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -463,9 +463,10 @@ bool CBattlefield::RemoveEntity(CBaseEntity* PEntity, uint8 leavecode)
                 if (static_cast<CPetEntity*>(PEntity)->isAlive() && PEntity->PAI->IsSpawned())
                     static_cast<CPetEntity*>(PEntity)->Die();
 
-                m_AllyList.erase(std::remove_if(m_AllyList.begin(), m_AllyList.end(), check), m_AllyList.end());
-                GetZone()->DeletePET(PEntity);
-                delete PEntity;
+                if (m_AllyList.size() > 0)
+                {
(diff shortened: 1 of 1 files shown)


src/map/battlefield.cpp
```diff
@@ -463,9 +463,10 @@ bool CBattlefield::RemoveEntity(CBaseEntity* PEntity, uint8 leavecode)
                 if (static_cast<CPetEntity*>(PEntity)->isAlive() && PEntity->PAI->IsSpawned())
                     static_cast<CPetEntity*>(PEntity)->Die();
 
-                m_AllyList.erase(std::remove_if(m_AllyList.begin(), m_AllyList.end(), check), m_AllyList.end());
-                GetZone()->DeletePET(PEntity);
-                delete PEntity;
+                if (m_AllyList.size() > 0)
```
ShelbyZ: You should not need to check the size of the object as erase with begin == end will result in no-op - https://en.cppreference.com/w/cpp/container/vector/erase

src/map/battlefield.cpp
```diff
@@ -463,9 +463,10 @@ bool CBattlefield::RemoveEntity(CBaseEntity* PEntity, uint8 leavecode)
                 if (static_cast<CPetEntity*>(PEntity)->isAlive() && PEntity->PAI->IsSpawned())
                     static_cast<CPetEntity*>(PEntity)->Die();
 
-                m_AllyList.erase(std::remove_if(m_AllyList.begin(), m_AllyList.end(), check), m_AllyList.end());
-                GetZone()->DeletePET(PEntity);
-                delete PEntity;
```
ShelbyZ: Are you sure it is not needed to delete the PEntity when removed from the battlefield?

src/map/battlefield.cpp
Safhaven: I did leave it off but it was crashing on it without the if statement..  i can try it again

src/map/battlefield.cpp
Safhaven: correct because the zone_entities.cpp handles the deletion already.
but I just saw that this function can be called from Lua.
so i will add a line "PEntity->status = STATUS_DISAPPEAR".
This will allow the zone_entities.cpp to pick up the deletion correctly and remove the Pet.

kaincenteno:
Latest commit (311b1a29e111f4f391b76a8c6b3b73cdbf9c0878) doesn't resolve the issue with pets on dead/releasing crashing BCNMs.

kaincenteno:
@TeoTwawki Tested today, on release/dead works fine when applying this fix. Unsure if it was fixed after with the help of the code from https://github.com/DarkstarProject/darkstar/pull/6294 but it works great now =) The fiance will be happy to finally be able to enter BCNMs as a Summoner.

PR #6214, opened 2019-09-06
Add mission / quest NATION.MISSION / AREA.QUEST table references

ibm2431:
`I was about to muse on the PR that since lua tables are by reference, if we wanted NATION.MISSION to also be a valid option, could assign dsp.mission.id.NATION = dsp.mission.id.nation in the global without losing any real performance (since we wouldn't be creating a second copy in memory)`

### DIFF
--- a/scripts/globals/missions.lua
+++ b/scripts/globals/missions.lua
@@ -640,6 +640,20 @@ dsp.mission.id =
     [dsp.mission.area[dsp.mission.log_id.CAMPAIGN]] = {},
 }
 
+-- Additional optional uppercase references to the mission ID tables (assigned by reference)
+dsp.mission.id.SANDORIA = dsp.mission.id.sandoria
+dsp.mission.id.BASTOK   = dsp.mission.id.bastok
+dsp.mission.id.WINDURST = dsp.mission.id.windurst
+dsp.mission.id.ZILART   = dsp.mission.id.zilart
+dsp.mission.id.COP      = dsp.mission.id.cop
+dsp.mission.id.TOAU     = dsp.mission.id.toau
+dsp.mission.id.WOTG     = dsp.mission.id.wotg
+dsp.mission.id.ACP      = dsp.mission.id.acp
+dsp.mission.id.AMK      = dsp.mission.id.amk
+dsp.mission.id.ASA      = dsp.mission.id.asa
+dsp.mission.id.SOA      = dsp.mission.id.soa
+dsp.mission.id.ROV      = dsp.mission.id.rov
+
 -- Assault and Campaign IDs deliberately left out of mission table
 -- due to their nature not being the same as other missions,
 -- and to allow the content the freedom to develop systems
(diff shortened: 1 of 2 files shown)


zynjec:
That's duplicating some pretty big tables for not much reason.

ibm2431:
https://www.lua.org/pil/2.5.html

zynjec:
Huh, guess I'm just retarded.

ibm2431:
Nah, Lua is pretty weird.

Cryptexx:
He's right in that it's referenced. 

I often use local t = some.stupidly.deep.hierarchy.of.a.table and it doesn't create a deep copy of the data.  Just stores the reference to that table in local t. As an interesting note too, I found that localizing tables and functions actually speed things up when it's referenced multiple times in the same function.

Kinda like using local ran = math.random is better than calling math.random multiple times in sequence.

In this case though, no performance gain or loss, just naming conventions/preferences.

ibm2431:
To be fair, I can't state _how_ useful it would be.

Mainly just a, "hey, we could do this if we're concerned about case sensitivity". And so here we are.

My best articulation would be:
"Since we use UPPERCASE for the final suffix for nations and mission/quest log areas (`dsp.nation.BASTOK`, `getQuestStatus(JEUNO,`) , it's easy to get into a habit of capitalizing those as well. This way there wouldn't be any issues if that happens either through scripter habit or preference."

Of course, on the _other_ hand:
"We should stick to the same convention for all tabled enums so there isn't confusion for people using _other_ tables."

This was a 5-minute commit; I wouldn't be offended if people decide one way or the other...!

nfaletra:
Ha. Reading is hard. I lean towards consistency over supporting multiple formats. I haven’t looked at these lua files in awhile, is there a reason it’s defined in a lower case manner to begin with?

ibm2431:
Just because the lower.lower.FINAL is the style elsewhere (Nations, Key Items, Magic Elements, etc).

For mission IDs, areas being lower does make the FINAL_REFERENCE visually prominent though. It might just be a "where do we want to apply visual weight" thing:

"If someone is looking through a script, do we want to draw their attention to `mission.WINDURST.THE_HEART_OF_THE_MATTER` or `mission.windurst.THE_HEART_OF_THE_MATTER`?"

TeoTwawki:
we had a convention collision during some major renovations. Prior to the mass tabling of all the things the global variables were all in cap. all the tables were done in lowercase and new non tabled variables are being done lowercase... so new things in general started being done full lowercase and mingling has begun. We didn't set a prominent guideline on how to do this stuff soon enough and its catching up to us now.

unrelated: also stuff became very wide thanks to nested tables, [which has lead to some ultra short variables to compensate](https://github.com/DarkstarProject/darkstar/blob/817225f1eda7dc68ada684090183a4ba96dafb04/scripts/globals/bcnm.lua#L443) (2 letter variables bother me, because I expect them to accidentally get used twice at some point)

takhlaq:
SNAKE_CASE_CAPS for enums since we have them all that way in cpp already

TeoTwawki:
> SNAKE_CASE_CAPS for enums since we have them all that way in cpp already

but now we have ENUMs in.a.table.

TeoTwawki:
tl;dr snake has a fat ass

PR #6220, opened 2019-09-14
Zone Mixin

Omnione:
This should enable any zone wide mixins to be applied,

I've set it up so the user can create a mixin as normal and place the file in the mixin folder.

To apply the mixin the user will have to create a script in the zones folder under mixins/zones (if the script does not allready exist) and the script must be named the same as the zone is named under scripts/zones/zonename, if it does exist then the user will just be able to add a new line.
I am not to sure on how deep the complexity can be with this setup as fo far i have only been using the line: mixins = {require("scripts/mixins/spawn_casket")} to apply the mixin.

Hopefully this is ok.


teschnei:
make sure it doesn't apply to non-mobs in the zone, like npcs and pets
and it's probably cleaner to just have the logic in ApplyMixins instead of its own function, since it is "applying" mixins, if that can work

Omnione:
This should be ok now, ive tested it to see if it runs on player pets or npc's and it does not, I have combined everything into ApplyMixins and removed any remaining ApplyZoneMixins refrences.

PR #6223, opened 2019-09-21
Cleaned up caskets in npc_list.sql ready for Caskets.

Omnione:
Prep for later addition of treasure caskets.


TeoTwawki:
Diff kinda big I’ll try and check it out sunday if nobody does before then. Just letting you know its not in limbo

Omnione:
I noticed after doing some testing a few lines were incorrect i.e. 'Treasure Casket','Treasure Casket', should have been, 'Treasure_Casket','Treasure Casket', fixed that now.

Omnione:
Sorry if i just made this longer than it was before, found more entries with duff info, should all be good now, and all up to date with master too.

PR #6229, opened 2019-09-28
ZM14 Offset: Return from local function instead of initializing

Cryptexx:
The ZM14 SHIMMERING_CIRCLE_OFFSET depends on the La'Loff Amphitheater zone being loaded in the current map server instance. 

If multiple map servers are being used, the arkCircleOffset variable in bcnm.lua attempts to localize using an unknown reference when La'Loff isn't running on that server. Moving it to the return of a local function avoids this conflict as it is only called when the correct zone needs this info from registerReqs, thus ensuring the zone IDs are loaded when called.

### DIFF
--- a/scripts/globals/bcnm.lua
+++ b/scripts/globals/bcnm.lua
@@ -459,7 +459,10 @@ function checkReqs(player, npc, bfid, registrant)
     local stc = player:hasCompletedMission(SANDORIA, mi.sandoria.SAVE_THE_CHILDREN)
     local dm1 = player:getQuestStatus(OUTLANDS, dsp.quest.id.outlands.DIVINE_MIGHT)
     local dm2 = player:getQuestStatus(OUTLANDS, dsp.quest.id.outlands.DIVINE_MIGHT_REPEAT)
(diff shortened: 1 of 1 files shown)


ibm2431:
I'll admit I can't find any other instances of this issue, but I still have this niggling feeling that it might look nicer to have a generalized function for returning an offset (`getOffset(zone, npc, offset)`) rather than a BC-specifically-named one. But I'm coming up blank on how to pass the right NPC in with those arguments without essentially "detabling" the original NPC reference.

(Ideally, you could just have the `full NPC reference + offset` inside the requirement check itself, since those also serve as thunks, but I imagine line length in the requirements is a concern.)

I'd muse that splitting the offset into the appropriate AA circles (`zones[dsp.zone.LALOFF_AMPHITHEATER].npc.AA_TT_CIRCLE`) inside the zone's ID file gets around the _need_ for an offset in `bcnm.lua`, while adding additional clarity to the requirement check. But that still might make the requirement checks too long.

...However, after skimming the list, I see that no other BCNM seemingly needs to check the current NPC ID. I'll admit unfamiliarity with BCNM operation, but is there a reason that the AA fights need to check the NPC ID when others don't?

PR #6230, opened 2019-10-02
Created new global file to be used for past event watcher NPCs

kaincenteno:
For now only added 1 Goblin before having to make huge
changes. Will work on other Goblins once it is approved
and merged, planning on at least getting the base game
area ones scripted.


scripts/globals/pasteventwatcher.lua
```diff
@@ -0,0 +1,60 @@
...
+    local bit10 = (1-options[9])*512
+    local bit9  = (1-options[8])*256
+    local bit8  = (1-options[7])*128
+    local bit7  = (1-options[6])*64
+    local bit6  = (1-options[5])*32
+    local bit5  = (1-options[4])*16
+    local bit4  = (1-options[3])*8
+    local bit3  = (1-options[2])*4
+    local bit2  = (1-options[1])*2
+    local bit1  = 1-1*1
```
zynjec: https://www.youtube.com/watch?v=U1UtRnGn5hc

scripts/globals/pasteventwatcher.lua
```diff
@@ -0,0 +1,60 @@
...
+
+    local gobbiePacket = bit32 + bit31 + bit30 + bit29 + bit28 + bit27 + bit26 + bit25 + bit24 +
+    bit23 + bit22 + bit21 + bit20 + bit19 + bit18 + bit17 + bit16 + bit15 + bit14 + bit13 + bit12 + bit11 +
+    bit10 + bit9 + bit8 + bit7 + bit6 + bit5 + bit4 + bit3 + bit2 + bit1
```
ibm2431:
```
local packet = 0
for i = 1, 31, do
    if options[i] then
        packet = packet + 2^i
    end
end
return packet
```

(I don't remember if the `hasCompleted` quests return actual true/false or 1/0. If 1/0, then just make the check `if options[i] > 0`)

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
```diff
@@ -0,0 +1,66 @@
...
+--  NPC: Goblin Footprint
+-- Type: NPC
+-- !pos  111.806 -24.007 57.266 167
+-----------------------------------
+local ID = require("scripts/zones/RoMaeve/IDs")
+require("scripts/globals/pasteventwatcher")
+require("scripts/globals/quests")
+require("scripts/globals/missions")
+-----------------------------------
+---
```
ibm2431: Don't need the ---

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
```diff
@@ -0,0 +1,66 @@
...
+    end
+    if player:hasCompletedMission(WINDURST,dsp.mission.id.windurst.MOON_READING) == true then
+        options[2] = 1
+    end
+    if player:getQuestStatus(ZILART,dsp.quest.id.outlands.DIVINE_MIGHT) == QUEST_COMPLETED then
+        options[3] = 1
+    end
+    if player:hasCompletedMission(ASA,dsp.mission.id.asa.FOUNTAIN_OF_TROUBLE) == true then
+        options[4] = 1
+    end
```
ibm2431:
```
local options =
{
    player:hasCompletedMission(WINDURST,dsp.mission.id.windurst.VAIN),
    player:hasCompletedMission(WINDURST,dsp.mission.id.windurst.MOON_READING),
    player:hasCompletedQuest(ZILART,dsp.quest.id.outlands.DIVINE_MIGHT),
    player:hasCompletedMission(ASA,dsp.mission.id.asa.FOUNTAIN_OF_TROUBLE)
}
```

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
```diff
@@ -0,0 +1,66 @@
...
+    if not next(options) == nil then
+        player:startEvent(6, eventList, 4294967294, 4294967294, 4294967294)
+    end
+end
+
+function onTrigger(player,npc)
+    if player:hasCompletedMission(WINDURST,dsp.mission.id.windurst.VAIN) == true or
+        player:hasCompletedMission(WINDURST,dsp.mission.id.windurst.MOON_READING) == true or
+        player:getQuestStatus(ZILART,dsp.quest.id.outlands.DIVINE_MIGHT) == QUEST_COMPLETED or
+        player:hasCompletedMission(ASA,dsp.mission.id.asa.FOUNTAIN_OF_TROUBLE) == true then
```
ibm2431: It'd be nice to have a helper function that accepts a player and table of quests/missions, and then depending on a passed in flag, either builds the full options table (above) for the trade, or simply goes through the checks and returns true on the first successful check (for checking if the goblin should ask for items).

I don't know if it'd be best to come up with a way of describing quest/mission checks in a table, so a table could be defined for each specific goblin:
```
local checks =
{
    {mission, windurst, vain},
    {mission, windurst, moon_reading}
}
```
And then have the helper function build options tables or onTrigger check as appropriate; or

Just use a table of lambdas (kind of ugly, but easier to implement):
```
local checks =
{
    function(player) player:hasCompletedMission(WINDURST,dsp.mission.id.windurst.VAIN) end,
    function(player) player:hasCompletedMission(WINDURST,dsp.mission.id.windurst.MOON_READING) end
}
```

And then in the helper function, just invoke each of the lambdas in order (`checks[i](player)`) and store the result into a table. If the "on trigger checking" flag is passed into the helper, abort on the first invokation that's `true`, otherwise fill out the table and return that.

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
```diff
@@ -0,0 +1,66 @@
...
+        end
+        if option == 2 then
+            player:startEvent(4)            -- Moon Reading
+        end
+        if option == 3 then
+            player:startEvent(7,917,1408)   -- Divine Might
+        end
+        if option == 4 then
+            player:startEvent(71)           -- Fountain of Trouble
+        end
```
ibm2431:
```
local option_cses =
{
    [1] = function(player) player:startEvent(2) end,
    [2] = function(player) player:startEvent(4) end,
    [3] = function(player) player:startEvent(7,917,1408) end
}

onFinish
    if csid == 6
        if goblin_cses[option] then
            -- Bonus points if you use the option checks from before to double-check if the player meets the requirement before playing the event (ie: make sure client isn't attempting to forge packet with an option they're not qualified for)
            goblin_cses[option](player)
        end
    end
end
```

You can get a cleaner result if you're willing to play with `unpack` to handle event params, letting you kill the lambdas outright.

ibm2431:
To whoever might know:

Will the onFinish for "zone events" still trigger if the finish is defined in the zone script, but the event is started outside of it (these NPCs)?

Is this a potential exploit we should be wary of?

scripts/globals/pasteventwatcher.lua
nfaletra: All of this should be using bitwise operations. They are infinitely faster. Brush up on them and how to do them in lua.

nfaletra: wtf. just set it to 0.

scripts/globals/pasteventwatcher.lua
nfaletra: This should all be using bitwise operations. It's already shifted into place. Just & them all together.

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
```diff
@@ -0,0 +1,66 @@
...
+    if player:hasCompletedMission(WINDURST,dsp.mission.id.windurst.MOON_READING) == true then
+        options[2] = 1
+    end
+    if player:getQuestStatus(ZILART,dsp.quest.id.outlands.DIVINE_MIGHT) == QUEST_COMPLETED then
+        options[3] = 1
+    end
+    if player:hasCompletedMission(ASA,dsp.mission.id.asa.FOUNTAIN_OF_TROUBLE) == true then
+        options[4] = 1
+    end
+    local eventList = dsp.pasteventwatcher.goblinFootprint({options[1], options[2], options[3], options[4]})
```
nfaletra: Can't you just pass `options` here?

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
```diff
@@ -0,0 +1,66 @@
...
+    if not next(options) == nil then
+        player:startEvent(6, eventList, 4294967294, 4294967294, 4294967294)
```
nfaletra: This looks like you're hardcoding garbage data from a packet capture. Remember that params are not zeroed out for every packet on retail. You may be getting bad data for unused params based on what packet was sent prior to this one.

scripts/globals/pasteventwatcher.lua
```diff
@@ -0,0 +1,60 @@
...
+        if not options[i] then
+            options[i] = 0
+        end
+    end
+
+    -- 32 bits represents a cutscene being watched or not
+    -- The first bit defaults to 0 to show the "Another time maybe" choice to let user get out from the dialog
+    -- The first scene starts at bit2, and it has to be 0 if it becomes part of the list, otherwise it should be a 1.
+    -- If you need to know the correct order of the scenes use !cs <scene number> for the gobbie to list them all
+    local bit32 = (1-options[31])*2147483648
```
nfaletra: I'm a little confused here. Where are these ever being used? You only ever pass in an options array with 4 elements. Are there missing NPCs that aren't implemented yet?

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
```diff
@@ -0,0 +1,66 @@
...
+    else
+        player:messageSpecial(ID.text.NOTHING_OUT_OF_ORDINARY)
+    end
+end
+
+function onEventUpdate(player,csid,option)
+end
+
+function onEventFinish(player,csid,option)
+    local spokeWithGobbie = 1
```
nfaletra: use true/false, man.

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
```diff
@@ -0,0 +1,66 @@
...
+        if option == 2 then
+            player:startEvent(4)            -- Moon Reading
+        end
+        if option == 3 then
+            player:startEvent(7,917,1408)   -- Divine Might
+        end
+        if option == 4 then
+            player:startEvent(71)           -- Fountain of Trouble
+        end
+    elseif spokeWithGobbie == 1 and csid == 2 or csid == 4 or csid == 7 or csid == 71 then
```
nfaletra: Wait. isn't `spokeWithGobbie` always 1? What is the point of it?

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
nfaletra: this ^

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
```diff
@@ -0,0 +1,66 @@
...
+        player:messageSpecial(ID.text.GOBLIN_SLIPPED_AWAY)
+    end
+    spokeWithGobbie = 0
```
nfaletra: What is this accomplishing?

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
nfaletra: you would break out of the loop if it returns false.

TeoTwawki:
> To whoever might know:
>
> Will the onFinish for "zone events" still trigger if the finish is defined in the zone script, but the event is started outside of it (these NPCs)?
>
> Is this a potential exploit we should be wary of?
 
If started by an npc nothing in the zone script matters. If started by a non npc, for example a gm command..then stuff in the zone script executes which can be annoying if you need to explore options codes using `!cs`

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
kaincenteno: you are right, is not need it. I'll remove it. Thank you 😸

scripts/globals/pasteventwatcher.lua
Wiggo32: How bout an example of what you'd like? I'm sure if he knew what was expected it would have already been done that way. thanks to @ibm2431 for being very useful to people who just want to help and contribute

zynjec: >  I'm sure if he knew what was expected it would have already been done that way

Common sense is a good place to start, if it needs to be zero, use a zero.

Wiggo32: So what I am taking from all of this is: What this overly bloated and unnecessarily demeaning review should have said is:
```
use 0 here:
local bit1  = 1-1*1
```

zynjec: Take your white knight shit elsewhere.

Wiggo32: Gladly! Kindly point me in the direction of somewhere else where they demean, bully, and ridicule for fun... :thinking:

Wiggo32: @teschnei you condone all of this?

Wiggo32: a DSP OP bans me from discord for this thread

Wiggo32: you deleted everything except the video mocking this poor guys mistake? @teschnei

teschnei: for the bitwise functions, we automatically require the "bit" library in every Lua script, so you can use `bit.and` etc. without needing to require it (this one, I think: http://bitop.luajit.org/)

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
ibm2431: Can make an assignment to the result (`blah = checks[i](player)`), and then check the assigned var to determine if you should keep going through your loop (`while ((not blah) or build_options) and i <= #checks`), rather than invoking each check directly and hitting a return that breaks you out.

(And if you should keep going, pop the result into the table for when the function is called with the "build options" flag)

```
function checkOptions(player, checks, build_options)
    local options = {}
    local check = false
    local i = 1
    while ((not check) or build_options and i <= #checks ) do
        check = checks[i](player)
        if check and not build_options then
            return true
        else
            options[i] = check
        end
    end
    if build_options then
        return options
    else
        return false
    end
end

local checks = {
    -- I'll spare the lambdas
}

onTrigger
    if not checkOptions(player, checks, false) then
        player:startEvent(6) -- You see nothing
    end
end

onTrade
    local options = checkOptions(player, checks, true)
end
```

Alternatively, if you're willing to sacrifice being able to abort early on the first true result for your onTrigger if check, you could just always build the table up front (without any sort of flag) and use a dual return.

```
function checkOptions(player, checks)
    local options = {}
    local has_options = false
    local i = 1
    while (i <= #checks ) do
        options[i] = checks[i](player)
        if options[i] then
            has_options = true
        end
    end
    return has_options, options
end

onTrigger
    local has_options, options = checkOptions(player, checks)
    if has_options then
        -- Items la la
    end
end

onTrade
    local has_options, options = checkOptions(player, checks)
    if has_options then
        player:startEvent(X, options)
    end
end
```

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
kaincenteno: removed

scripts/globals/pasteventwatcher.lua
kaincenteno: set to zero 😸

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
kaincenteno: I removed the local var as it was not needed 😸

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
kaincenteno: Hi @nfaletra  thanks for taking time on reviewing my PR 😸

The three parameters of 0xFFFFFFFE are needed to be included for the goblins else it won't populate the table. I'm attaching 3 packets for gobbies in Zilart and Base game area. (Thanks to @UynGH 's hard work)
gobbie
gobbie3
gobbie2

scripts/globals/pasteventwatcher.lua
kaincenteno: Hi @nfaletra this PR is for the function logic to be scrutinized with 1 goblin before I start writing the other logic for all the other goblins as none has been scripted in DSP, have already tested it in other 3 areas but my logic looks horrible for the other ones but wanted to get the go ahead on this first before spending more time and having to rewrite it, since theres a ton of gobbies lol  😺 notes are on the 1st comment of PR (sorry if i wasn't clear, English is my second language)

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
kaincenteno: resolving this as I have removed the local var.

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
ibm2431: Does it need to be 0xFFFFFFFE, or just a certain bit that the currently used param happens to have set? (Ie: Is it really something like 255, or just 2?).

TeoTwawki: > Also, do we have a hard "no hex in params" rule? 

hex is actually preferred ***if*** its as a bitmask in the script. we de-hexed the csid themselves though.

kaincenteno:
@itsMaru its not ready yet, i need to change the bits and rebase =( will take me a bit

kaincenteno:
Made changes now it works with a nice looking global

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
TeoTwawki: looks like one of your followup commits may have put it back? there is also another above the other onTrade

kaincenteno: @TeoTwawki  thanks for catch, yeah i PR'd wrong branch lol xD
It's fixed now =)

scripts/zones/RoMaeve/npcs/Goblin_Footprint.lua
kaincenteno: Hi @nfaletra  I made the changes and now it looks nice  😸

kaincenteno:
I wanted to say thank you to @ibm2431 and @cocosolos for the help 😸

TeoTwawki:
> scripts/globals/pasteventwatcher.lua:35:74: (E011) expected ')' near 'RELIC'

says travis

zynjec:
Sorry, you'll need to update your text ID's with Decembers patch now.

scripts/globals/pasteventwatcher.lua
```diff
@@ -0,0 +1,220 @@
...
+-- Past Event Watcher NPCs
+-----------------------------------
+
+-----------------------------------
+
+require("scripts/globals/quests")
+require("scripts/globals/missions")
+require("scripts/globals/zone")
+
+pasteventwatcher = {}
```
zynjec: It'd be nice if we kind if we followed what's become the "standard" of loading everything in to the DSP table.

```lua
dsp = dsp or {}
dsp.goblinfootprint = dsp.goblinfootprint or {}

etc
```

scripts/globals/pasteventwatcher.lua
```diff
@@ -0,0 +1,220 @@
...
+    [102] = 126, -- La Theine Plateau
+    [106] = 247, -- North Gustaberg
+    [107] = 905, -- South Gustaberg
+    [119] = 35,  -- Meriphataud Mountains
+    [122] = 6,   -- Ro'Maeve
+    [162] = 10,  -- Castle Zvahl Keep
+    [167] = 100, -- Bostaunieux Oubliette
+    [198] = 67,  -- Maze of Shakhrami
+    [200] = 61,  -- Garlaige Citadel
+}
```
zynjec: Is there a reason we're not just passing the csid through the function?

If there is a reason this is needed, can we do away with magic numbers please?

```lua
local cutscenes = 
{
    [dsp.zone.LA_THEINE_PLATEAU] = 126,
    [dsp.zone.NORTH_GUSTABERG] = 247,
    ...etc
}
```

scripts/globals/pasteventwatcher.lua
```diff
@@ -0,0 +1,220 @@
...
+
+function pasteventwatcher.rewatch(player, trigger)
+    local zone = player:getZoneID()
+    local ID = zones[zone]
+    local options = 1
+    local cs = -- add checks to this table
```
zynjec: Can we move this outside of the function? The table doesn't need to be constructed every time someone watches a cutscene, just reference it.

scripts/globals/pasteventwatcher.lua
```diff
@@ -0,0 +1,220 @@
...
+    {
+        [102] = -- La Theine Plateau
```
zynjec: Death to magic numbers in this whole table.

```lua
local cutscenesList = 
{
    [dsp.zone.LA_THEINE_PLATEAU] = {},
    [dsp.zone.NORTH_GUSTABERG] = {},
    etc
}
```

scripts/globals/pasteventwatcher.lua
```diff
@@ -0,0 +1,220 @@
...
+            [6] = function() return (player:hasCompletedMission(COP, dsp.mission.id.cop.THE_MOTHERCRYSTALS)) end,
+            [7] = function() return (player:hasCompletedMission(COP, dsp.mission.id.cop.THREE_PATHS)) end,
+            [8] = function() return (player:completeQuest(OTHER_AREAS_LOG, dsp.quest.id.otherAreas.WAKING_THE_BEAST)) end,
+            [9] = function() return (player:completeQuest(OTHER_AREAS_LOG, dsp.quest.id.otherAreas.WAKING_THE_BEAST)) end,
+            [10] = function() return (player:hasCompletedQuest(JEUNO, dsp.quest.id.jeuno.CHOCOBO_ON_THE_LOOSE)) end,
+            [11] = function() return (player:hasCompletedQuest(ABYSSEA, dsp.quest.id.abyssea.A_GOLDSTRUCK_GIGAS)) end,
+            [12] = function() return (player:hasCompletedQuest(ABYSSEA, dsp.quest.id.abyssea.A_GOLDSTRUCK_GIGAS)) end,
+            [13] = function() return (player:hasCompletedQuest(ABYSSEA, dsp.quest.id.abyssea.AN_OFFICER_AND_A_PIRATE)) end,
+            [14] = function() return (player:hasCompletedQuest(ABYSSEA, dsp.quest.id.abyssea.TENUOUS_EXISTENCE)) end,
+            [15] = false,               -- QUEST NOT IMPLEMENTED (ROV)
```
zynjec: If you actually have all these bits figured out, go ahead and add them. 

Don't worry about the quests not currently working, your script will work once they do.

scripts/globals/pasteventwatcher.lua
```diff
@@ -0,0 +1,220 @@
...
+            [24] = false,               -- QUEST NOT IMPLEMENTED (ROV)
+            [25] = false,               -- QUEST NOT IMPLEMENTED (ROV)
+            [26] = false,               -- QUEST NOT IMPLEMENTED (ROV)
+
+        },
+        [106] = -- North Gustaberg
+        {
+            [1] = function() return (player:hasCompletedQuest(WINDURST, dsp.quest.id.windurst.AS_THICK_AS_THIEVES)) end,
+            [2] = function() return (player:hasCompletedQuest(BASTOK, dsp.quest.id.bastok.THE_GUSTABERG_TOUR)) end,
+            [3] = false, --function() return (player:hasCompletedQuest(APOCALIPSE RELIC WEAPON no way to track)) end,
```
zynjec: Could we not abuse `hasItem` for this check?

```lua
function () return player:hasItem(18306) end,
```

Granted, it could get cumbersome to check for all 10 potential versions from 75 to 119-3.

TeoTwawki: I think thats the only way we can here is loop a hasItem through IDs. Maybe make a helper function  in utils because checking if players owns a version of a given relic will be needed elsewhere in the future as well I am almost sure. the quests for those aren't real quests on DSP and I doubt retail has a hidden quest entry that they just never display so its that or permanent playervars and player vars will make me grind you up and feed you to the chocobos..

scripts/globals/pasteventwatcher.lua
cocosolos: I helped with these functions and I put these here so the startEvent function could send the GOBLIN_SLIPPED_AWAY text and so all the npc files would be uniform, as well as having all the csid's related to this feature in the same place.

zynjec: > and I put these here so the startEvent function could send the GOBLIN_SLIPPED_AWAY text
I'm not following, what is stopping you from doing this without the table?

cocosolos: You could, but then the npc file would need to be something like
```lua
if csid ~= gobcs then
    player:messageSpecial(ID.text.GOBLIN_SLIPPED_AWAY)
else
    dsp.goblinfootprint.startEvent(player, gobcs, option)
end
```
where gobcs is different for every npc file. I figured keeping the npc files as simple as possible would be worth tabling these id's here.

scripts/globals/pasteventwatcher.lua
cocosolos: I based this on the checkReqs function in bcnm.lua. It can be moved out but I'm not sure how to make it look cleaner than doing something like
```lua
    [dsp.zone.MERIPHATAUD_MOUNTAINS] =
    {
        reqs = function(player, i)
            return switch (i): caseof
            {
                [1] = function() return (player:hasCompletedQuest(SANDORIA, dsp.quest.id.sandoria.THE_HOLY_CREST)) end,
            }
        end,
    },
```
to use the player.

scripts/globals/pasteventwatcher.lua
cocosolos: I've also since rewritten that function as
```lua
function dsp.goblinfootprint.startEvent(player, csid, option)
    local zone = player:getZoneID()
    local ID = zones[zone]
    if csid == gobCS[zone] and cutscene[zone][option] ~= nil then
        player:startEvent(unpack(cutscene[zone][option]))
        return
    else
        player:messageSpecial(ID.text.GOBLIN_SLIPPED_AWAY)
    end
end
```
so that the text message is displayed for the goblin cs as well if the player decides to cancel.

scripts/globals/pasteventwatcher.lua
kaincenteno: fixed 😸

scripts/globals/pasteventwatcher.lua
kaincenteno: fixed 😸

scripts/globals/pasteventwatcher.lua
kaincenteno: @zynjec  don't have the correct CS ID for those ones. left them as false and with a note once those quest are implemented they can be added. (CS for ROV are weird)

scripts/globals/pasteventwatcher.lua
kaincenteno: I am leaving it as false until someone finds a better way to implement a check if the char has ever completed the quest (since if it is trade from char to char the check for it would not be 100% accurate [like if he traded it from a mule but that mule didn't really do the quest])

scripts/globals/pasteventwatcher.lua
kaincenteno: removed magic numbers.

scripts/globals/pasteventwatcher.lua
kaincenteno: cutscene moved outside function.

scripts/globals/pasteventwatcher.lua
TeoTwawki: uh..you can't trade relics to your mule bro.

kaincenteno: thanks @TeoTwawki  made changes =) wasn't aware that relics couldn't be traded =)

kaincenteno:
Hi DSP friends =)
Anything I need to change on this? I have already corrected the text IDs after December Update. (trying to get this merge before January's update)

scripts/globals/goblinfootprint.lua
```diff
@@ -0,0 +1,218 @@
...
+    },
+}
+
+function dsp.goblinfootprint.rewatch(player, trigger)
+    local zone = player:getZoneID()
+    local ID = zones[zone]
+    local options = 1
+
+    for i, check in pairs(csReq[zone]) do
+        if check(player) == true then
```
nfaletra: take off the == true

scripts/globals/goblinfootprint.lua
```diff
@@ -0,0 +1,218 @@
...
+            options = bit.bor(options, bit.lshift(1, i))
+            if trigger then
+                player:messageSpecial(ID.text.ITEMS_ITEMS_LA_LA)
+                return
+            end
+        end
+    end
+    if options > 1 then
+        player:startEvent(gobCS[zone], bit.bnot(options), -2, -2, -2)
```
nfaletra: What are these -2s? params are unsigned values. Is this from a retail capture or just unimplemented params? If it's unimplemented params, just leave a comment. If it's from the capture, it could just be garbage data. Retail does not zero out packets before sending them (dsp does). I would recommend trying to remove these extra params and see if the Goblins still work.

cocosolos: These -2's are the 4294967294 value he had there before that are [apparently required](https://github.com/DarkstarProject/darkstar/pull/6230#discussion_r330840136). I just figured -2 looked cleaner.

scripts/globals/goblinfootprint.lua
```diff
@@ -0,0 +1,218 @@
...
+    {
+        [1] = function(player) return (player:hasCompletedQuest(BASTOK, dsp.quest.id.bastok.DARK_PUPPET)) end,
+        [2] = function(player) return (player:hasCompletedQuest(WINDURST, dsp.quest.id.windurst.HITTING_THE_MARQUISATE)) end,
+        [3] = function(player) return (player:hasCompletedQuest(WINDURST, dsp.quest.id.windurst.I_CAN_HEAR_A_RAINBOW)) end,
+        [4] = function(player) return (player:hasCompletedQuest(SANDORIA, dsp.quest.id.sandoria.A_TIMELY_VISIT)) end,
+        [5] = function(player) return (player:hasCompletedQuest(SANDORIA, dsp.quest.id.sandoria.A_TIMELY_VISIT)) end,
+        [6] = function(player) return (player:hasCompletedMission(COP, dsp.mission.id.cop.THE_MOTHERCRYSTALS)) end,
+        [7] = function(player) return (player:hasCompletedMission(COP, dsp.mission.id.cop.THREE_PATHS)) end,
+        [8] = function(player) return (player:completeQuest(OTHER_AREAS_LOG, dsp.quest.id.otherAreas.WAKING_THE_BEAST)) end,
+        [9] = function(player) return (player:completeQuest(OTHER_AREAS_LOG, dsp.quest.id.otherAreas.WAKING_THE_BEAST)) end,
```
cocosolos: Should these be calling completeQuest?

kaincenteno: oh crap good catch @cocosolos I musta been asleepy when i wrote this. since it was a massive number of copy and paste lol. Going to change it now

scripts/globals/goblinfootprint.lua
kaincenteno: fixed 🐱  thank you for feedback

scripts/globals/goblinfootprint.lua
TeoTwawki: these are the 0xFFFFFFFE he explained previously, which is preferred if it must be that precise value.

https://github.com/DarkstarProject/darkstar/pull/6230#discussion_r330846095

and not just certain bits that value happened to cover

kaincenteno: @nfaletra those extra params are needed for the choices to populate on the !cs just like @cocosolos explained. He helped me make the function look cleaner as when I originally when I started this PR i had a horrible bit table as an array (you can check how horrible it was if you see first commit lol)

kaincenteno: @TeoTwawki thanks for the reminder :)
Yeah those -2 are a static value, it's not used or change in the script to present less or more choices

nfaletra: A quick comment above that line is fine, then. Thanks for breaking it down. I may investigate on retail further at some point.

Issue #6222, opened 2019-09-21
 can't get moongate pass


tourmaline3333:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **30190731_2** 

**_Source Branch_** (master/stable) **master** 

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

I entered moongate when full moon.and I found "???" on A-8.
I checked "???" but no motion.
Of course I did not get moongate pass

kaincenteno:
@tourmaline3333 It looks like the logic of it hasn't been implemented for all 4 locations

Can you get the X, Y, Z location of the pass possible locations?  (https://ffxiclopedia.fandom.com/wiki/Moongate_Pass_Quest)

From the wiki it says that it will dissapear right after being obtained by a person and will go on a respawn timer of 30 minute. Can you also check if the  "???"  moves around if not being touched?

Once I get that info I can try and script it 😸

tourmaline3333:
Thunks for your response.

???(Moongate pass)'s location is  "X:-277.7171 Y:-3.5752 Z:-17.6135"

Im only player on my server. So no one got "???".
and there is "???" on same location yesterday

kaincenteno:
Thanks for the Coordinates.

Ah @tourmaline3333  you don't have access to retail? Let me check on the table maybe the coordinates are there, it's just a bit of a pain to go online and mark em with a character lol.

I'm currently on the DSP discord chat. Feel free to send me a direct message there. My handle is Kain

tourmaline3333:
It's so difficult for me.I don't know "retail", and I started play DSP only 3 days ago.

kaincenteno:
Which server do you plat @tourmaline3333 ?

tourmaline3333:
Umm..what do you mean?
DSP Server: I build Server myself. Code -> releases -> 0.0.8318 -> darkstar_setup.exe
Official FF11 : I played on Bismark ten years ago

kaincenteno:
@UynGH  I checked the file you shared with me and couldnt find enough points containing ??? (some of them are outside the moon gates, for example I couldnt find the one at 0-8 accourding to https://ffxiclopedia.fandom.com/wiki/Moongate_Pass_Quest)

[...]

PR #6232, opened 2019-10-06
Added Moongate Pass Quest

kaincenteno:
Question mark will vanish for 30 minutes if someone that needs the pass
touches it. If someone that has the key item touches it, it won't do
anything.

This resolves https://github.com/DarkstarProject/darkstar/issues/6222

### DIFF
--- a/scripts/zones/RoMaeve/npcs/qm3.lua
+++ b/scripts/zones/RoMaeve/npcs/qm3.lua
@@ -1,0 +1,38 @@
+-----------------------------------
+-- Area: Ro'Maeve
+--  NPC: qm3 (Moongate Pass QM)
+-- !pos -277.651,-3.765,-17.895 122 and many <pos>
+-----------------------------------
+local ID = require("scripts/zones/RoMaeve/IDs")
+require("scripts/globals/npc_util")
+require("scripts/globals/settings")
+require("scripts/globals/keyitems")
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end
+
+function onTrigger(player,npc)
+    if player:hasKeyItem(dsp.ki.MOONGATE_PASS) then
+        player:messageSpecial(ID.text.NOTHING_OUT_OF_ORDINARY)
+    else
+        local moongateQMLocations =
+        {
+            {-277.651,-3.765, -17.895},
+            { 264.681, 4.000, -52.630},
+            { 278.402, 4.993,  -3.200},
+            { 151.779, 4.719,  68.553},
+            {-134.518, 4.000, 106.042}
+        }
+        npcUtil.giveKeyItem(player, dsp.ki.MOONGATE_PASS)
+        npc:hideNPC(1800)
+        local newPosition = npcUtil.pickNewPosition(npc:getID(), moongateQMLocations, true)
+        npc:setPos(newPosition.x, newPosition.y, newPosition.z)
+    end
+end
+
+function onEventUpdate(player,csid,option)
+end
+
+function onEventFinish(player,csid,option)
+end


scripts/zones/RoMaeve/IDs.lua
```diff
@@ -53,6 +53,15 @@ zones[dsp.zone.ROMAEVE] =
...
         MOONGATE_OFFSET = 17277195,
         BASTOK_7_1_QM   = 17277207,
+        MOONGATE_PASS_QM_POS =
+        {
+            {-277.651,-3.765, -17.895},
+            { 264.681, 4.000, -52.630},
+            { 278.402, 4.993,  -3.200},
+            { 151.779, 4.719,  68.553},
+            {-134.518, 4.000, 106.042},
+        },
```
ibm2431: Do we usually put Pos in these NPC ID tables? Would they not go in the NPC script?

scripts/zones/RoMaeve/npcs/qm3.lua
```diff
@@ -0,0 +1,30 @@
...
+
+function onTrade(player,npc,trade)
+end
+
+function onTrigger(player,npc)
+    if player:hasKeyItem(dsp.ki.MOONGATE_PASS) then
+        player:messageSpecial(ID.text.NOTHING_OUT_OF_ORDINARY)
+    else
+        npcUtil.giveKeyItem(player, dsp.ki.MOONGATE_PASS)
+        npc:hideNPC(3000)
```
ibm2431: I _assume_ `hideNPC` goes in seconds: 3000 seconds != 30 minutes

scripts/zones/RoMaeve/npcs/qm3.lua
```diff
@@ -0,0 +1,30 @@
...
+        local newPosition = npcUtil.pickNewPosition(npc:getID(), ID.npc.MOONGATE_PASS_QM_POS, allowCurrentPosition)
```
ibm2431: `allowCurrentPosition` not defined

scripts/zones/RoMaeve/IDs.lua
kaincenteno: @ibm2431  I have seen them for other mobs, but let me know if I should put it back inside 😸

scripts/zones/RoMaeve/npcs/qm3.lua
kaincenteno: a crap it's 1,800 seconds. Thanks @ibm2431  😸

scripts/zones/RoMaeve/IDs.lua
kaincenteno: @ibm2431  in the same file bastok quest QM also stores it there.

scripts/zones/RoMaeve/npcs/qm3.lua
kaincenteno: ah crap that was meant to be true! thank you for catch @ibm2431  😸

ibm2431:
Would anyone care to chime in on the pos-in-ID-Lua thing?

@kaincenteno is correct in that we have apparently done it with other QMs in the past, but I don't know if we _should_.

scripts/zones/RoMaeve/npcs/qm3.lua
```diff
@@ -0,0 +1,30 @@
...
+end
+
+function onTrigger(player,npc)
+    if player:hasKeyItem(dsp.ki.MOONGATE_PASS) then
+        player:messageSpecial(ID.text.NOTHING_OUT_OF_ORDINARY)
+    else
+        npcUtil.giveKeyItem(player, dsp.ki.MOONGATE_PASS)
+        npc:hideNPC(1800)
+        local newPosition = npcUtil.pickNewPosition(npc:getID(), ID.npc.MOONGATE_PASS_QM_POS, true)
+        GetNPCByID(ID.npc.MOONGATE_PASS_QM):setPos(newPosition.x, newPosition.y, newPosition.z)
```
TeoTwawki: don't have the npc "get" its own ID. It only even needs stored if some other script is going to use it.. Like say the zone script assigning this npc a random position at server start. When the NPC needs to reference ITSELF you do NOT "getbywhatever" it. You don't "get" at all.
```lua
npc:setPost(shit here)
```

scripts/zones/RoMaeve/IDs.lua
TeoTwawki: Since they are not "IDs" at all, I think those should eventually be moved to the zone helper functions
https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/PsoXja/globals.lua
like that. We use those as sort of zone specific globals - its not game wide, but it is intended for stuff used by multiple things in the zone.

For the purpose of this PR..Currently exactly 1 script will use those. They may as well be in that one script as a local table instead. Alternatively, leave them where they are but also set the position at server startup in the zone.lua. If in the zone lua, you _will_ need to "GetByID". In the NPC's own script, you _will not_ need to do that.

TeoTwawki: tl;dr don't global anything that does not need to be global.

scripts/zones/RoMaeve/npcs/qm3.lua
kaincenteno: ah thank you 😸

scripts/zones/RoMaeve/IDs.lua
kaincenteno: that makes total sense, since this is only used by one NPC it doesn't need to be global. Thank @TeoTwawki  making the changes now 😸

TeoTwawki:
Re-posting since this is hidden away by default in a resolved conversation:

>Would anyone care to chime in on the pos-in-ID-Lua thing?
> 
> @kaincenteno is correct in that we have apparently done it with other QMs in the past, but I don't know if we _should_.

Since they are not "IDs" at all, I think those should eventually be moved to the zone helper functions
https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/PsoXja/globals.lua
like that. We use those as sort of zone specific globals - its not game wide, but it is intended for stuff used by multiple things in the zone.

PR #6238, opened 2019-10-12
Adding Survival Guide Teleport System

Brian-Wuest:
Hi all, this is my first pull request for this project!

I noticed that the Survival Guides are in the game (some of them at least), but they didn't do anything. After some digging I found out that it wasn't implemented yet so I hunkered down and implemented it (about 90% of it at least). I've tried to get this to work as close to retail as possible. There are few things that don't work which I will mention below.

**What Works!**

1. Clicking on a Survival Guide for the first time registers it as a character variable in the char_vars table.
    1. I removed the extra click you have to do when registering so you can get right into the menu!
1. Any Survival Guide you have registered you can teleport too.
1. You can order by region or content. 
1. All expansions work and if the server admin has decided that the expansion isn't available then those options don't show in the teleportation options.
1. A new option was added to the "settings.lua" which will allow the players to pay the 100 gil instead of the 1000 gil since the ROV stuff isn't working yet.
1. All 98 zones have Survival Guides and can be teleported too once you register them.
    1. Yes, that's ALL of them 😄.
1. Your current zone doesn't show up in the list (just like retail).

**What Doesn't Work!**

1. Using "tabs" to pay for teleporting.
    1. This is because you can only pass a single number value to the various parameter values of the "startEvent" method and the value we pass is the "gil" value. I *think* the C++ method would have to be updated to allow a table or array to be passed but I am not sure. Someone would have to review the actual retail code to figure out what needs to be passed here.
    1. Please see the code comments for more detail about what parameters are used (all 8!) and what they are used for.
2. Favorites
    1. I tried a lot of different combinations of values but I have no idea how favorites work with the Survival Guides, I have tried to disable them. 

**Somethings to watch out for**
1. I tried to get the new Survival Guide positions correct but some of them won't be in the same place as retail or their rotation is wrong/weird. I was more concerned with getting everything working and getting the npc in mostly the correct spot.

This was a really fun and sometimes frustrating thing to implement, but I hope you like it! I learned about Lua during all this too (never used it before)!

I tried to document as much as I could inside the main processing script (survival_guide.lua inside the globals folder).


UynGH:
Thank you very much for this, was waiting for this feature to be implemented.

"I removed the extra click you have to do when registering so you can get right into the menu!"
Bring it back, retail accuracy!

And as for the Guides positions, you can find them on KnowOne's GitHub here: https://github.com/KnowOne134/NPC-MOB-Logger/tree/master/Tables (I think I'm only missing one currently).

kaincenteno:
awesome work @Brian-Wuest  and welcome 😸

Brian-Wuest:
> Thank you very much for this, was waiting for this feature to be implemented.
> 
> "I removed the extra click you have to do when registering so you can get right into the menu!"
> Bring it back, retail accuracy!
> 
> And as for the Guides positions, you can find them on KnowOne's GitHub here: https://github.com/KnowOne134/NPC-MOB-Logger/tree/master/Tables (I think I'm only missing one currently).

I can add back retail accuracy, just a minute :)

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+
+        -- This turns negative in the database for int32 max
+        hasRegisteredGuide = mask == (guide.groupMask * -1)
+    else
+        mask = masks[group]
+        hasRegisteredGuide = (bit.band(mask, guide.groupMask) ~= 0)
+    end
+
+    if (not hasRegisteredGuide) then
+        player:PrintToPlayer(
```
TeoTwawki: We don't use PrintToPlayer for this. That message will either be part of a cs event, a messageSpecial, or another type of message in client accessed by its ID.

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+dsp.survival_guide.onTrigger = function(player, csid, index)
+    local guide = survivalGuides[index]
+
+    if (not (guide == nil) or not (ENABLE_ROV == 1)) then
+
+        -- Contains the zone to include, this is a bit-wise value.
+        local expansions = getEnabledExpansions()
+
+        local masks = {
+            [1] = player:getVar("SgTeleportMask1"),
```
TeoTwawki: Absolutely not. This happened with homepoints and I refuse to allow it a second time. This needs stored in char_unlock table. (as do homepoints). char_vars are ONLY for temporary data that will eventually be eliminated.

TeoTwawki:
Aside form the two items I mentioned..Looks golden. Thank you very much.

scripts/globals/survival_guide.lua
Brian-Wuest: That makes sense, can you point me towards where I can figure out the appropriate ID? The only CSIDs I could fine for Survival Guides are 8500 and 8501 and they both say the same thing in the UI (basically the first line to go into the teleportation sub-menu)

scripts/globals/survival_guide.lua
Brian-Wuest: Oh, wasn't aware of the history behind this. I followed how homepoints are registering. If you want me to push it over to char_unlock I will have to add a new column to that table and figure out how to get that to work.

Brian-Wuest:
> Aside form the two items I mentioned..Looks golden. Thank you very much.

Thank you, I posted questions to your comments.

scripts/globals/survival_guide.lua
TeoTwawki: all strings zone specific can be located via polutils. This type of things is likely to be the zone specific `messageSpecial()`. In menu, I believe the ones you don't have yet don't even show to be selected (so no message at all).

scripts/globals/survival_guide.lua
TeoTwawki: Dear everyone: see? I told it would happen! If we let bad examples happen they get copied! /sulk


@Brian-Wuest definitely add new columns there as needed. 1 new column in that table can hold all the data form every variable you used here. you may either make it a binary blob (recommended), or just a very large number if it doesn't exceed mysql's maximum datatype there (existing examples are just very large integers).

Brian-Wuest: I was looking at the C++ code and in the "lua_baseentity.cpp" and the "luautils.cpp" I didn't see a way to get/save character unlocks. There is a call to SaveCharUnlocks but it's only done in the "addNationalTeleport" method.

I am not strong at all at C++ (mainly C# and Java developer). So I don't know what else to do here :(.

scripts/globals/survival_guide.lua
Brian-Wuest: Cool, I got the tool and am going through it. Will probably take a while (day or 2) to get all of the IDs together since they are different per zone.

TeoTwawki:
Brian if you have not yet joined us on Discord yet, you should. We have memes and there is cake in the #welcome channel. https://discord.gg/4dgAzfC
image

sql/npc_list.sql
```diff
@@ -22574,22 +22455,22 @@ INSERT INTO `npc_list` VALUES (17613278,'HomePoint#2','Home Point #2',0,102.000,
...
-INSERT INTO `npc_list` VALUES (17617192,'Treasure_Casket','Treasure Casket',0,0.000,0.000,0.000,7,40,40,0,4,4,2,3,0x0000C50300000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17617193,'Treasure_Casket','Treasure Casket',0,0.000,0.000,0.000,7,40,40,0,4,4,2,3,0x0000C50300000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17617194,'Treasure_Casket','Treasure Casket',0,0.000,0.000,0.000,7,40,40,0,4,4,2,3,0x0000C50300000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17617195,'Treasure_Casket','Treasure Casket',0,0.000,0.000,0.000,7,40,40,0,4,4,2,3,0x0000C50300000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17617196,'Treasure_Casket','Treasure Casket',0,0.000,0.000,0.000,7,40,40,0,4,4,2,3,0x0000C50300000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17617197,'Treasure_Casket','Treasure Casket',0,0.000,0.000,0.000,7,40,40,0,4,4,2,3,0x0000C50300000000000000000000000000000000,0,NULL,0);
+-- NC: INSERT INTO `npc_list` VALUES (17617182,'NOT_CAPTURED','Treasure Casket',0,0.000,0.000,0.000,0,50,50,0,0,0,0,0,0x0000320000000000000000000000000000000000,0,NULL,0);
```
TeoTwawki: I think you need to rebase onto current master when you get a chance, and make sure unrelated NPC's aren't accidentally being reverted.

Brian-Wuest: doh! Yeah, I can fix that.

scripts/globals/survival_guide_map.lua
```diff
@@ -0,0 +1,1181 @@
+survivalGuides = {
+    [6] = {
+        value = "West Ronfaure",
+        zoneId = 100,
+        group = 1,
+        groupMask = 1,
```
ibm2431: Isn't groupMask always going to be 2 ^ (groupIndex - 1)? Do we need to define all these powers of 2 in our table?

scripts/globals/survival_guide_map.lua
```diff
@@ -0,0 +1,1181 @@
...
+        groupIndex = 2,
+        posX = -51,
+        posY = 0,
+        posZ = -120,
+        posRot = 64
+    }
+}
+
+zoneIdToMenuID = {
+    [231] = {menuId = 0},
```
ibm2431: But if it does end up being used, zone enums, like before.

scripts/globals/survival_guide_map.lua
```diff
@@ -0,0 +1,1181 @@
...
+        groupMask = 2,
+        groupIndex = 2,
+        posX = -51,
+        posY = 0,
+        posZ = -120,
+        posRot = 64
+    }
+}
+
+zoneIdToMenuID = {
```
ibm2431: I'm not seeing where this table is used...!

(But if we're okay with a translation table, I would totally go for a optionToZone in the opposite direction, so we can just use zone IDs for the table indexes above and inside functions.)

scripts/globals/survival_guide_map.lua
```diff
@@ -0,0 +1,1181 @@
...
+    [231] = {menuId = 0},
+    [234] = {menuId = 1},
+    [240] = {menuId = 2},
+    [243] = {menuId = 3},
+    [26] = {menuId = 4},
+    [50] = {menuId = 5},
+    [100] = {menuId = 6},
```
ibm2431: (And maybe some spacing to line everything up. :wink:

scripts/globals/survival_guide_map.lua
```diff
@@ -0,0 +1,1181 @@
...
+        posX = -51,
+        posY = 0,
+        posZ = -120,
+        posRot = 64
+    }
+}
+
+zoneIdToMenuID = {
+    [231] = {menuId = 0},
+    [234] = {menuId = 1},
```
ibm2431: Actually, why tables instead of just returning the ID?

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+    return 1
+end
+
+dsp.survival_guide.onEventFinish = function(player, csid, eventId, option)
+    if (option >= 1 and not (option == 7) and not (option == 65539) and csid ==
+        eventId) then
+        local selectedMenuId = bit.rshift(option, 16)
+
+        if (selectedMenuId <= 97) then
+            local guide = survivalGuides[selectedMenuId]
```
ibm2431: If using an optionToID translation table (which I greatly advocate, this becomes `local guide = survivalGuides[optionToGude[option]]

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+        player:PrintToPlayer('Survival guides are not enabled!')
+    end
+
+    return 1
+end
+
+dsp.survival_guide.onEventFinish = function(player, csid, eventId, option)
+    if (option >= 1 and not (option == 7) and not (option == 65539) and csid ==
+        eventId) then
+        local selectedMenuId = bit.rshift(option, 16)
```
ibm2431: Is selectedMenuID used for anything other than as a key in a table? Can option, or a translated option->zone ID, be used as a table instead?

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+        -- param 8 = expansions available.
+        -- NOTE: To allow teleportation to all zones, set all of the mask values to -1.
+        player:startEvent(csid, -1, index, player:getGil(), masks[1] + masks[5],
+                          masks[2] + masks[6], masks[3] + masks[7], masks[4],
+                          expansions)
+    else
+        player:PrintToPlayer('Survival guides are not enabled!')
+    end
+
+    return 1
```
ibm2431: Don't really need this return :wink:

scripts/globals/settings.lua
```diff
@@ -145,6 +145,7 @@ DIG_GRANT_BURROW = 0; -- Set to 1 to grant burrow ability
 DIG_GRANT_BORE = 0; -- Set to 1 to grant bore ability
 ENM_COOLDOWN = 120;  -- Number of hours before a player can obtain same KI for ENMs (default: 5 days)
 FORCE_SPAWN_QM_RESET_TIME = 300; -- Number of seconds the ??? remains hidden for after the despawning of the mob it force spawns.
+CHEAPER_SURVIVAL_GUIDE_TELEPORT = 0; -- Determines if the survival guide teleport cost is like if you had a Rhapsody in White key item. Does not affect UI! (Default: 0)
```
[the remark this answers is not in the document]
Brian-Wuest: So always have it at original price or 100 or 200 gil?

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
+require("scripts/globals/settings")
+require("scripts/globals/survival_guide_map")
+
+dsp = dsp or {}
+dsp.survival_guide = dsp.survival_guide or {}
```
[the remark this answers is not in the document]
Brian-Wuest: yeah, I can change the name of the file and the associated type

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+
+-- 1 = Original zones
+-- 2 = Zilart Zones
+-- 4 = Chains of Promethia Zones
+-- 8 = Treasures of Aht Urgahn
+-- 16 = Wings of the Goddess
+-- 2048 = Seekers of Auldoin
+local function getEnabledExpansions()
+    -- Original areas and zilart are always available.
+    local returnValue = 1 + 2
```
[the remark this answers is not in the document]
Brian-Wuest: I guess..... :P

yeah I can change that.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Survival_Guide.lua
```diff
@@ -0,0 +1,12 @@
+require("scripts/globals/survival_guide")
+
+local eventId = 8500
+local guideIndex = 5
```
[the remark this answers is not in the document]
Brian-Wuest: The guide index is the the actual index (option value) that get's passed around to the event handlers so that's what I'm indexing off of. I do have a separate map where it's zone id to index so I could use that so there is less to pass in.

scripts/globals/survival_guide_map.lua
Brian-Wuest: Because I'm used to object oriented and self-describing code.

scripts/globals/survival_guide_map.lua
Brian-Wuest: This is how VS code lined everything up which is based on this projects editor config.

scripts/globals/survival_guide.lua
Brian-Wuest: Saw it in other places so figured I needed it :)

Brian-Wuest: Saw it in other places so I thought I needed it :)

scripts/globals/settings.lua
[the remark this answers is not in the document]
ibm2431: Original price of 1000, if this setting ends up making it in (`SURVIVAL_TELEPORT_PRICE = 1000`)

scripts/globals/survival_guide.lua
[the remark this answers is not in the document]
ibm2431: I know this will sound contradictory, but keep the file name the same as you have it. 😅 

(Our naming conventions are a little weird at times.)

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+        masks = checkForRegisteredSurvivalGuide(player, guide, masks)
+
+        -- param 1 = Favorites, this is broken yet, not sure how to fix this, don't really care right now either, the rest of it works (except for tabs).
+        -- param 2 = current area - this is the zone index passed in, this means that the area won't show in the list of areas to teleport too.
+        -- param 3 = gil, NOTE: The startEvent method probably needs to be updated to allow for a table so gil and valor_point (tabs) can be passed in too.
+        -- param 4 = zones unlocked (group 1), set to -1 to enable all zones in the group.
+        -- param 5 = Zones unlocked (group 2), set to -1 to enable all zones in the group.
+        -- param 6 = Zones unlocked (group 3), set to -1 to enable all zones in the group.
+        -- param 7 = zones unlocked (Zehrun mines and Eastern Adoulin), set to -1 to enable all zones in the group.
+        -- param 8 = expansions available.
```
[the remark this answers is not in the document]
Brian-Wuest: yes....

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+            [5] = player:getVar("SgTeleportMask1a"),
+            [6] = player:getVar("SgTeleportMask2a"),
+            [7] = player:getVar("SgTeleportMask3a")
+        }
+        -- If this survival guide hasn't been registered yet (saved to database) do that now.
+        masks = checkForRegisteredSurvivalGuide(player, guide, masks)
+
+        -- param 1 = Favorites, this is broken yet, not sure how to fix this, don't really care right now either, the rest of it works (except for tabs).
+        -- param 2 = current area - this is the zone index passed in, this means that the area won't show in the list of areas to teleport too.
+        -- param 3 = gil, NOTE: The startEvent method probably needs to be updated to allow for a table so gil and valor_point (tabs) can be passed in too.
```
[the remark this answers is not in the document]
Brian-Wuest: Due to other conversations I think I have a way to show tabs (valor points)!

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+        -- param 5 = Zones unlocked (group 2), set to -1 to enable all zones in the group.
+        -- param 6 = Zones unlocked (group 3), set to -1 to enable all zones in the group.
+        -- param 7 = zones unlocked (Zehrun mines and Eastern Adoulin), set to -1 to enable all zones in the group.
+        -- param 8 = expansions available.
+        -- NOTE: To allow teleportation to all zones, set all of the mask values to -1.
+        player:startEvent(csid, -1, index, player:getGil(), masks[1] + masks[5],
+                          masks[2] + masks[6], masks[3] + masks[7], masks[4],
+                          expansions)
+    else
+        player:PrintToPlayer('Survival guides are not enabled!')
```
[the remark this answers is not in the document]
Brian-Wuest: Just following how the Field/Grounds tomes are done

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+end
+
+local function checkForRegisteredSurvivalGuide(player, guide, masks)
+    local group = guide.group
+    local currentZoneId = player:getZoneID()
+    local mask = nil
+    local variableSuffix = ''
+    local hasRegisteredGuide = false
+
+    if (guide.groupIndex == 32) then
```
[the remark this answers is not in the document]
Brian-Wuest: Oh, can you update the contributing.md then? Didn't see that in there (unless I'm blind)

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+    if (ENABLE_WOTG == 1) then returnValue = returnValue + 16 end
+
+    if (ENABLE_SOA == 1) then returnValue = returnValue + 2048 end
+
+    return returnValue
+end
+
+local function checkForRegisteredSurvivalGuide(player, guide, masks)
+    local group = guide.group
+    local currentZoneId = player:getZoneID()
```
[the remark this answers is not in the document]
Brian-Wuest: yeah, unfortunate dead code trying to play with things, will address

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+        mask = masks[group]
+        hasRegisteredGuide = (bit.band(mask, guide.groupMask) ~= 0)
+    end
+
+    if (not hasRegisteredGuide) then
+        player:PrintToPlayer(
+            'This guide hasn\'t been registered yet, registering this Survival Guide.',
+            13)
+
+        local updatedValue = mask + guide.groupMask
```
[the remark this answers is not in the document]
Brian-Wuest: bit.band actually doesn't work in all cases (i've tried for a few days).

Adding things together ended up working in all cases.

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+end
+
+-------------------------------------------------
+-- public functions
+-------------------------------------------------
+
+dsp.survival_guide.onTrigger = function(player, csid, index)
+    local guide = survivalGuides[index]
+
+    if (not (guide == nil) or not (ENABLE_ROV == 1)) then
```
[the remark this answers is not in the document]
Brian-Wuest: Good points, being new to Lua I didn't know how that would evaulate.

I think I copied the wrong thing. Was looking at regimes for this type of idea. Since there isn't a setting specifically for Survival Guides and there is a stigma against new settings I will just make sure a guide was found.

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,136 @@
...
+    if (option >= 1 and not (option == 7) and not (option == 65539) and csid ==
+        eventId) then
+        local selectedMenuId = bit.rshift(option, 16)
+
+        if (selectedMenuId <= 97) then
+            local guide = survivalGuides[selectedMenuId]
+
+            if (not (guide == nil)) then
+                -- This is the standard fare unless they have the Rhapsody in white which isn't implemented yet.
+                local teleportCost = 1000
```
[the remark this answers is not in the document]
Brian-Wuest: cool

sql/npc_list.sql
Brian-Wuest: This should be fixed now.

scripts/globals/survival_guide.lua
Brian-Wuest: Latest checkin removed this.

scripts/globals/survival_guide.lua
Brian-Wuest: Updated how this is keyed

scripts/globals/survival_guide.lua
Brian-Wuest: Updated how this is keyed

scripts/globals/survival_guide_map.lua
TeoTwawki: re-spacing this would be optional for you. editorconfig mostly gets this right, but it'll never be perfection (a lot is dependent on an editors interpretation of editirconfig) :smiley:

scripts/globals/survival_guide.lua
```diff
@@ -69,14 +87,60 @@ local function checkForRegisteredSurvivalGuide(player, guide, masks)
...
 
-dsp.survival_guide.onTrigger = function(player, csid, index)
-    local guide = survivalGuides[index]
+dsp.survivalGuide.onTrigger = function(player)
+    local currentZoneId = player:getZoneID()
+    local tableIndex = zoneIdToMenuID[currentZoneId]
+    local guide = survivalGuides[tableIndex]
 
-    if (not (guide == nil) or not (ENABLE_ROV == 1)) then
+    if (guide) then
```
TeoTwawki: `if guide then`

scripts/globals/survival_guide.lua
```diff
@@ -90,47 +154,63 @@ dsp.survival_guide.onTrigger = function(player, csid, index)
...
             [7] = player:getVar("SgTeleportMask3a")
         }
+
         -- If this survival guide hasn't been registered yet (saved to database) do that now.
-        local foundRegisteredGuide = checkForRegisteredSurvivalGuide(player, guide, masks)
+        local foundRegisteredGuide = checkForRegisteredSurvivalGuide(player,
+                                                                     guide,
+                                                                     masks)
 
         if (foundRegisteredGuide) then
```
TeoTwawki: `if foundRegisteredGuide then`

Brian-Wuest:
@TeoTwawki - Yes it is, however that's an extra jump across the Lua-> C++ barrier. Especially since it's not exposed as a property/field. 

Since I already got the zone id on line 138 it made sense to just pass it in instead of "getting" it again.

TeoTwawki:
> @TeoTwawki - Yes it is, however that's an extra jump across the Lua-> C++ barrier. Especially since it's not exposed as a property/field.
> 
> Since I already got the zone id on line 138 it made sense to just pass it in instead of "getting" it again.

@teschnei do we want to start doing that kind of optimization? All through the trunk up to now, we've given zero f\*\*\* and thrown gets willy nilly. I’ve thought of the cost of that, just got used to it.

ibm2431:
In this case, you already have the zone ID from your `guide` table! Just use the `guide.zoneId` you set up for yourself. :wink:

Brian-Wuest:
> In this case, you already have the zone ID from your `guide` table! Just use the `guide.zoneId` you set up for yourself. 😉

Doh! Good point! Changing!

teschnei:
It costs essentially nothing to get something from c++ from Lua anyways, so it's not really a concern. Anyways, why can't you implement tabs? I didn't understand the explanation, startEvent can take up to 8 parameters already

Brian-Wuest:
@teschnei I eventually got some help to figure out the tabs thing. Apparently the currency has to be passed in the same parameter as the current zone parameter. Survival guides pretty much use all 8 parameters.

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,219 @@
...
+local function teleportMenuUpdate(player, option, teleportType, optionMap, var)
+    local choice = bit.band(option, 0xFF)
+
+    if choice == optionMap.SET_MENU_LAYOUT then
+        if bit.rshift(option, 16) == 1 then
+            -- Select by content/expansion
+            player:setVar(var, 1)
+        else
+            -- Select by region
+            player:setVar(var, 0)
```
kaincenteno: player:setVar changed to player:setCharVar. Righ @ibm2431 ?

ibm2431: It did, but I wouldn't worry about updating this just yet due to wanting to change these to use char_unlock blobs.

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+    return true
+end
+
+local function teleportMenuUpdate(player, option)
+    local choice = bit.band(option, 0xFF)
+
+    if choice >= optionMap.SET_MENU_LAYOUT and choice <= optionMap.TELEPORT_MENU then
+        local favorites = player:getTeleportMenu(travelType)
+
+        local index = bit.rshift(bit.lshift(option, 8), 24)
```
ibm2431: Is this equivalent to `bit.rshift(option, 16)`?

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+
+    if ENABLE_COP == 1 then returnValue = returnValue + 4 end
+
+    if ENABLE_TOAU == 1 then returnValue = returnValue + 8 end
+
+    if ENABLE_WOTG == 1 then returnValue = returnValue + 16 end
+
+    if ENABLE_SOA == 1 then returnValue = returnValue + 2048 end
+
+    return returnValue
```
ibm2431: While it's technically possible for these settings to be changed while the server is already booted, I wonder if enabledExpansions should be calculated and defined just once upon the first require of this file (`dsp.survivalGuide.enabledExpansions = getEnabledExpansion()`), and then that value gets passed into the CS each time, instead of the onTrigger having to go through this function every time a player talks to a guide.

ibm2431:
I forgot a Review Comment, but just so you know, I do like what you've done with this! 😄

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+
+        local index = bit.rshift(bit.lshift(option, 8), 24)
+        if (not (choice == optionMap.TELEPORT_MENU)) then
+            if choice == optionMap.ADD_FAVORITE then
+                local temp = 0
+                for x = 1, 9 do
+                    temp = favorites[x]
+                    favorites[x] = index
+                    index = temp
+                end
```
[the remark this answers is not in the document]
Brian-Wuest: This is the same code as the homepoint.lua file. The favorites work the same between them.

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+            -- param 8 = expansions available.
+            player:startEvent(cutsceneID, 0, param, player:getGil(), G1, G2, G3, G4, expansions)
+        end
+    else
+        player:PrintToPlayer('Survival guides are not enabled!')
+    end
+end
+
+dsp.survivalGuide.onEventUpdate = function(player, csid, option)
+    teleportMenuUpdate(player, option)
```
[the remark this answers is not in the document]
Brian-Wuest: Not sure if this is a question for me or for the group as a whole? Do you want me to change something here?

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+                end
+
+                if bit.rshift(bit.lshift(option, 16), 24) == 1 then
+                    -- Paying with tabs
+                    player:delCurrency('valor_point', teleportCostTabs)
+                else
+                    player:delGil(teleportCostGil)
+                end
+
+                player:setPos(guide.posX, guide.posY, guide.posZ, guide.posRot, guide.zoneId)
```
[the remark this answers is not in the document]
Brian-Wuest: The ui won't let you even select the option if you don't have enough of the appropriate currency.

If you try to select tabs and you don't have enough if that low tone sounds and nothing happens and no events are triggered.

scripts/globals/survival_guide_map.lua
```diff
@@ -0,0 +1,1283 @@
...
+        groupIndex = 2,
+        posX = -51,
+        posY = 0,
+        posZ = -120,
+        posRot = 64,
+        messageId = 12341
+    }
+}
+
+zoneIdToMenuID =
```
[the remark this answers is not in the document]
Brian-Wuest: Like what? zone to menu id converter?

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+
+    if ENABLE_SOA == 1 then returnValue = returnValue + 2048 end
+
+    return returnValue
+end
+
+local function checkForRegisteredSurvivalGuide(player, guide)
+    local group = guide.group
+    local mask = nil
+    local variableSuffix = ''
```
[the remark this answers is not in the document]
Brian-Wuest: Fixed

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+end
+
+local function teleportMenuUpdate(player, option)
+    local choice = bit.band(option, 0xFF)
+
+    if choice >= optionMap.SET_MENU_LAYOUT and choice <= optionMap.TELEPORT_MENU then
+        local favorites = player:getTeleportMenu(travelType)
+
+        local index = bit.rshift(bit.lshift(option, 8), 24)
+        if (not (choice == optionMap.TELEPORT_MENU)) then
```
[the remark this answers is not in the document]
Brian-Wuest: Fixed

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+                            favorites[x] = favorites[x+1]
+                        end
+                        favorites[9] = -1
+                        break
+                    end
+                end
+            elseif choice == optionMap.REPLACE_FAVORITE then
+                favorites[bit.rshift(option, 24) + 1] = index
+            elseif choice == optionMap.SET_MENU_LAYOUT then
+                favorites[10] = bit.rshift(option, 16) == 1 and 1 or 0
```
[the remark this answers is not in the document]
Brian-Wuest: Updated

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+            -- Menu options can be organized by Region or Content.
+            -- Default (0) is region.
+            local teleportMenu = player:getTeleportMenu(travelType)
+            teleportMenu = teleportMenu[10]
+
+            if teleportMenu == 1 then
+                param = bit.bor(param, 0x0800)
+            end
+
+            if player:hasKeyItem(dsp.keyItem.RHAPSODY_IN_WHITE) == 1 then
```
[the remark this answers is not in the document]
Brian-Wuest: Fixed

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+        if foundRegisteredGuide then
+            local param = bit.bor(tableIndex, bit.lshift(player:getCurrency("valor_point"), 16))
+
+            -- Get the teleport menu option.
+            -- Menu options can be organized by Region or Content.
+            -- Default (0) is region.
+            local teleportMenu = player:getTeleportMenu(travelType)
+            teleportMenu = teleportMenu[10]
+
+            if teleportMenu == 1 then
```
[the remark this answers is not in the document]
Brian-Wuest: Fixed

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+        player:PrintToPlayer('Survival guides are not enabled!')
+    end
+end
+
+dsp.survivalGuide.onEventUpdate = function(player, csid, option)
+    teleportMenuUpdate(player, option)
+end
+
+dsp.survivalGuide.onEventFinish = function(player, eventId, option)
+    if cutsceneID == eventId and not (option >= 1 and option <= 7)
```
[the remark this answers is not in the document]
Brian-Wuest: Fixed

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+            local currentZoneId = player:getZoneID()
+
+            if guide and not (guide.zoneId == currentZoneId) then
+                local teleportCostGil = SURVIVAL_GUIDE_TELEPORT_COST_GIL
+                local teleportCostTabs = SURVIVAL_GUIDE_TELEPORT_COST_TABS
+
+                -- If the player has the rhapsody in white, the cost is 10% of original gil or 20% of original tabs.
+                -- GIL: 1000 -> 200
+                -- TABS: 100 -> 10
+                if player:hasKeyItem(dsp.keyItem.RHAPSODY_IN_WHITE) == 1 then
```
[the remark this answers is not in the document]
Brian-Wuest: Fixed

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,200 @@
...
+                    teleportCostGil = teleportCostGil * .2
+                    teleportCostTabs = teleportCostTabs * .1
+                end
+
+                if bit.rshift(bit.lshift(option, 16), 24) == 1 then
```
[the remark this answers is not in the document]
Brian-Wuest: Fixed

scripts/globals/survival_guide_map.lua
```diff
@@ -0,0 +1,1283 @@
...
+    {
+        zoneId = dsp.zone.WEST_RONFAURE,
+        group = 1,
+        groupMask = 2^0,
+        groupIndex = 1,
+        posX = -451,
+        posY = -19,
+        posZ = -217,
+        posRot = 64,
+        messageId = 12341
```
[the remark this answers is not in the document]
Brian-Wuest: Fixed

scripts/globals/survival_guide_map.lua
```diff
@@ -0,0 +1,1283 @@
...
+    [102] = 10,
+    [103] = 11,
+    [108] = 12,
+    [193] = 13,
+    [196] = 14,
+    [2] = 15,
+    [104] = 16,
+    [105] = 17,
+    [149] = 18,
+    [1954] = 19,
```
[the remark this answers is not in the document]
Brian-Wuest: Fixed

scripts/globals/survival_guide.lua
Brian-Wuest: Fixed

scripts/globals/survival_guide.lua
Brian-Wuest: Fixed

scripts/globals/survival_guide_map.lua
```diff
@@ -0,0 +1,1283 @@
...
+        posZ = -120,
+        posRot = 64,
+        messageId = 12341
+    }
+}
+
+zoneIdToMenuID =
+{
+    [231] = 0,
+    [234] = 1,
```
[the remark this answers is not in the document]
Brian-Wuest: Fixed

sql/npc_list.sql
```diff
@@ -13102,8 +13102,7 @@ INSERT INTO `npc_list` VALUES (17220207,'blank','     ',0,0.000,0.000,0.000,0,50
 -- NC: INSERT INTO `npc_list` VALUES (17220209,'NOT_CAPTURED','Geomagnetic Fount',0,0.000,0.000,0.000,0,50,50,0,0,0,0,0,0x0000320000000000000000000000000000000000,0,NULL,0);
 INSERT INTO `npc_list` VALUES (17220209,'Ergon_Locus','Ergon Locus ???',234,220.041,24.451,309.234,1,50,50,0,0,0,0,3,0x0000340000000000000000000000000000000000,0,NULL,1);
 -- NC: INSERT INTO `npc_list` VALUES (17220211,'NOT_CAPTURED','Geomantic Reservoir',0,0.000,0.000,0.000,0,50,50,0,0,0,0,0,0x0000320000000000000000000000000000000000,0,NULL,0);
--- NC: INSERT INTO `npc_list` VALUES (17220214,'NOT_CAPTURED','Survival Guide',0,0.000,0.000,0.000,0,50,50,0,0,0,0,0,0x0000320000000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17220214,'blank','     ',0,0.000,0.000,0.000,0,50,50,0,0,0,2,2051,0x0000340000000000000000000000000000000000,0,NULL,1);
```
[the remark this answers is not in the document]
Brian-Wuest: Fixed

scripts/globals/survival_guide.lua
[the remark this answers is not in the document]
Brian-Wuest: Should clarify, the player doesn't get teleported and they get a message that they don't have enough gil/tabs.

scripts/globals/survival_guide.lua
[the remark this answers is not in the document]
ibm2431: Nah, so long as it functions as is, I _think_ you're fine.

Sometimes I have a habit of pointing things out just in case it becomes relevant in the future.

scripts/globals/survival_guide.lua
Ivaar: It's not the same, it was correct before.

scripts/globals/survival_guide.lua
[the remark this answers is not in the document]
Ivaar: Option 8 is used for homepoints, the survival guide equivalent is option 7. That said the function does respond to it with the favorites.

scripts/globals/survival_guide.lua
ibm2431: Was the idea to give the option padding bits on the right?

Ivaar: Lua bit ops return signed 32 bit values. The lshift truncates the upper byte 
so the next byte gets extracted by the shift right.

e.g.
```
bit.lshift(0x01020304, 8)
0x02030400
bit.rshift(0x02030400, 24)
0x02
```
i.e.
`bit.rshift(bit.lshift(option, 8), 24)`
is the same as
`bit.rshift(bit.band(option, 0xFF0000), 16)`

scripts/globals/survival_guide.lua
```diff
@@ -0,0 +1,188 @@
...
+    return true
+end
+
+local function teleportMenuUpdate(player, option)
+    local choice = bit.band(option, 0xFF)
+
+    if choice >= optionMap.SET_MENU_LAYOUT and choice <= optionMap.TELEPORT_MENU then
+        local favorites = player:getTeleportMenu(travelType)
+
+        local index = bit.rshift(option, 16)
```
TeoTwawki:
```
ibm2431 4 days ago Contributor
Is this equivalent to bit.rshift(option, 16)?

Brian-Wuest 3 days ago Author
Fixed

Ivaar 11 hours ago Contributor
It's not the same, it was correct before.

ibm2431 6 hours ago Contributor
Was the idea to give the option padding bits on the right?

Ivaar 4 hours ago Contributor
Lua bit ops return signed 32 bit values. The lshift truncates the upper byte
so the next byte gets extracted by the shift right.

e.g.

bit.lshift(0x01020304, 8)
0x02030400
bit.rshift(0x02030400, 24)
0x02
i.e.
bit.rshift(bit.lshift(option, 8), 24)
is the same as
bit.rshift(bit.band(option, 0xFF0000), 16)
```

COpy and pasted form comment thread of the prior commit, as github isn't letting me reply to it.

I think 
`bit.rshift(bit.band(option, 0xFF0000), 16)`
is clearer

TeoTwawki: Once we're collectively sure that bitshift is correct and that this has been tested in game by somebody, its merge ready I think.

Brian-Wuest: I verified that the change worked properly before committing.

I trusted what was suggested to me initially but I verified first 😸 .

TeoTwawki: yeah, sadly I'll need someone to test after all the iterations you put in.

scripts/globals/bcnm.lua
```diff
@@ -494,11 +497,11 @@ function checkReqs(player, npc, bfid, registrant)
...
-        [ 290] = function() return ( roz == mi.zilart.ARK_ANGELS and rozStat == 1 and npcid == arkCircleOffset + 2 and not player:hasKeyItem(dsp.ki.SHARD_OF_ENVY)          ) end, -- ZM14: Ark Angels (Mithra)
-        [ 291] = function() return ( roz == mi.zilart.ARK_ANGELS and rozStat == 1 and npcid == arkCircleOffset + 3 and not player:hasKeyItem(dsp.ki.SHARD_OF_ARROGANCE)     ) end, -- ZM14: Ark Angels (Elvaan)
-        [ 292] = function() return ( roz == mi.zilart.ARK_ANGELS and rozStat == 1 and npcid == arkCircleOffset + 4 and not player:hasKeyItem(dsp.ki.SHARD_OF_RAGE)          ) end, -- ZM14: Ark Angels (Galka)
+        [ 288] = function() return ( roz == mi.zilart.ARK_ANGELS and rozStat == 1 and npcid == getZM14Offset(0) and not player:hasKeyItem(dsp.ki.SHARD_OF_APATHY)           ) end, -- ZM14: Ark Angels (Hume)
+        [ 289] = function() return ( roz == mi.zilart.ARK_ANGELS and rozStat == 1 and npcid == getZM14Offset(1) and not player:hasKeyItem(dsp.ki.SHARD_OF_COWARDICE)        ) end, -- ZM14: Ark Angels (Tarutaru)
```
TeoTwawki: I'm not sure why the diff is showing these changes that already exist on master, it shouldn't hurt anything but it defo makes the diff look confusing if you didn't know where these were from (could easily mistake them for unrelated changes in the pr, which they are not).

when you update your PR do you pull from master or do you rebase your branch onto the latest master commit? Neither one should result in this so I wanna say github is mishandling the display of merge commits.

Brian-Wuest: I re-based from latest from Darkstar/master

TeoTwawki: does this commit exist on your branch? https://github.com/DarkstarProject/darkstar/commit/f25aae86965d8a8bbd11d22295b57d18bb185077

it should. I see this though
https://github.com/DarkstarProject/darkstar/pull/6238/commits/36e4c84eebf321deca02d5b86eb8b6ce6d4e2161
which shouldn't exist in a rebase scenario.

kaincenteno:
@TeoTwawki  i tested it with several zone areas (past, present wing of the goddess, aht ughan) and works correctly after the last 3 commits that were made. It has my blessing lol

TeoTwawki:
@kaincenteno & @Brian-Wuest 
thanks guys. waiting on apveyor now

scripts/globals/survival_guide.lua
```diff
@@ -142,8 +142,8 @@ dsp.survivalGuide.onEventUpdate = function(player, csid, option)
 end
 
 dsp.survivalGuide.onEventFinish = function(player, eventId, option)
-    if cutsceneID == eventId and not (optionMap.TELEPORT > 1 and optionMap.TELEPORT_MENU <= 7)
-        and not (option == 65539)  then
+    if cutsceneID == eventId and not (option > optionMap.TELEPORT and option <= optionMap.TELEPORT_MENU)
```
Ivaar: `if bit.band(option, 0xFF) == optionMap.TELEPORT then`

TeoTwawki: I completely missed that this was doing a greater and less than check on that.

#qualityRevieweingTeo

PR #6242, opened 2019-10-13
Added Casket text Id's to zones and cleared up some comments.

Omnione:
Added text ID's for Caskets.

Cleaned up some comment alignments.
Sorry i know its a big diff.


Omnione:
Why you error on me travis >.>, must hate me

ibm2431:
`combination is greater than ≺RANDOM LOCK NUMBER≻ and less than ≺RANDOM LOCK NUMBER≻`

Are these comments taken as they are from POLUtils?

Omnione:
I may changed them slightly to make more sense as the polutils ones were a bit hard to undestand the values

Omnione:
hmm thinking about it, i should have probly just had number rather than random lock number

Omnione:
i can change them if you prefer, i dont mind, any input would be great

ibm2431:
They usually need to match for the batch tools to pick them up. But I think there might be a little bit of "fuzziness" involved where it can pick up based on partial strings. It's been a while since I've looked at the text ID batch tool though, so someone correct me if I'm wrong and/or how exact the comments need to be.

Omnione:
Right all the npc ID's should be in order now.

wrenffxi:
ibm is right about the comment matching having some fuzziness.  I've done the last few client updates, and don't even use the batch tools to do them.  My code also has some fuzziness, that's different from the batch tools.  For mine, most of the "tags" can be replaced with <player>, <number>, <item>, or <keyitem>.  But if they're not right in a PR, it's not a big deal; I'll fix them to match when I do the next client update.

Omnione:
Cheers Wren, I appreciate it.

But if you need me to change it i can, no bigge.

Omnione:
Gonna baby sit AppVeyor and travis so all conflicts are resloved, moved most message ID's moved to the global script im working to minimise conflicts.

Omnione:
OK this should finaly be done now, /crosses fingers.

As disscused with Teo,
I have re added NO_COMBINATION text ID to all the casket related zones ID.lua,
This was as Teo pointed out, even though some of the text ID's i added were close to the casket text ID's i should not rely on them not changing and so i need to add a Text ID that is directly associated to caskets.
Now i can use an offset in the global script i am currently working on to grab any other text ID.

Omnione:
Hmm thats odd, why do they all seem like the spacing is wrong, damn you Notepad++ making me look bad, this is why i use sublime, the one time i use notepad++  -.-;

ibm2431:
Because your editor is using tabs instead of spaces...!

Omnione:
yep, just noticed that myself, stupid notepad, i never set it up

Omnione:
gonna fix it, one sec

Omnione:
thats odd, its not picking up my changes, i bet its because i ticked amended last commit

TeoTwawki:
the spacing is off because of tabs. get the editorconfig plugin or configre it to make the tabkey insert 4 spaces instead. (just get editorconfig)

if you are doing amends, make sure you are force-pushing and NOT pulling down again which is where most ppl go wrong the 1st time they get a push error after amending.

Omnione:
yeah, i think i made TortioseGit angry lol

TeoTwawki:
ps
it is also possible to make them visible
image

Omnione:
I usualy use sublime so its not a issue, but this time i was copy pasta'ing stuff and didnt wanna cause issuse so i opened one file in sublime and the other in notedpad++, stupid me thought it was a good idea lol, guess i learnt something today lol.

Omnione:
ill be glad to see the end of this PR, editing 59 zones is not a fun task

TeoTwawki:
I have this so its never a prob for me
image

Omnione:
yeah same in sublime, i just never added it to notepad++ because i stopped using after we was doing Legion, that was a fun mess >.<

PR #6246, opened 2019-10-16
Prepare to AxeMurder competing comment formats in item mods..

TeoTwawki:
I HAVE HAD IT WITH THESE ~~\<CENSORED>SNAKES~~ _INCONSISTENT FORMATS_ ON THIS ~~\<CENSORED>PLANE!~~ _SQL TABLE!_


TeoTwawki:
What the PR does:
- adds a varchar column for the item name and the modifier comment (or any other comment needed)

Why it does it:
- I got sick of everyone changing the format of the commenting and seeing multiple versions of them in the same file, triggering my OCD and generally giving me a migraine.
- Comments don't survive redumps anyway, and redumping via mysqldump is a good way to force the other formatting to be consistent uniform and also _correct._

What this doesn't do:
- Migrate the existing comments. I'm not doing this myself. They still exist.

What server owners can do to fix their custom before complaining about this:
- If using live DB: Adding empty text column isn't hard no matter what editor you use. Learn to google.
- If using a file: this is also not hard. `VALUES (` -> `VALUES ('FixMe',` in any plain text or code editor, such as Notepad++

zynjec:
I don't think we need yet another table with item name for the purposes of human readability.

If they really need the item name, join item_basic in a query.

wrenffxi:
I agree that adding columns for descriptions of stuff that's already described in related tables is overkill.  If anything, I'd recommend making a separate table with modId and description, that can be joined if people need descriptions for mods in any other table.

TeoTwawki:
We've done this to several other tables already. Its not a new table, its a new column. I agree its not the greatest of solutions to retaining information. I do ***not*** want yet another place to watch over making sure everyone that adds mods keeps them in sync so I am not in favor of an sql table for modifier names. In fact I'll dig my heels in pretty hard on that.

TeoTwawki:
That is also 100% useless for reviewers.

TeoTwawki:
I'll be doing something different instead of intentional making it look as bad as possible to make a point. ***coughs***

TeoTwawki:
will open a new one when ready. Joins are still not a solution BTW because nobody is going to do that just to review.

zynjec:
Joins are a perfectly reasonable, we're not adding another column for just human readability.

TeoTwawki:
na they ain't. a notes field would be handy and I don't just mean for modifiers. We have human readable fields already and they have no performance or maintenance impacts. but I ain't doing item name field so chill. Ima just enforce one ducking comment style.

zynjec:
Most of the columns we have already have for human readability only exist for automation tools, as of yet item mods does not have a (public) one.

TeoTwawki:
we don't automate mob stuff but we have name columns in multiple tables - in theory you could join on pools and not use names in either groups or spawn points, but it'd get a little difficult to read when not looking at query results then. I'm just saying such fields are not the devil like that floozeball, altho I generally don't like have extraneous fields laying about. the last one we allowed in was for mob groups I think because of it going from to per-zone groupings creating a ton of new rows at once.

PR #6251, opened 2019-10-20
Casket base ID's Swapped in East Ronfaure and West Ronfaure IDs.lua a…

Omnione:
Casket base ID's Swapped in East Ronfaure and West Ronfaure IDs.lua as was mixed up.

Now corrected.

### DIFF
--- a/scripts/zones/West_Ronfaure/IDs.lua
+++ b/scripts/zones/West_Ronfaure/IDs.lua
@@ -61,9 +61,9 @@ zones[dsp.zone.WEST_RONFAURE] =
     },
     npc =
     {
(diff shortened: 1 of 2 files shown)


scripts/zones/West_Ronfaure/IDs.lua
```diff
@@ -63,7 +63,7 @@ zones[dsp.zone.WEST_RONFAURE] =
     {
         SIGNPOST_OFFSET = 17187537,
         OVERSEER_BASE   = 17187557,
-        CASKET_BASE     = 17191481,
+        CASKET_BASE     = 17187499,
```
ibm2431: Order~!

PR #6253, opened 2019-10-20
Temenos

cocosolos:
North, West, East, Central 1, 2, 3, 4, and Basement


TeoTwawki:
Thank you for taking the time to start fixing up what is my favorite content of the 75 cap days of the game. can you say what is being done and what it fixes when you post pr's tho? Like the change form messageBasic to messageSpecial - without loading up the exact message being used there I don't know if you are fixing a mistake or making a mistake. If you detail in your initial commit messages and pr comments then we'd know what you found and corrected there. just "Temenos" tells us almost nothing.

scripts/zones/Temenos/IDs.lua
```diff
@@ -16,6 +16,7 @@ zones[dsp.zone.TEMENOS] =
         KEYITEM_OBTAINED        = 6391, -- Obtained key item: <keyitem>.
         CHIP_TRADE_T            = 7028, -- What do you wish to do? Show me the cutscene again. Skip the cutscene and continue.
         CONDITION_FOR_LIMBUS_T  = 7055, -- You have clearance to enter Limbus, but cannot enter while you or a party member is engaged in battle.
+        TIME_EXTENDED           = 7372, -- Yur time in Limbus has been extended...
```
TeoTwawki: typo I presume

scripts/zones/Temenos/bcnms/central_temenos_2nd_floor.lua
```diff
@@ -2,43 +2,39 @@
...
+function onBattlefieldInitialise(battlefield)
+    battlefield:setLocalVar("loot", 1)
+    HideTemenosDoor(Central_Temenos_2nd_Floor)
+    HideArmouryCrates(Central_Temenos_2nd_Floor, TEMENOS)
+    if GetMobByID(16929039):isSpawned() then DespawnMob(16929039) end
+    if GetMobByID(16929040):isSpawned() then DespawnMob(16929040) end
+    if GetMobByID(16929041):isSpawned() then DespawnMob(16929041) end
+    if GetMobByID(16929042):isSpawned() then DespawnMob(16929042) end
+    if GetMobByID(16929043):isSpawned() then DespawnMob(16929043) end
+    if GetMobByID(16929044):isSpawned() then DespawnMob(16929044) end
```
TeoTwawki: I can has for loop? Also why do we do this onInit instead of when the battlefield is over with? (I realize the old implementation basically did the same thing. To be fair I hated the old implementation. If it has to be this way just state the reasoning.)

TeoTwawki: p.s. IDs.lua is where we should stuff NPC or MOB IDs whenever possible, and then just call them into other scripts by their then human readable variable names.

sql/bcnm_treasure_chests.sql
```diff
@@ -421,3 +421,16 @@ INSERT INTO bcnm_treasure_chests VALUES (1152,3,17051654);
 INSERT INTO bcnm_treasure_chests VALUES (1153,1,17051660); -- shadows_of_the_mind
 INSERT INTO bcnm_treasure_chests VALUES (1153,2,17051666);
 INSERT INTO bcnm_treasure_chests VALUES (1153,3,17051672);
+
+-- ----------------------------
+-- Temenos
+-- ----------------------------
+INSERT INTO bcnm_treasure_chests VALUES (1298,3,16928785); -- Temenos - Western Tower
```
TeoTwawki: I don't know if this changed with the BCNM rewrite, but I think the reason these were never there was because of the BCNM system counting the battlefield as "over with" on chest open when limbus lets you continue to roam around after until the time runs out. I f this isn't a problem, ignore+"resolve" this comment :+1:

scripts/zones/Temenos/bcnms/central_temenos_2nd_floor.lua
cocosolos: I plan on cleaning this all up more after I look at Apollyon. This commit is basically just hooking it up with the new bcnm stuff to get it working. I put that stuff in init because I was having issues where a party member could reset the doors by zoning and you'd get stuck. The messageSpecial thing just worked and the only place the function is being used is Limbus. I admittedly don't really know the difference. The IDs is something I wanna change soon but just wanted to get this out of the way first.

sql/bcnm_treasure_chests.sql
cocosolos: I was getting missing treasure chest errors without them. I also pulled the core stuff out of the bcnm loot rolling into the Armoury Crate file to get it to actually make the loot. Without doing that it would only give loot on the last chest on 7th floor. Not sure why, since it's doing the exact same thing. It does currently warp you out to the lobby after you open the last chest.

scripts/zones/Temenos/bcnms/central_temenos_2nd_floor.lua
cocosolos: Also, these are being despawned because they're the avatar mimic mobs. They get spawned on init because I needed to put them into bcnm_battlefield to be able to attack them when they pop but they need to be despawned until they're popped by the player.

TeoTwawki: Not sure I understand. ...Wait, are these those a weird set where its using both an NPC -and- a mob for the same object (single object), rather than depopping the NPC and Spawning a mob in the same place (2 separate objects, same location)? I think I remember seeing that before.

Normally we'd just not spawn the mob until needed in the first place - it wouldn't be up before hand unless it wasn't added to the battlefield proper which would cause it to still be up after the battlefield was over.

sql/bcnm_treasure_chests.sql
TeoTwawki: Someone more familiar with how this should work is going to have to review then. I'm not confident I would guide you to the correct solution.

scripts/zones/Temenos/bcnms/central_temenos_2nd_floor.lua
cocosolos: I don't think so. The treasure chest despawns and the avatar pops correctly, but unless the avatar is in bcnm_battlefield it doesn't get added to the moblist for the battlefield and you're unable to attack it. A side effect of adding it to bcnm_battlefield is that it pops on init, so it needs to be despawned and then it can be popped and attacked normally.

TeoTwawki: > A side effect of adding it to bcnm_battlefield is that it pops on init

That not how that is supposed to work: you can have things be part of the battlefield without them being immediately spawned. We had that working prior to the rewrite, but limbus was never updated.

TeoTwawki:
```c++
enum BATTLEFIELDMOBCONDITION : uint8
{
    CONDITION_NONE = 0x00,
    CONDITION_SPAWNED_AT_START = 0x01,
    CONDITION_WIN_REQUIREMENT = 0x02
};
```
If you have them in the table with a value of 1, they are spawned with the bcnm (or a value of 3 because that is spawned + required kill for win condition). If they are in the table but the value was 0 or 2, they should not auto spawn.

scripts/zones/Apollyon/mobs/Jidra.lua
```diff
@@ -41,13 +41,13 @@ function onMobDespawn(mob)
...
 
  if (
-IsMobDead(16932882)==true and
-IsMobDead(16932883)==true and
-IsMobDead(16932884)==true and
-IsMobDead(16932885)==true and
-IsMobDead(16932886)==true and
-IsMobDead(16932887)==true and
-IsMobDead(16932888)==true
+limbus.isMobDead(16932882)==true and
```
TeoTwawki: `==true` can be removed

cocosolos: This is just from a mass function rename, still gotta go through all the mobs and fix styling and replace hard coded IDs

sql/bcnm_treasure_chests.sql
cocosolos: I checked out some recorded streams and it looks like retail does warp you out when you open the last chest.

scripts/globals/battlefield.lua
```diff
@@ -83,7 +84,7 @@ function dsp.battlefield.onBattlefieldTick(battlefield, timeinside, players)
                 canLeave = true
             end
         end
-        if canLeave and cutsceneTimer >= 15 then
+        if canLeave and cutsceneTimer >= 15 and not dontKick then
```
TeoTwawki: `local dontKick = battlefield:getLocalVar("dontKick")`
` if canLeave and cutsceneTimer >= 15 and dontKick == 0 then` 
seems saner to me

wrenffxi:
Thanks for taking lead on this!  No idea when I'd get around to it, since I'm focused on Dynamis first.  Glad you found my draft PR.  You may also find useful this spreadsheet created by my friend Bri.
https://docs.google.com/spreadsheets/d/1UyMlLNgvB1quY1StNsiQLwFnXNOOZNEEIh6SeoYZgrA/edit#gid=0, which contains our combined effort to sort out the various chests.

cocosolos:
@wrenffxi thanks for the spreadsheet! I'm currently running through the zones on retail and I've noticed a few things I still need to change. Still have no idea how to handle the doors in Temenos. There's IDs 16929197~16929220 which are positioned where the actual doors are. That may explain why all the z positions are off noted in the spreadsheet, but I can't get any of them to actually do anything.

Brierre:
I have the Temenos doors running in our server mostly through the mob lua files. Apollyon too.  We had a pretty good limbus situation going on, but it utilizes the old system of BCNM crap.  This spreadsheet was preparing for the alignment with new BCNM crap.  LMK if I can help you out any.

cocosolos:
@Brierre How are you opening the doors to the teleporter rooms? That's pretty much the biggest thing I'm missing, getting the Temenos doors to open and the teleporter pads to light up in both zones.

scripts/globals/battlefield.lua
```diff
@@ -229,7 +229,7 @@ function dsp.battlefield.ExtendTimeLimit(battlefield, minutes, message, param, p
     if message then
         players = players or battlefield:getPlayers()
         for _, player in pairs(players) do
-            player:messageSpecial(message, param or minutes)
+            player:messageBasic(message, param or minutes)
```
TeoTwawki: we might need 2 different versions as I am pretty sure some battlefields use a zone specific messageSpecial while others use the game-wide messageBasic, and IDs between the 2 are not the same.

If we wanna be fully retail accurate that is. I dunno if any currently use message Special, if none are I'd personally say screw those lets use the global one like you have this right now >.>

cocosolos: Yeah this is currently only being used in Limbus, but I'm not sure where it would be used in the future so I just reverted it back to messageBasic and made a new function specific to Limbus. What other battlefields even have time extensions? Limbus uses a zone specific message "Your time in Limbus has been extended." followed by a Limbus specific time-remaining message.

TeoTwawki: github seems to have taken away my ability to jump to the line we are talking on from the comment here, but I initially mistook this for the time remaining general message. Which..every bf has in some form, inc every bcnm isnm dyna limbo etc etc _everything._ I think they don't all use the same one normally (which is kinda annoying).

TeoTwawki:
rebase please :heart:

scripts/globals/mobskills/citadel_buster.lua
```diff
@@ -0,0 +1,86 @@
...
+    if (phase == 4) then
+        mob:setLocalVar("citadelBuster", 1)
+        mob:SetMobAbilityEnabled(false)
+        mob:SetMagicCastingEnabled(false)
+        mob:SetAutoAttackEnabled(false)
+        mob:setMobMod(dsp.mobMod.DRAW_IN, 1)
+        local battlefield = mob:getBattlefield()
+        local players = battlefield:getPlayers()
+        for _, member in pairs(players) do
+            member:messageSpecial(ID.text.CITADEL_BASE)
```
TeoTwawki: exactly zero of this belongs in the skillcheck, and should happen before the mob is even trying to activate the skill instead. Then when your countdown is over, force call the skill via the function we have for that purpose.

scripts/globals/mobskills/nuclear_waste.lua
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    return 1
+end
+
+function onMobWeaponSkill(target, mob, skill)
+    mob:setLocalVar("nuclearWaste", 1)
+    local typeEffect = dsp.effect.ELEMENTALRES_DOWN
+    target:addStatusEffectEx(dsp.effect.ELEMENTALRES_DOWN, 0, 50, 0, 60)
+    skill:setMsg(msg)
```
TeoTwawki: `msg` is nil in this context. I understand why you didn't use `MobStatusEffectMove()` but you need to add the logic for what message should be set here (and anything else we would normally be doing that this should use) -or- just modify the function to let you use the subid and Ex version for adding your status effect.

cocosolos: oh whoops this was a last minute change. It should be giving no message and at one point was set to dsp.msg.basic.none but I was messing around with different ways to do this and ended up with just making it 100% apply for now and forgot to change the message back.

scripts/globals/mobskills/citadel_buster.lua
cocosolos: I'm not sure where/how to move this without needing to manually set when to use the skill. It should be used randomly like any other tp move, but pulling these messages out of the skillcheck means I would need to manually call that at a certain hp percent or rely on random, which ends up just basically proccing over and over.

scripts/globals/mobskills/nuclear_waste.lua
TeoTwawki: okies :+1: 

I resume there is also no resist calc to be done. so if there is no message at all with it, that'll do.

scripts/globals/mobskills/citadel_buster.lua
TeoTwawki: I was about to explain the logic for timing it right, but then I realized the mixin for the mob 1hr/2hr moves can be hijacked for this to make ita  lot simpler. Unlock at 25% hp, allow it to get used more than once. mixin will handle the rest,

scripts/globals/mobskills/nuclear_waste.lua
cocosolos: It actually should be able to resist so I'm gonna have to edit the MobStatusEffectMove function.

scripts/globals/mobskills/citadel_buster.lua
TeoTwawki: 1 min will talk in discord

scripts/globals/mobskills/nuclear_waste.lua
TeoTwawki: a reist calc is small enough you could do it here if you want, there a function for the resist check that can be tossed to a variable, then you just check the resist value against the target tier or use it as a modifier (depending on what resistance does for this particular effect.

It's in the form of a float value because dsp's resist system is convoluted af.

For hit or mis style or resists its:
```lua
    local resist = applyPlayerResistance(whatever);
    if resist >= 0.25 then
        --stuff
    end
```
for reduced duration we have:
```lua
    local resist = applyPlayerResistance(whatever)
    local totalDuration = utils.clamp(duration * resist, 1);
```

and for dmg reduction we have:
```lua
    local resist = applyPlayerResistance(whatever)
    dmg = dmg * resist
```

TeoTwawki: if you do choose to edit MobStatusEffectMove move tho that will help future additions

cocosolos: Awesome, thanks! I'm pretty sure it's hit or miss. It doesn't do any damage and the duration I'm using right now is completely made up anyway since it's pretty hard to see how this effect actually works. I'm even clearing it immediately if you get hit by the elemental ability that comes next.

TeoTwawki: Just in case, see if `handleNinjutsuDebuff()` already does what you need

cocosolos: That doesn't look like I can use it to lower all elements at the same time, also doesn't seem to resist at all.

TeoTwawki: well in theory you'd just call it for each element. but if you need a resist check thats no good.

scripts/globals/mobskills/citadel_buster.lua
KnowOne134: You could move it to the mobs script under onMobWeaponskill(mob, target, skill) I believe, then make it do the messages if skill = c. Buster

cocosolos: The problem with that is onMobWeaponskill executes after the skill charging animation starts. The skill check it the only place with lua access between the ai picking which skill to use and the start of the charge animation.

Issue #6250, opened 2019-10-19
Pathing NPCs can (fairly consistently) crash server


ibm2431:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 301907xx_x

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Was working on some addons, and figured I would use [Red Ghost](https://github.com/DarkstarProject/darkstar/blob/master/scripts/zones/Port_Jeuno/npcs/Red_Ghost.lua) in Port Jeuno to test. Problem is he somewhat consistently crashes the server if he's spoken to a couple times after the first (after which his pathing breaks and he stops walking).

I suspect all NPCs set to path could cause this same crashing behavior.

I do use Enternity, so that may (or may not) be involved.

**Debugger Output:**

DEBUG_BREAK_IF thrown in [CAIContainer, Line 160](https://github.com/DarkstarProject/darkstar/blob/master/src/map/ai/ai_container.h#L160)

**Call Stack:**
- DSGame-server.exe!CAIContainer::ForceChangeState<CInactiveState,CBaseEntity * &,std::chrono::duration<__int64,std::ratio<1,10000000>> &,bool &>(CBaseEntity * & <args_0>, std::chrono::duration<__int64,std::ratio<1,10000000>> & <args_1>, bool & <args_2>) Line 160
- DSGame-server.exe!CAIContainer::Inactive(std::chrono::duration<__int64,std::ratio<1,10000000>> _duration, bool canChangeState) Line 165
- DSGame-server.exe!CLuaBaseEntity::wait(lua_State * L) Line 1945
- DSGame-server.exe!Lunar<CLuaBaseEntity>::thunk(lua_State * L) Line 173
- [External Code]	
- lua51.dll![Frames below may be incorrect and/or missing, no symbols loaded for lua51.dll]
- DSGame-server.exe!parse(char * buff, unsigned int * buffsize, sockaddr_in * from, map_session_data_t * map_session_data) Line 596
- DSGame-server.exe!do_sockets(fd_set * rfd, std::chrono::duration<__int64,std::ratio<1,10000000>> next) Line 376
- DSGame-server.exe!main(int argc, char * * argv) Line 270
[...]

PR #6255, opened 2019-10-21
Fixing patrol for Red Ghost in Lower Jeuno

kaincenteno:
Only lines 10 to 28 have been changed, all other line that shows as
changed are only removal of end-line semicolons to align with style
guide.

This fix was grabbed in part from https://github.com/DarkstarProject/darkstar/blob/c78bd68becd20a7250d78af03fcc7b2cf9972a78/scripts/zones/Port_Jeuno/npcs/Red_Ghost.lua
from @Wiggo32 

Tested and fixed it. This fix only resolves the NPC stopped patroling
after reaching the last point in path, it does not solve the NPC from
jumping all over the place to catch up to the current location the
server thinks where it should be at (if it wasn't stopped)

If this gets approved I'll go ahead and do the same for NPCs that
currently do not have patrol set up, like several in Windurst
which should make the areas more lively (I know its something cosmetic
but it's a good to have, specially for new people in server)


Wiggo32:
the "wait" command is going to cause the issues discussed here: https://github.com/DarkstarProject/darkstar/issues/6250 if dsp is gung ho on keeping broken tools i'd suggest at least commenting it out so people can "decide" to enable borked npc's whilst keeping a more stable alternative by default.

TeoTwawki:
@wiggo32 its not being “gung ho” on keeping anything broken, but I would like a better solution that `wait()` even if it didn’t crash. It seems awkward to me to be saying from lua _”hey I know I tol you do this, but hol up a sec!”_

`wait()` itself does what its supposed to but we did something wrong core side I think and now it can’t do its job. That seems like it should be doing the job of pausing pathing when ontrigger/ontrades happen in the first place as it halts the entire script (for a set duration or single tick, ad near as I can tell).

If you want to open a pr commenting them all out I am good with that until its fixed. Not sure what happens if the npc moves out of range mid event though.. the issue did not receive much attention - we have squeakier wheels and a lack of manpower. Presumably we hadn’t heard much more due to the major servers all having worked around it already or people just bot frequently triggering these NPCs.

TeoTwawki:
@kaincenteno need ti rebase it bud

Wiggo32:
@TeoTwawki I know its nothing you had anything to do with but I've offered to comment out "waits" long ago until it could be effectively fixed and instead i was guided into changing them into a different wait format which was a lot of work that still didn't solve the issue.... so, not much either of us could do. And I know we probably shouldn't have these types of lengthy discussions in github but this is the option dsp has left me with.

scripts/zones/Port_Jeuno/npcs/Red_Ghost.lua
```diff
@@ -3,59 +3,53 @@
...
+end
 
 function onEventFinish(player,csid,option,npc)
     if (csid == 314) then
-        player:setMaskBit(player:getVar("WildcatJeuno"),"WildcatJeuno",15,true);
+        player:setMaskBit(player:getVar("WildcatJeuno"),"WildcatJeuno",15,true)
     end
 
-    npc:wait(0);
+    npc:wait(0)
```
TeoTwawki: https://github.com/DarkstarProject/darkstar/issues/6250#issuecomment-544594726
take the zero out and tell me if he crashes.

TeoTwawki:
> @TeoTwawki I know its nothing you had anything to do with but I've offered to comment out "waits" long ago until it could be effectively fixed and instead i was guided into changing them into a different wait format (#4152) which was a lot of work that still didn't solve the issue.... so, not much either of us could do. And I know we probably shouldn't have these types of lengthy discussions in github but this is the option dsp has left me with.

Yeah I get it. Ima just ignore anyone that said not to and gd do it tho, unless that person volunteers to fix it immediately :P

I actually expect all these `wait(0)` I see hanging about to be a crash..its designed to expect a time input OR do 4 seconds, sending it zero seems like a bad idea.. with no arg it should be 4 seconds tho so if that crashes some interaction with it is causing it, like not being able to pause after an event start command is running (I see that all start after the startEvent call has happened)

kaincenteno:
@TeoTwawki  i tried with the wait() and the char still went all over the place.
changed wait() to use speed(0) and once event is done back to speed(12) [walk]
there's a timer set in case that the PC DCs

The only bug in which this wouldn't be 100% retail accurate would be that the NPC sometimes skips the next coordinate/or doesn't go all the way to last coordinate. This would have to be fixed inside the pathfind.lua

If this is approved I can go ahead and start implementing patrolling for npc's that are lacking it; then since the paths and logic is already in place would be a matter of someone taking ownership of how patrolling works. Unfortunately I do not have the knowledge of C++ to fix that matter but could be set as a confirmed issue. (at least this PR would fix one of the issues not properly but getting there since currently NPCs with patrol stop moving once they reach their last coordinate which is just as bad)

Forced pushed my changes, waiting on checks to complete

Wiggo32:
You need more than 3 points for an NPC to "path" on otherwise the distance is too great and the NPC will appear to "slide" between points.

kaincenteno:
Will add the extra points once back home in 6 hours

Wiggo32:
Don't do it on *MY* suggestion... I'm not the one you have to please... for all we know you may be facing a giant rejection. I was merely giving insight on how to take care of that "bug" you described. Figured you might get some guidance from someone more important than I...

TeoTwawki:
I'd have to see this version active in game to know but I trust him to test it, we spoke in discord about it a day or so ago. In the mean time, try not to wig out wiggo. :)

General guide: This NPC has a small area he basically paces in but in general you want nodes far enough apart that they don't stutter-step but not so far that they appear to teleport or glide, and for npc's that loop back to their starting point the order those nodes come up in when looped (you want the full set to include leading back to its beginning so it doesn't teleport there).  I have no idea how many path nodes this NPC should have at the moment. The older version being replaced here had lots of nodes pretty close together looks like.

kaincenteno:
@TeoTwawki  Red_Ghost also waits rights next to bastok and sandoria's port entrance for 6 seconds, looking towards [Sauromugue Champaign]. This I cannot do the way pathfinding is done. Would be good just to close for now until pathfinding is fixed/improved.

TeoTwawki:
I couldn't get it working proper either, the cores pathing for NPCs is busted even if you never click on them at all. https://yadi.sk/i/f3nj0XQ2Ov857A 

BTW I played about with the node spacing and 3-5 yalms seems optimal. but once it reaches the last node it will get stuck. If you drag the NPC away with GM commands it will look like it has resumed moving - its really just your client moving it back to its actual position it is stuck at. it will always go back to that exact spot, because its not really moving and your client is receiving that pos from the servers entity packet.

I was also unable to trigger a crash with the `wait()` commands.
_wheres the kaboom? there was supposed to be an earth shattering kaboom!_

It also didn't appear to even really stop. on my screen the NPC stops moving, but in the server while I'm in the event looked like it was still walking. The source code indicates that wait called with no value will default to 4 seconds. I'm unsure if feeding it zero will actually clear it when it is running. I suspect its been doing nothing except the occasional crash from overlapping things that I apparently can't get red here to do.

TeoTwawki:
```lua
-- Area: Port Jeuno
--  NPC: Red Ghost
-- Standard Info NPC
require("scripts/globals/pathfind")
require("scripts/globals/quests")

local path =
{
    -97, 0.001, -10, -- we're supposed to stop for a moment here and look out over the bridge each time we reach this pos
    -97, 0.001, -5,
    -97, 0.001, 0,
    -97, 0.001, 5,
    -97, 0.001, 10,
    -97, 0.001, 5,
    -97, 0.001, 0,
    -97, 0.001, -5, -- go back to 1st pos after this
}

function onSpawn(npc)
    npc:initNpcAi()
    npc:setPos(dsp.path.first(path))
    onPath(npc)
end

function onPath(npc)
    dsp.path.patrol(npc, path, dsp.path.flag.WALK)
    if not npc:isFollowingPath() then
        npc:pathThrough(dsp.path.first(path), dsp.path.flag.WALK)
    end
end

end

    local WildcatJeuno = player:getCharVar("WildcatJeuno")

    if
        player:getQuestStatus(JEUNO,dsp.quest.id.jeuno.LURE_OF_THE_WILDCAT) == QUEST_ACCEPTED
        and not player:getMaskBit(WildcatJeuno, 15)
    then
        player:startEvent(314)
    else
        player:startEvent(34)
    end

    -- Default wait with no value is 4 seconds.
    npc:wait() -- Not sure this is even stopping the path
end

end

function onEventFinish(player,csid,option,npc)
    if csid == 314 then
        player:setMaskBit(player:getCharVar("WildcatJeuno"), "WildcatJeuno", 15, true)
    end

    npc:wait(0) -- I have no idea if this actually clears the timer or not
    onPath(npc) -- Resume pathing. Tested and without this NPC does not walk again.
end
```

PR #6258, opened 2019-10-23
Added afterZoneIn and style guide fix

kaincenteno:
This zone has to change music while holding key item MESSAGE TO JEUNO (BASTOK)
Unfortunately this cannot be done in onZoneIn, tried it and music will stay the default.
If added to on afterZoneIn, this will work but will make the default metalwork music 
be heard for about .5 seconds before switching to correct one.
Besides this, went ahead and removed semicolons and parenthesis in if statements
according to style guide preference.


TeoTwawki:
We're going to need the music changed back to normal after the next cs finishes, because otherwise it will continue to be that track until the player re-zones.

As much as this is not retail accurate we don't have a better way to handle it atm.

PR #6262, opened 2019-10-25
Nagi - Augments Mijin Gakure

neuromancerxi:
modifier.h and status.lua: Adds MOD 957 MIJIN_NAGI

item_mods.sql:  Adds MOD 957 with value of 1 to all variants of Nagi with
'Augments "Mijin Gakure"' text.

abilities/mijin_gakure.lua: Adds check for MIJIN_NAGI > 0 and remove any
existing reraise and apply a reraise effect if so.

map/entities/charentity.cpp: Adds a step to check if Mijin Gakure was
used AND a Nagi is equipped, when processing reraise.

### DIFF
--- a/scripts/globals/abilities/mijin_gakure.lua
+++ b/scripts/globals/abilities/mijin_gakure.lua
@@ -25,6 +25,10 @@ function onUseAbility(player,target,ability)
     target:takeDamage(dmg, player, dsp.attackType.SPECIAL, dsp.damageType.ELEMENTAL)
 
     player:setLocalVar("MijinGakure", 1)
+    if (player:getMod(dsp.mod.MIJIN_NAGI) > 0) then
+        player:delStatusEffect(dsp.effect.RERAISE)
+        player:addStatusEffect(dsp.effect.RERAISE,1,0,3600)
+    end
     player:setHP(0)
     return dmg
 end
--- a/scripts/globals/abilities/mijin_gakure.lua
+++ b/scripts/globals/abilities/mijin_gakure.lua
@@ -25,9 +25,8 @@ function onUseAbility(player,target,ability)
     target:takeDamage(dmg, player, dsp.attackType.SPECIAL, dsp.damageType.ELEMENTAL)
 
     player:setLocalVar("MijinGakure", 1)
-    if (player:getMod(dsp.mod.MIJIN_NAGI) > 0) then
-        player:delStatusEffect(dsp.effect.RERAISE)
-        player:addStatusEffect(dsp.effect.RERAISE,1,0,3600)
+    if (player:getMod(dsp.mod.MIJIN_RERAISE) > 0) then
+        player:sendReraise(1)
     end
     player:setHP(0)
     return dmg
--- a/scripts/globals/abilities/mijin_gakure.lua
+++ b/scripts/globals/abilities/mijin_gakure.lua
@@ -25,9 +25,6 @@ function onUseAbility(player,target,ability)
     target:takeDamage(dmg, player, dsp.attackType.SPECIAL, dsp.damageType.ELEMENTAL)
 
     player:setLocalVar("MijinGakure", 1)
-    if (player:getMod(dsp.mod.MIJIN_RERAISE) > 0) then
-        player:sendReraise(1)
-    end
     player:setHP(0)
     return dmg
 end
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -1538,12 +1538,13 @@ dsp.mod =
     BERSERK_DURATION                = 954, -- Berserk Duration
     AGGRESSOR_DURATION              = 955, -- Aggressor Duration
     DEFENDER_DURATION               = 956, -- Defender Duration
+    MIJIN_NAGI                      = 957, -- Nagi - Reraise no weakness full HP
 
     -- The spares take care of finding the next ID to use so long as we don't forget to list IDs that have been freed up by refactoring.
     -- 570 - 825 used by WS DMG mods these are not spares.
-    -- SPARE = 957, -- stuff
     -- SPARE = 958, -- stuff
     -- SPARE = 959, -- stuff
+    -- SPARE = 960, -- stuff
 };
 
 dsp.latent =
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -1105,7 +1105,7 @@ dsp.mod =
     DEATHRES                        = 255,
     AFTERMATH                       = 256,
     PARALYZE                        = 257,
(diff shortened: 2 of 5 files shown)


scripts/globals/abilities/mijin_gakure.lua
```diff
@@ -25,6 +25,10 @@ function onUseAbility(player,target,ability)
     target:takeDamage(dmg, player, dsp.attackType.SPECIAL, dsp.damageType.ELEMENTAL)
 
     player:setLocalVar("MijinGakure", 1)
+    if (player:getMod(dsp.mod.MIJIN_NAGI) > 0) then
+        player:delStatusEffect(dsp.effect.RERAISE)
+        player:addStatusEffect(dsp.effect.RERAISE,1,0,3600)
```
TeoTwawki: What if you already had a reraise with a higher power than this? Is replacing with rr 1 what is supposed to happen?

neuromancerxi: I searched the youtubes for a reference video of how this is supposed to behave generally, and for the RR effect specifically, without success.

That said, I considered multiple ways to handle it (creating reraise Tier 4, creating a completely separate buff, etc), but without a reference it is what it actually does, it's a just a best guess. I'd like to believe that a reraise that gives you full HP would be at a higher tier of effectiveness than RR1-3, so I planned for it that way.

Alternately, without the removal, if you have any reraise effect active it will show two reraise icons momentarily, but still process the buffed reraise due to nagi first based on the order of operations.

This just seemed like the cleanest way to present it without having a working example.

TeoTwawki: well what you have here would delete my reraise3 and then apply reraise1, and your core check is looking specifically for reraise1 only. Since mijin prevents the exp loss, if it consumes any existing reraise anyway its not going to matter I guess. If it lets reraise status stay then this might get complicated.

We should probably just leave m_hasRaise out of the check entirely, you'll get the result without a blinky status re-add.

TeoTwawki: p.s. you also have to change the line that checks for rr >0 to check for this.

TeoTwawki: I'd also prefer we call this mod MIJIN_RERAISE if since we're not reusing existing mods for equipped item reraise.

(if we do need additional RR tiers, this is where you'd edit for it: https://github.com/DarkstarProject/darkstar/blob/ba4d44dc6039b61c0b217596f9176939930949a9/src/map/entities/charentity.cpp#L1666 and not mess with the status effect)

TeoTwawki: Just noticed the existing mod for mijin appears to have been refactored out and replaced with a localvar.

https://github.com/DarkstarProject/darkstar/commit/bd91bc2583dc03e8007c236d8a9f9eccf9435943

https://github.com/DarkstarProject/darkstar/blob/a110d870d57587c7bcf6ba29f711a1744aa88eec/src/map/modifier.h#L292

so I guess modID 258 can become MIJIN_RERAISE ?

scripts/globals/abilities/mijin_gakure.lua
```diff
@@ -25,9 +25,8 @@ function onUseAbility(player,target,ability)
     target:takeDamage(dmg, player, dsp.attackType.SPECIAL, dsp.damageType.ELEMENTAL)
 
     player:setLocalVar("MijinGakure", 1)
-    if (player:getMod(dsp.mod.MIJIN_NAGI) > 0) then
-        player:delStatusEffect(dsp.effect.RERAISE)
-        player:addStatusEffect(dsp.effect.RERAISE,1,0,3600)
+    if (player:getMod(dsp.mod.MIJIN_RERAISE) > 0) then
+        player:sendReraise(1)
```
TeoTwawki: if you do the core right you don't need to force send the menu. also this would arrive at the player before they finished falling down.

TeoTwawki: > 
> ```c++
> // Check if we have a form of reraise
> if (m_hasRaise > 0 || (getMod(Mod::MIJIN_NAGI) != 0 && GetLocalVar("MijinGakure") != 0))
> ```
> 
> ```c++
>         // Check for Nagi equipped (MOD) and Mijin Gakure was used.
>         if (GetLocalVar("MijinGakure") != 0 && getMod(Mod::MIJIN_NAGI) != 0)
>         {
>             actionTarget.animation = 511;
>             hpReturned = (uint16)(GetMaxHP());
>         }
>         else if (m_hasRaise == 1)
> ```

https://github.com/DarkstarProject/darkstar/pull/6262/files#diff-b3a74064477a12e9d4c37b8668ffad22R1468

neuromancerxi: > if (m_hasRaise > 0 || (getMod(Mod::MIJIN_NAGI) != 0 && GetLocalVar("MijinGakure") != 0))

Prior to submitting the changes in 8bc76a5, I tried using your example at charentity.cpp:1468, and it does not generate a Raise event (after using Mijin with Nagi and this check, no raise menu pops). There's likely another place that generates the Raise event. (OnRaise?). 

I opted for the compromise of pushing a reraise event directly if Nagi was present and tested Mijin with Nagi and RR3, With Nagi and RR1, With Nagi and no RR, With RR and No Nagi, and without RR and without Nagi, all of which returned the anticipated correct results.

If you can point me to another place to generate the raise event without either pushing Reraise directly or pushing the reraise effect, I'd be curious to see where it resides and how to leverage it to accomplish the task.

TeoTwawki: > Prior to submitting the changes in 8bc76a5, I tried using your example at charentity.cpp:1468, and it does not generate

And I don't know you've tried it till you tell me :) however, while my oversight in my sample code neglects the issue of needing call to onRaise to happen first, another place I pointed at just happens to cover this:
```cpp	
void CCharEntity::Die(duration _duration)
{
    m_deathSyncTime = server_clock::now() + death_update_frequency;
    PAI->ClearStateStack();
    PAI->Internal_Die(_duration);

    // reraise modifiers
    if (this->getMod(Mod::RERAISE_I) > 0)
        m_hasRaise = 1;

    if (this->getMod(Mod::RERAISE_II) > 0)
        m_hasRaise = 2;

    if (this->getMod(Mod::RERAISE_III) > 0)
        m_hasRaise = 3;

    // Check if we have MIJIN_RERAISE
    if (m_hasRaise == 0 && this->getMod(Mod::MIJIN_RERAISE) > 0)
        m_hasRaise = 1;

    CBattleEntity::Die();
}
```
The only requirement is the value be non zero before we hit the dirt, and I believe this would satisfy that.

Issue #6264, opened 2019-10-27
MNK unarmed has no weaponskills


ffxijuggalo:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30191004_0

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** Switched to mnk at lvl 1 and after gaining 5 lvls of hand to hand skill without a weapon equipped, I received the chat message that Combo had been learned. So I try to select it from the weaponskills menu on the next mob I was fighting, and it wasn't listed. Not listed from the main menu / abilities / weaponskills menu either. I went to the weapon shop in port windurst and bought a lvl 1 h2h weapon, equipped it, and sure enough, Combo shows up in the list of weaponskills. Unequip the weapon, and Combo disappears again. This is occuring on a branch built clean from master that is current as of writing this as well as an up to date client.

Cryptexx:
PR #6265

zynjec:
Fixed with https://github.com/DarkstarProject/darkstar/pull/6265

PR #6265, opened 2019-10-28
Set Unarmed Default Skill to 1, not 0

Cryptexx:
In response to issue #6264 , while unarmed, PItem initializes to null and therefore initializes skill to a default of 0. Since H2H weapon skills are Type ID 1, the for loop has nothing to iterate over within the global weapon skill list.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -2310,7 +2310,7 @@ namespace charutils
 
         //add in melee ws
         PItem = dynamic_cast<CItemWeapon*>(PChar->getEquip(SLOT_MAIN));
-        uint8 skill = PItem ? PItem->getSkillType() : 0;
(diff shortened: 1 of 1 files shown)


ShelbyZ:
Is there a situation where a player would gain access to H2H weaponskills when the job/sub-job combo does not natively have H2H skill?

Cryptexx:
> Is there a situation where a player would gain access to H2H weaponskills when the job/sub-job combo does not natively have H2H skill?

Those limitations are defined elsewhere. I tested this particular PR by changing jobs, though, and verified that whm, rdm, rdm, and other jobs have no access to H2H weaponskills, but Monk and Pup do. They also swing with one arm only, not two, on the jobs without access.

teschnei:
at the very least, the "1" should probably be the hand-to-hand skill enum

PR #6267, opened 2019-10-28
A Short Story of Hope

ibm2431:
Once upon a time there was a mob.

This mob looked like an Opo-opo.

However, [its true inner identity was that of something else](https://github.com/KnowOne134/NPC-MOB-Logger/blob/master/Tables/Temple%20of%20Uggalepih.lua#L10).

This poor mob was confused about what it was supposed to be. [It felt it was one thing, but it looked like another.](https://github.com/DarkstarProject/darkstar/blob/5ab3f15abb42ae0ba07b5b393cc2f5f3c437172e/sql/mob_spawn_points.sql)

Back in 2016, society tried to deny it of this inner identity, [instead attempting to force upon it the label of Opo-opo](https://github.com/DarkstarProject/darkstar/pull/3083).

But this mob would not be deterred, and it [kept proudly declaring its preferred label, even when it didn't look the part](https://user-images.githubusercontent.com/13112942/67663776-27b62000-f95e-11e9-8ecf-052712e87c3d.jpg).

Now in the year 2019, society should be willing to accept it for what it truly wishes to be. [And it will finally have both the label, and look, of its desired identity - that of a Wespe.](https://user-images.githubusercontent.com/13112942/67664076-d2c6d980-f95e-11e9-8176-6d93326ce8d0.jpg)

A story of hope for all: stay strong - eventually you will be accepted for who you are.


kaincenteno:
"it all began with a wespe, or so the legend said"

PR #6269, opened 2019-10-30
Anti-cheating system

NiQ1:
Implemented an anti-cheating system. The system logs all cheating attempts into a new table called "cheating_incidents", containing character ID, cheat ID (according to enum), the time of the incident, the severity of the cheat, an optional argument and an optional text description.
There is also an option to automatically jail offenders according to a configurable option in the map server configuration file.

Also created a first cheating countermeasure - If the player attempts to inject packets to speed up crafting (i.e. sending the synth done 0x59 packet before the synth animation is finished) it will be caught and reported (player may also be jailed according to the configuration) and also the synth will fail.
EDIT: This has been tested with the "fastcraft" Ashita addon and successfully blocked it.

It's my first contribution so please don't be too harsh.

Cheers.


src/map/utils/synthutils.cpp
```diff
@@ -835,6 +836,29 @@ int32 startSynth(CCharEntity* PChar)
 int32 doSynthResult(CCharEntity* PChar)
 {
     uint8 m_synthResult = PChar->CraftContainer->getQuantity(0);
+    if (map_config.anticheat_enabled)
+    {
+        std::chrono::duration animationDuration = server_clock::now() - PChar->m_LastSynthTime;
+        if (animationDuration < 5s)
+        {
```
ibm2431: Is this not [already being covered in map/packet_system.cpp](https://github.com/DarkstarProject/darkstar/blob/master/src/map/packet_system.cpp#L3776-L3780)?

conf/map_darkstar.conf
```diff
@@ -199,3 +199,9 @@ healing_tick_delay: 10
 #Central message server settings (ensure these are the same on both all map servers and the central (lobby) server
 msg_server_port: 54003
 msg_server_ip: 127.0.0.1
+
+#Set to 1 to enable server side anti-cheating measurements
+anticheat_enabled: 1
+#Set to change the cheating severity level that will cause players to be automatically jailed.
+#Defaults to 4 (severe). Set to 0 to disable auto-jailing.
+#anticheat_jail_threshold: 4
```
TeoTwawki: Should prolly default these off for now, but we'll need the team weigh in on the system in general. May be awhile, everybody has been busy with IRL stuff lately. Thank you in advance for your patience.

src/map/utils/synthutils.cpp
NiQ1: No, apparently it only checks how long after you finish synth before you can start synth again, it doesn't check if you inject a synth end packet right after beginning.
I've verified that by using the said "fastcraft" addon. Without the patch I was able to craft without waiting for the animation.

ibm2431:
> It's my first contribution so please don't be too harsh.

If you're not already on DSP's Discord, you should come say hi~!

(And if you are, you can still say hi anyway! Or not. Up to you!)

conf/map_darkstar.conf
NiQ1: OK, I'll disable it once the CI tests end (pass or fail). Apparently I messed up some makefiles so I had to push again anyway.
@TeoTwawki I may need some assistance if they don't pass, I compile on Visual Studio so makefiles are different. Sorry about that.

NiQ1:
> 
> 
> > It's my first contribution so please don't be too harsh.
> 
> If you're not already on DSP's Discord, you should come say hi~!
> 
> (And if you are, you can still say hi anyway! Or not. Up to you!)

Could you paste the link please?

ibm2431:
https://discord.gg/Xuu5YX7 !

src/map/anticheat.h
```diff
@@ -0,0 +1,81 @@
...
+#ifndef _CANTICHEAT_H
+#define _CANTICHEAT_H
+
+#include "../common/cbasetypes.h"
+#include "entities/charentity.h"
+
+namespace anticheat
+{
+    // Cheat type as appears in the cheat id column in the cheat_incidents table.
+    enum CHEAT_ID
```
zynjec: These should be enum classes.

Also not sure why we need STUB/LAST.

NiQ1: Could you please elaborate what do you mean by enum class?
Regarding your second comment, I can remove the STUB if you wish though I would like to reserve the value zero for edge cases. I can start the enum from 1.
Regarding the LAST, it's for the possibility of using if (cheatid >= CHEAT_ID_LAST) { error }, it's a coding convention that I (and my company) use.

src/map/utils/synthutils.cpp
```diff
@@ -835,6 +836,29 @@ int32 startSynth(CCharEntity* PChar)
...
+            #endif
+            PChar->CraftContainer->setQuantity(0, synthutils::SYNTHESIS_FAIL);
+            m_synthResult = SYNTHESIS_FAIL;
+            doSynthFail(PChar);
+            // And report the incident (will possibly jail the player)
+            anticheat::ReportCheatIncident(PChar,
+                anticheat::CHEAT_ID_FASTSYNTH,
+                anticheat::CHEAT_SEVERITY_PROBABLE,
+                (uint32)std::chrono::duration_cast<std::chrono::milliseconds>(animationDuration).count(),
+                "Player attempted to bypass synth animation by injecting synth done packet.");
```
zynjec: Do we really need `CHEAT_ID` if we're passing a verbose string to the table? Or at least have it tied in some fashion to the ID.

NiQ1: It's intended to allow statistics SQL queries, e.g. find all players who used a specific cheat to see how widespread it is. In this case I put a constant line in the (optional) description but maybe others will make more versatile use of it.

src/map/anticheat.h
TeoTwawki: [example of an enum class](https://github.com/DarkstarProject/darkstar/blob/56ba22932de4855c04cabc42c7bde748c1dd07e8/src/map/packets/char_emotion.h#L31)

conf/map_darkstar.conf
NiQ1: OK, disabled by default. Makefiles also seem to work (CI tests passed).

src/map/anticheat.h
NiQ1: Removed the STUB.

ibm2431:
Some thoughts:
1) Should we consider a DB table linking cheat ID to severity and string? That way a server can define which cheats they consider more severe than others.

2) Instead of comparing against severity as an integer value for when to jail, should there be a bitmask of what actions to take for a given cheat ID? Like Log, Notify GM, Jail, Ban, Feed to Vrtra?

(I'm only 80% joking on Vrtra)

src/map/utils/synthutils.cpp
zynjec: Do we need the description column at all then? CHEAT_ID saying exactly what they were jailed for?

src/map/anticheat.h
zynjec: https://en.cppreference.com/w/cpp/language/enum
http://www.stroustrup.com/C++11FAQ.html#enum

TeoTwawki:
> 2. Instead of comparing against severity as an integer value for when to jail, should there be a bitmask of what actions to take for a given cheat ID? Like Log, Notify GM, Jail, Ban, Feed to Vrtra?
> 
> (I'm only 80% joking on Vrtra)

Needs at least 50% more setting homepoint to Jormy while simultaneously setting Jormy in rage state

NiQ1:
> 
> 
> Some thoughts:
> 
>     1. Should we consider a DB table linking cheat ID to severity and string? That way a server can define which cheats they consider more severe than others.
> 
>     2. Instead of comparing against severity as an integer value for when to jail, should there be a bitmask of what actions to take for a given cheat ID? Like Log, Notify GM, Jail, Ban, Feed to Vrtra?
> 
> 
> (I'm only 80% joking on Vrtra)

Haven't thought of that... that could be a good idea to let server admins decide on action for each violetion. I don't have time for large refactors now. Hopefully tomorrow I will.
Do we already have mechanics for notifying GMs and banning?

src/map/anticheat.h
NiQ1: OK then I can't use them as the code is right now because I actually cast them to int when writing to DB. I actually thought of using an ENUM type in DB but apparently there's a considerable performance hit.
Assuming I'll have time for @TeoTwawki's suggestion (probably tomorrow) I'm going to add a cheat types table anyway and I can use that to associate id and name.

src/map/utils/synthutils.cpp
NiQ1: Just to give maximum flexibility. Yes, right now I used a constant string but it doesn't have to be that way. You can provide whatever information you want.

TeoTwawki:
> Do we already have mechanics for notifying GMs and banning?

Just server console and log messages not all GMs would currently have access to. I planned something similar to what you are doing (thank you) and intended eventually (say..3 or 4 years down the road..) to make a GM command that could pass parameters to a function that would execute a query on the table and then phase out all those print messages.

NiQ1:
> 
> 
> > Do we already have mechanics for notifying GMs and banning?
> 
> Just server console and log messages not all GMs would currently have access to. I planned something similar to what you are doing (thank you) and intended eventually (say..3 or 4 years down the road..) to make a GM command that could pass parameters to a function that would execute a query on the table and then phase out all those print messages.

Then how about I lay the grounds by creating the said bitmask table, for now the only supported actions will be log and jail and as we develop more actions that could be used (e.g. notify GM) it will be very easy to add.

src/map/anticheat.h
zynjec: There is nothing stopping you from casting it, enum classes just don't implicitly cast the enum to integer.

```cpp
enum class CheatReason : uint8 
{
    SYNTH_ANIMATION = 0,
    ZONE_EXPLOIT = 1
}

static_cast<uint8>(CheatReason::ZONE_EXPLOIT)
```

src/map/utils/synthutils.cpp
zynjec: I don't feel like we need both, if you want verbosity like that for every jail event stick with just the description column and scrap `CHEAT_ID`'s. You can just as easily query for a specific (or fuzzy) string as you can a number.

src/map/anticheat.h
NiQ1: Done.

src/map/utils/synthutils.cpp
NiQ1: I've accepted @ibm2431's proposal and will refactor this tomorrow to include a cheat_types table that allows server admins to configure the punitive measures for each one. Cheat ID will then serve as the primary key for that table. Let's continue this discussion tomorrow after I push the change.

TeoTwawki:
>> Should we consider a DB table linking cheat ID to severity and string? That way a server can define which cheats they consider more severe than others.

.

> Then how about I lay the grounds by creating the said bitmask table, for now the only supported actions will be log and jail and as we develop more actions that could be used (e.g. notify GM) it will be very easy to add.
EDIT: Maybe can also add a "warn player" option, which sends a system message to the player telling them they've been caught cheating and that it will be reported.

If we did do a table for severity separately from the logging table, you can orient it by severity of response rather than assuming severity of cheat, make everything default do do nothing or do warn, and server ops can just edit the table to change their default responses instead of having settings config entries. Just hypothetically.

NiQ1:
> 
> 
> > > Should we consider a DB table linking cheat ID to severity and string? That way a server can define which cheats they consider more severe than others.
> 
> .
> 
> > Then how about I lay the grounds by creating the said bitmask table, for now the only supported actions will be log and jail and as we develop more actions that could be used (e.g. notify GM) it will be very easy to add.
> > EDIT: Maybe can also add a "warn player" option, which sends a system message to the player telling them they've been caught cheating and that it will be reported.
> 
> If we did do a table for severity separately from the logging table, you can orient it by severity of response rather than assuming severity of cheat, make everything default do do nothing or do warn, and server ops can just edit the table to change their default responses instead of having settings config entries. Just hypothetically.

Yeah, I'm gonna completely remove the severity field during the refactor, it's won't be needed anymore anyway.

teschnei:
can you update the issue description with the actual detection methods/what
they're looking for?

teschnei:
okay nevermind, I looked at the header and it's only fast synth so far

I don't mind the idea for the framework, but why don't we just fix the exploit instead of messing with people and making their crafts fail?

TeoTwawki:
> okay nevermind, I looked at the header and it's only fast synth so far
> 
> I don't mind the idea for the framework, but why don't we just fix the exploit instead of messing with people and making their crafts fail?

~~Wheres the fun in that?~~ er I mean, lets fix the things, obviously, but a system in place to log questionable events someplace other than the map server's log file is also nice.

NiQ1:
> 
> 
> okay nevermind, I looked at the header and it's only fast synth so far
> 
> I don't mind the idea for the framework, but why don't we just fix the exploit instead of messing with people and making their crafts fail?

Hi! Thanks for your comment.
I'm not sure how it works on retail (and I sure don't want to find out the hard way, even though with all the RMT going on there I'm not even sure they give a f**k about cheaters anymore) but checking for the very few places in the DSP project where there is some sort of cheating protection (only 2 places so far) the general idea was to foil the cheating attempt so that's what I did.

NiQ1:
Refactored the anti-cheating framework as suggested by @ibm2431 :
* Added cheat_types table with the following columns -
- cheatid - Primary key (and link from cheat_incidents).
- name - Short textual name for readability (just so admins don't have to remember numbers).
- description - Description of the cheat (again, intended so admins don't have to dig for information).
- argument - Not used in the only cheat currently supported but will be used in the future, e.g. if we make a cheat type for huge gil transfers this will indicate the threshold.
- action_bitmask - What to do when a player is caught, see further description below.
- warning_message - Warning message to send to the player (if action_bitmask indicates it).

action_bitmask is a bitmask specifying what punitive measures to take against offenders. Currently supported actions are -
* Log (0x01) - Log the incident in the cheat_incidents table.
* Block (0x02) - Block the cheating attempt. How this is done is dependent on the cheat itself. In the case of the fast synth cheat it causes all such synths to fail. Added the ability to disable blocking per @teschnei's recommendation as this allows silent logging of cheaters / speculative actions.
* Warn (0x04) - Send a message specified in the warning_message column to the offender as a system message.
* Jail (0x08) - One way trip to Mordion Gaol. Note that auto-jailing can be globally disabled from the map configuration file so the offender will only be sent to jail if both this bit is enabled and if jailing is not disabled in the global configuration.

PLEASE NOTE - I haven't had the time to test these changes yet so please don't merge even if you approve. I will test it later today. Already uploading so I can already get your comments and speed things up when I do have access to the game.

Thank you all.
@ibm2431 @TeoTwawki @teschnei @zynjec

KnowOne134:
If you speed craft in retail it auto full fails. No skillups

NiQ1:
> 
> If you speed craft in retail it auto full fails. No skillups

Thank you. I hope you did not learn that the hard way.
Sounds like my default config (reporting and breaking synth but not jailing) is close to the retail behavior, is that correct?

KnowOne134:
Tested lastnight by someone. No jail yet. It breaks the crystal and lose all mats without even the synth animation starting

NiQ1:
> Tested lastnight by someone. No jail yet. It breaks the crystal and lose all mats without even the synth animation starting

I suggest to be careful when testing cheats on retail. I know they usually don't give a f**k anymore but wouldn't trust that 100%.
Anyway I don't want to mess with the code too much just to make sure you lose all mats. If no objections I'll leave it as a regular synth fail (=always lose the crystal, possibly other masts).

src/map/anticheat.cpp
```diff
@@ -0,0 +1,131 @@
...
+        // Check what we should do
+        char warningmsg[256] = { 0 };
+        CheatActionBitmask action = GetCheatPunitiveAction(cheatid, warningmsg, sizeof(warningmsg));
+        if (action & CHEAT_ACTION_LOG)
+        {
+            // Log intgo cheat_incidents table
+            const char* fmtQuery = "INSERT INTO cheat_incidents SET charid = %u, incident_time = '%s', cheatid = %u, cheatarg = %u, severity = %u, description= '%s';";
+            char strIncidentTime[128];
+            time_t timeNow = time(NULL);
+            strftime(strIncidentTime, sizeof(strIncidentTime), "%Y:%m:%d %H:%M:%S", gmtime(&timeNow));
```
zynjec: Is there a reason we're doing this in C++ instead of just leveraging MySQL to deal with it?

Set `incident_time`'s type to `TIMESTAMP`, default it to `CURRENT_TIMESTAMP`, if you don't pass a value in the query it'll default to the current time.

src/map/anticheat.h
```diff
@@ -0,0 +1,86 @@
...
+#ifndef _CANTICHEAT_H
+#define _CANTICHEAT_H
+
+#include "../common/cbasetypes.h"
+#include "entities/charentity.h"
+
+namespace anticheat
+{
+    // Cheat type as appears in the cheat id column in the cheat_incidents table.
+    enum class CheatID : uint32
```
zynjec: Probably don't need to waste 4 bytes on this, if we have 4 billion cheats we need to track I'll just delete the repo instead.

src/map/anticheat.h
```diff
@@ -0,0 +1,86 @@
...
+        // Player attempted to speed up synth (craft) by injecting a synth done
+        // (0x59) packet. If anti cheating is enabled this will cause all cheated
+        // crafts to fail.
+        // Argument is the time spent on synth animation.
+        CHEAT_ID_FASTSYNTH = 1,
+        // For boundary checks
+        CHEAT_ID_LAST
+    };
+
+    enum CheatActionBitmask
```
zynjec: enum class

src/map/anticheat.h
```diff
@@ -0,0 +1,86 @@
...
+    {
+        // Do nothing, also indicates failure
+        CHEAT_ACTION_NOTHING = 0,
```
zynjec: If they're being used for bitwise operations, probably makes sense to have these in hex/base 6.

src/map/anticheat.cpp
NiQ1: I just looked at it again. incident_time's type is DATETIME because on Windows and AMD64 Linux time_t is 64-bit. CURRENT_TIMESTAMP returns a TIMESTAMP value, which is only 32-bit and also apparently may cause issues with older versions of MySQL / MariaDB if attempting to store as DATETIME.
I know that 2038 is pretty far away but I'd rather not write new code that doesn't take it into account. If there's any MySQL function that returns a proper 64-bit DATETIME I'd be glad to get rid of that ugly (I admit that!) code.

src/map/anticheat.h
NiQ1: I'd rather not use enum class for bitmasks, causes the code to become uglier when doing bitwise operation. I only use enum because C++ doesn't really have a proper bitmask type that I know of (I think that using #define in CPP is even worse).
Regarding the second comment, changed (will push soon).

src/map/anticheat.cpp
TeoTwawki: @zynjec @niq1 its the same thing I did in 2 of our other tables that take the same thing. I vaguely recall talking about it with some mysql guys over at their forums and there was an alternative I was going to test out, but I never got around to it..and don't remember wtf it was now.

This current method is basically the answer you'd get if you ask stack overflow.

zynjec: Looks like `DATETIME` still allows a default of `CURRENT_TIMESTAMP`. Sure it might break eventually, but I am okay with kicking the can down the hall for 19 years for clean(er) code now.

src/map/anticheat.h
NiQ1: Fixed.

src/map/anticheat.h
NiQ1: Fixed.

src/map/anticheat.cpp
NiQ1: @TeoTwawki I'd also like to hear your opinion which is least bad.

TeoTwawki: doing it right would probably annoyingly involve not using either of them, but instead do something with `UTC_TIMESTAMP()` and epochs. I was lazy so we wound up with copypasta on our current columns that are using `DATETIME` and a few bad practices like `NOT NULL DEFAULT '0000-00-00 00:00:00'`

src/map/anticheat.h
NiQ1: Pushed.

src/map/anticheat.cpp
NiQ1: That still does not solve the 32-bit time issue...

NiQ1: Changed even though I don't really like it.

NiQ1:
As far as I see I've completed working on all review comments. Please let me know if there are further issues. If it's OK then please authorize and let me know :)

PR #6272, opened 2019-11-04
Fix ACTION_ITEM_FINISH to include Item ID in packet

ibm2431:
Per Windower documentation:
https://github.com/Windower/Lua/wiki/Action-Category-05
And retail capture:
paralyze-potion-retail


kaincenteno:
what's the Item Finish bit for?

ibm2431:
For the sake of accurate server emulation towards the client.

For example, client addons that work on retail may expect the item ID in this field. But on DSP, the addon would have stopped working because the server didn't send out the complete packet information that retail does.

(Which is exactly how I found that this information wasn't being set and sent on DSP.)

PR #6273, opened 2019-11-06
Added enable caskets and droprate multiplier to settings.lua

Omnione:
This is a prerequisit of the future addition of caskets, it will do nothing right now.
Added the ability to enable caskets through settings and also the abilty to adjust base droprate of caskets through a multiplier if so desired.

### DIFF
--- a/scripts/globals/settings.lua
+++ b/scripts/globals/settings.lua
@@ -33,7 +33,7 @@ ENABLE_GROUNDS_TOMES  = 1; -- Enables Grounds of Valor
 REGIME_WAIT = 1; -- Make people wait till 00:00 game time as in retail. If it's 0, there is no wait time.
 
 -- TREASURE CASKETS
-ENABLE_CASKETS = 0; -- Set to 1 to enabled caskets to spawn on mob death in in applicable zones, 0 to disable.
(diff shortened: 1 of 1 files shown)


Omnione:
I don't think travis likes me lol, 2 lines changed and he died

ibm2431:
Don't need `ENABLE_CASKETS` if `DROP_RATE` is 0~!

TeoTwawki:
as IBM is pointing out, we can default to what we think the retail rate is and check for 0 to disable it rather than making the setting itself a multiplier,

TeoTwawki:
I also don't see why you are preemptively pr'ing the setting before it really exists

Omnione:
tbh I don’t know why I pr’ed this first, close it if you want, I’ll include it when I actually feel brave enough to pr the global script, in regards to the enable caskets setting, I was using that to mainly disable the mixins from being added to mobs if it’s turned off, unless it’s not a problem letting the mixins being added to every mob in 59 zones if you don’t want caskets.

Omnione:
Gonna close this, no point in it being PR'ed yet, sorry for doing that, and i have taken the advice onboard and going to remove the enable caskets setting and go with the setting rate to 0 to disable.

thanks everyone for commenting

PR #6286, opened 2019-11-16
Remove hard-coded entity IDs from luas that are not IDs.lua

wrenffxi:
Let's keep NPC IDs only in the IDs.lua files in each zone.  The IDs can shift over time, and I only keep in sync the IDs.lua files when doing client updates.

(Note: There are still lots of hard-coded IDs in Dynamis, Limbus, and Abyssea.  And a couple in Ilrusi Atoll.  Will get to those eventually.)

### DIFF
--- a/scripts/zones/Beadeaux/Zone.lua
+++ b/scripts/zones/Beadeaux/Zone.lua
@@ -16,12 +16,12 @@ require("scripts/globals/zone")
 
 function onInitialize(zone)
     -- The Afflictor System (RegionID, X, Radius, Z)
-    zone:registerRegion(1, -163, 10, -137, 0,0,0) -- 17379798 The Afflictor
-    zone:registerRegion(2, -209, 10, -131, 0,0,0) -- 17379799 The Afflictor
-    zone:registerRegion(3, -140, 10,   20, 0,0,0) -- 17379800 The Afflictor
-    zone:registerRegion(4,  261, 10,  140, 0,0,0) -- 17379801 The Afflictor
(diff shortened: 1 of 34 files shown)


scripts/zones/Beadeaux/Zone.lua
```diff
@@ -16,12 +16,12 @@ require("scripts/globals/zone")
 
 function onInitialize(zone)
     -- The Afflictor System (RegionID, X, Radius, Z)
-    zone:registerRegion(1, -163, 10, -137, 0,0,0) -- 17379798 The Afflictor
-    zone:registerRegion(2, -209, 10, -131, 0,0,0) -- 17379799 The Afflictor
-    zone:registerRegion(3, -140, 10,   20, 0,0,0) -- 17379800 The Afflictor
-    zone:registerRegion(4,  261, 10,  140, 0,0,0) -- 17379801 The Afflictor
-    zone:registerRegion(5,  340, 10,  100, 0,0,0) -- 17379802 The Afflictor
-    zone:registerRegion(6,  380, 10,   60, 0,0,0) -- 17379803 The Afflictor
+    zone:registerRegion(1, -163, 10, -137, 0,0,0)
```
TeoTwawki: Lets keep a small note mentioning that these 6 are for Afflictors on the 1st registerRegion line

wrenffxi: The comment is just above the registerRegion lines (it was already there; I just left it)

TeoTwawki: /fail

I completely missed it

PR #6289, opened 2019-11-18
Mob level QC zones 1-100

wrenffxi:
All these changes are based on recent retail captures done by Nyu.

Broke commits up into batches of five zones each.  If a block of five is missing a commit, that means I found no discrepancies in those five zones.

Mostly it's mob ranges that were off by a level or two.  In a few cases, I created new groups for mobs that were split into level tiers.

*~~Note: I'm excluding Jugner Forest [S] (zone 82) from this PR because I don't trust the retail capture of this zone.  It contains many low level mobs I don't see on any wiki, and many level-tiers for the same mob -- sometimes up to four or five tiers for one mob!~~ Edit: fixed!*


ibm2431:
>  I'm excluding Jugner Forest [S] (zone 82) from this PR because I don't trust the retail capture of this zone.

_checks Nyu's DB file for Jugner, sees a Colibri in there_
"That doesn't look right."

It looks like the DB file for Jugner [S] got non-Jugner [S] mobs in it. The DB starts at ID 17105248, but Jugner [S] ID block _should_ start at 17113088. The block goes to 17114112, so the file is fine up until its end, after where it "truly" begins for Jugner [S] at 17113094

ibm2431:
There seem to be multiple batches of Campaign coffers in the file, so:

The file starts at 17105248, which is in zone 80.

It then runs until zone 81, which starts at 17108992.

And zone 81 has a max mob ID of 17110016. (The IDs from 17111446 to 17111452 are trusts in zone 81).

ibm2431:
My best guess is that the file tried to write and failed, so the zone database wasn't flushed out (because it error'd out mid-write), and the mobs in memory were carried over into another zone or two until a hard reload was done.

(I saw in one of the files that a non-ASCII NPC name was inside, which can at times stop a write, so some of these files might have been generated before I began "escaping" those.)

[Here's a list of mob blocks in case it might have happened more than once.](https://github.com/DarkstarProject/darkstar/files/3860329/mob_blocks.txt)

wrenffxi:
Nyu recaptured Jugner Forest [S] (zone 82) and I included it in the PR.  All set!

PR #6292, opened 2019-11-20
Mob level QC zones 101-300

wrenffxi:
Here are all the remaining zones.  All these changes are based on recent retail captures done by Nyu.

Broke commits up into batches of five zones each.  If a block of five is missing a commit, that means I found no discrepancies in those five zones.

Mostly it's mob ranges that were off by a level or two.  In a few cases, I created new groups for mobs that were split into level tiers.


sql/mob_groups.sql
```diff
@@ -7369,23 +7369,23 @@ INSERT INTO `mob_groups` VALUES (14,1738,119,'Goblin_Tinkerer',330,0,1032,0,0,16
 INSERT INTO `mob_groups` VALUES (15,1666,119,'Goblin_Gambler',330,0,1077,0,0,21,25,0);
 INSERT INTO `mob_groups` VALUES (16,1643,119,'Goblin_Butcher',330,0,1032,0,0,16,20,0);
 INSERT INTO `mob_groups` VALUES (17,4516,119,'Zombie',330,1,2809,0,0,16,26,0);
-INSERT INTO `mob_groups` VALUES (18,5727,119,'Chonchon',3600,0,2992,3154,0,60,60,0);
```
TeoTwawki: I KNEW IT. I FKING KNEW that lil s.o.b. was stronger than I remembered.
/vindicated

PR #6294, opened 2019-11-22
BCNM mob superlinking

### LINKED ISSUE #6119: Mobs not superlinking in BCNMs

cocosolos:
Set m_bcnmID and m_battlefieldID when inserting mob into battlefield. Fixes #6119.

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -326,6 +326,8 @@ bool CBattlefield::InsertEntity(CBaseEntity* PEntity, bool enter, BATTLEFIELDMOB
                 BattlefieldMob_t mob;
                 mob.PMob = static_cast<CMobEntity*>(PEntity);
                 mob.condition = conditions;
+                mob.PMob->m_bcnmID = this->GetID();
+                mob.PMob->m_battlefieldID = this->GetArea();
 
                 if (mob.condition & CONDITION_WIN_REQUIREMENT)
                     m_RequiredEnemyList.push_back(mob);


Hozu:
Aren't there certain BCNM fights where this behaviour would be undesirable? Such as Promy Vahzl (linking the next boss in order when the current one's HP gets low enough).

TeoTwawki:
@hozu they should still be added to the battlefield - we’d just override the behavior rather than leaving them out. Also, pretty much all of limbus we don’t want superlinked either (its technically a bcnm under the hood, so was old style dyna before we killed it).

cocosolos:
@Hozu Promy zones are marked as ZoneType Dungeon not Battlefield, and there's a few Battlefields that are specifically marked to not superlink, including the one you're talking about I think (desires of emptiness). I plan on adding Limbus to those zones in my Limbus PR.

TeoTwawki:
@cocosolos limbus might have its own type (or use/ take over the ild dynamis type) it it doesn’t already use one.

PR #6295, opened 2019-11-22
Erroneously used the signed i vs unsigned I in the tobytes() conversion.

Cryptexx:


TeoTwawki:
Pssst this is why we test things before we pr them

Cryptexx:
Worked on my rig haha

TeoTwawki:
“But it works on my machine”

“Then we’ll ship your machine”

This is probably how comer and heroku were born :rofl:

PR #6296, opened 2019-11-22
Remove superlinking from BCNM mob LUAs

wrenffxi:
Now that the core is properly superlinking BCNM mobs, we don't need the mob LUAs doing it.

Also removing the ID offsets that were only used in the superlink calculations.


scripts/zones/Horlais_Peak/IDs.lua
```diff
@@ -28,8 +28,6 @@ zones[dsp.zone.HORLAIS_PEAK] =
     },
     mob =
     {
-        HELLTAIL_HARRY_OFFSET = 17346645,
-        FIGHTING_SHEEP_OFFSET = 17346684,
```
TeoTwawki: Were these ever used? I can't seem to locate their purpose in the commit history.

TeoTwawki: (I know they aren't now, thats why you are gutting them. wandering wtf they were doing there in the first place tho)

zynjec: https://github.com/DarkstarProject/darkstar/commit/452de08317343a0a7537cac7acfdc50330af3b55

TeoTwawki: ty @zynjec

PR #6298, opened 2019-11-23
Jugner Forest LUA cleanup/style

wrenffxi:
* signposts identified by ID/offset rather than X/Y
* Meteormauler Zhagtegg added 2HR and damage reduction
* npcUtilified trades
* style guide

### DIFF
--- a/scripts/zones/Jugner_Forest/mobs/Meteormauler_Zhagtegg.lua
+++ b/scripts/zones/Jugner_Forest/mobs/Meteormauler_Zhagtegg.lua
@@ -2,6 +2,12 @@
 -- Area: Jugner Forest
 --   NM: Meteormauler Zhagtegg
 -----------------------------------
+mixins = {require("scripts/mixins/job_special")}
+-----------------------------------
+
+function onMobInitialize(mob)
+    mob:setMod(dsp.mod.DMG, -50) -- "Has an innate damage reduction trait, about 50%"
+end
 
 function onMobDeath(mob, player, isKiller)
 end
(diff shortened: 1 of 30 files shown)


TeoTwawki:
we'd rather you didn't put multiple things in one pr, much less in the same commit (plus style changes to boot), when doing these things. its makes it harder to review.

Also using coordinates in and of itself for the signposts isn't necessarily a bad thing - future proofs them from ID changes.

scripts/zones/Jugner_Forest/mobs/Meteormauler_Zhagtegg.lua
```diff
@@ -2,6 +2,12 @@
 -- Area: Jugner Forest
 --   NM: Meteormauler Zhagtegg
 -----------------------------------
+mixins = {require("scripts/mixins/job_special")}
+-----------------------------------
+
+function onMobInitialize(mob)
+    mob:setMod(dsp.mod.DMG, -50) -- "Has an innate damage reduction trait, about 50%"
```
TeoTwawki: clopededia is wrong here. he had magic dmg redux only (as bg correctly shows)
https://www.bg-wiki.com/bg/Meteormauler_Zhagtegg

TeoTwawki: Incidentally, the entire line of similar NMs all have either phys or mag dmg reduction and a posse (auto follows them) that does not always pop (separate timers maybe, who knows)

the quadav gets -50 physical

TeoTwawki: (which I can't prove atm wiki's don't list, just that one anecdotal comment on every NM..from a blu)

TeoTwawki: possibly their resistance is related to their posse being up, no way to know atm

wrenffxi:
will split into multiple PRs

PR #6299, opened 2019-11-23
BCNM allies and multi-phase fights

cocosolos:
This fixes some errors trying to spawn allies into a battlefield and updates some renamed BCNM functions.
This includes:
Prishe and Selh'teus in Dawn
Ajido-Marujido in Moon Reading
Karababa in Shield of Diplomacy
Trion in The Heir to the Light
Volker in Where Two Paths Converge

Also addressed is an issue where multi-phase fights will trigger win conditions if the target mob isn't spawned at the start.
These fights are:
Dawn
Moon Reading
The Heir to the Light
Where Two Paths Converge
One to be Feared
The Celestial Nexus
Shadow Lord

Fight logic wasn't checked other than spawning allies and stopping fights from ending prematurely. Allies currently will cause a crash just like pets until that is fixed. Working fine with #6212 with the only exception being the client (not server) crashes if you try to cure a dead Prishe.

### DIFF
--- a/scripts/globals/spells/raise.lua
+++ b/scripts/globals/spells/raise.lua
@@ -16,6 +16,9 @@ function onSpellCast(caster,target,spell)
         if (target:getName() == "Prishe") then
             -- CoP 8-4 Prishe
             target:setLocalVar("Raise", 1)
+            target:entityAnimationPacket("sp00")
+            target:addHP(target:getMaxHP())
+            target:addMP(target:getMaxMP())
         end
     end
     spell:setMsg(dsp.msg.basic.MAGIC_CASTS_ON)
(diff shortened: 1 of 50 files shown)


scripts/globals/spells/raise.lua
```diff
@@ -16,6 +16,9 @@ function onSpellCast(caster,target,spell)
         if (target:getName() == "Prishe") then
             -- CoP 8-4 Prishe
             target:setLocalVar("Raise", 1)
+            target:entityAnimationPacket("sp00")
```
TeoTwawki: @teschnei wasn't that animation already being handled elsewhere when you implemented these originally?

TeoTwawki: er, on prishe and aern's I mean , not the spell

cocosolos: One of the few things I changed was moving this stuff from Prishe to the spell because she wasn't accepting the raise before. It was in onMobFight which wasn't getting called while she was dead.

TeoTwawki: onMobFight seems like a terrible place to have it

PR #6314, opened 2019-12-02
Fixed order of cutscenes

zynjec:
Opening cutscene should always have top priority.

Also fixed check for ENABLE_SOA, have to check the actual value as lua will return true so long as the variable exists.

### DIFF
--- a/scripts/zones/Bastok_Markets/Zone.lua
+++ b/scripts/zones/Bastok_Markets/Zone.lua
@@ -17,22 +17,20 @@ end
 function onZoneIn(player,prevZone)
     local cs = -1
 
-    -- SOA 1-1 Optional CS
-    if 
-        ENABLE_SOA and 
-        player:getCurrentMission(SOA) == dsp.mission.id.soa.RUMORS_FROM_THE_WEST and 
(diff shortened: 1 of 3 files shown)


zynjec:
Hmm, true.

I'll throw something together later, or finally do what I've wanted for awhile with zone-in cutscenes and just return the cs value in-line instead of a value to avoid that garbage.

ibm2431:
> just return the cs value in-line instead of a value to avoid that garbage

💯 👍

scripts/zones/Bastok_Markets/Zone.lua
```diff
@@ -17,24 +17,22 @@ end
...
-    then
-        cs = 22
-    end
-
     -- FIRST LOGIN (START CS)
-    if (player:getPlaytime(false) == 0) then
-        if (OPENING_CUTSCENE_ENABLE == 1) then
-            --cs = 0
-        end
+    if player:getPlaytime(false) == 0 and OPENING_CUTSCENE_ENABLE == 1 then
```
TeoTwawki: don't we still want to set pos and homepoint even if opening cs is disabled? by merging those 2 if's that won't happen anymore.

PR #6319, opened 2019-12-05
Fix Ancient Vows CutScene order.

Doytoend:
Remove incorrect cutscene when triggering Spatial Displacement. (CS
involved Tenzen)
Add trigger to zone player to RA01 when on Ancient Vows.
On ZoneIn to RA01 the correct cutscence plays.

### DIFF
--- a/scripts/zones/Misareaux_Coast/npcs/Spatial_Displacement.lua
+++ b/scripts/zones/Misareaux_Coast/npcs/Spatial_Displacement.lua
@@ -14,7 +14,7 @@ function onTrigger(player,npc)
     if (player:hasCompletedMission(COP,dsp.mission.id.cop.SHELTERING_DOUBT)) then
         player:startEvent(551); -- Access to Sites A & B
     elseif (player:getCurrentMission(COP) == dsp.mission.id.cop.ANCIENT_VOWS and player:getCharVar("PromathiaStatus") == 1) then
-        player:startEvent(8);
+        player:setPos(732.55,-32.5,-506.544,90,30); -- Go to Riverne #A01 {R}
     else
         player:startEvent(550); -- Access to Site A Only
     end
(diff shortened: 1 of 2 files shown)


scripts/zones/Misareaux_Coast/npcs/Spatial_Displacement.lua
```diff
@@ -14,7 +14,7 @@ function onTrigger(player,npc)
     if (player:hasCompletedMission(COP,dsp.mission.id.cop.SHELTERING_DOUBT)) then
         player:startEvent(551); -- Access to Sites A & B
     elseif (player:getCurrentMission(COP) == dsp.mission.id.cop.ANCIENT_VOWS and player:getCharVar("PromathiaStatus") == 1) then
-        player:startEvent(8);
+        player:setPos(732.55,-32.5,-506.544,90,30); -- Go to Riverne #A01 {R}
```
TeoTwawki: while the cs event that was here may have been incorrect, I am almost certain retail does indeed have an event here that does this move on its event finish, before the event that happens after the zone change, even if all that event does is play an animation of you being sucked through the portal.

This can be confirmed (or proven wrong) via a retail capture of the mission here.

scripts/zones/Riverne-Site_A01/Zone.lua
```diff
@@ -18,11 +18,11 @@ end;
 
 function onZoneIn(player,prevZone)
     local cs = -1;
-
-    if (player:getXPos() == 0 and player:getYPos() == 0 and player:getZPos() == 0) then
+    if (player:getCurrentMission(COP) == dsp.mission.id.cop.ANCIENT_VOWS and player:getCharVar("PromathiaStatus") == 1) then
```
TeoTwawki: outer parenthesis is not needed. a lot of older scripts still have it and we haven't cleaned all of them yet, sorry for the numerous bad examples.

Also instead of making this the 1st check followed by an else, we probably need this to just be its own check after the original ends. That position 0 check is a safety for landing in the zone without a legit position. If you met the requirements for this conditional but somehow still landed with invalid pos data, you'd get stuck. (I'm aware of your change in the other file setting the pos - but a GM could teleport themselves in without triggering that, as just one example)

scripts/zones/Misareaux_Coast/npcs/Spatial_Displacement.lua
Doytoend: Noted: I'll addthe event for the option window to travel to RA01 and move stuff back to oneventfinish.

Doytoend: Just watched video there is indeed an option to enter the site. Will add it back and test it and send it up when ready.

Issue #6321, opened 2019-12-09
Windurst Mission 8-1 VAIN CS event missing parameters (needs !release to get out)


kaincenteno:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 
30191102_1

**_Source Branch_** (master/stable) **:** 
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Try trying to complete mission 9-1 for Windurst.
When visiting the Qu_Hua_Spring, the mysterious npc won't pop, making the entire CS broken and needing to have !release.

```
    if (CurrentMission == dsp.mission.id.windurst.VAIN and MissionStatus >= 1) then
        player:startEvent(2);
```

You can also test by calling the cs using !cs 2 while in Ro'Maeve

Not sure if this worked before but if it did then SE must of done something that broke it

PR #6322, opened 2019-12-09
update cs npc info - windurst 8-1 vain

cocosolos:
#6321


TeoTwawki:
another npc in the list has this exact position, did id shifts fubar this? the id your using I'm assuming is correct via a capture and the other is...who knows what but in wrong place now.

```
INSERT INTO `npc_list` VALUES (17277194,'no_name_2','     ',64,-0.053,-29.606,82.054,0,40,40,0,0,0,6,27,0x0000490000000000000000000000000000000000,32,NULL,1);
```

PR #6324, opened 2019-12-10
C++: Added code to make sure pets claim on spells

m241dan:
I swapped the order of the checkcallforhelp and checkplayer.
The reason for this is that the checkplayer change needs to be two
parts now.

Checks if its called for help, then if the caster of a spell was
a player, and if it is a player, it behaves like it used to.
Then checks to see if the caster was not a player, was it a pet
whose master was a player. In this case, it sets owner and targid
to the master.

### DIFF
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -669,6 +669,17 @@ void CMobEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
             PTarget->PAI->EventHandler.triggerListener("WEAPONSKILL_TAKE", PTarget, this, PSkill->getID(), state.GetSpentTP(), &action);
         }
 
+        if (objtype == TYPE_PET && PMaster && PMaster->objtype == TYPE_PC )
+        {
+            auto mob = dynamic_cast<CMobEntity *>(PTarget);
+            if (mob && !mob->CalledForHelp())
+            {
+                mob->m_OwnerID.id = PMaster->id;
+                mob->m_OwnerID.targid = PMaster->targid;
+                mob->updatemask |= UPDATE_STATUS; //This can go here because we only wanna call the updatemask if this happens
+            }
+        }
+
         if (msg == 0)
         {
             msg = PSkill->getMsg();
--- a/src/map/ai/states/magic_state.cpp
+++ b/src/map/ai/states/magic_state.cpp
@@ -305,12 +305,14 @@ void CMagicState::ApplyEnmity(CBattleEntity* PTarget, int ce, int ve)
 
                 if (!(m_PSpell->isHeal()) || m_PSpell->tookEffect())  //can't claim mob with cure unless it does damage
                 {
-                    if (m_PEntity->objtype == TYPE_PC)
+                    if (m_PEntity->objtype == TYPE_PC || (m_PEntity->PMaster && m_PEntity->PMaster->objtype == TYPE_PC))
                     {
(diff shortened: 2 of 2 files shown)


cocosolos:
looks like status should still update if it's CFH, I suggest this

```cpp
// magic_state.cpp L308-316
	if (m_PEntity->objtype == TYPE_PC)
	{
		if (!mob->CalledForHelp())
		{
			mob->m_OwnerID.id = m_PEntity->id;
			mob->m_OwnerID.targid = m_PEntity->targid;
		}
		mob->updatemask |= UPDATE_STATUS;
	}
	else if (m_PEntity->PMaster != nullptr && m_PEntity->PMaster->objtype == TYPE_PC)
	{
		if (!mob->CalledForHelp())
		{
			mob->m_OwnerID.id = m_PEntity->PMaster->id;
			mob->m_OwnerID.targid = m_PEntity->PMaster->targid;
		}
		mob->updatemask |= UPDATE_STATUS;
	}
```

something similar needs to be done in ability_state.cpp as well

m241dan:
Hmmmm, k.

TeoTwawki:
no tabs only spaces. 4 per indention level. also pls don't `if(  squish the parens and pad the inside   )` 

`if (we like this instead)`
`if (this is also (ok) )`

TeoTwawki:
`if (it kinda feels (like (whispering (getting quieter the farther I (go)))))`

m241dan:
Fixed the formatting stuff, and got the things in there for abilities.

However, I had to take a much different approach for abilities because the avatar (or even bst pets), don't use ability state. They use the mob_skill_state. Which seems like a bad place to put claiming mechanics. So, I shimmed it into the lua binding for updateEnmityFromDamage.

It makes sense to keep magic claiming in the magic AI state because there are some skills that pets can cast that don't immediately do damage (so they will still need to claim). There are no abilities, that i can think of, that pets use, that don't do damage and claim a mob.

m241dan:
Okay, pets claiming should work in the "best possible" manner now.

src/map/entities/mobentity.cpp
```diff
@@ -669,6 +669,17 @@ void CMobEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
             PTarget->PAI->EventHandler.triggerListener("WEAPONSKILL_TAKE", PTarget, this, PSkill->getID(), state.GetSpentTP(), &action);
         }
 
+        if (objtype == TYPE_PET && PMaster && PMaster->objtype == TYPE_PC )
+        {
+            auto mob = dynamic_cast<CMobEntity *>(PTarget);
+            if (mob && mob->objtype == TYPE_MOB && !mob->CalledForHelp())
```
nfaletra: If it casts to a CMobEntity, we already know it's a TYPE_MOB. Don't need to do both checks.

m241dan: ha, totally fair, if its not a mob dynamic cast would return nullptr, right?

nfaletra: Correct!

cocosolos: I'm also not sure if this should even check for TYPE_PET, just if it has a PC master, in case of a charmed mob killing an unclaimed link with an aoe mobskill.

m241dan:
Status on this?

nfaletra:
Status is we all have jobs and it's mid week. I'll take another look this weekend.

m241dan:
Understood, was just checking in...

PR #6327, opened 2019-12-16
Set phase 1 mobs of 2 phase fights to be required kills.

cocosolos:
Fixes an issue where party members can't enter these battlefields


kaincenteno:
have tested Shadowlord battle with Jameson and the fight went well with this fix 😸

jamesontai:
Yup, Shadow Lord battle was successful with this fix. =) Thanks.

PR #6332, opened 2019-12-21
move global stuff into dsp table, make function return bool instead of bitmask

TeoTwawki:
~~I failed to notice we were redefining global entries inside of a local table when I merged 63944e1d449f7ce5f6174f935778a104c8647dd0 ..But I don't think that needs to be global now that I think about it.~~

Teo is wrong. ignore the stuff Teo said.


scripts/globals/utils.lua
```diff
@@ -447,6 +380,73 @@ local relics =
 -- 10 = 119 III (ammo dispensing)
 -------------------------------------------------------
 function utils.getRelicStages(player, relic)
```
wrenffxi: How do you actually pass a meaningful relic parameter into this function, though?

cocosolos: Yeah this doesn't work for me unless I use strings.
```["SPHARAI"] =```
```utils.getRelicStages(player, "SPHARAI")```
etc

Comment (TeoTwawki) [reply scripts/globals/utils.lua]: > How do you actually pass a meaningful relic parameter into this function, though?

The way he wrote it, you don't. the function returns a bitmask of which version of a relic they have. so you don't specify which aegis you are looking for. you get a bitmask of which aegis they they have.

Comment (TeoTwawki) [reply scripts/globals/utils.lua]: > Yeah this doesn't work for me unless I use strings.
> `["SPHARAI"] =`
> `utils.getRelicStages(player, "SPHARAI")`
> etc
> or get rid of the relics table for a bunch of smaller tables
> 
> ```
> function utils.getRelicStages(player, relic)
>         SPHARAI =
>         {
>             18264, 18265, 18637, 18651, 18665, 19746, 19839, 20480, 20481, 21509
>         }
> ...
> for i, id in pairs(relic) do
> ```

the index is already technically a string and I forgot that

Comment (TeoTwawki) [reply scripts/globals/utils.lua]: @cocosolos man I am such a moron lately. it _was_ fine as you had it written and the problem I was looking at only happens after converting it to dsp.whatever.whatever _which the same table name_

Comment (TeoTwawki) [reply scripts/globals/utils.lua]: and then at the last minute I decided it was better off local _anyway_

Comment (cocosolos) [reply scripts/globals/utils.lua]: still think this should probably be
```lua
function utils.hasRelic(player, relic, stage)
    SPHARAI =
    {
        18264, 18265, 18637, 18651, 18665, 19746, 19839, 20480, 20481, 21509
    }
...
    if stage ~= nil then
        if player:hasItem(relic[stage]) then
            return true
        else
            return false
        end
    end
    for i, id in pairs(relic) do
        if player:hasItem(id) then
            return true
        end
    end
    return false
end
```

scripts/globals/utils.lua
```diff
@@ -364,94 +364,37 @@ function utils.getSystemStrengthBonus(attacker, defender)
...
+-- 9  = 119 II
+-- 10 = 119 III
+-- 11 = 119 III (ammo dispensing)
 -------------------------------------------------------
-function utils.getRelicStages(player, relic)
-    local stages = 0
-    for i, id in pairs(relics[relic]) do
-        if player:hasItem(id) then
-            stages = bit.bor(stages, bit.lshift(1, i - 1))
+function utils.getRelic(player, relic, tier)
```
TeoTwawki: if/when someone need to get mythics instead, they can edit this. If this works again, i am done with it..

scripts/globals/utils.lua
```diff
@@ -364,94 +364,37 @@ function utils.getSystemStrengthBonus(attacker, defender)
...
-    local stages = 0
-    for i, id in pairs(relics[relic]) do
-        if player:hasItem(id) then
-            stages = bit.bor(stages, bit.lshift(1, i - 1))
+function utils.hasRelic(player, relic, tier)
+    if tier ~= nil then
+        if player:hasItem(dsp.relicTiers[relic][tier]) then
+            return true
+        else
+            return false
```
zynjec:
```lua
return player:hasItem(dsp.relicTiers[relic][tier])
```

hasItem returns a boolean, no need for any of those if's

scripts/globals/utils.lua
```diff
@@ -364,94 +364,37 @@ function utils.getSystemStrengthBonus(attacker, defender)
...
         end
     end
-    return stages
+
+    for i, itemID in pairs(dsp.relicTiers[relic]) do
+        if player:hasItem(itemID) then
+            return true
+        end
```
zynjec: Same thing here.

cocosolos: Don't wanna return false on this one until after the loop is done

cocosolos: Actually I may be wrong about that, originally I wanted this to return the highest tier but I looks like I changed it at some point.

TeoTwawki: what he said

PR #6333, opened 2019-12-22
Matrimonial Coffers

kaincenteno:
Created global for matrimonial coffers locates in
Sandoria, Windurst, and Bastok.
Fixed 3 NPCs in Metalworks that were not up in sql file.
Added 3 Matrimonial Coffers scripts.

### DIFF
--- a/scripts/globals/matrimonialcoffer.lua
+++ b/scripts/globals/matrimonialcoffer.lua
@@ -1,0 +1,53 @@
+-----------------------------------
+-- Matrimonial Coffer NPCs
+-----------------------------------
+require("scripts/globals/npc_util")
+
+dsp = dsp or {}
+dsp.matrimonialcoffer = dsp.matrimonialcoffer or {}
+
+function dsp.matrimonialcoffer.startEvent(player)
+    local playerGender = player:getGender()
+    if playerGender == 1 then
+        player:startEvent(2000, playerGender, 200000, 15848, 100000, player:getGil())   -- Male Dialog
+    else
+        player:startEvent(2000, playerGender, 400000, 15847, 80000, player:getGil())    -- Female Dialog
+    end
+end
+
+function dsp.matrimonialcoffer.finishEvent(player, csid, option)
+    local zone = player:getZoneID()
+    local ID = zones[zone]
+    local playerGender = player:getGender()
+    if playerGender == 1 and csid == 2000 then
+        if option == 1 and player:getGil() >= 200000 and
+        (not player:hasItem(16364) or not player:hasItem(11309)) then
+            if npcUtil.giveItem(player, {16364, 11309}) then
+                player:delGil(200000)
+            end
+        elseif option == 1 and player:getGil() < 200000 then
+            player:messageSpecial(ID.text.NOT_HAVE_ENOUGH_GIL)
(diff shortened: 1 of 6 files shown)


scripts/globals/matrimonialcoffer.lua
```diff
@@ -0,0 +1,54 @@
...
+-- Matrimonial Coffer NPCs
+-----------------------------------
+require("scripts/globals/zone")
+require("scripts/globals/npc_util")
+
+dsp = dsp or {}
+dsp.matrimonialcoffer = dsp.matrimonialcoffer or {}
+
+function dsp.matrimonialcoffer.startEvent(player)
+    local race = player:getRace()
```
zynjec: Seems like `getGender` might be a better fit?

scripts/globals/matrimonialcoffer.lua
```diff
@@ -0,0 +1,54 @@
+-----------------------------------
+-- Matrimonial Coffer NPCs
+-----------------------------------
+require("scripts/globals/zone")
```
wrenffxi: unused require

kaincenteno: removed, thank you @wrenffxi  😸

scripts/globals/matrimonialcoffer.lua
kaincenteno: awesome @zynjec  didn't know how to use that since only saw it being used for dancer but this is hella awesome. Appreciate it.

kaincenteno:
Hi DSP friends :cat: 
Any other change I should make to matrimonial coffers or is this good for merge? :smile_cat:

TeoTwawki:
instead of so many calls to getGil(), how about variable it to use in those compares. I'm also thinking this could be structured to have fewer compares and use an else or 2 instead of so many elseif's repeating the same calls.

scripts/globals/matrimonialcoffer.lua
```diff
@@ -0,0 +1,53 @@
...
+            if npcUtil.giveItem(player, {15848}) then
+                player:delGil(100000)
+            end
+        elseif option == 2 and player:getGil() < 100000 then
+            player:messageSpecial(ID.text.NOT_HAVE_ENOUGH_GIL)
+        end
+    elseif playerGender == 0 and csid == 2000 then
+        if option == 1 and player:getGil() >= 400000 and (not player:hasItem(13933) or
+            not player:hasItem(14386) or not player:hasItem(14251) or not player:hasItem(14126)) then
+            if npcUtil.giveItem(player, {13933, 14386, 14251, 14126}) then
```
zynjec: I don't think you have to check with items using npcUtil.giveItem, it'll just fail (and then not remove the gil).

kaincenteno: Hi @zynjec  thanks for reviewing :cat: 
I'm kind of confused at your statement (must be my ESL).

making the check is necessary because npcUtil.giveItems is unaware of having to delete items from the player if it gives the item successfully this is why it has to be inside an if statement. so if it's true (able to give item because player has enough space) then it goes ahead and deletes the gil, if not true then it doesnt give the item and doesn't delete the gil.

PR #6336, opened 2019-12-22
Bcnm orb fix

m241dan:
Two things:

1. All orbs don't break on entrance now, just the trader's
2. Added the proper value for not finding an item in createWornItem(). I think due to the nature of unsigned 8 bit this was still doing the same thing, but I did see crashes where this block was moving forward with a nullptr. Maybe this fixes that, who knows? It was only happening in situations that shouldn't happen anyway. Either way, this is a slightly cleaner version of createWornItem().

### DIFF
--- a/scripts/globals/bcnm.lua
+++ b/scripts/globals/bcnm.lua
@@ -967,13 +967,15 @@ function EventFinishBCNM(player, csid, option)
             local stat = player:getStatusEffect(dsp.effect.BATTLEFIELD)
             local bfid = stat:getPower()
             local item = getItemById(player, bfid)
+            local initiatorId, initiatorName = player:getBattlefield():getInitiator()
+
             if item ~= 0 then
                 -- remove limbus chips
                 if zone == 37 or zone == 38 then
                     player:tradeComplete()
 
                 -- set other traded item to worn
-                elseif player:hasItem(item) then
+                elseif player:hasItem(item) and player:getName() == initiatorName then
                     player:createWornItem(item)
                 end
             end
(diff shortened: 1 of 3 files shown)


scripts/globals/bcnm.lua
```diff
@@ -967,13 +967,16 @@ function EventFinishBCNM(player, csid, option)
             local stat = player:getStatusEffect(dsp.effect.BATTLEFIELD)
             local bfid = stat:getPower()
             local item = getItemById(player, bfid)
+            local initiatorName = ""
+            _, initiatorName = player:getBattlefield():getInitiator()
```
zynjec: You don't have to initialize `initiatorName` before assigning it a value, this also leads to `_` being a global.

```lua
local initiatorId, initiatorName = player:getBattlefield():getInitiator()
```

m241dan: Okay, I'll do it that way.

Forgot to add Shami properly checking broken orbs anyway.

m241dan: Done.

scripts/globals/bcnm.lua
```diff
@@ -967,13 +967,15 @@ function EventFinishBCNM(player, csid, option)
             local stat = player:getStatusEffect(dsp.effect.BATTLEFIELD)
             local bfid = stat:getPower()
             local item = getItemById(player, bfid)
+            local _, initiatorName = player:getBattlefield():getInitiator()
```
TeoTwawki: you really don't want a variable named `_` do you? ..what is that being used for?

TeoTwawki: Zyn suggested `initiatorId` instead of `_` I'd go with that if you don't have any better ideas. Descriptive vars are better than throwaways. But I didn't see where its being used.

m241dan: The whole point of the _ var in lua is that it is not stored. It is best practice with lua. I don't ever use it, it is literally a throw away.

m241dan: Will change if you feel it is 100% necessary.

However, this is standard practice for things in Lua and even Python when you are dealing with  multiple returns and you don't want them.

TeoTwawki: didn't realize it was returning 2 things. name it what it is in case we use it later.

m241dan: Done

TeoTwawki:
I think you need to rebase onto current master, you have duplicate commits of recently merged things in here.

m241dan:
Rebased

PR #6340, opened 2019-12-27
Added Kindlix shop in Port Jeuno

kaincenteno:
Added Kindlix script and updated Port Jeuno IDs for her dialog

### DIFF
--- a/scripts/zones/Port_Jeuno/npcs/Kindlix.lua
+++ b/scripts/zones/Port_Jeuno/npcs/Kindlix.lua
@@ -1,0 +1,51 @@
+-----------------------------------
+-- Area: Bastok Markets
+--  NPC: Kindlix
+-- Standard Merchant NPC
+-- !pos -18.820 4.000 23.302 246
+-----------------------------------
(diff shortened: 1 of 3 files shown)


scripts/zones/Port_Jeuno/npcs/Kindlix.lua
```diff
@@ -0,0 +1,51 @@
...
+    end
+end
+
+function onEventUpdate(player,csid,option)
+end
+
+function onEventFinish(player,csid,option)
+    if csid == 348 and option == 0 then
+        dsp.shop.general(player, stock)
+        player:setCharVar("spokeKindlix", 1)
```
TeoTwawki: GDI, noticed right after I hit merge. this leaves a var forever. I HATE forever vars. Set it to 1 in character creation, then have the NPC check for and set 0 instead.

TeoTwawki: Pyropox has the same behavior btw.

PR #6343, opened 2019-12-30
Fix for Dynamis Mask check in Xarcabard

nfaletra:
* Check specific flag of the dynamis mask instead of the entire mask like a jerk

Dynamis_Status can change if you click on Trail Markings or other Dynamis NPCs prior to getting the cutscene in Xarcabard. This caused the cutscene to never happen when the player became level and mission eligible to receive the cutscene. Checking the specific bit fixes it.

### DIFF
--- a/scripts/zones/Xarcabard/Zone.lua
+++ b/scripts/zones/Xarcabard/Zone.lua
@@ -27,7 +27,7 @@ function onZoneIn(player, prevZone)
         player:setPos(-136.287, -23.268, 137.302, 91)
     end
 
-    if not player:hasKeyItem(dsp.ki.VIAL_OF_SHROUDED_SAND) and player:getRank() >= 6 and player:getMainLvl() >= 65 and player:getCharVar("Dynamis_Status") == 0 then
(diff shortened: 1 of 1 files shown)


scripts/zones/Xarcabard/Zone.lua
```diff
@@ -27,7 +27,7 @@ function onZoneIn(player, prevZone)
         player:setPos(-136.287, -23.268, 137.302, 91)
     end
 
-    if not player:hasKeyItem(dsp.ki.VIAL_OF_SHROUDED_SAND) and player:getRank() >= 6 and player:getMainLvl() >= 65 and player:getCharVar("Dynamis_Status") == 0 then
+    if not player:hasKeyItem(dsp.ki.VIAL_OF_SHROUDED_SAND) and player:getRank() >= 6 and player:getMainLvl() >= 65 and bit.band(player:getCharVar("Dynamis_Status"), 1) == 0 then
         player:setCharVar("Dynamis_Status", 1)
```
TeoTwawki: And its not a problem that those npc's are changing the var? or that immediately following this the mask is overwritten? (should we be setting just one bit here?)

My understanding is those other NPCs shouldn't be doing anything until you've gotten this cs. In the case of cop dyna's ones I thought you had to clear the original dyna to enter first.

TeoTwawki: resolving per our discord chat

PR #6346, opened 2020-01-03
Added Pyropox shop in Port Jeuno and fixed Kindlix charvar

kaincenteno:
Added Pyropox script and updated Port Jeuno IDs for her dialog.
Added initial charvar for both Kindlix and Pyropox on char creation.
Once player talks to them for first time, charvar is set to zero.


TeoTwawki:
```
Found item 4256 with buy price 25 which is lower than sell price 27 in script /home/travis/build/DarkstarProject/darkstar/tools/../scripts/zones/Port_Jeuno/npcs/Pyropox.lua
Found 1 errors
The command "python3 ./tools/price_checker.py" failed and exited with 1 during .
```

Issue #2569, opened 2016-01-06
Unable to kick people from LS


Deadwing888:
When using the "kick" command as a sack or shell-holder it will remove someone from linkshell group, but will not break their pearl. All they have to do to get back in is re-equip it.

It should be replacing their pearl in inventory with an identically colored broken one that is unequipable for shell chat.

I also believe in retail you had to have a sack to a linkshell to trade a pearl, even if you had an extra one in inventory. I believe darkstar will currently let pearls be traded freely once they are created.

Edit: the first part of this seems to be working now even though I saw no pull request about it or even a maint that would have implemented it. Closing the issue as the second half is not important

FrankWHansen:
I am not a sackholder, but can confirm it's possible to trade pearls freely. I was able to give my pearl from one of my characters to a new player on the server. On the other hand I'm not sure if that was possible on retail or not.

spectre99:
Just a quick poke/bump.  This is still an issue and has the potential to be annoying/disruptive.

PR #6350, opened 2020-01-05
Fix various issues with linkshells

cocosolos:
This adds several features to make linkshells match retail behavior. The biggest change is using the 9th byte of a linkshell items extra data to hold the linkshell type, rather than using an offset of the item id as is currently done (this appears to be how retail does it). Also added is the ability to change /lsmes privilege level to LS holder only, pearlsack holders, or everyone. These changes should require no input to existing shells for compatibility. Breaking the linkshell group has also been added. This will kick and break the pearls/sacks of all equipped members, then set the linkshell as broken in the database. Any offline or unequipped members will receive a message when they next try equipping which will then break the item. There has been a !breaklinkshell command added for GMs to break any linkshell group. ***This change DOES require that you add a column to the linkshells table.***

* break linkshell and all pearls/sacks by dropping linkshell or with gm command
* pearlsack kick breaks equipped pearl, linkshell kick breaks all pearls/sacks in all inventories
* lsmes privilege (/lsmes level ls, /lsmes level ps, /lsmes level all, /lsmes clear)
* equipable from any local inventory (inventory, satchel, sack, case)

Fixes pretty much everything in #2569 except trading pearls.

### DIFF
--- a/migrations/broken_linkshells.py
+++ b/migrations/broken_linkshells.py
@@ -1,0 +1,22 @@
+import array
+import mysql.connector
+
+def migration_name():
+	return "Adding broken column to linkshells table"
+
+def check_preconditions(cur):
+	return
+
+def needs_to_run(cur):
+	# Ensure broken column exists in linkshells
+	cur.execute("SHOW COLUMNS FROM linkshells LIKE 'broken'")
+	if not cur.fetchone():
+		return True
+	return False
+
+def migrate(cur, db):
+	try:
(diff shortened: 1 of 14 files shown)


TeoTwawki:
rebase required

TeoTwawki:
I'm wrong. gitub is stupidly putting out notices that you aren't up to date with master even when no conflict exists. it looks just like the conflict notice except with the actual words "edit conflicts" removed

PR #6354, opened 2020-01-06
Removed custom settings for Homepoints

zynjec:
Homepoint will always be enabled, per-retail, and no option to heal the player,

### DIFF
--- a/scripts/globals/homepoint.lua
+++ b/scripts/globals/homepoint.lua
@@ -147,7 +147,7 @@ local selection =
 local travelType = dsp.teleport.type.HOMEPOINT
 
 local function getCost (from, to, key)
-    
+
     if HPs[from].group == HPs[to].group and HPs[to].group ~= 0 then
         return 0
     else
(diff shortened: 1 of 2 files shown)


TeoTwawki:
I'm ok with killing the heal setting ( I mean, its kinda broken to the absurd with the new placements) and all our default behaviors should be current retail, but there is no reason to just give the finger to the old school servers entirely by stripping away the stuff they use. In the case of stupidly simple edits that didn't need a setting, explain in the commit that the setting isn't needed to achieve it.

zynjec:
> no reason to just give the finger to the old school servers entirely by stripping away the stuff they use

They can add it back themselves, we're targeting retail which has them enabled, and the setting didn't even properly disable them, the event can't be sent at all for 100% disabled.

TeoTwawki:
1 line misplaced caused it to not work. correct me if I am wrong but I believe this a simple edit to fix, I just want you to not willy nilly go nuking everything, ALL settings are non retail. so thats not a good reason for removal.

Areas we agree on:
- target retail. default behavior should be retail
- they can add it back

Areas we may not agree on:
- kill all the settings coz they aren't retail

pls reassure you won't head for the latter.

zynjec:
> 1 line misplaced caused it to not work. correct me if I am wrong but I believe this a simple edit to fix

Without adding a check for `HOMEPOINT_TELEPORT` in a bunch more functions, the only way to 100% disable it is no event at all.

The options for keeping the setting around are just having it PrintToPlayer (making HP unusable entirely), or making them work like they did in the past and mapping the TextID in every zone for the "Home point set" message.

TeoTwawki:
I don't have time to do this right now, but you are wrong. I have them disabled right now.

Hokuten85:
Pretty sure you can technically just send the appropriate packets to the server and completely bypass the menu, and there are addons out there that do this. So additional work is probably required to 100% disable them.

I don't contribute much to the project so I don't expect my opinion to hold much weight, but I do try to maintain my own custom server. I completely understand wanting the default to be retail, but removing stuff that's already built in with the justification of "they can just add it back" is pretty shitty. It hurt nothing to leave that in.

TeoTwawki:
> Pretty sure you can technically just send the appropriate packets to the server and completely bypass the menu, and there are addons out there that do this. So additional work is probably required to 100% disable them.

Yes, additional things need done to protect against injections - which still needs done anyways. Its not ok to keep allowing teleports to the ones the player never unlocked, even if SE hasn't seen fit to fix that yet on retail.

PR #6355, opened 2020-01-08
Fix True Will NM kill requirements

nfaletra:
The code that used to be here was checking if a nil value is dead in each NM script. It was impossible to complete the quest.

Also, made it retail accurate. If you're clever enough, you can kill only 1 NM and despawn the other two and still complete the quest. popFromQM does most of the heavy lifting for checking to make sure the mobs are despawned before attempting to spawn them again, etc.


TeoTwawki:
> you must wait for the other two Kappas to de-spawn before defeating the one you choose to fight.

TeoTwawki:
> you must wait for the other two Kappas to de-spawn before defeating the one you choose to fight."

This is what I was trying to explain but you thought I just didn't read your comment. And like I said then, someone else can re-add it later.

PR #6356, opened 2020-01-09
PR: Useful command setrank (sets mission rank)

m241dan:
Just found this useful

### DIFF
--- a/scripts/commands/setrank.lua
+++ b/scripts/commands/setrank.lua
@@ -1,0 +1,36 @@
+---------------------------------------------------------------------------------------------------
+-- func: setrank
+-- desc: Sets the players rank.
+---------------------------------------------------------------------------------------------------
+
+cmdprops =
+{
+    permission = 1,
+    parameters = "si"
+}
+
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!setrank <player> <new rank>")
+end
+
+function onTrigger(player, target, rank)
+    if target == nil then
+        error(player, "Set who's rank?")
+        return
+    end
+
+    local targ = GetPlayerByName(target)
+
+    if targ == nil then
+        error(player, string.format("Cannot find player: %s.", target))
+        return
+    end
+
+    if rank == nil or ( rank < 1 or rank > 10 ) then
+        error(player, "Improper rank passed. Valid Values: 1 to 10")
+        return
+    end
+    player:PrintToPlayer(string.format("You set %s's rank to %d", target, rank))
+    targ:setRank(rank)
+end


kaincenteno:
That's an awesome command, really handy 😸

scripts/commands/setrank.lua
```diff
@@ -0,0 +1,37 @@
...
+
+    targ = GetPlayerByName(target)
+
+    if targ == nil then
+        error(player, string.format("Cannot find player: %s.", target))
+        return
+    end
+
+    if rank == nil or ( rank < 1 or rank > 10 ) then
+        error(player, "Improper rank passed.\nValid Values: 1 to 10")
```
TeoTwawki: why not just 1 line? save muh precious vertical space. the game has wrapping built in for long lines / narrow user defines widths on menu.

m241dan: Just thought it was cleaner, but if you want it changed I will.

TeoTwawki: I think
```lua
error(player, "Improper rank passed. Valid values: 1 to 10")
```
Looks just as well.
If our binding wasn't hard coded to "whatever nation you presently are on" I'd say make which rank your adjusting an option too.

m241dan: Done

scripts/commands/setrank.lua
```diff
@@ -0,0 +1,37 @@
...
+
+cmdprops =
+{
+    permission = 1,
+    parameters = "si"
+};
+
+function error(player, msg)
+    player:PrintToPlayer(msg);
+    player:PrintToPlayer("!setrank <player> <amount>");
```
TeoTwawki: should probably be `<new rank>` instead of `<amount>`

scripts/commands/setrank.lua
```diff
@@ -0,0 +1,37 @@
+---------------------------------------------------------------------------------------------------
+-- func: setrank
+-- desc: Sets the players rank.
+---------------------------------------------------------------------------------------------------
+
+cmdprops =
+{
+    permission = 1,
+    parameters = "si"
+};
```
TeoTwawki: we dropped semicolons awhile back too, just fyi (they do nothing in lua)

m241dan: copy and paste left over

scripts/commands/setrank.lua
```diff
@@ -0,0 +1,38 @@
...
+    parameters = "si"
+}
+
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!setrank <player> <new rank>")
+end
+
+function onTrigger(player, target, rank)
+    local targ
```
TeoTwawki:
```lua
    if target == nil then
        error(player, "Set who's rank?")
        return
    end

    local targ = GetPlayerByName(target)
```
should be fine since targ isn't used before this line

TeoTwawki: sorry for having a small nitpicky looking thing every time you've made a change, just been browsing between support calls tonight. I think this is the last one.

m241dan: kk

PR #6357, opened 2020-01-09
PR: Lua Bugfix: Book Shell is way too powerful

m241dan:
It wasn't being divided properly, so book shell was giving waayyy too much.

### DIFF
--- a/scripts/globals/regimes.lua
+++ b/scripts/globals/regimes.lua
@@ -1138,15 +1138,14 @@ dsp.regime.bookOnEventFinish = function(player, option, regimeType)
             local power = 0
 
             if mLvl < 37 then
-                power = 24
+                power = 9
             elseif mLvl < 57 then
-                power = 36
+                power = 14
             elseif mLvl < 68 then
-                power = 48
+                power = 19
             else
-                power = 56
+                power = 22
             end
-
             player:delStatusEffectSilent(dsp.effect.SHELL)
             player:addStatusEffect(dsp.effect.SHELL, power, 0, 1800)
 


kaincenteno:
isn't the power the same as the highest shell casted by a WHM at the level of the target?

TeoTwawki:
kept reading as "book shelf" got very confused for a minute.

And yeah its supposed to be the same as whichever tier of shell you would have from whm at whatever level you are on receiving it from the book, but we never implemented that. Shell 2 potency starting at lv37 for example. If you trust wiki:
> Potency of the Shell is that of the highest tier spell cast by a White Mage of your current level, however it will not grant Shell V

Best make sure its the right scale though - looking at the whm spell script sit was changed from 256ths to 100ths and the book didn't get the memo.

m241dan:
So its good? Or is there a change, because I just copied the current shell spell scripts.

scripts/globals/regimes.lua
```diff
@@ -1146,7 +1146,7 @@ dsp.regime.bookOnEventFinish = function(player, option, regimeType)
             else
                 power = 56
             end
-
+            power = ( power * 100 ) / 256
```
nfaletra: Current shell spell scripts use whole numbers for the power. Copy those.

E.G. Shell = 9, Shell II = 14, etc.

m241dan:
Made changes as I believe Teo and I discussed.

TeoTwawki:
image
?????????

m241dan:
hmm I messed something up... sorry

m241dan:
ah, grabbed the wrong hash, yikes I'm tired.

PR #6360, opened 2020-01-10
Hasim shop added spells

kaincenteno:
Added scrolls (previously merited)


kaincenteno:
ah didnt see it was already added, closing this

PR #6361, opened 2020-01-11
Ephemeral Moogles

nfaletra:
Added support for Ephemeral Moogle crystal storage

- Global file handles triggers, trades, and failures
- Works exactly like retail. If you're at the cap, the failure event is sent to the client. If you trade any amount of crystals that can be stored, those crystals will be stored but the rest will be ignored and returned to the player's inventory. If the player trades a crystal amount that will put them above the cap, enough crystals to hit the cap will be traded while the rest are returned to the player
- Renamed the moogles in zones that have more than 1 moogle so csids can be contained to the npc scripts and IDs do not have to be updated outside the databse
- Updated addItem to handle adding a quantity of items that would take up more than a single slot of inventory space. Tested with the moogles adding crystals to the player and also tested using !additem command.
- Added to the python migration script to add the new *_crystals columns to char_points

EDIT: Only thing I didn't touch but kinda wanted to was their Content Tags. Should they be SOA or ROV, maybe?


tagban:
This appears to be causing loops when an itemid that doesn't exist in database is added using either additem (or a steal item with id of 0) (0 is used on a few mobs where the itemid isn't confirmed yet it appears). Basically causes game server to crash. 

[28/Jan] [10:56:16][Warning] charplugin::AddItem: Item <0> is not found in a database
[28/Jan] [11:12:11][Warning] charplugin::AddItem: Item <23733> is not found in a database

nfaletra:
I’ll add item verification to fix it. But the change is actually needed to allow adding multiple stacks of an item at once.

nfaletra:
Fixed. https://github.com/DarkstarProject/darkstar/pull/6403

PR #6362, opened 2020-01-11
No reward for True Will if any of the NMs are still up

nfaletra:
So I forgot to check to make sure no NMs were up before allowing the player to grab the reward. Now it correctly checks that you killed at least one of them and the rest have despawned

### DIFF
--- a/scripts/zones/Yhoator_Jungle/npcs/qm3.lua
+++ b/scripts/zones/Yhoator_Jungle/npcs/qm3.lua
@@ -15,7 +15,8 @@ end
 
 function onTrigger(player,npc)
     if player:getQuestStatus(OUTLANDS, dsp.quest.id.outlands.TRUE_WILL) == QUEST_ACCEPTED and not player:hasKeyItem(dsp.ki.OLD_TRICK_BOX) then
-        if player:getCharVar("trueWillKilledNM") > 0 then
+        local mobsStillUp = GetMobByID(ID.mob.KAPPA_AKUSO):isSpawned() or GetMobByID(ID.mob.KAPPA_BONZE):isSpawned() or GetMobByID(ID.mob.KAPPA_BIWA)
+        if not mobsStillUp and player:getCharVar("trueWillKilledNM") > 0 then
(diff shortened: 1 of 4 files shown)


TeoTwawki:
This still isn't retail. Retail won't let you kill one and then depsawn them after - they have to depop before the one you kill or you have to fight them again. This is already confirmed behavior.

nfaletra:
Hmm. Okay. I’ll have to think about that, then.

nfaletra:
That should do it. Similar to what was there before but grabbing the mobs correctly.

scripts/zones/Yhoator_Jungle/npcs/qm3.lua
```diff
@@ -15,7 +15,8 @@ end
 
 function onTrigger(player,npc)
     if player:getQuestStatus(OUTLANDS, dsp.quest.id.outlands.TRUE_WILL) == QUEST_ACCEPTED and not player:hasKeyItem(dsp.ki.OLD_TRICK_BOX) then
-        if player:getCharVar("trueWillKilledNM") > 0 then
+        local mobsStillUp = GetMobByID(ID.mob.KAPPA_AKUSO):isSpawned() or GetMobByID(ID.mob.KAPPA_BONZE):isSpawned() or GetMobByID(ID.mob.KAPPA_BIWA)
```
TeoTwawki: This is still just attempting to confirm all 3 are not currently up.

There is zero way the npc can tell if the mob was killed or just depopped, also that last one in the var just checks if the mob exists. basically whats needs to happen is checking isSpawned at time of death on the others - whether they are dead or depopped won't matter, so long as its the last of the 3 to be killed by the player. the scripts you changed in the 1st pr mistakenly check on nil entities they had not defined and was checking the wrong state (I think - maybe a despawned mob still counts as "dead" tho). That author was using for `KAPPA_AKUSO` instead of `GetMobByID(ID.mob.KAPPA_AKUSO)` which is why it was nil.

TeoTwawki: nvm jithoob was showing me old version I had to refresh sorry

PR #6364, opened 2020-01-12
Update incorrect D value upper caps in Blue Magic.

mrhappyasthma:
I did a quick audit of the spells. I found a few that had mismatching D
upper cap values.

The new values come from:
http://wiki.ffxiclopedia.org/wiki/Calculating_Blue_Magic_Damage

### DIFF
--- a/scripts/globals/spells/bluemagic/claw_cyclone.lua
+++ b/scripts/globals/spells/bluemagic/claw_cyclone.lua
@@ -32,7 +32,7 @@ function onSpellCast(caster,target,spell)
         params.tp150 = 1.4375
         params.tp300 = 1.4375
         params.azuretp = 1.4375
-        params.duppercap = 9
+        params.duppercap = 23
         params.str_wsc = 0.0
         params.dex_wsc = 0.3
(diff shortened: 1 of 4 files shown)


TeoTwawki:
I'm glad someone is finally taking a look at blu spell values, but bear in mind that the ffxiclopedia page was last updated in 2010 and any changes SE did after that will not be reflected - whenever possible look for a more recent source than 'clopedia as most of its pages are woefully out of date.

BG wikwi's page was last updated in 2018 https://www.bg-wiki.com/bg/Calculating_Blue_Magic_Damage

Though that doesn't necessarily mean its more or less accurate. Hopefully the 2 agree on most things. Blue magic has been in dire need of retail accuracy adjustments for ages.

mrhappyasthma:
Comparing the two docs, they seem to be in agreement for the D values. But that doesn't necessarily mean that those accurately reflect the retail environment.

Unfortunately I don't have a retail subscription anymore.

TeoTwawki:
yeah I just wanted to make sure you knew - this applies to most things because neither wiki is complete or fully up to current retail info.

PR #6366, opened 2020-01-12
Add a simple flow to allow changing user passwords.

mrhappyasthma:
This is a prerequisite for
https://github.com/DarkstarProject/xiloader/issues/7.

This change adds support for a new command in the login code to support changing a users password. It simply verifies the users login similarly to the existing flow, and then makes another request over the socket for the new password. Then it updates the SQL.

Tested locally. Have a set of changes ready for xiloader as well, but this work is required first.

NOTE: There is definitely some refactoring that could be done in this file to
reduce the amount of nested braces and to reduce duplicated code. I tried
to touch as little existing code as possible in this change to reduce
the risk of introducing a regression.

### DIFF
--- a/src/login/login_auth.cpp
+++ b/src/login/login_auth.cpp
@@ -102,6 +102,9 @@ int32 login_parse(int32 fd)
             return -1;
         }
 
+        Sql_EscapeString(SqlHandle, escaped_name, name.c_str());
+        Sql_EscapeString(SqlHandle, escaped_pass, password.c_str());
+
         switch (code)
         {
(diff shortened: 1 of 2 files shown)


mrhappyasthma:
Let me know what folks think of this change.

mrhappyasthma:
Here's how things look on the xiloader side: https://github.com/mrhappyasthma/xiloader/commit/93077b8e426006a73d0828085642808a7917a690

TeoTwawki:
I'll get with you in a bit so we can launch the changes to the loader at the same time

PR #6370, opened 2020-01-12
Rename email fields in accounts.sql for clarity

### LINKED ISSUE #6369: Rename "email2" to something more descriptive in `accounts` SQL table
I think this would be a nice cleanup request. Currently there are two emails fields: `email` and `email2`. But it's not clear what should go in either one.

We could leave `email` as is, to indicate the current email. (Or we could rename to `current_email`.)

But we should consider renaming `email2`. Perhaps `registration_email` to make it clear that this was the from registration and could be used for account recovery purposes.

mrhappyasthma:
We plan to update xiloader to support adding emails. Currently there are two email fields, but it's not clear what gets stored in which.

Update `email` to `current_email`. This can store the user's latest email, in case it changed. Or an empty string if they never provided an email.

Update `email2` to `registration_email`. This should store the email the user used during the account creation process (if any). And it should not change. This could be used for account recovery purposes.

NOTE: If you have an existing `accounts` table, you shouldn't re-run this command as it will drop the existing table. Instead you can update your table by using the following commands in MySQL:

```
ALTER TABLE accounts CHANGE email current_email varchar(64);
ALTER TABLE accounts CHANGE email2 registration_email varchar(64);
```

fixes #6369


nfaletra:
In the future, stuff like this needs to be added to the migration script. Otherwise, existing servers have to parse through commits to figure out what changed.

TeoTwawki:
are we going forward on that for all future trivial sql changes as well? who decided that?

TeoTwawki:
I mean, coz we haven't been. people have always been told to watch the sql changes when they pull. We've only mad emigrations for things like switching something to/from a blob that was something else.

nfaletra:
The migration script is meant for any DB structural change. Whoever merged the initial scripts decided it was a good idea. We should always seek to improve the process. Some may see this as a “trivial” change but not all server operators are familiar with writing their own SQL.

The question now becomes, do we support making these kinds of transitions as easy as possible for server ops or not? There’s a lot of push for custom settings. Do we support those but allow db structural changes without an easy way to update existing tables?

TeoTwawki:
see edit

TeoTwawki:
I have this lousy tendency to magically hit the right hotkeys to jump half of what I am saying into a new post instead of the line break I intended for readability.

mrhappyasthma:
Sorry for any confusion. Still pretty new to the project.

Is there a link to the migration script that should be appended to for future pull requeststo SQL files?

Or do you mean including a custom script with each change as well? If so, what file path should it go under?

TeoTwawki:
we have a migrations folder that we've been adding new scripts to for major schema changes and dear lord why can't I type right I'm managing to add wrong words to spell check and even open new tabs as I reply

mrhappyasthma:
Ah, I see. You mean the scripts here: https://github.com/DarkstarProject/darkstar/tree/master/migrations

Got it! Will keep this in mind.

PR #6371, opened 2020-01-13
Exp Fixes

nfaletra:
I missed a bunch of stuff in core when I updated the exp table. Also, the exp table still needed a little work.

- Added EMobDifficulty enum and charutils::CheckMob to eliminate all the places that did separate exp checks. All places now check against the difficulty and only needs editing in a single place.
- Updated Charm Chance code to make more sense. It was calculating the chance of failure and then inverting it.
- Added new rows to exp_table and updated the code that loads them in. I still think this data/code needs work to be truly retail accurate, but with the lack of existing data and the amount of work that would go into collecting new data, this is a fine stop gap.
- Also did some basic code cleanup while I was in the area.


nfaletra:
I did do a fair amount of charm testing since I changed that code around. Charm chances feel pretty good despite the crazy amount of guesstimating we’re doing.

Issue #2900, opened 2016-03-12
 /nominate command


TeoTwawki:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
non relevant

**_Server Version_** (type `@revision` in game) **:**
d44a781f97f467cfaa021ad0d457e4cf0767f259

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
I assume the reason `/nominate` does not work is because its such an under used feature of retail that nobody thought to capture and implement that.

mrhappyasthma:
This could be a fun starter bug. Will look into this.

mrhappyasthma:
The current behavior for this command seems a bit wonky. Also I couldn't easily test casting a vote, since it's limited on trial commands.

The lines of text feel reversed to me, but here's what I got so far:

- Proposing a question: https://i.imgur.com/zqho9Od.jpg- Receiving a vote, updates tally for you (not sure what the vote caster sees): https://i.imgur.com/kK91I3A.jpg

- Calling `propose` again ends the tally: https://i.imgur.com/zb0O4ZX.jpg

Some nuances:

- Zoning ends the vote but doesn't seem to show the final tally.
- /vote is documented on FFXIclopediate to take two arguments <name> <vote#>. I found a volunteer to vote for my question and they said that format didn't work. Only `/vote <vote#>` without the username. Its possible a typo by them, since I cant see their screen.
- Max of 8 options allowed on the proposal. If you have more, it just truncates and ignores the 9th+ elements. No error is shown or anything.

mrhappyasthma:
Commands seem to be:

Propose = `0x0A0`
Vote = `0x0A1`

https://github.com/DarkstarProject/darkstar/blob/7661e2f74218fb5baa1e531894f0bd2f7960b413/src/map/packet_system.cpp#L6112-L6113

mrhappyasthma:
Strings seem to be here: https://github.com/DarkstarProject/darkstar/blob/a142cd9fe6e85bf8d66f8354e5ed13a7c69685f5/documentation/MessageSystemIDs.log#L141-L155

[...]

PR #6374, opened 2020-01-13
Clean up message_standard.h and add /nominate IDs.

mrhappyasthma:
This CL cleans up the file a bit (removed stale comments, aligned everything, etc).

It also adds the message IDs for the /propose or /nominate command. This will be needed for https://github.com/DarkstarProject/darkstar/issues/2900

### DIFF
--- a/src/map/packets/message_standard.h
+++ b/src/map/packets/message_standard.h
@@ -31,49 +31,69 @@
 /*
  * Valid MessageIDs
  * Complete list at 1/27/76.DAT
+ *
+ * NOTE: System messages do not specify the color in the packet explictly. Instead there is a different message ID
+ * for each of the supported chat colors (system (yellow), say (white), shout (red), party (blue), linkshell (green)).
  */
 enum class MsgStd
 {
-    InvitationDeclined      = 11, /* Your invitation was declined */
-    CannotBeProcessed       = 14, /* That request cannot be processed */
-    NotPartyLeader          = 21, /* You are not the party leader. */
-    CannotInvite            = 23, /* You cannot invite that person at this time. */
-    WaitLonger              = 38, /* You must wait longer to perform that action. */
+    // Used as a sentinal value. This should not be used as part of a packet.
+    Unknown = -1,
 
-    DiceRoll                = 88, /* Dice Roll! <player> rolls <roll>. */
-    Examine                 = 89, /* <name> examines you. */
-    /* Linkshells */
-    LinkshellUnavailable    = 112, /* The linkshell name you entered is already in use or otherwise unavailable. */
-    TellNotReceivedOffline  = 125, /* Your tell was not received.  The recipient is currently away. */
-    ThrowAway               = 180, /* You throw away <item>. */
-    TellNotReceivedAway     = 181, /* Your tell was not received.  The recipient is currently away. */
-    ItemEx                  = 220, /* You cannot possess more than one of that item. */
-    BlockingAid             = 225, /* Target is currently blocking outside magical assistance */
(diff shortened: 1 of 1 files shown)


src/map/packets/message_standard.h
```diff
@@ -31,49 +31,69 @@
 /*
  * Valid MessageIDs
  * Complete list at 1/27/76.DAT
+ *
+ * NOTE: System messages do not specify the color in the packet explictly. Instead there is a different message ID
+ * for each of the supported chat colors (system (yellow), say (white), shout (red), party (blue), linkshell (green)).
```
TeoTwawki: Technically the color can be whatever I specify in my client settings. Its more accurate to say the types each have different default colors and that said coloring is a client side thing.

src/map/packets/message_standard.h
```diff
@@ -31,49 +31,69 @@
...
-    TrustCannotLFP          = 296, /* You cannot use Trust magic while seeking a party. */
-    WaitParty               = 297, /* While inviting a party member, you must wait a while before using Trust magic. */
-    TrustLimit              = 298, /* You have called forth your maximum number of alter egos. */
-    TrustSame               = 299, /* That alter ego has already been called forth. */
-    TrustEnmity             = 300, /* You cannot use Trust magic while having gained enmity. */
-    TrustSoloOrLeader       = 301  /* You cannot use Trust magic unless you are solo or the party leader. */
+    // Used as a sentinal value. This should not be used as part of a packet.
+    Unknown = -1,
+
+    InvitationDeclined             = 11, /* Your invitation was declined */
```
TeoTwawki: I realize a past author did this, but this seems a good chance to clean up that comment style.

```cpp
// This is a single line comment
```
```cpp
/* 
this is a
multi-line or "block"
comment
*/
```

mrhappyasthma: Is this the preferred style?

I tend to like doxygen style comments since they can also be parsed by many tools. But I'm happy to do whatever is consistent with the project.

--

Also is there a preferred naming convention? It seems to be a mix of `camelCase` , `CamelCase`, `snake_case`, and `everythinglowercase`.

I tend to try to apply Google's C++ style whenever possible, since it's documented and widely accepted. But the code style in DSP seems a bit inconsistent across functions and files.

TeoTwawki: style guide has some of it. other parts need new style guide entries. We defo don't follow google conventions here. We have different function styles to indicate different things.

UPPER_SNAKE_CASE for ENUM
UpperCamelCase for namespaced functions and classes
lowerCamelCase for pure lua functions (not in cpp at all)

If you want to pitch using a tool that requires those, talk to the team on discord and if we're on board with it, we can go the other route instead and convert them all to allow for that tool.Right now though most files are using a lot of non-block comments

TeoTwawki: If you want to pitch using a tool that requires those block comments, talk to the team on discord and if we're on board with it, we can go the other route instead and convert them all to allow for that tool.Right now though most files are using a lot of non-block comments

mrhappyasthma: I'm not too concerned about using a tool. Some IDEs just pick up doxygen style comments by default.

However I'm happy to just match the comment style you mentioned. Will update.

src/map/packets/message_standard.h
mrhappyasthma: Reworded

src/map/packets/message_standard.h
mrhappyasthma: Updated to match the recommended style.

mrhappyasthma:
Please take another look. Thanks!

PR #6375, opened 2020-01-13
GetCurrentContinent: Check if zone is middle lands ship

pyromaniackeca:
- Not sure if any zones are still missing
- Not sure if any added zones should be omitted

P.S.
Any comments on approach and style will be greatly appreciated as I haven't written any serious cpp since university.

### DIFF
--- a/src/map/utils/zoneutils.cpp
+++ b/src/map/utils/zoneutils.cpp
@@ -962,8 +962,34 @@ REGIONTYPE GetCurrentRegion(uint16 ZoneID)
 *                                                                       *
 ************************************************************************/
 
+bool IsMiddleLandsShip(uint16 ZoneID) {
+    switch (ZoneID)
+    {
+        case ZONE_SHIP_BOUND_FOR_SELBINA:
+        case ZONE_SHIP_BOUND_FOR_MHAURA:
+        case ZONE_SHIP_BOUND_FOR_SELBINA_PIRATES:
+        case ZONE_SHIP_BOUND_FOR_MHAURA_PIRATES:
+        case ZONE_SAN_DORIA_JEUNO_AIRSHIP:
+        case ZONE_BASTOK_JEUNO_AIRSHIP:
+        case ZONE_WINDURST_JEUNO_AIRSHIP:
+        case ZONE_KAZHAM_JEUNO_AIRSHIP:
+            return true;
+    }
+    return false;
+}
+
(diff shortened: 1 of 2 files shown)


TeoTwawki:
the only reason they aren't considered part of the continent is they don't have a region so the switch hits "REGION_UNKNOWN" which is the default. Anything that is that region is considered not middle lands. from the look of it currently all zones with a region are being counted as middle lands, which is wrong.. https://github.com/DarkstarProject/darkstar/blob/2c2d4f737bb4534e8366741e4d74c0d67c99bcf4/src/map/utils/zoneutils.cpp#L956

```cpp
        case ZONE_CEIZAK_BATTLEGROUNDS:
        case ZONE_FORET_DE_HENNETIEL:
        case ZONE_SIH_GATES:
        case ZONE_MOH_GATES:
        case ZONE_CIRDAS_CAVERNS:
        case ZONE_CIRDAS_CAVERNS_U:
        case ZONE_YAHSE_HUNTING_GROUNDS:
        case ZONE_MORIMAR_BASALT_FIELDS:
            return REGION_EAST_ULBUKA;
    }
    return REGION_UNKNOWN;
```
```cpp
CONTINENTTYPE GetCurrentContinent(uint16 ZoneID)
{
    return GetCurrentRegion(ZoneID) != REGION_UNKNOWN ? THE_MIDDLE_LANDS : OTHER_AREAS;
}
```

TeoTwawki:
I think we should fix our region debacle and then give these a region that counts as middle lands, instead of making a new function to bypass that system

zynjec:
Before we go any further, could we have some proof that they're subject to signet effects on retail?

nfaletra:
Confirmed signet effects work on ferry ships on retail.

pyromaniackeca:
The PR was opened as a response to this: https://github.com/EdenServer/community/issues/2408

I am still unfamiliar with the bigger context of this change, so am open to suggestions on how to improve this or any discussions I could read about the region debacle @TeoTwawki mentioned to get a better understanding of the issue.

TeoTwawki:
The reason these aren’t counting is the function that checks if its “the middle land” doesn't do what its name implied - its just checking the region is “unknown” or not and since these zones were never assigned a region and the defaultis unknown, they are and fail that check. Additionally a ton of zones that are NOT in middle lands are counted as if they are.

Better?

PR #6377, opened 2020-01-14
Fix an issue with null player being passed to death listener

cocosolos:
Also added a check in mixins that use the player.


kaincenteno:
awesome! does this fix the issue with caskets?

cocosolos:
Soft of. It fixes the crashing but I didn't touch the caskets mixin because of the open PR so it will still give a lua error in certain circumstances.

PR #6380, opened 2020-01-15
Installer script

chad-ermacora:
Interactive install script for darkstar on Debian, Ubuntu & Raspbian.  Only need to enter a username, password and IP to get up and running.  The user is used to start the servers and access the SQL database.

### DIFF
--- a/darkstar_install_script.sh
+++ b/darkstar_install_script.sh
@@ -48,16 +48,16 @@ if [ "${RUN_CHOICE}" == "1" ]; then
     git clone --recursive https://github.com/DarkstarProject/darkstar/ /opt/darkstar
   fi
   cd /opt/darkstar || exit
-  sh autogen.sh
+  sh autogen.sh || exit
   if [ "${OS}" == "debian" ] || [ "${OS}" == "raspbian" ]; then
     printf "\nRunning Debian or Raspbian Install\n"
     apt-get -y install libmariadb-dev-compat
-    ./configure CXXFLAGS=" -pthread"
+    ./configure CXXFLAGS=" -pthread" || exit
   else
     printf "\nRunning Default Install\n"
-    ./configure
+    ./configure || exit
   fi
-  make -j4
+  make -j4 || exit
   chown -R "${DARKSTAR_USER}":root /opt/darkstar/
   printf "\nEnter the MySQL root password\nDefault: No password\n"
   mysql -h "localhost" -u "root" -p -e "CREATE USER '${DARKSTAR_USER}'@'localhost' IDENTIFIED BY '${DARKSTAR_PASSWORD}';CREATE DATABASE dspdb;USE dspdb;GRANT ALL PRIVILEGES ON dspdb.* TO '${DARKSTAR_USER}'@'localhost';"
@@ -157,7 +157,7 @@ if [ "${RUN_CHOICE}" == "3" ] || [ "${RUN_CHOICE}" == "2" ]; then
   systemctl restart darkstar-game
   systemctl restart darkstar-search
 elif [ "${RUN_CHOICE}" == "4" ]; then
-  printf "\nShutting Down & Disabling Darkstar Server\n\n"
+  printf "\nShutting Down Darkstar Server & Disabling Auto Start\n\n"
   systemctl disable darkstar-connect
   systemctl disable darkstar-game
   systemctl disable darkstar-search
@@ -165,7 +165,7 @@ elif [ "${RUN_CHOICE}" == "4" ]; then
   systemctl stop darkstar-game
   systemctl stop darkstar-search
 elif [ "${RUN_CHOICE}" == "5" ] || [ "${RUN_CHOICE}" == "1" ]; then
-  printf "\nStarting & Enabling Darkstar Server\n\n"
+  printf "\nStarting Darkstar Server & Enabling Auto Start\n\n"
   systemctl enable darkstar-connect
   systemctl enable darkstar-game
   systemctl enable darkstar-search
--- a/darkstar_install_script.sh
+++ b/darkstar_install_script.sh
@@ -1,0 +1,178 @@
+#!/bin/bash
+# Make sure its running with root
+if [[ $EUID != 0 ]]; then
+  printf "Root permissions required, starting with sudo\n\n"
+  chmod +x "$0"
+  sudo "$0" "$@"
+  exit $?
+fi
+# Static script Options
+APT_GET_INSTALL_ONE="nano git mariadb-client mariadb-server libmariadb-dev libmariadbclient-dev libssl-dev"
+APT_GET_INSTALL_TWO="luajit lua5.1 liblua5.1-dev libzmq3-dev autoconf pkg-config gcc build-essential"
+INSTALL_DIR="/opt/darkstar"
+# User set options (These are replaced by user input later in the script)
+DARKSTAR_USER="darkstar"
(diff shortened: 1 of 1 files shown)


darkstar_install_script.sh
```diff
@@ -0,0 +1,178 @@
...
+  # shellcheck disable=SC2086
+  apt-get update && apt-get -y install ${APT_GET_INSTALL_ONE} ${APT_GET_INSTALL_TWO}
+  adduser "${DARKSTAR_USER}" --gecos "Darkstar Server" --disabled-password
+  if [[ ${GIT_VERSION} == "1" ]]; then
+    git clone --branch stable --recursive https://github.com/DarkstarProject/darkstar/ /opt/darkstar
+  else
+    git clone --recursive https://github.com/DarkstarProject/darkstar/ /opt/darkstar
+  fi
+  cd /opt/darkstar || exit
+  sh autogen.sh
```
whasf: No check for if this is successful or not?

darkstar_install_script.sh
```diff
@@ -0,0 +1,178 @@
...
+  if [ "${OS}" == "debian" ] || [ "${OS}" == "raspbian" ]; then
+    printf "\nRunning Debian or Raspbian Install\n"
+    apt-get -y install libmariadb-dev-compat
+    ./configure CXXFLAGS=" -pthread"
```
whasf: Check if it completes successfully?

darkstar_install_script.sh
```diff
@@ -0,0 +1,178 @@
...
+  else
+    printf "\nRunning Default Install\n"
+    ./configure
```
whasf: Check if it completes successfully?

darkstar_install_script.sh
```diff
@@ -0,0 +1,178 @@
...
+  fi
+  make -j4
```
whasf: Check if it completes successfully?

darkstar_install_script.sh
```diff
@@ -0,0 +1,178 @@
...
+  chown -R "${DARKSTAR_USER}":root /opt/darkstar/
+  printf "\nEnter the MySQL root password\nDefault: No password\n"
+  mysql -h "localhost" -u "root" -p -e "CREATE USER '${DARKSTAR_USER}'@'localhost' IDENTIFIED BY '${DARKSTAR_PASSWORD}';CREATE DATABASE dspdb;USE dspdb;GRANT ALL PRIVILEGES ON dspdb.* TO '${DARKSTAR_USER}'@'localhost';"
```
whasf: Is the service running? (Does the install automatically start it?)

darkstar_install_script.sh
chad-ermacora: for the apt-get install or the ./configure?

darkstar_install_script.sh
chad-ermacora: If you mean the 3 darkstar services (connect, game, search)they are automatically enabled & started at the end of the script.  The SQL database is also up and running right after it's installed with apt-get above.

darkstar_install_script.sh
chad-ermacora: Do the ./configure, make and autogen.sh return a none 0 value when they fail? I should be able to add a `|| exit` if they do.

darkstar_install_script.sh
chad-ermacora: Now exits on failure.

darkstar_install_script.sh
chad-ermacora: Now exits on failure.

darkstar_install_script.sh
chad-ermacora: I was able to test it, and it does indeed exit on fail with `|| exit` and continues on success. 
Now set to exit on failure.

darkstar_install_script.sh
chad-ermacora: Now exits on failure.

darkstar_install_script.sh
chad-ermacora: I have clarified the descriptive text for services, explaining how it set's the server to auto-start or disables auto-start.

darkstar_install_script.sh
```diff
@@ -0,0 +1,178 @@
...
+  echo
+  if [[ ${ZONEIP} =~ ^[Yy]$ ]]; then
+    printf "\nYour current systems IPv4 Address are as follows\n"
+    ip -4 addr | grep -oP '(?<=inet\s)\d+(\.\d+){3}'
+    echo
+    read -r -p "Enter the IP address you wish to use with this Server: " DARKSTAR_IP
+    printf "\nEnter the MySQL root password\nDefault: No password\n"
+    mysql -h "localhost" -u "root" -p -e "USE dspdb;UPDATE zone_settings SET zoneip = '${DARKSTAR_IP}';"
+  fi
+  nano ${INSTALL_DIR}/conf/login_darkstar.conf
```
ShelbyZ: Is the purpose of opening these files to allow the users to make changes?

ShelbyZ:
There are some potential problems with what this script will offer:

- We are not providing the ability for a user that already has this installed to update branches or code via a 'friendly' option
- How do we plan on handling dependency changes
- How do we plan on handling SQL changes

Feels like there should be an option for updating the server from GIT and figuring out how those changes would be applied.  Code would be easy, SQL would require usage of migration scripts.

darkstar_install_script.sh
chad-ermacora: That is correct.  You have to set the newly created user and password for the SQL database access.

chad-ermacora:
> There are some potential problems with what this script will offer:
> 
> * We are not providing the ability for a user that already has this installed to update branches or code via a 'friendly' option
> * How do we plan on handling dependency changes
> * How do we plan on handling SQL changes
> 
> Feels like there should be an option for updating the server from GIT and figuring out how those changes would be applied. Code would be easy, SQL would require usage of migration scripts.

I was told to not allow upgrades, due to SQL change issues that are currently not automated. So if install is chosen but it detects it has been built already, it exits the script.  Configurations, such as Zone IP and the 3 main configuration files can still be accessed after setup by running the script and choosing edit config option. 

Dependency changes will have to be updated in the script by adding or removing it in one of the 2 APT_GET_INSTALL variables near the top. If they are Distro specific, they can be added in the section shown below. Adding other distro's should be relatively easy, as you add in an extra else if statement checking for the lowercase name of the distro.  

`if [ "${OS}" == "debian" ] || [ "${OS}" == "raspbian" ]; then
    printf "\nRunning Debian or Raspbian Install\n"
    apt-get -y install libmariadb-dev-compat
    ./configure CXXFLAGS=" -pthread" || exit
  else
    printf "\nRunning Default Install\n"
    ./configure || exit
  fi`

If we did have a SQL script in the main repository that would update SQL entries based on whatever has changed, I could link directly to it in the script and download it at runtime.  I was thinking as long as the SQL script is updated before pushing to Stable, upgrading between stable releases should be a viable option.

TeoTwawki:
as long as they can git pull after they can get code changes and then rebuild for themselves. 

we can't really get a good way to automate sql changes at this time.

darkstar_install_script.sh
chad-ermacora: I have now added text to explain why said files will be opened and how to use the text editor.

chad-ermacora:
-Added fix to work on Ubuntu desktop (I was testing on server, which enables required repository by default but desktop does not)
-Misc text output improvements.

PR #6383, opened 2020-01-16
suggested addition regarding migrations

TeoTwawki:
Proposal to make things simpler for people updating after we've made or merged schema changes

@DarkstarProject/core 

https://github.com/DarkstarProject/darkstar/pull/6370#issuecomment-574982499

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -116,3 +116,8 @@ Clang-Format is also an option for C++
   ```
   insert into table_name
   ```
+  
+  #### SQL Migrations for Schema changes
+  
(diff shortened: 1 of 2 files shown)


TeoTwawki:
suggestions on more specifically what is wanted instead of what I wrote are welcome. If we want to go pure python only its gonna get annoying on the simpler things, and we should also phase out python 2.7 its not being updated anymore (he's dead Jim) in favor of 3.x (latest release is 3.8.1)

zynjec:
sike, not getting added

TeoTwawki:
> why not stick to one python version?

That was one the points I was making. We should use one - the one that's still getting fixes and not the old discontinued one.

We should also have a mention in the contrib guide for what we actually want people to do. I talked to Whasf about this but never never found time to comment here. I invited the entire team to share their thoughts about what they actually want in the contrib guide. This and deletion are the only response I've gotten.

PR #6388, opened 2020-01-19
 Addition of White Coney/Black Coney/Rabbit Footprint (By Caelic)

Xaver-DaRed:
- Addition of 2 mobs, the trigger NPC and their spawn mechanics.
- Corrected issues from original PR.
- Tabs converted to 4 spaces.
Credit goes to Caelic. Permission was asked for and granted.
Original PR #6325


scripts/zones/Uleguerand_Range/npcs/Rabbit_Footprint.lua
```diff
@@ -0,0 +1,105 @@
...
+function onEventUpdate(player,csid,option)
+end
+
+function onEventFinish(player,csid,option)
+end
+
+-- https://ffxiclopedia.fandom.com/wiki/Black_Coney
+-- The Rabbit Footprint is located at (K-11) in the wooded area, and moves around every 7-10 seconds between various spots in that locale.
+-- It will spawn 9-15 minutes after the New Moon phase starts and will last until the end of the phase.
+-- NOTE: i think 7-10 seconds is WAY too fast, through testing on a GM character with 255 speed i can't even get a carrot into the trade menu before it moves so i used a much higher number
```
TeoTwawki: having done this on retail, that is how it normally is. I always had to make a macro to trade the carrot because its to fast.

```
/item "item name" <t>
```

Xaver-DaRed: Want me to lower the respawn time of the NPC to match wiki?

TeoTwawki: yeah

TeoTwawki:
> Made it less player friendly

We not setting out to just hamper the player.  Retail accuracy is default before any settings confs or custom adjustments. And by the way, the page for the opposite coney says an even shorter time. One page says 5 seconds the other 7. _so possibly its even shorter than what I just set._

Xaver-DaRed:
> We not setting out to just hamper the player._

I know, I was just joking.^^

Caelic:
Thanks for doing this Xaver, been pretty busy lately.

Caelic:
Something to note since the time interval is way lower now, the server is spamming ```npc:setLocalVar("activeTime", 0)
            npc:setLocalVar("currentPoint", 0)
            npc:setStatus(dsp.status.DISAPPEAR)```
every 3 vana-minutes instead of every 16 vana-minutes so at some point as a to-do may wanna have two different time intervals, one for checking if it's time to shift into moon-phase mode and one that actually moves the footprint just to lower spam/server lag.

Xaver-DaRed:
I wonder if some short of function similar to
onZoneWeatherChange
exists, but for moonphases.

PR #6391, opened 2020-01-21
Docking Message when traveling to Nashmau

kaincenteno:
Added ID for message (as it's not part of CS)
When loading into the boat, it will store the Vanadiel hour and minute.
It will check the time it should be arriving depending on the boat schedule.
It will subtract the time the boat will dock and the time the char zoned in then use that as
a timer so the text shows to the player on the correct time.
Also removed semicolons


kaincenteno:
Hi darkstar friends 🐱 

Let me know if there's something i should change to this message or if everything is okay so I can make the changes to all the boats and submit a new PR 😸

PR #6392, opened 2020-01-21
Update synthutils.cpp

Xaver-DaRed:
Initial work on crafting.
- Corrected being able to HQ without all skills over the needed value.
- Corrected unfair handicap on Lightning crystals.
- Very profesinal comments added.
- Even more profesional google translate performed on russian comments

THIS IS UNTESTED. For several reasons, I can't setup a DSP server at the time, so I need some1 to test this for me.

Crafting still has other issues. For now this just attempts to correct the HQ problem and make some crafters angry.

### DIFF
--- a/src/map/utils/synthutils.cpp
+++ b/src/map/utils/synthutils.cpp
@@ -1,4 +1,4 @@
-/*
+﻿/*
 ===========================================================================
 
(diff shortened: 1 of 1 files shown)


zynjec:
> THIS IS UNTESTED
Please don't do this, everything should be tested before PRing
It doesn't even compile

PR #6393, opened 2020-01-21
Fix a whitespace

Xaver-DaRed:
Fix a whitespace caused by me that was preventing the file to be run by HeidiSQL

### DIFF
--- a/sql/mob_skill_lists.sql
+++ b/sql/mob_skill_lists.sql
@@ -1,4 +1,4 @@
-﻿-- phpMyAdmin SQL Dump
+-- phpMyAdmin SQL Dump
 -- version 3.3.8
 -- http://www.phpmyadmin.net
 --


TeoTwawki:
I see nothing here that would have prevented importing the table via heidi or any other program

Xaver-DaRed:
I know, but HeidiSQL was returning an error in Line 1.
Editing the file through the browser actually showed a "dot" character at the very beggining of the file.

zynjec:
It's not invisible characters, it's HeidiSQL's terrible parsing engine and that file being saved as UTF8 with BOM.

Another tally mark for "GUIs are shit".

TeoTwawki:
na it is on githubs side at least. I copied contents local and there is a non standard line break right at the top that github can't display but my editor can. githubs all "we know there's some kinda change so we gonna mark it, but we're still gonna show same line" - that's why I called it invisible.
image

TeoTwawki:
byte order marks aren't _usually_ a problem either.

Issue #5810, opened 2019-03-26
blu spell pincone bomb has no spell script may be others like it


ziaxe427:
**_I have:_**

- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 
version 30190305_8

**_Source Branch_** (master/stable) **:** 
dont know what this is first time posting an issue lol

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
i did this on the dsp oldschool server where i am currently lvling blu wheneveri go to cast pinecone bomb at an enemy it cause ziaxe cannot cast or something like that.

also on a side note i cant confirm if this is a bug or an issue or not but sometimes blu spells do 0 dmg for no reason on  eenemys of any lvl and other times will do dmg. i cant confirm if this is an issue or an intended fature lol im new to this game so yea lol.

whasf:
A missing script isn't a bug, it's just not implemented.

zynjec:
Just because it's not implemented doesn't mean there can't be an issue, it's not exclusively for bugs.

TeoTwawki:
if I recall correctly nobody has tackled the prob of it trying to apply its damage after it applies the sleep effect without giving up partway into the attempt yet.

helixhamin:
Can you just apply the damage then apply sleep after to get around that issue?

TeoTwawki:
The issue of effects getting removed by the damage exists precisely because you don’t have that ability in scripts - the damage happens at the end. We need to refactor how these work to let you do that.

mrhappyasthma:
Are there any code pointers for this? I'd be interested in looking into this as a starter task.

Seems like this is mostly just a refactor and then implementing the script in `darkstar/scripts/globals/spells/bluemagic/`.

But I'm still quite unfamiliar with the code.

mrhappyasthma:
Just by scanning through the code, seems like the damage is handed by calls to `BlueFinalAdjustments` here:
https://github.com/DarkstarProject/darkstar/blob/58539ad5a2c612b0041135b2349d99824a51e1b6/scripts/globals/bluemagic.lua#L247

Howeverany sleep status effect would be removed after that Lua script has executed, via this line:
https://github.com/DarkstarProject/darkstar/blob/2effc1ae16220a1a29faa7e489da7f37e91a12ee/src/map/entities/battleentity.cpp#L1304

[...]

PR #6396, opened 2020-01-22
Refactor the order of operations for how damage is applied from spells.

mrhappyasthma:
This primarily refactors the order of operations such that any magic
spell that deals damage now handles the effects from the damage
immediately after.

This is necessary as a prerequisite for implementing spells like
Pinecone Bomb, which reapply a sleep effect after the damage has been
applied.

The prior implementation would apply the damage and then apply the sleep
all in the Lua code before returning back to
CBattleEntity::OnCastFinished(), which would then remove the sleep since
damage was taken.

After this refactor the damage and debuff behavior from the damage all
occur on the Lua side. That way Lua scripts for spells can reapply
sleeps and have it function properly.

This was modeled after how TakeWeaponskillDamage() was implemented to
achieve a similar order of operations.

NOTE: I tested manually by sleeping mobs and applying the various types of
spells to verify that sleep is canceled correctly. Also tested a pinecone_bomb
script (not included in this PR) to verify that re-applying sleep would work.

I also audited every Lua script that includes a spell that deals damage to
cover all possible code paths.

### DIFF
--- a/scripts/globals/bluemagic.lua
+++ b/scripts/globals/bluemagic.lua
@@ -216,12 +216,12 @@ function BlueMagicalSpell(caster, target, spell, params, statMod)
     local magicAttack = 1.0;
     local multTargetReduction = 1.0; -- TODO: Make this dynamically change, temp static till implemented.
     magicAttack = math.floor(D * multTargetReduction);
-    
+
     local rparams = {};
     rparams.diff = dStat;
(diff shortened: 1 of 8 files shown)


mrhappyasthma:
This is a refactor necessary to address https://github.com/DarkstarProject/darkstar/issues/5810. But it does not close the issue.

PR #6402, opened 2020-01-27
Added Omens quest (BLU AF2)

kaincenteno:
Changed BEGINNINGS to check job to be level 40
Position of flans not accurate to retail as there's no captures for a full alliance fighting BCNM


scripts/zones/Navukgo_Execution_Chamber/bcnms/omens.lua
```diff
@@ -0,0 +1,45 @@
...
+function onBattlefieldTick(battlefield, tick)
+    dsp.battlefield.onBattlefieldTick(battlefield, tick)
+end
+
+function onBattlefieldRegister(player, battlefield)
+end
+
+function onBattlefieldEnter(player, battlefield)
+    local players = battlefield:getPlayers()
+    if #players ==  4 then GetMobByID(17039375):spawn(); end
```
TeoTwawki: my total is 5. now what? where is your Altana now, Kain.

kaincenteno: Praise Altana in all Her might and glory. 🙏 

As the allied forces of Altana enter the battlefield, the script will make a check once each of them has enter the battlefield popping the enemies. In this case if five holy warriors join the battle on Her name, then 2 mobs will be in the area.

Now follow the sacred words sung on Altana's Hymn and hit that "merge pull request."

kaincenteno:
@TeoTwawki  dont merge yet, i need to remove the master pudding, wasn't aware that 'Master...' monsters were not really mobs but used by retail to clone mobs

scripts/zones/Navukgo_Execution_Chamber/bcnms/omens.lua
```diff
@@ -0,0 +1,46 @@
...
+end
+
+function onBattlefieldEnter(player, battlefield)
+    local mobOffset = (battlefield:getArea() - 1) * 7  -- offset to spawn correct mob depending on battlefieldNumber
+    local players = battlefield:getPlayers()
+    if #players ==  4 then GetMobByID(17039375 + mobOffset):spawn(); end
+    if #players ==  8 then GetMobByID(17039376 + mobOffset):spawn(); end
+    if #players == 12 then GetMobByID(17039377 + mobOffset):spawn(); end
+    if #players == 14 then GetMobByID(17039378 + mobOffset):spawn(); end
+    if #players == 16 then GetMobByID(17039379 + mobOffset):spawn(); end
```
zynjec: Entity ID's should not be in anything other than `ID.lua`.

Also, I feel like the conditions you're checking are incorrect the way they are. If you're in a party/alliance of 5, 6, 7, 9, 10, 11, 13, 15 no additional mobs will spawn?

scripts/zones/Navukgo_Execution_Chamber/bcnms/omens.lua
TeoTwawki: gonna need to be made clearer still

scripts/zones/Navukgo_Execution_Chamber/bcnms/omens.lua
```diff
@@ -0,0 +1,46 @@
...
+        player:startEvent(32002)
+    end
+end
+
+function onEventUpdate(player, csid, option)
+end
+
+function onEventFinish(player, csid, option)
+    if csid == 32001 then
+        player:setCharVar("BluAf2CS",2)
```
OmegaBison: Should be checking for `BluAF2CS == 1` before setting it to 2 here.

scripts/zones/Navukgo_Execution_Chamber/bcnms/omens.lua
cocosolos: onBattlefieldEnter gets called once per person as they enter, so the second mob gets spawned when the 4th person enters, nothing happens when player 5, 6, 7 enter, then the third mob spawns when player 8 enters. This would be a problem if it were in onBattlefieldInitialise, which just gets called only once when the battlefield is created.

TeoTwawki: @cocosolos the way its written tho makes for this question to keep popping up every time a new person comes across it. I think the logic could be made to be clearer about whats happening so that future editors don't try to "fix" it.

scripts/zones/Navukgo_Execution_Chamber/bcnms/omens.lua
kaincenteno: nice catch, changed it. Thank you @OmegaBison

scripts/zones/Navukgo_Execution_Chamber/bcnms/omens.lua
kaincenteno: Made changes so it doesn't go through every if statement. Thank you for your input 🐱

PR #6403, opened 2020-01-28
Fix infinite loop failure in charutils::AddItem

nfaletra:
Break out of add item loop if we have an invalid item (like when people add bad item ids like a fucking idiot)

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -3481,6 +3481,7 @@ inline int32 CLuaBaseEntity::addItem(lua_State *L)
             else
             {
                 ShowWarning(CL_YELLOW"charplugin::AddItem: Item <%i> is not found in a database\n" CL_RESET, id);
+                break;
             }
         }
     }


tagban:
This appears to still be causing an issue, I just updated to this and re-built, still getting errors:
[28/Jan] [14:56:59][Warning] charplugin::AddItem: Item <0> is not found in a database

This is happening on a fresh pull, using Steal on Bark Tarantula in the Boyhada tree. It also still crashes when using !additem and accidentally picking an item not valid.

TeoTwawki:
your issue is that shit should not ever have item id zero. the only thing wrong in this code was missing that break to prevent it from repeating the message forever.

tagban:
Okay so the issue is however the bark Tarantula is getting an item ID 0. THis never happened prior to this change, before it would still display _ added successfully, but didn't actually add anything since it didn't actually exist. So something was clearly changed that adjusted that sanity check (or lack there of sanity check)

TeoTwawki:
go find that line in your drop table and comment it out.

tagban:
This change unfortunately makes it so if any ID is called that for some reason doesnt' exist (IE User error, etc.) it will crash the server currently in an infinite loop.

nfaletra:
Missed the second loop. my bad.

nfaletra:
Add a break locally after the warning at ln3559 if you really need it fixed right this second. I don't have time to get to it right now.

tagban:
@nfaletra  Thanks! I had just mentioned that in another discord when I saw this email come thru about this comment that I thought that might be it on 3559. Thanks again!

nfaletra:
Was this in the discord where you're talking shit?

tagban:
Where have I been talking shit? Was on FFXIPrivateServers asking if other owners had seen it, And bouncing some ideas off of people before I posted that it was having issues? Or do you mean my discord where someone came in and bashed my server because of something someone else said?

TeoTwawki:
https://gist.github.com/TeoTwawki/1a452e6eb162816f8c7ffa3b0c750e5d

thats the entire convo, and I was never aware you left Autkast. theres no shit talking in this that is word for word the entire thing typos and all.

Issue #6225, opened 2019-09-22
BLU Magical Damage Spells Treated as attackType.PHYSICAL


ibm2431:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** n/a

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Before I forget to publicly note this (again): Damage-dealing Blue Magic that goes through `BlueFinalAdjustments` is always treated as `attackType.PHYSICAL`, even if the damage type should be Magical.

Some examples of Blue Spells that should be treated as Magical - but aren't due to `BlueFinalAdjustments` - are:

Death Ray
> https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/spells/bluemagic/death_ray.lua#L42

Ice Break
> https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/spells/bluemagic/ice_break.lua#L50

Mind Blast
> https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/spells/bluemagic/mind_blast.lua#L44

When they get into `BlueFinalAdjustments`, after calculating reductions due to Phalanx and Stoneskin, `takeDamage` is always called with `attackType.PHYSICAL`:

> https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/bluemagic.lua#L232-L247

mrhappyasthma:
I can go through and audit/correct these.

I'm doing some refactors to `BlueFinalAdjustments` in https://github.com/DarkstarProject/darkstar/pull/6396/files anyway. 

This could be a good follow up.

mrhappyasthma:
Looks like the same issue applies to Breath spells as well.

mrhappyasthma:
This shouldn't affect too much though, from the looks of it. The `attackType` is only referenced in `takeDamage` to trigger an event. But no code seems to be listening for it.

https://github.com/DarkstarProject/darkstar/blob/b923615c2cbd5543b5baf16e6c5494cefcc3c089/src/map/entities/battleentity.cpp#L522

PR #6411, opened 2020-02-09
Remove unused attack type `breath` enum.

mrhappyasthma:
This enum value is unused.

Breath spells seem to just be classified as  `magical` with the appropriate element.

Removing this should alleviate some ambiguity/confusion when trying to use the appropriate value.

(Found this during an audit of BLU spells for https://github.com/DarkstarProject/darkstar/issues/6225)


TeoTwawki:
might possibly have been done with the intent to use it during implementation of how breaths actually should work (based on current/max hp) and just never got done.

https://www.bg-wiki.com/bg/Damage_Taken#Breath_Damage_Taken
oddly MDT doesn't effect it unless it was from an augment. possibly SE's guy adding augments didn't realize, or SE has an unfixed bug.
https://www.ffxiah.com/forum/topic/52145/finally-releasing-this-addon-gearinfo/8/#3344204

mrhappyasthma:
Would you prefer to leave this value? And, if so, should it be used for BLU breath spells instead of `MAGICAL`?

Whatever we decide I can implement it as part of the clean ups.

TeoTwawki:
Nothing stopping us from re-adding it if its actually needed later I spoze, but I'm not decision making either way here. I don't even know if that is needed to implement it, just thought I'd give the score on breath damage.

mrhappyasthma:
Understood.

I'll leave this PR open then for now.

During my audit I'll continue with the existing convention of annotating breath BLU spells as `MAGICAL`. We can always correct them whenever we add appropriate breath spell handling.

Issue #4240, opened 2017-11-18
claim/enmity issues for charmed pets


Wiggo32:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:** 30170905_5

**_Server Version_** (type `!revision` in game) **:** unknown

**_Source Branch_** (master/stable) **:** master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Just tested out retail BST here are some issues needing addressing for dsp:

1. When pets 'uncharm' they should come completely off the enmity list of the mob which they were attacking. The 'attacked' mob does not retain memory of enmity toward a mob that was once charmed.

2. a BST and his PET cannot both hold claim on their own mob. Only 1 mob can be claimed between the two of them. How I watched this working was: BST attacking a mob while PET attacking a different mob, when the pet hit the mob it turned red and the BST mob turned yellow, when the BST hit, his mob turned red and the PET's mob turned yellow. There was never two mobs red at the same time even for a split second. 

3. PET has its own enmity list and acts off of it according to which mob generates the most enmity. 
 - first scenerio: PET ordered to fight a mob, then ordered to fight a different mob, pet would switch focus between the two mobs depending on who hit the hardest or generated the most enmity basically. and still only 1 mob could be red at a time.
- second scenerio: PET ordered to fight a mob, then was attacked by a link without any other provocation. as linked mob generated enough enmity it caused PET to switch focus from the originally attacked mob to the linked mob. and still only 1 mob could be red at a time.

TeoTwawki:
fixing no.2 may require taking care of the issues with claiming extant elsewhere as well. per issue #2109 right now one player can have multiple claims, which is also wrong.

Issue #3923, opened 2017-06-28
charm bugs out when it lands during raise animation


TeoTwawki:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30170427_0

**_Server Version_** (type `@revision` in game) **:**
4ca87a3dd8b29dda4f8b2a11443d917eff30ed48

**_Source Branch_** (master/stable) **:**
Master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
1. Die
2. Raise or reraise
3. Get hit with a charm move like Gala Macabre (its AoE) before your animation completes
4 Grats, you are now a forever charmed corpse with HP. You will remain on the ground but follow your master (he'll drag the body around) **and charm will never wear off.**

**Additional issue with charm:** 
Other players who were charmed but not stuck between life and death were completely unable to claim the mob after their charm effect wore off.

Issue #3876, opened 2017-05-16
Mob claimed/unclaimed status


Jessland86:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30170304_1

**_Server Version_** (type `@revision` in game) **:**
581b5e7da03d7e1f088057abccf4e5af4334a1c5

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
If a party or alliance has claim on a mob and someone outside the alliance/party gets on the hate list and the party/alliance wipes the mob will not go unclaimed till every one on the hate list is dead or zones.

Deadwing888:
This is more or less correct behavior. Some of the ways mob can go white (disengaging weapon for instance) are not handled properly, but claim should hold on whoever has it irrespective of the hate list.

Jessland86:
I worded it incorrectly then, the mob will stay red claimed to a dead alliance and not go white until everyone is dead. Allowing an alliance to zombie kill things without chance of the mob ever going white and allowing other players a chance to kill it.

I also notice another interesting thing. Just attempting to claim a mob puts you on the hate list. even if the spell or JA. Heck even clicking "attack" and getting the message "Cannot attack. Your target is already claimed." puts you on the hate list. 

So on servers that actually have lots of competition to claim things the actual party that gets claim is basically safe to zombie the mob to death because lots of other players are on the hate list and it will stay read till everyone wipes or zones it.

Issue #2789, opened 2016-02-22
Missed weaponskills don't claim mobs


Hozu:
Mostly an issue for Jump since that was refactored to be like a weaponskill.

PR #6413, opened 2020-02-09
Claim updates

cocosolos:
Add a var to char entity to keep track of currently claimed mob and centralized all places that currently set claim/owner to battleutils::ClaimMob.
* For battleutils::ClaimMob to claim successfully, someone from the attacking players alliance must be top of the hate list.
* A killing blow does not cause anything to go unclaimed.
* When a mob is claimed, any mob owned by the attacker will go unclaimed.
* Claiming a mob or attacking a claimed mob will update the players claimed mob var and unclaim anything owned by that player. Each action performed on a mob updates the actual owner.
* The claimed mob var is cleared on load, zone, disengage, mob or player death, when the player is charmed, when the mob is charmed, and when a claiming action is performed.
* When current owner disengage, is charmed, or dies, will try to pass claim to alliance member, otherwise will go unclaimed. For the claim to pass, the member must have performed at least 1 claiming action on the mob and must not have done anything (like disengage) to clear their claimed mob var.
* Clean mob owner after exp/loot is distributed to deal with an issue with Aern reraising causing problems. Also refactored Aern mixin to work correctly with these changes. Aern will now reraise claimed and unclaim anything else that may have been claimed in the mean time.
* Popped mobs using entity:updateClaim will likewise unclaim any mobs owned by the popping players party. In cases where multiple mobs pop from a qm, only 1 will pop claimed.
* All aoe spells, abilities, and weaponskills end with claim being on targeted mob.

Issues:
#2109 functionality based mostly on these comments
#4240 point 2
#3923 claim issue
#3876
#2789
#1738
#1900

### DIFF
--- a/scripts/mixins/families/aern.lua
+++ b/scripts/mixins/families/aern.lua
@@ -11,34 +11,34 @@ g_mixins = g_mixins or {}
 g_mixins.families = g_mixins.families or {}
 
 g_mixins.families.aern = function(mob)
-    mob:addListener("DEATH", "AERN_DEATH", function(mob)
-        local reraises = mob:getLocalVar("AERN_RERAISE_MAX")
-        local curr_reraise = mob:getLocalVar("AERN_RERAISES")
-        if reraises == 0 then
-            if math.random() < 0.4 then
-                reraises = 1
-            end
-        end
-        if curr_reraise < reraises then
-            local dropid = mob:getDropID()
-            mob:setDropID(0)
-            local target = mob:getTarget()
-            local targetid = 0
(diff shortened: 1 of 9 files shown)


KnowOne134:
i beleive that a party can have 1 mob claimed that party for however many people are in the party, pets do not count.
so a party of 3 can all be attacking different mobs and all 3 are claimed to that party

cocosolos:
Interesting. It looks like you're right, I'll see what I can do about it. I think it should be as easy as not unclaiming party member mobs in the claim function.

cocosolos:
needs work, will revise

PR #6419, opened 2020-02-18
Charm Fix

nfaletra:
Fixed charm level and charisma ratio calculations. They were backwards.
Also fixed IncrediblyEasyPrey typo.


whasf:
travis-ci completed, but not reported back :(

PR #6424, opened 2020-04-12
Tredecim Scythe

Caelic:
the 13th hit guaranteed critical is not coded yet so here it is. wiki says it shouldnt count WS hits and this code follows that, this is just for autoattacks. wiki says it should save hit count when zoning but that will require adding an extra field within the database and i didn't want to clutter the database more to add "cross-zone memory" just seems unimportant. upon zoning the count is just set to 0.

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -1426,6 +1426,12 @@ bool CBattleEntity::OnAttack(CAttackState& state, action_t& action)
     actionList_t& list = action.getNewActionList();
 
     list.ActionTargetID = PTarget->id;
+    
+    bool tredecim = false;
+	if (this->objtype == TYPE_PC && (((CCharEntity*)this)->getEquip(SLOT_MAIN))->getID() == 18052)
+	{
+		tredecim = true;
+	}
 
     /////////////////////////////////////////////////////////////////////////
     //	Start of the attack loop.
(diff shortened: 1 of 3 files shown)


src/map/entities/battleentity.cpp
```diff
@@ -1426,6 +1426,12 @@ bool CBattleEntity::OnAttack(CAttackState& state, action_t& action)
     actionList_t& list = action.getNewActionList();
 
     list.ActionTargetID = PTarget->id;
+    
+    bool tredecim = false;
+	if (this->objtype == TYPE_PC && (((CCharEntity*)this)->getEquip(SLOT_MAIN))->getID() == 18052)
```
zach2good, line 1431: It's not good to hard code specific item IDs in the core, it would be much better to handle this with an item mod

src/map/entities/battleentity.cpp
```diff
@@ -1514,7 +1523,11 @@ bool CBattleEntity::OnAttack(CAttackState& state, action_t& action)
             {
                 // Set this attack's critical flag.
                 attack.SetCritical(dsprand::GetRandomNumber(100) < battleutils::GetCritHitRate(this, PTarget, !attack.IsFirstSwing()));
-
+                if (tredecim && ((CCharEntity*)this)->m_hitCounter > 12)
```
zach2good, line 1526: If the mod was CRIT_ON_HIT_X, you could do:
```
if (hasMod(CRIT_ON_HIT_X))
    (((CCharEntity*)this)->m_hitCounter)++;

...

if (hasMod(CRIT_ON_HIT_X) && counter % modValue == 0)
    ((CCharEntity*)this)->m_hitCounter = 0;
	 attack.SetCritical(true);
```

Caelic:
this is the only item in the game that tracks x hits. if you want to generalize it then you have to set up a memory system so that every item with the mod will remember its own hit counter and not interfere with another item's hit counter. it's an unnecessary amount of work to generalize it if it's the only item that will ever use the latent.

zach2good:
I wonder if it's worth adding a new member to `CCharEntity` when only the population that equips the scythe at least once will ever need it, there's nothing else as item-specific in `CCharEntity`. As much as forever-CharVars are frowned upon, they could be another option to track the hit counter.

Issue #3425, opened 2016-10-09
Problems potentially arrising from lack of retail delivery box queue limtations


TeoTwawki:
**_I have:_**

- [x] searched existing issues (http://github.com/darkstarproject/darkstar/issues/) to see if the issue I am posting has already been addressed or opened by another contributor
- [x] checked the commit log to see if my issue has been resolved since my server was last updated

**_Client Version_** (type `/ver` in game) **:**
30160831-1

**_Server Version_** (type `@revision` in game) **:**
ac26b2eac9bea1caee27ee7c7325cdcec2dfef74

**_Source Branch_** (master/stable) **:**
master

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**
On retail you can only jam so much into a characters delivery box from stuff mailed to self and returned form AH. You can hit return to sender to stuff more than 8 things form mailing to yourself in there, since that will free up the send box, but there is still eventually a point where the game will cut you off with an error message and ever prevent you form listing more at the AH until you go reduce your clutter.

Well, on darkstar server I'm not seeing where such a limit is imposed. A character can queue up enough things to break their own delivery box and lose items. Combine this with the still unfixed issue #2673 allowing a single player to infinitely list thousands of something (even when not returned, each sale uses its own slot for the gil...lets face it, a lot of players are just massive idiots) and you've a recipe for a lot of annoyance.

On the upside trying to exploit these 2 bugs on servers where a bot buys the items wind up punishing themselves! 😈

Djcocha:
idk if helpful or related but issue #3350 is effected by this on Oldschool. With the same duplication method applying to things sent to deliv box from AH, like if I have 100 wind xtals posted for 100gil each, and someone buys them, but then buys just one Vulcan's Staff for lets say 50k, my DBox will them have 50k in one single slot, and I will then be able to go back into 100 more times(for the amount of crystal cash sent) for 50k each time (referencing #3350 to see why)

TeoTwawki:
Sounds like the overloaded mailbox may be the cause of #3350

maxtherabbit:
the way it's written currently, the dbox should support up to 65535-7 pending items

do you really have people getting up that high? or is there a smaller limit somewhere I'm missing?

TeoTwawki:
I'm not sure exactly, I was helping someone figure out why they were seeing errors when I noticed there seems to be no limit in place and figured that there *should be*.

[...]

PR #6426, opened 2020-04-15
Fix AH 7-item limit not being enforced correctly

### LINKED ISSUE #2673: AH allowing more than 7 listings
Logged into whasf's oldschool server to verify. I was able to list far more than 7 slots worth.

Please note that if you at all check your sales status, the client will impose its own restriction for the duration of the session. If you never check sales history, client relies on the server to restrict it. The server restriction is somehow broke but used to work.

Kreidos:
AH listing count is now checked at every sale,
and not just the cached value from the last time
player viewed their Sale History.

As previously discussed, the limit is enforced on the client so there's no reason to make it a configurable value.

Fixes #2673

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -2348,7 +2348,20 @@ void SmallPacket0x04E(map_session_data_t* session, CCharEntity* PChar, CBasicPac
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


Kreidos:
This would also alleviate a common cause of #3425, where players can list way too many items and overflow their mailboxes.

PR #6428, opened 2020-04-25
Fix COP 6-4 Airship Fight

RunAwayDSP:
Add missing var player:setCharVar("bcnm_instanceid", inst) to Airship Door so you can progress through COP 6-4


RunAwayDSP:
I guess topaz fixed this with a different method, I am unsure which is preferred. I am simply uploading a fix that someone requested some help with.

nfaletra (approved):

